PR #143, opened 2020-10-29
Refactor server boot code

jage:
* First round of refactoring, split up server setup code into `AvalancheMQ::Launcher`.
* Separate into more methods, hopefully easier to follow startup code and stack traces.
* Should make it easier to test some of this code (separate PR)


src/avalanchemq/launcher.cr
```diff
@@ -0,0 +1,233 @@
+require "file"
+require "systemd"
+
+module AvalancheMQ
+  class Launcher
+    @tls_context : (OpenSSL::SSL::Context::Server|Nil)
```
carlhoerberg, line 6: Can add question mark instead of union of nil in crystal

src/avalanchemq/launcher.cr
```diff
@@ -0,0 +1,233 @@
...
+    @first_shutdown_attempt = true
+    @log : Logger
+    @lock : (File|Nil)
```
carlhoerberg, line 9:
```suggestion
    @lock : File?
```

src/avalanchemq/launcher.cr
jage: Ah, TIL

jage:
Have done some tests locally (with SSL as well), seems to work. There's probably more we can do, but unless it's very weird I'd prefer to merge and do another pass later on.

carlhoerberg (approved):

PR #168, opened 2020-12-13
Without systemd

carlhoerberg:
Do it here or in systemd.cr?


jage:
Most SystemD-stuff was removed in #303 but we still report status.
Revisit the "make libsystemd optional" part of this PR?

jage:
This might not be needed, check if we can run without libsystemd-dev

jage:
Yes, solved in systemd.cr https://github.com/84codes/systemd.cr/commit/d39f3ebf6ac37e1871b5cd3bc3e6b428149b05b7

PR #230, opened 2021-01-27
Prevent closed channel frames

carlhoerberg:


src/avalanchemq/client/client.cr
```diff
@@ -191,6 +191,10 @@ module AvalancheMQ
 
     def send(frame : AMQP::Frame) : Bool
       return false if closed?
+      unless frame.channel.zero? || @channels[frame.channel]?.try &.running?
```
carlhoerberg, line 194: Inefficient, in the cases where we deliver from a `Channel`, then we can check `@running` already there, this implementation always requires an extra hash lookup.

src/avalanchemq/client/client.cr
```diff
@@ -231,6 +235,11 @@ module AvalancheMQ
     @write_lock = Mutex.new(:checked)
 
     def deliver(frame, msg)
+      return false if closed?
+      unless frame.channel.zero? || @channels[frame.channel]?.try &.running?
```
antondalgren, line 239: I find it easier reading unless statements if they are only depending on one boolean value and not a mix.
```suggestion
      unless channel_closed?(frame)
```
``` crystal
     private def channel_closed?(frame)
        frame.channel.zero? || @channels[frame.channel]?.try &.running?
      end
```

src/avalanchemq/client/client.cr
```diff
@@ -432,7 +440,7 @@ module AvalancheMQ
       else
         send AMQP::Frame::Channel::Close.new(frame.channel, code, text, 0, 0)
       end
-      @channels[frame.channel].running = false
+      @channels.delete(frame.channel).try &.close
```
carlhoerberg, line 443: This means that if we close the channel (most often due to precondition failures), we ignore all the rest of the frames for that channel that might be coming in, i think that's the way to do it, but it's different from before.

src/avalanchemq/client/client.cr
```diff
@@ -297,7 +305,7 @@ module AvalancheMQ
             @log.debug { "Discarding #{frame.class.name}, waiting for Close(Ok)" }
             frame.body.skip(frame.body_size)
           else
-            yield ch
+            @log.debug { "Discarding #{frame.inspect}, waiting for Close(Ok)" }
```
carlhoerberg, line 308: This means that we ignore all frames coming in for a channel that's already been closed, before we could still process them, which seems wrong

PR #231, opened 2021-01-27
Improving federation reliability.

baelter:
Apart from refactoring links to keep references to the upstream and downstream connections this PR makes sure we send publish confirms before we deliver messages to consumers.


carlhoerberg (approved):
lgtm

PR #234, opened 2021-01-28
Delete entities wasn't saved to definitions.amqp

### LINKED ISSUE #209: Deleted queues got recreated on server crash
Seen on:  v1.0.0-alpha.22 
What happened:
1. queue got created through amqpcat
2. queue got manually deleted through UI 
3. Server crashed due to GC error (Maybe restart does it aswell?)
4. On startup queue got recreated

Jan 20 13:54:46 <server>[59652]: vhost=****** queue=testqueue: (messages=0) Deleted
Jan 20 14:58:48 <server> [66463]: vhost=****** queue=testqueue: Restoring index
Jan 20 14:58:48 <server> [66463]: vhost=****** queue=testqueue: 0 messages

carlhoerberg:
In `def save` we tried to find the queue/exchange that was going to be
deleted and decied wheather to write or not if it was durable, but the
queue was already deleted from the hash so was never written. Resulting
in that if a crash occured deleted queues/exchanges reappeared.

Fixes #209

### DIFF
--- a/src/avalanchemq/vhost.cr
+++ b/src/avalanchemq/vhost.cr
@@ -29,7 +29,7 @@ module AvalancheMQ
 
     @exchanges = Hash(String, Exchange).new
     @queues = Hash(String, Queue).new
-    @save = Channel(AMQP::Frame).new(32)
+    @save = Channel(AMQP::Frame).new(128)
     @write_lock = Mutex.new(:checked)
     @wfile : MFile
     @log : Logger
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Tested the fix by `kill -9` the server process, that's harder to do in a spec... as the spec and the server runs in the same process.

oskgu360 (approved):
Cant say I grasp all the details/conditions, but looks good as far as i can tell! Good catch!

PR #238, opened 2021-02-02
Set the correct version in generated HTTP API docs

dentarg:
Get the version from git so we only need to bump it in one place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Lint and build HTTP API documentation
         run: npm install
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
dentarg, line 39: Had to add this here and in two more workflows, checkout step now takes 2s instead of 1s, so we should still be fine.

carlhoerberg:
you could also have used `shards version` instead of `git describe`

openapi/openapi.yaml
```diff
@@ -1,7 +1,8 @@
 ---
 openapi: 3.0.3
 info:
-  version: 1.0.0
+  version:
+    "$ref": "./avalanchemq_version.yaml#/version"
```
carlhoerberg, line 5: Why not use `"$ref": "./shards.yml#/version"`? then you don't need `create_avalanchemq_version_yaml` et al.

PR #244, opened 2021-02-11
Let read_loop handle cleanups.

### LINKED ISSUE #243: Connection closed is double the connection opened
image
`Client#cleanup` is called twice every time a connection is closed.

antondalgren:
Since the read_loop will continuously read the socket, any error originating from the socket will cause the cleanup to happen. If the read_loop is either sending or receiving a close, it will cleanup the client and close the socket.

image

Closes #243

### DIFF
--- a/src/avalanchemq/client/client.cr
+++ b/src/avalanchemq/client/client.cr
@@ -211,9 +211,11 @@ module AvalancheMQ
       true
     rescue ex : IO::Error | OpenSSL::SSL::Error
       @log.debug { "Lost connection, while sending (#{ex.inspect})" } unless closed?
+      close_socket
       false
     rescue ex : IO::TimeoutError
       @log.info { "Timeout while sending (#{ex.inspect})" }
+      close_socket
       false
     rescue ex
       @log.error { "Unexpected error, while sending: #{ex.inspect_with_backtrace}" }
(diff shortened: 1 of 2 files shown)


src/avalanchemq/client/client.cr
```diff
@@ -271,11 +267,9 @@ module AvalancheMQ
       true
     rescue ex : IO::Error | OpenSSL::SSL::Error | AMQ::Protocol::Error::FrameEncode
       @log.debug { "Lost connection, while sending (#{ex.inspect})" }
```
baelter, line 269: Will `AMQ::Protocol::Error::FrameEncode` raise something on `from_io` in `read_loop` as well??

baelter (commented):
Kinda tricky to follow what will happen in all error cases. Left some thoughts...

src/avalanchemq/client/client.cr
carlhoerberg: no it won't

carlhoerberg (changes requested):
I'm not sure this covers all cases, there are cases where sending a frame will yield an exception and the read loop won't be notified. But instead of running `cleanup` we can just close the socket instead, and should raise an exception in `read_loop`

PR #261, opened 2021-03-11
reuse buffers for restore_index!

snichme:
This is an idea to fix the 
```
GC Warning: Repeated allocation of very large block (appr. size 18599936):
	May lead to memory leak and poor performance
```
warnings from `restore_index! `

It happens since each file is 44mb which results in allocating a lot of arrays and deque. 

The idea is to reuse one array and one ready queue for each vhost. 
The array that holds the acked SPs are just clear for each queue and the ReadyQueue we to a shallow copy of and set to `@ready` on the queue. Shallow copy is enough since the SP is a struct. 

Things to consider: 

* Size of `@ack_buffer` in VHost
* Size of `@ready_buffer` in VHost
* Not sure if `purge` is correct to use for `ReadyQueue` as it might allocate a new `Deque` if it has been resized 

Also, the code isn't very pretty, but not sure how to share the buffers otherwise. Putting them on the vhost I think is the best place, no locks needed.

### DIFF
--- a/src/avalanchemq/exchange/exchange.cr
+++ b/src/avalanchemq/exchange/exchange.cr
@@ -141,7 +141,7 @@ module AvalancheMQ
         next if ms <= 0
         args["x-message-ttl"] = ms
       end
-      @persistent_queue = PersistentExchangeQueue.new(@vhost, q_name, args)
+      @persistent_queue = QueueFactory.make_persistent_queue(@vhost, q_name, args)
       @vhost.queues[q_name] = @persistent_queue.not_nil!
     end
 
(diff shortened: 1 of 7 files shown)


snichme:
ohh, compact_index! is "disabled" so it doesn't destroy my testing data :)

baelter:
Some concerns: We have to be really careful not to `Fiber.yield` during index restoration. This code may also be problematic if we enable multi-threading in the future. Maybe we should have locks even if it's not strictly necessary right now?

snichme:
Maybe we should lock the whole vhost when avalanchemq is starting? 

We could also have a special case for Queue::Declare, instead of using the `apply` method we move the `restore_index!` method to the vhost. This would keep the code more contained.

baelter:
> Also, can we be sure that the definitions.amqp file is sorted, so all Queue::Declare comes after each after?

Think it's in the same order as the frames arrived. So there can be other definitions in between, bindings, declare exchange etc.

snichme:
Moved some stuff around.
All new queues are created the same way as before, it's only for durable queues that is being loaded that we actually run the restore_index. 

The method restore_index is moved to queue_factory and the buffers it's using is owned by Vhost and I have a mutex around it per vhost. 

Cleaner code and easier to understand. 

Performance wise it's the same as before but it's doesn't complain about the "Repeated allocation of very large block "

snichme:
Moved the restore index code to QueueFactory

Would like some suggestions on how to make the `make()` methods in QueueFactory nicer, a lof of `if` s there now

baelter (approved):
Looks good,

Regarding ifs in the factory I think it's ok. Keep all the ifs there instead of spread out through all code.

PR #266, opened 2021-04-21
Delayed message exchange bug

snichme:


src/avalanchemq/vhost.cr
```diff
@@ -40,8 +40,7 @@ module AvalancheMQ
     @connections = Array(Client).new(512)
     @segments : Hash(UInt32, MFile)
     EXCHANGE_TYPES = %w(direct fanout topic headers
-      x-federation-upstream x-delayed-message
```
baelter, line 43: This makes it incompatible with rabbitmq, no?

snichme: No, that list only controls which types are listen in the UI. The API looks the same. 

In the UI you instead select which exchange you want and check "Delayed"  checkbox

src/avalanchemq/queue/queue.cr
```diff
@@ -536,19 +536,18 @@ module AvalancheMQ
...
-          if expire_in > 0
-            @log.debug { "No more message to expire" }
-            break
-          end
-
-          expire_msg(sp, :expired)
-          if (i += 1) == 8192
-            Fiber.yield
-            i = 0
+          if expire_in < 0
```
carlhoerberg, line 539:
```suggestion
          if expire_in <= 0
```

src/avalanchemq/queue/queue.cr
```diff
@@ -536,19 +536,18 @@ module AvalancheMQ
...
+              i = 0
+            end
+            true
+          else
+            false
           end
-          true
         else
-          break
+          false
```
carlhoerberg, line 550: we won't break early now? so all messages in the ready queue will always be tested? or does returning false break?

src/avalanchemq/vhost.cr
carlhoerberg: shouldn't `EXCHANGE_TYPES` then be in the http controller if it only affects the UI?

src/avalanchemq/queue/queue.cr
snichme: `shift`  method in ReadyQueue breaks if block returns false

baelter (approved):

Issue #201, opened 2021-01-20
Message bodies not deleted from disk on queue purge/delete


johanrhodin:
To reproduce:
1. Generate a file with somewhat larger messages
2. Send them to an AvalancheMQ server (for example `dd if=file | amqpcat --producer --uri=$CLOUDAMQP_URL --exchange testing --routing-key blah`)
3. Purge the queue.
At this point I expect disk usage to drop down to zero. It does not.
4. Delete the queue.
At this point I expect disk usage to drop down to zero. It does not.

johanrhodin:
I filled up a node with messages. Purged the queue. Deleted the queue. Still can't publish to it:

```ubuntu@test-drafty-bobcat-01:/var/lib/avalanchemq$ df -h
Filesystem       Size  Used Avail Use% Mounted on
/dev/root        7.6G  2.8G  4.8G  37% /
devtmpfs         458M     0  458M   0% /dev
tmpfs            477M     0  477M   0% /dev/shm
tmpfs             96M  976K   95M   2% /run
tmpfs            5.0M     0  5.0M   0% /run/lock
tmpfs            477M     0  477M   0% /sys/fs/cgroup
/dev/loop0        49M   49M     0 100% /snap/core18/1949
/dev/loop1        27M   27M     0 100% /snap/snapd/10709
/dev/nvme0n1p15   98M  274K   98M   1% /boot/efi
/dev/nvme1n1     4.0G  2.0G  2.0G  51% /var/lib/avalanchemq
tmpfs             96M     0   96M   0% /run/user/1000```

```du -sh /var/lib/avalanchemq/*
16K     /var/lib/avalanchemq/42099b4af021e53fd8fd4e056c2568d7c2e3ffa8
141M    /var/lib/avalanchemq/f09fca61ab45f00f00434923787f4d212159b528
4.0K    /var/lib/avalanchemq/users.json
4.0K    /var/lib/avalanchemq/vhosts.json```

After purge+delete I get a: `PRECONDITION_FAILED - Server low on disk space", @classid=60, @methodid=40`

carlhoerberg:
@johanrhodin is this reproducible in v1.0.0-alpha.25?

johanrhodin:
Yes.

antondalgren:
[...]

PR #267, opened 2021-04-21
Trigger GC loop when queue is purged or deleted

snichme:


snichme:
Fixes #201

antondalgren (approved):

spec/server_spec.cr
```diff
@@ -141,15 +141,6 @@ describe AvalancheMQ::Server do
     end
   end
 
-  it "should mark the vhost as dirty when a queue is purged" do
```
carlhoerberg, line 144: Can we write another test for this? but one that checks that bytes are actually recovered from disk or something?

src/avalanchemq/queue/queue.cr
```diff
@@ -825,6 +826,7 @@ module AvalancheMQ
       count = @ready.purge
       @log.debug { "Purged #{count} messages" }
       @vhost.dirty = true
+      @vhost.gc_loop.send nil # trigger gc for message clean up
```
carlhoerberg, line 829: how about a new method called `trigger_gc` that sets the dirty flag and sends nil to the channel?

src/avalanchemq/vhost.cr
```diff
@@ -672,32 +673,39 @@ module AvalancheMQ
...
-        unless @dirty
-          sleep Config.instance.gc_segments_interval
-          next
+        select
+        when @gc_loop.receive
+          run_gc(referenced_sps)
+        when timeout interval
+          next unless @dirty
+          run_gc(referenced_sps)
         end
```
carlhoerberg, line 686:
```suggestion
        select
        when @gc_loop.receive
        when timeout interval
        end
        run_gc(referenced_sps)
```

src/avalanchemq/vhost.cr
```diff
@@ -672,32 +673,39 @@ module AvalancheMQ
...
+                   "killing process #{ex.inspect_with_backtrace}")
+        exit 1
       end
-    rescue ex
-      @log.fatal("Unhandled exception in #gc_segments_loop, killing process #{ex.inspect_with_backtrace}")
-      exit 1
+    end
+
+    private def run_gc(referenced_sps)
+      return if @closed
```
carlhoerberg, line 695:
```suggestion
      return if @closed
      return unless @dirty
```

spec/server_spec.cr
snichme: Yes, some type of test that verifies that GC has run. Iäll take a look at it.

PR #275, opened 2021-09-08
Display the timestamp of the first & last message in a Queue.

### LINKED ISSUE #250: Show oldest and newest message timestamp in queue UI and API
This will make it  easier to spot consumer lag

antondalgren:
By displaying the timestamp, it's easier to observe if the consumer is lagging behind.

image

closes #250


jage (approved):
Awesome, thanks! 🎉

PR #279, opened 2021-09-26
Select queues

Christina-84:
Work in progress on a feature to select multiple queues
Skärmavbild 2021-09-26 kl  18 38 14

### DIFF
--- a/static/queues.html
+++ b/static/queues.html
@@ -46,8 +46,10 @@
             </thead>
             <tbody></tbody>
           </table>
-          <div id="selected-controls" style="visibility: hidden;"> 
-            <button>Delete</button>
+          <div id="selected-controls" style="visibility: hidden; padding: 5px; border: solid; border-width: thin; display: inline-block;"> 
+              <button>Delete</button>
+              <button>Purge</button>
+              <button>Pause</button>
           </div>
         </div>
       </section>
@@ -144,20 +146,20 @@
           const view = item.internal ? 'queue_internal' : 'queue'
           queueLink.href = '/' + view + '?vhost=' + encodeURIComponent(item.vhost) + '&name=' + encodeURIComponent(item.name)
           queueLink.textContent = item.name
+
           var checkbox = document.createElement('input')
           checkbox.type='checkbox'
-          checkbox.id='selected-checkbox'
+
           checkbox.addEventListener('change', function() {
-            //Needs fixing
-            if (document.getElementById('selected-checkbox').checked) {
-              document.getElementById('selected-controls').style.visibility="visible"
-              console.log(document.getElementById('selected-checkbox'))
+            if((this.hasAttribute('data-select'))){
+              this.removeAttribute('data-select')
             } else {
-              document.getElementById('selected-controls').style.visibility="hidden"
-              console.log(document.getElementById('selected-checkbox').checked == 'selected-checkbox')
+              this.setAttribute('data-select', '')
             }
-          }); 
-          
+
+            selectedCounter();
+          });
+      
           avalanchemq.table.renderCell(tr, 0, checkbox) 
           avalanchemq.table.renderCell(tr, 1, item.vhost)
           avalanchemq.table.renderCell(tr, 2, queueLink)
@@ -180,6 +182,15 @@
         avalanchemq.table.renderCell(tr, 11, avalanchemq.helpers.formatNumber(item.message_stats.deliver_details.rate), 'right')
         avalanchemq.table.renderCell(tr, 12, avalanchemq.helpers.formatNumber(item.message_stats.ack_details.rate), 'right')
       })
+
+      function selectedCounter() {
+        var selectedList = document.querySelectorAll('[data-select]');
+        if (selectedList.length >= 1) {
+              document.getElementById('selected-controls').style.visibility="visible"
+            } else {
+              document.getElementById('selected-controls').style.visibility="hidden"
+            }
+      } 
 
       document.querySelector('#declare').addEventListener('submit', function (evt) {
         evt.preventDefault()
--- a/static/queues.html
+++ b/static/queues.html
@@ -29,6 +29,7 @@
           <table id="table" class="table">
             <thead>
               <tr>
+                <th>Selected</th>
                 <th data-sort-key="vhost">Virtual host</th>
                 <th data-sort-key="name" class="left">Name</th>
                 <th>Features</th>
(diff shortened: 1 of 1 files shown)


static/queues.html
```diff
@@ -140,9 +160,26 @@ <h3>Add queue</h3>
...
-          avalanchemq.table.renderCell(tr, 0, item.vhost)
-          avalanchemq.table.renderCell(tr, 1, queueLink)
-          avalanchemq.table.renderCell(tr, 2, features, 'center')
+
+          var checkbox = document.createElement('input')
+          checkbox.type='checkbox'
+          checkbox.setAttribute('data-vhost', encodeURIComponent(item.vhost))
+          checkbox.setAttribute('data-name', encodeURIComponent(item.name))
+
+          checkbox.addEventListener('change', function() {
```
carlhoerberg, line 169: doesn't checkboxes have a native checked attribute? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#attr-checked

static/queues.html
```diff
@@ -151,17 +188,26 @@ <h3>Add queue</h3>
...
+        avalanchemq.table.renderCell(tr, 12, avalanchemq.helpers.formatNumber(item.message_stats.ack_details.rate), 'right')
       })
 
+      function selectedCounter() {
+        var selectedList = document.querySelectorAll('[data-select]');
+        if (selectedList.length >= 1) {
+              document.getElementById('selected-controls').style.visibility="visible"
+            } else {
+              document.getElementById('selected-controls').style.visibility="hidden"
+            }
```
carlhoerberg, line 208: indentation seems off here

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
       document.querySelector('#dataTags').onclick = e => {
         window.avalanchemq.helpers.argumentHelperJSON("arguments", e)
       }
+
+      document.querySelector("#deleteQueues").addEventListener('submit', function (evt) {
+        evt.preventDefault()
+
+        if (window.confirm('Are you sure? The selected queues are going to be deleted. Messages cannot be recovered after deletion.')) {
+          var selectedList = document.querySelectorAll('[data-select]')
```
carlhoerberg, line 238: always use `const` or `let`, never `var`. var has weird/outdated scoping rules. in this case you want to use `const` as the variable will never change.

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+          var vhost; 
+          var name; 
+
+          for (let i = 0; i < selectedList.length; i++) {
+            vhost = selectedList[i].dataset.vhost
+            name = selectedList[i].dataset.name
```
carlhoerberg, line 244:
```suggestion
            const vhost = selectedList[i].dataset.vhost
            const name = selectedList[i].dataset.name
```

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
         window.avalanchemq.helpers.argumentHelperJSON("arguments", e)
       }
+
+      document.querySelector("#deleteQueues").addEventListener('submit', function (evt) {
+        evt.preventDefault()
+
+        if (window.confirm('Are you sure? The selected queues are going to be deleted. Messages cannot be recovered after deletion.')) {
+          var selectedList = document.querySelectorAll('[data-select]')
+          var vhost; 
+          var name;
```
carlhoerberg, line 240:
```suggestion
```

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+          var selectedList = document.querySelectorAll('[data-select]')
+          var vhost; 
+          var name; 
+
+          for (let i = 0; i < selectedList.length; i++) {
+            vhost = selectedList[i].dataset.vhost
+            name = selectedList[i].dataset.name
+            const url = '/api/queues/' + vhost + '/' + name
+            avalanchemq.http.request('DELETE', url)
+            .then(() => { window.location = '/queues' })
```
carlhoerberg, line 247: won't this cause a page reload, for each delete?

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+            name = selectedList[i].dataset.name
+            const url = '/api/queues/' + vhost + '/' + name + '/contents'
+            avalanchemq.http.request('DELETE', url)
+            .then(() => { avalanchemq.dom.toast('Queues purged!') })
+            .catch(avalanchemq.http.standardErrorHandler)
+          }
+        }
+      })
+
+      document.querySelector("#pauseQueues").addEventListener('submit', function (evt) {
```
carlhoerberg, line 273: what about unpause/resume?

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+          for (let i = 0; i < selectedList.length; i++) {
+            vhost = selectedList[i].dataset.vhost
+            name = selectedList[i].dataset.name
+            const url = '/api/queues/' + vhost + '/' + name
+            avalanchemq.http.request('DELETE', url)
+            .then(() => { window.location = '/queues' })
+            .catch(avalanchemq.http.standardErrorHandler)
+          }
+        }
+
```
carlhoerberg, line 251:
```suggestion
```

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+            const url = '/api/queues/' + vhost + '/' + name + '/pause'
+            avalanchemq.http.request('PUT', url)
+            .then(() => {
+              avalanchemq.dom.toast('Queues paused!')
+            })
+            .catch(avalanchemq.http.standardErrorHandler)
+          }
+        }
+      })
+
```
carlhoerberg, line 293:
```suggestion
```

static/queues.html
```diff
@@ -184,6 +230,67 @@ <h3>Add queue</h3>
...
+          var name; 
+
+          for (let i = 0; i < selectedList.length; i++) {
+            vhost = selectedList[i].dataset.vhost
+            name = selectedList[i].dataset.name
+            const url = '/api/queues/' + vhost + '/' + name + '/pause'
+            avalanchemq.http.request('PUT', url)
+            .then(() => {
+              avalanchemq.dom.toast('Queues paused!')
+            })
```
carlhoerberg, line 288:
```suggestion
            .then(() => avalanchemq.dom.toast('Queues paused!'))
```

static/queues.html
snichme: We talked about this and it will be changed so that the pages isn't reloaded at all.

PR #294, opened 2021-10-25
Allow Get/Reject from paused queue via the API

### LINKED ISSUE #293: Pause queue should only pause the consumers on the queue
Pause queue should pause the consumers on the queue but it should be possible to get/reject messages on the queue from the UI and API

The idea behind this feature is to be able to remove messages from a queue without the need to stop the consumers.
When pause all consumers should be stopped and all messages should be moved from unacked to ready.

snichme:
This allows the workflow of pausing all the consumers and pick the messages in the queue without the
need of stopping any clients

Fixes #293

### DIFF
--- a/src/avalanchemq/queue/queue.cr
+++ b/src/avalanchemq/queue/queue.cr
@@ -681,7 +681,7 @@ module AvalancheMQ
     end
 
     def basic_get(no_ack, force = false) : Envelope?
-      return nil unless @state.running? || @state.paused? && force
+      return nil if !@state.running? && (@state.paused? && !force)
       @last_get_time = Time.monotonic
       @get_count += 1
       if env = get(no_ack)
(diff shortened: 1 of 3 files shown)


jage:
@carlhoerberg you got time to review this? Would be nice to get this into main. IMO it makes sense, but would be nice to have another set of eyes on this.

carlhoerberg (approved):

src/avalanchemq/queue/queue.cr
```diff
@@ -680,8 +680,8 @@ module AvalancheMQ
       true
     end
 
-    def basic_get(no_ack) : Envelope?
-      return nil unless @state.running?
+    def basic_get(no_ack, force = false) : Envelope?
+      return nil unless @state.running? || @state.paused? && force
```
carlhoerberg, line 684: avoid double negation with unless

jage:
👍 Please add an entry in the Changelog when you have time. :)

PR #301, opened 2021-11-01
rm 2GB limit on mfile

carlhoerberg:
No need for a Slice/Bytes, use a Pointer instead. Just had to add a
constructor to IO::Memory that took a Pointer as well.

### DIFF
--- a/src/avalanchemq/message.cr
+++ b/src/avalanchemq/message.cr
@@ -33,7 +33,7 @@ module AvalancheMQ
       rk = AMQP::ShortString.from_io io, format
       pr = AMQP::Properties.from_io io, format
       sz = UInt64.from_io io, format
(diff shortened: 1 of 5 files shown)


carlhoerberg:
compacting the 100M msgs index takes ~25 seconds though... and then GC segments takes like 6, so the server is mostly blocked when consuming from that queue

jage:
> compacting the 100M msgs index takes ~25 seconds though... and then GC segments takes like 6, so the server is mostly blocked when consuming from that queue

Hehe, yeah noticed that too. A bit faster with more IOPS/bandwidth to the disk, but consume performance is getting worse the longer the queue right now. But better slow than crashing the connection (which this PR fixes). :)

jage (approved):

PR #304, opened 2021-11-02
Reference count segments and delete when zero references

carlhoerberg:
Way faster than collecting references and hole punch segments.
But of course not as space efficent. Can be somewhat mitigated with
smaller segment sizes, and in the future compact segments (somehow).


jage:
Looks reasonable, and I've been running this with some synthetic traffic for a day without issues.
I'm thinking about exceptions, if there's a risk of missing a inc/dec in case of an unexpected error, but haven't fully thought this through if it's possible.

carlhoerberg:
yeah, somethings is up because the durable exchange specs fails..

carlhoerberg:
Fixed the durable exchange problem. It's much faster consuming a 100M msgs queue now. Previously bytesize was recaluclated each time, now we have a counter that keeps adds/substracts when messages are added to the ready deque

carlhoerberg:
The only problem is that it's "relatively" easy to exhaust disk space now, as now compaction of individual segments is being made. Publish and consume from one queue very fast, only publish to another queue very slowly. No segments will be deleted as at least one message from the slow queue is in each segment.

src/avalanchemq/queue/durable_queue.cr
```diff
@@ -110,16 +107,23 @@ module AvalancheMQ
...
       super
-      begin
-        @ack_lock.synchronize do
+      @ack_lock.synchronize do
+        begin
           @log.debug { "writing #{sp} to ack" }
           @ack.write_bytes sp
+        rescue IO::EOFError
+          half_enq_size = @enq.size // 2
+          if @ack.size < half_enq_size && @ack.size < 512 * 1024**2
```
carlhoerberg, line 116: hard coded limit so it won't grow indefintly

carlhoerberg:
todo: when consuming with manual ack and then with auto ack sometimes `Unexpected exception in deliver_loop: Missing key 7 (KeyError)` can happen, or `Invalid memory access (signal 11) at address 0x7fb15cd45fed` 
```
bin/avalanchemqperf throughput -z 5 -a
bin/avalanchemqperf throughput -z 5
```

carlhoerberg:
Fixed now in https://github.com/cloudamqp/avalanchemq/pull/304/commits/53491c42ba5b814404b4a7dd94a70123af972cbd

carlhoerberg:
Big question is if only deleting segments when no messages resides in them is enough or if too much disk space will be wasted...

PR #305, opened 2021-11-03
Only allow guest through localhost

tbroden84:
Controls if the username is set to 'guest' and if the remote_address is a loopback address both HTTP server (port 15672) and AMQPConnection (5672).


src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -58,6 +58,11 @@ module AvalancheMQ
         return false if user.password.to_s.empty?
         user.password.not_nil!.verify(password)
       end
+
+      private def guest_localhost?(context, user)
+        return true unless user.name == "guest"
+        context.request.remote_address.as(Socket::IPAddress).loopback?
```
carlhoerberg, line 64: can "RemoteAddress" be somthing else? could this raise an exception at some point?

src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -58,6 +58,11 @@ module AvalancheMQ
         return false if user.password.to_s.empty?
         user.password.not_nil!.verify(password)
       end
+
+      private def guest_localhost?(context, user)
+        return true unless user.name == "guest"
```
carlhoerberg, line 63: Should we make it configurable? or is it good enough to hardcode at this point?

carlhoerberg (commented):
Should we make it configurable? either with the username or to not do the check at all

src/avalanchemq/http/handler/basic_auth.cr
jage: Maybe it could be nice to override, just realised I actually did use `guest/guest` over network, but a private network. (used my Tailscale cert to test HTTPS/AMQPS on my machine https://tailscale.com/blog/tls-certs/).

... or make it easier to get a non-guest default user (create random username/password, print URL to stdout for quick usage).

Will we block easy Docker test usage? (not localhost).

</rambling>

carlhoerberg: Yes, been thinking the same, witht the random password for default user. And yes, the current hard block first for a non local connections is a blocker for docker adaption. Rabbitmq disables the protection in their image, and allow guest

tbroden84: So at least make the check enabled/disabled in the configuration? Looking in to the user store, as for now there is no way to detect a user as default what I can see. Could another tag or field be used for this?

jage: > So at least make the check enabled/disabled in the configuration? 

Yes, and default disabled. Guess a config file and CLI option would be good (CLI option could be used for container).

> Looking in to the user store, as for now there is no way to detect a user as default what I can see. Could another tag or field be used for this?

Correct, It's "dynamic", first user with tag Admin IIRC. There was a field in `VHost` that I removed recently since it wasn't updated correctly see https://github.com/cloudamqp/avalanchemq/pull/290 and https://github.com/cloudamqp/avalanchemq/issues/276 

Yes maybe we need to rethink that, not sure what's the best way forward.

Feel free to split up into two PRs, i.e. block guest/guest (with option to disable) here, then another PR that will create random the password (or username + password).

src/avalanchemq/http/handler/basic_auth.cr
tbroden84: HTTP Server binds to TCP, TCP with TLS or unix socket. So unless a family check is done I assume it could fail with Socket::IPAddress cast if the connection is done through the unix socket. Also isn't the connection done through a unix socket always on the same host?

extras/config.ini
```diff
@@ -7,6 +7,7 @@ socket_buffer_size = 16384
 ;tls_cert = /etc/avalanchemq/cert.pem
 ;tls_key = /etc/avalanchemq/key.pem
 ;tls_compability_mode = false
+block_default_user_remotely = false
```
jage, line 10: We only block "guest" in this PR, so a default user could be another user. 
Are we calling it `block_default_user_remotely` as this might change, or should we call it `block_guest_user_remotely`?

jage (commented):
Looks good! Do you think we could add a spec?

extras/config.ini
tbroden84: As I understood it, we would like to be able to change the guest user to default user. Otherwise it better be called `block_guest_user_remotely`.

jage (approved):
Code looks great!
Specs would be nice, but if it's too tricky we can add it later on.

tbroden84:
Noticed all spec calls using IPv6 loopback address `[::1]`, so not sure if it's possible to simulate non loopback remote address.

extras/config.ini
```diff
@@ -7,6 +7,7 @@ socket_buffer_size = 16384
 ;tls_cert = /etc/avalanchemq/cert.pem
 ;tls_key = /etc/avalanchemq/key.pem
 ;tls_compability_mode = false
+allow_guest_user_remotely = false
```
carlhoerberg, line 10: what about? 

```suggestion
guest_only_loopback = true
```

src/avalanchemq/client/amqp_connection.cr
```diff
@@ -158,5 +159,11 @@ module AvalancheMQ
       end
       nil
     end
+
+    private def self.allow_guest_user?(remote_address, user)
```
carlhoerberg, line 163:
```suggestion
    private def self.guest_only_loopback?(remote_address, user) : Boolean
```

extras/config.ini
tbroden84: Yes probably better to reflect what the check actually does.

src/avalanchemq/client/amqp_connection.cr
```diff
@@ -88,10 +88,11 @@ module AvalancheMQ
       end
     end
 
-    def self.authenticate(socket, users, start_ok, log)
+    def self.authenticate(socket, remote_address, users, start_ok, log)
       username, password = credentials(start_ok)
       user = users[username]?
-      return user if user && user.password && user.password.not_nil!.verify(password)
+      return user if user && user.password && user.password.not_nil!.verify(password) &&
+        allow_guest_user?(remote_address, user)
```
tbroden84, line 95:
```suggestion
        guest_only_loopback?(remote_address, user)
```

src/avalanchemq/client/amqp_connection.cr
```diff
@@ -158,5 +159,11 @@ module AvalancheMQ
       end
       nil
     end
+
+    private def self.allow_guest_user?(remote_address, user)
+      return true unless user.name == "guest"
+      return true if Config.instance.allow_guest_user_remotely
```
tbroden84, line 165:
```suggestion
      return true unless Config.instance. guest_only_loopback
```

src/avalanchemq/config.cr
```diff
@@ -39,6 +39,7 @@ module AvalancheMQ
     property segment_size : Int32 = 8 * 1024**2 # bytes
     property raise_gc_warn : Bool = false
     property data_dir_lock : Bool = true
+    property allow_guest_user_remotely : Bool = false
```
tbroden84, line 42:
```suggestion
    property guest_only_loopback : Bool = true
```

src/avalanchemq/config.cr
```diff
@@ -84,6 +85,7 @@ module AvalancheMQ
         when "tls_key"              then @tls_key_path = v
         when "tls_ciphers"          then @tls_ciphers = v
         when "tls_min_version"      then @tls_min_version = v
+        when "allow_guest_user_remotely" then @allow_guest_user_remotely = true?(v)
```
tbroden84, line 88:
```suggestion
        when "guest_only_loopback" then @guest_only_loopback = true?(v)
```

src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -26,7 +26,7 @@ module AvalancheMQ
             begin
               username, password = Base64.decode_string(base64).split(":")
               if user = @user_store[username]?
-                if valid_auth?(user, password)
+                if valid_auth?(user, password) && allow_guest_user?(context, user)
```
tbroden84, line 29:
```suggestion
                if valid_auth?(user, password) && guest_only_loopback?(context, user)
```

src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -58,6 +58,18 @@ module AvalancheMQ
         return false if user.password.to_s.empty?
         user.password.not_nil!.verify(password)
       end
+
+      private def allow_guest_user?(context, user)
```
tbroden84, line 62:
```suggestion
      private def guest_only_loopback?(context, user) : Boolean
```

src/avalanchemq/server_cli.cr
```diff
@@ -53,6 +53,9 @@ module AvalancheMQ
         parser.on("-h", "--help", "Show this help") { puts parser; exit 1 }
         parser.on("-v", "--version", "Show version") { puts AvalancheMQ::VERSION; exit 0 }
         parser.on("--build-info", "Show build information") { puts AvalancheMQ::BUILD_INFO; exit 0 }
+        parser.on("--allow-guest-user-remotely=ALLOW", "Allow guest user remotely") do |v|
```
tbroden84, line 56:
```suggestion
        parser.on("--guest-only-loopback=BOOL", "Limit guest user to only connect from loopback address") do |v|
```

src/avalanchemq/server_cli.cr
```diff
@@ -53,6 +53,9 @@ module AvalancheMQ
...
+          config.allow_guest_user_remotely = {"true", "yes", "y", "1"}.includes? v.to_s
```
tbroden84, line 57:
```suggestion
          config.guest_only_loopback = {"true", "yes", "y", "1"}.includes? v.to_s
```

src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -58,6 +58,18 @@ module AvalancheMQ
...
+        return true unless user.name == "guest"
+        return true if Config.instance.allow_guest_user_remotely
```
tbroden84, line 64:
```suggestion
        return true unless Config.instance.guest_only_loopback
```

carlhoerberg:
yes, better!

src/avalanchemq/client/amqp_connection.cr
```diff
@@ -158,5 +159,11 @@ module AvalancheMQ
       end
       nil
     end
+
+    private def self.guest_only_loopback?(remote_address, user) : Boolean
```
carlhoerberg, line 163: typo:

```suggestion
    private def self.guest_only_loopback?(remote_address, user) : Bool
```

src/avalanchemq/http/handler/basic_auth.cr
```diff
@@ -58,6 +58,18 @@ module AvalancheMQ
         return false if user.password.to_s.empty?
         user.password.not_nil!.verify(password)
       end
+
+      private def guest_only_loopback?(context, user) : Boolean
```
tbroden84, line 62:
```suggestion
      private def guest_only_loopback?(context, user) : Bool
```

src/avalanchemq/config.cr
```diff
@@ -40,7 +40,8 @@ module AvalancheMQ
     property raise_gc_warn : Bool = false
     property data_dir_lock : Bool = true
     property tcp_keepalive : Tuple(Int32, Int32, Int32)? = { 60, 10 , 3 } # idle, interval, probes/count
-
+    property guest_only_loopback : Bool = true
+
```
tbroden84, line 44:
```suggestion

```

PR #309, opened 2021-11-11
Abort specs if AMQP server running (specs will fail)

jage:
Could be helpful as it's not uncommon to have another AMQP server
running, and some specs will work, some specs will not.


jage:
@antondalgren did a quick attempt with TCPServer listen but didn't manage to get a nice solution, is it more reasonable with this refactoring?

antondalgren (approved):
Perfect!

PR #315, opened 2022-01-13
Message safety

carlhoerberg:
put sp in unacked only when delivered to consumer and ack or unack basic_get msgs after they've been deliviered

### DIFF
--- a/src/avalanchemq/queue/queue.cr
+++ b/src/avalanchemq/queue/queue.cr
@@ -359,7 +359,7 @@ module AvalancheMQ
 
     private def deliver_to_consumer(c)
       # @log.debug { "Getting a new message" }
-      if env = get(c.no_ack)
+      get(c.no_ack) do |env|
         sp = env.segment_position
         # @log.debug { "Delivering #{sp} to consumer" }
         if c.deliver(env.message, sp, env.redelivered)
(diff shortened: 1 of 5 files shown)


carlhoerberg:
Extracted from #304

src/avalanchemq/queue/queue.cr
```diff
@@ -696,25 +697,27 @@ module AvalancheMQ
...
     end
 
-    def basic_get(no_ack, force = false) : Envelope?
-      return nil if !@state.running? && (@state.paused? && !force)
+    def basic_get(no_ack, force = false, &blk : Envelope -> Nil) : Bool
+      return false if !@state.running? && (@state.paused? && !force)
       @last_get_time = Time.monotonic
       @get_count += 1
       if env = get(no_ack)
+        yield env
```
carlhoerberg, line 705: this might raise, and then the msg have to be requeued

src/avalanchemq/queue/queue.cr
```diff
@@ -359,12 +359,14 @@ module AvalancheMQ
 
     private def deliver_to_consumer(c)
       # @log.debug { "Getting a new message" }
-      if env = get(c.no_ack, c)
+      if env = get(c.no_ack)
```
carlhoerberg, line 362: rescue exceptions in this block and requeue env from `get`

carlhoerberg (commented):
todo

jage (approved):
Changes look good, if it's possible to cook up some specs it would be great.

carlhoerberg:
yeah.. i think that's tricky because it's it would require the TCP socket to reset exactly while the server is sending the message, not sure to go about that..

carlhoerberg:
Found a way to test it!

PR #317, opened 2022-01-13
Let Ready have a bytesize counter instead of sum:ing

carlhoerberg:
Way faster for long queues


carlhoerberg:
Extracted from #304

snichme (approved):

jage (commented):
Specs to ensure we don't break this in the future (and thus break overflow `max-length-bytes`.

jage (approved):

PR #318, opened 2022-02-11
Close Queue if an unrecoverable ReadError is raised

jage:
Otherwise the read/expire loop will just continue to loop.

### DIFF
--- a/spec/api/definitions_spec.cr
+++ b/spec/api/definitions_spec.cr
@@ -46,18 +46,22 @@ describe AvalancheMQ::HTTP::Server do
     # https://github.com/cloudamqp/avalanchemq/issues/276
     context "if default user has been replaced" do
       before_each do
-        s.users.delete(s.users.default_user.name)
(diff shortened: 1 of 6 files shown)


jage:
TODO:

* [ ] How should `Exchange` handle a closed `Queue`?
* [ ] Consider adding a few more specs,  i.e. break it in other ways
* [ ] Fix default_user spec having side-effects

carlhoerberg:
Maybe distinguish between Segment file read error and Index read error?

jage:
> Maybe distinguish between Segment file read error and Index read error?

I think many cases are hard to distinguish, i.e. is it corrupt data in mfile or reading from wrong position, so most will probably be IndexError. But IO errors from the mfile could be marked as as Segment errors.

jage:
> * [x]  How should `Exchange` handle a closed `Queue`

We will handle it as deleted queues, just skip. nack on closed/errored queues will be a future improvement.

snichme (approved):

PR #320, opened 2022-02-17
RockyLinux 8.5 instead of CentOS 8 for RPM build

### LINKED ISSUE #319: GitHub Action: Build RPM broken
```
...
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
...
```

https://github.com/cloudamqp/avalanchemq/runs/5201205520?check_suite_focus=true

https://github.com/cloudamqp/avalanchemq/blob/f28eacf0cbda3923181a2939ccd458469d434369/.github/workflows/rpm.yml#L35

I think this is due to CentOS 8 being EOL and repository data removed/moved. Should we try to use CentOS 8 Stream or switch use another OS to build RPM?

snichme:
CentOS 8 has reach EOL

Fixes #319

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['centos:8']
+        os: ['rockylinux:8.5']
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


jage (approved):
TIL about Rocky Linux.

PR #321, opened 2022-02-17
Reset vhost

snichme:
Make a copy of the vhost data directory and purge all queues in the vhost

image


jage:
Would be nice with a sub-command in `avalanchemqctl` as well, i.e. just calling the API.

src/avalanchemq/vhost.cr
```diff
@@ -930,5 +930,23 @@ module AvalancheMQ
...
+
+    def purge_all_queues!(max_count : Int? = nil) : UInt32
+      sum = 0_u32
+      @queues.each_value do |q|
+        sum += q.purge(max_count)
+      end
+      sum
+    end
+
+    def reset!(backup_data = true)
```
jage, line 942: Could we write some specs?

snichme: specs are coming, just wanted to see your input first.

jage (commented):
Got seconds thoughts about what we call this method.
I.e. maybe we should call this something more specific "purge" or "purge_all_queues", reset is a bit ambiguous (will it remove queues, users etc?).

src/avalanchemq/vhost.cr
jage: ... and I think this makes sense, just want to be sure what will actually happens to consumers with unacked messages etc. :)

snichme: unacked will right now just be putted back in the queue, so need to find all consumers with unacked and clear them. Or maybe we can do something with just clear the `@unacked`  in `Queue` and when a consumers acks check if the tag is in @unacked.

jage (approved):

jage:
@snichme hmm, maybe we shouldn't drop all consumer channels, i.e. it's a big change on the purge feature (if used by a user, not by reset). Maybe make that optional?

snichme:
> @snichme hmm, maybe we shouldn't drop all consumer channels, i.e. it's a big change on the purge feature (if used by a user, not by reset). Maybe make that optional?

If we don't close all consumer channels there might still be unacked messages (on the channel not the queue) that will be requeued later. those messages wont be GC:ed. 
I think that this is more correct way of doing it than before, even if it's a more destructive action. We could instead write/warn about what will happen.

jage:
> If we don't close all consumer channels there might still be unacked messages (on the channel not the queue) that will be requeued later. those messages wont be GC:ed. I think that this is more correct way of doing it than before, even if it's a more destructive action. We could instead write/warn about what will happen.

Yes, agree that it could be seen as more correct, and we really do want this for the reset vhost, but still not convinced it's the right thing to do for a standard queue purge. I'm thinking we want it like this:

* reset vhost -> purge queues + cancel consumer channels and purge unacked messages
* purge queue -> just like before, don't removed unacked messages

AMQP 0.9.1 the spec specifies acked messages should not be purged:

> Purge
> ...
> This method removes all messages from a queue which are not awaiting acknowledgment.

https://www.rabbitmq.com/amqp-0-9-1-reference.html#queue.purge.reserved-1

---

We can still break this, and change it again, we're still in alpha etc, but not 100% convinced it's a good thing to break queue purge in the final product. :)

PR #323, opened 2022-02-22
wip: one Channel per queue that consumers receive messages from

carlhoerberg:


carlhoerberg:
How to handle consumer priority? :thinking:

PR #324, opened 2022-02-22
use distroless-debian as base image for container

carlhoerberg:
Images size goes from 90 to 48MB. 

static-debian11 works too, and then the image size is 30MB, but then "localhost" won't resolve to 127.1 :/


jage (approved):

carlhoerberg:
I'm thinking again and oppose the idea, we do want the container to contain a shell and other simple debug tools so that users can debug/edit files etc.

PR #327, opened 2022-02-26
Limit number of consumers listed on Queue page

snichme:
Start by showing 20 consumers and a button below the table if there are more consumers available
By clicking the button 10 more is loaded onto the list. Button disappear when all are loaded

The parameter to the API is optional and defaults to All to be backward compatible

image


snichme:
Also moved all inline JS to a separate file, there is a lot of JS on that page so now with caching the page size should be reduced

src/avalanchemq/queue/queue.cr
```diff
@@ -886,12 +886,20 @@ module AvalancheMQ
...
-    def to_json(builder : JSON::Builder)
+    def to_json(builder : JSON::Builder, limit : Int32 = -1)
       builder.object do
         details_tuple.each do |k, v|
           builder.field(k, v) unless v.nil?
         end
-        builder.field("consumer_details", @consumers)
+        builder.field("consumer_details") do
+          builder.array do
+            @consumers.each do |c|
```
carlhoerberg, line 896: I think you could write `.each.take(limit)`

carlhoerberg:
Very nice 👍

src/avalanchemq/queue/queue.cr
snichme: `each` on ConsumerStore doesn't return an Iterable (returns Nil), it only yields each item inside each group.

jage (approved):
Nice! :)

PR #330, opened 2022-03-07
Queue multi action

snichme:
Continuation on the work that Christina did before starting the thesis. 

Popup is shown when a checkbox is selected, one request per queue

TODO: 
- Checkbox in header that checks all queues

image

### DIFF
--- a/src/avalanchemq/queue/queue.cr
+++ b/src/avalanchemq/queue/queue.cr
@@ -893,7 +893,6 @@ module AvalancheMQ
         end
         builder.field("consumer_details") do
           builder.array do
-            # pp @consumers.each
             @consumers.each do |c|
               c.to_json(builder)
               limit -= 1
--- a/src/avalanchemq/queue/queue.cr
+++ b/src/avalanchemq/queue/queue.cr
@@ -893,6 +893,7 @@ module AvalancheMQ
         end
         builder.field("consumer_details") do
           builder.array do
+            # pp @consumers.each
             @consumers.each do |c|
               c.to_json(builder)
               limit -= 1
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -20,15 +20,13 @@
     let performed = 0
     elems.forEach(el => {
       const data = el.dataset;
-      const urlEncodedQueue = encodeURIComponent(data.name)
-      const urlEncodedVhost = encodeURIComponent(data.vhost)
       let url;
       switch(action) {
       case "delete":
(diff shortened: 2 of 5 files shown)


snichme:
It's a very big diff since I've moved the js into it's own file but the changes are on this lines: https://github.com/cloudamqp/avalanchemq/pull/330/files#diff-61e046fd549fa0d149192cf50d76a706bb66e85185b34a59760be34d5e644875R49-R68

static/js/queues.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+  const performMultiAction = (el) => {
+    const action = el.target.dataset.action
+    const elems = document.querySelectorAll("input[data-name]:checked")
+    const totalCount = elems.length
+    let performed = 0
+    elems.forEach(el => {
+      const data = el.dataset;
+      const urlEncodedQueue = encodeURIComponent(data.name)
+      const urlEncodedVhost = encodeURIComponent(data.vhost)
```
jage, line 24: Data elements already encoded, we double-encode now (breaking vhost `/`).

jage (commented):
Nice!

* Inline comment about double-encoded name/vhost
* If I select all, then purge, the "select all" checkbox is still selected
* Would be nice to get a "Close" or "X" link to close the action window if you change your mind

jage:
Not prioritised, but couldn't help myself to google this https://stackoverflow.com/questions/659508/how-can-i-shift-select-multiple-checkboxes-like-gmail :-)

snichme:
Was trying out the shift + click but felt that it's better to get this out and then add the shift select later (keep it simpler). 

But I know that people want it and it is a very nice feature.

jage:
> Was trying out the shift + click but felt that it's better to get this out and then add the shift select later (keep it simpler).

100% agree :)

src/avalanchemq/queue/queue.cr
```diff
@@ -893,6 +893,7 @@ module AvalancheMQ
         end
         builder.field("consumer_details") do
           builder.array do
+            # pp @consumers.each
```
jage, line 896: Reminder to remove

static/queues.html
```diff
@@ -24,11 +24,19 @@ <h2>
     <aside></aside>
     <main>
       <section class="card">
+        <div id="multiselect-controls" class="popup-card">
```
jage, line 27:
```suggestion
        <div id="multiselect-controls" class="popup-card hide">
```

Accidentally removed during extraction to CSS class.

jage (approved):

PR #333, opened 2022-03-21
Use curl instead of wget in Makefile

carlhoerberg:


dentarg:
Isn't the automatic retires done by `wget` very good to have? Could avoid flaky CI etc. Maybe fallback to `curl` when `wget` isn't present?

carlhoerberg:
Yes, I agree, that's why I went with `wget`. However, wget can only retry/resume if you download to a directory (`-P`), not if piping the output. And they seem to retry differently. Wget by default retries/resumes only if a download has already started, not on timeouts or 50x errors. `curl --retry` does that, and even respect the `retry-after` header if such is present.

snichme (approved):

PR #336, opened 2022-03-22
Read cgroup memory.max if available

### LINKED ISSUE #283: Get containerized/cgroup memory maximum
Podman/cgroupV2 exposes `/sys/fs/cgroup/memory.max` (and others) inside the container. Use that instead, when available, when checking physical memory available. (It can change, so poll it)

memory.current might be a better alternative than RSS too

carlhoerberg:
If memory is limited by a container or a systemd job (both are using
cgroups for doing that) we can now read that value instead of the
physical memory of the host machine.

Fixes #283

Or maybe it should be implemented here instead:
https://github.com/cloudamqp/avalanchemq/blob/4e12153602deab3113a1f8df5ab5d550101d933f/src/avalanchemq/server.cr#L280


carlhoerberg:
Also no support for older cgroup v1. I don't know docker still uses cgroup v1?

carlhoerberg:
Works in rootless podman too, but if `--net host` is used then also `--cgroupns host` is required.

carlhoerberg:
Docker desktop uses cgroup v2 since 4.3, released in december: https://docs.docker.com/desktop/mac/release-notes/#bug-fixes-and-minor-changes-4

PR #344, opened 2022-03-24
Efficient RSS fetching

carlhoerberg:


carlhoerberg:
let's go with #336 instead

PR #346, opened 2022-03-25
Count channels for each connection

### LINKED ISSUE #340: Keep a counter of channels per vhost
To speed up metric collection when the server got a lot of connections: 
https://github.com/cloudamqp/avalanchemq/blob/183a65cf7ed6c32094bc54487f551266759a3b47/src/avalanchemq/http/controller/prometheus.cr#L174-L176

snichme:
Makes it more efficient to produce metrics if there are a lot of connections and/or
channels

Fixes #340

### DIFF
--- a/src/avalanchemq/http/controller/prometheus.cr
+++ b/src/avalanchemq/http/controller/prometheus.cr
@@ -173,8 +173,10 @@ module AvalancheMQ
           connections += vhost.connections.size
           vhost.connections.each do |conn|
             channels += conn.channels.size
+            conn.channels.each_value do |ch|
+              consumers += ch.consumers.size
+            end
           end
-          consumers += vhost.consumers.size
           queues += vhost.queues.size
         end
         writer.write({name:  "connections",
--- a/src/avalanchemq/http/controller/prometheus.cr
+++ b/src/avalanchemq/http/controller/prometheus.cr
@@ -172,7 +172,7 @@ module AvalancheMQ
           unacked += d[:messages_unacknowledged]
           connections += vhost.connections.size
           vhost.connections.each do |conn|
-            channels += conn.channel_count
+            channels += conn.channels.size
           end
           consumers += vhost.consumers.size
           queues += vhost.queues.size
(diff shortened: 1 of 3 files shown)


src/avalanchemq/http/controller/prometheus.cr
```diff
@@ -172,7 +172,7 @@ module AvalancheMQ
           unacked += d[:messages_unacknowledged]
           connections += vhost.connections.size
           vhost.connections.each do |conn|
```
carlhoerberg, line 174: I think the problem is this, iterating all quarter million connections. `channels.size` is already a "property" today so i don't think this makes much of a difference. What we want is a `channel_count` property in `Vhost`.

carlhoerberg (changes requested):

src/avalanchemq/http/controller/prometheus.cr
snichme: One option is to not count channels if there are to many connections. But I can look at how it can be stored on vhost level.

carlhoerberg: yes, that might be a quickfix!

src/avalanchemq/http/controller/prometheus.cr
```diff
@@ -172,7 +172,7 @@ module AvalancheMQ
...
-            channels += conn.channels.size
+            channels += conn.channel_count
           end
           consumers += vhost.consumers.size
```
snichme, line 177: Hidden slow code here as well: consumers is a method and does this: 
```
def consumers
  @connections.flat_map { |conn| conn.channels.each_value.flat_map &.consumers }
end
```

I guess it was this that ate all the RAM

carlhoerberg: ok! yes!

carlhoerberg (approved):
Good catch! did you perf test it?

snichme:
> Good catch! did you perf test it?

No, not sure how you created that many connections to an instance so no tests yet.

carlhoerberg:
ok, can do it. we have `avalanchemqperf connection-count -x 50000 --localhost` in #339

PR #350, opened 2022-03-30
TLS Proxy

carlhoerberg:
Speed up TLS connections by having a separate process that deals with
TLS encryption/decryption.  That way multi-processing is utilised.


carlhoerberg:
Need `fork` for this to work reliably, which crystal has opted to not support.

PR #354, opened 2022-04-12
Only allow administrators to get system information

snichme:
For cases where you dont want to expose the underlying system information to all users that has
access to the broker.
Page is still available but no data will be shown.


snichme:
It's also solved here: https://github.com/cloudamqp/avalanchemq/pull/355 and in a nicer way :)

jage:
> It's also solved here: #355 and in a nicer way :)

Ah yeah, so lets go with that solution? :)

PR #355, opened 2022-04-20
Rates per vhost

snichme:
Moving rate metrics from server to each vhost

Will use some additional memory since we store more data per vhost but it allows us to get show metrics per vhost which is good for limited users. 
This also includes prometheus endpoint where it only shows metrics for the vhosts that the user has access to, additionally the requester can limit which vhost to get the "normal" and the prometheus metrics for using `?vhost=some_vhost`

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -32,6 +32,7 @@ describe AvalancheMQ::HTTP::ConsumersController do
       response = get("/metrics/detailed?family=connection_churn_metrics")
       response.status_code.should eq 200
       lines = response.body.lines
+      pp lines
       lines.any?(&.starts_with? "lavinmq_detailed_connections_opened_total").should be_true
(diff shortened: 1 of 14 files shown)


snichme:
Nodes tab now looks like this is you dont have the `administrator`  tag 
image

CPU, mem and IO graphs are shown but empty, we could hide them but there no support for that now so I left it empty

jage:
Testing this now.

snichme:
@jage Last commit https://github.com/cloudamqp/avalanchemq/pull/355/commits/5c09c145355faf4197c92f5aea9554a213b535e3 is removing the channel per vhost. Test with that after the first round of testing to see if there is any differences

carlhoerberg:
Very nice! Code looks good and my local testing checked out just fine

carlhoerberg (approved):
:star2:

PR #358, opened 2022-05-03
Truncate sparse index file on queue index restore

carlhoerberg:


src/avalanchemq/queue/durable_queue.cr
```diff
@@ -228,6 +220,24 @@ module AvalancheMQ
...
 
+    # Reads all SPs in the index, truncate if there's only 0's left
+    # As we preallocate files, they are not truncated on ungraceful shutdown
+    # and thus we allocate too large memory structures in restore_index.
+    # TODO: optimize using SEEK_HOLE or compare allocated blocks * block size vs file size
+    private def truncate_sparse_file(path)
+      File.open(path, "W") do |f|
+        f.buffer_size = Config.instance.file_buffer_size
+        f.advise(File::Advice::Sequential)
+        sp = SegmentPosition.from_io file
```
snichme, line 231: Dont you need to loop here to go through the whole file? 

and the variable is called `file` here but `f` in the block

src/avalanchemq/queue/durable_queue.cr
```diff
@@ -228,6 +220,24 @@ module AvalancheMQ
...
+        if sp.zero?
+          size = file.pos - SegmentPosition::BYTESIZE
```
snichme, line 233: When we find the end, why back up SP bytesize and truncate, doesn't that remove the last SP in the file?

snichme (commented):
Some comments, but otherwise I like it!

src/avalanchemq/queue/durable_queue.cr
carlhoerberg: no, because we have just read a 0 SP, our pos is past that one, we don't need that one so truncate to just before the 0-SP

src/avalanchemq/queue/durable_queue.cr
carlhoerberg: thanks

snichme (approved):

src/avalanchemq/queue/durable_queue.cr
```diff
@@ -228,6 +220,34 @@ module AvalancheMQ
...
+    # and thus we allocate too large memory structures in restore_index.
+    # TODO: optimize using SEEK_HOLE
+    private def truncate_sparse_file(path : String) : Nil
+      {% if flag?(:unix) %}
+        # if number of allocated blocks (times block size) is smaller than
+        # the file size then the file must be sparse
+        stats = File.info(path).@stat
+        return if stats.st_blocks * stats.st_blksize >= stats.st_size
+      {% end %}
+      File.open(path, "W") do |f|
```
jage, line 234:
```suggestion
      File.open(path, "W+") do |f|
```

Since we're reading as well as truncating now.

jage:
Nice! Could we get a spec? :)

jage (approved):

carlhoerberg:
The `SEEK_HOLE` optimization shrinks the time taken to find the position to truncate to from 500ms to 0.3ms when there's 2M msgs in the queue

carlhoerberg:
spec added

PR #362, opened 2022-06-08
Queue size metrics

tbroden84:
Expand messages size rates in queue view, with average sizes. Introduce message size snapshot, with manual trigger expensive calculation of max/min over ready and unacked queues. Only display this data upon user/debug action.

- Adds counter for unacked bytesize, same as [#317](https://github.com/cloudamqp/lavinmq/pull/317)
- Adds average calculation on queue (ready, unacked, total). "queue.counter / queue.size"
- Updates "Messages" card in queue view.
- Adds max/min calculation on queue (ready, unacked, total)
   - Separated endpoint and only calculated when invoked.
- Displayed snapshot in separated card in queue view.

Messags card:
Screenshot 2022-06-07 at 14 52 25

Message size snapshot:
Screenshot 2022-06-08 at 09 58 54

### DIFF
--- a/spec/unacked_queue_spec.cr
+++ b/spec/unacked_queue_spec.cr
@@ -33,11 +33,11 @@ describe LavinMQ::Queue::UnackQueue do
   it "should calculate message sizes" do
     q = LavinMQ::Queue::UnackQueue.new
     sps = [
-      LavinMQ::SegmentPosition.new(10,10,5u32),
-      LavinMQ::SegmentPosition.new(10,10,1u32),
-      LavinMQ::SegmentPosition.new(10,10,10u32),
-      LavinMQ::SegmentPosition.new(10,10,3u32),
-      LavinMQ::SegmentPosition.new(10,10,1u32)
+      LavinMQ::SegmentPosition.new(10, 10, 5u32),
+      LavinMQ::SegmentPosition.new(10, 10, 1u32),
+      LavinMQ::SegmentPosition.new(10, 10, 10u32),
+      LavinMQ::SegmentPosition.new(10, 10, 3u32),
+      LavinMQ::SegmentPosition.new(10, 10, 1u32),
     ]
     sps.each { |sp| q.push(sp, nil) }
     q.bytesize.should eq 20u32
--- a/spec/unacked_queue_spec.cr
+++ b/spec/unacked_queue_spec.cr
@@ -1,6 +1,35 @@
 require "./spec_helper"
 
-describe LavinMQ::Queue::ReadyQueue do
+describe LavinMQ::Queue::UnackQueue do
+  it "keeps track of bytesize" do
+    q = LavinMQ::Queue::UnackQueue.new
+    q.bytesize.should eq 0
+
+    sp1 = LavinMQ::SegmentPosition.new(0u32, 0u32, bytesize: 1u32)
+    sp2 = LavinMQ::SegmentPosition.new(1u32, 0u32, bytesize: 3u32)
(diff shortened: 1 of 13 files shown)


spec/unacked_queue_spec.cr
```diff
@@ -0,0 +1,24 @@
+require "./spec_helper"
+
+describe LavinMQ::Queue::ReadyQueue do
```
jage, line 3:
```suggestion
describe LavinMQ::Queue::UnackQueue do
```

src/lavinmq/queue/queue.cr
```diff
@@ -840,7 +852,7 @@ module LavinMQ
 
     def rm_consumer(consumer : Client::Channel::Consumer, basic_cancel = false)
       deleted = @consumers.delete_consumer consumer
-      consumer_unacked_size = @unacked.sum { |u| u.consumer == consumer ? 1 : 0 }
+      consumer_unacked_size = @unacked.bytesize # { |u| u.consumer == consumer ? 1 : 0 }
```
jage, line 855: Keep the conditional sum here?

spec/unacked_queue_spec.cr
```diff
@@ -0,0 +1,24 @@
...
+      LavinMQ::SegmentPosition.new(10,10,10u32),
+      LavinMQ::SegmentPosition.new(10,10,3u32),
+      LavinMQ::SegmentPosition.new(10,10,1u32)
+    ]
+    sps.each { |sp| q.push(sp, nil) }
+    q.bytesize.should eq 20u32
+    q.avg_bytesize.should eq 4u32
+    q.max_bytesize(&.sp.bytesize).should eq 10u32
+    q.min_bytesize(&.sp.bytesize).should eq 1u32
+    sps.each { |sp| q.purge }
```
jage, line 18:
```suggestion
    q.purge
```

spec/unacked_queue_spec.cr
```diff
@@ -0,0 +1,24 @@
+require "./spec_helper"
+
```
jage, line 2: I think we should have a "keeps track of bytesize" test here as well, similar to what we have in ReadyQueue, to ensure we always test subtraction during delete etc.

spec/unacked_queue_spec.cr
jage, line 3:
```suggestion
describe LavinMQ::Queue::UnackQueue do
```

tbroden84:
Message snapshot, table hidden until invoked. 

Screenshot 2022-06-29 at 10 40 21

src/lavinmq/queue/queue.cr
jage: Hmm, maybe I didn't understand the code good enough here ... :)

tbroden84: As for my understanding the Unacked queue class have an internal unacked queue. Each queue item is a unack record containing segment position and a consumer.

```
unacked = @unacked
unack = Unack.new(sp, consumer)
if idx = unacked.bsearch_index { |u| u.sp > sp }
  unacked.insert(idx, unack)
else
  unacked << unack
end
```

The sum can the be feed a block, filtering out the consumer.
```
def sum(&blk : Unack -> _) : UInt64
  @unacked.sum(0_u64, &blk)
end
```

On the otherhand `consumer_unacked_size` are only used to log the message size of the unacked messages when removing a specific consumer. So perhaps not even needed.

jage: Thanks. Yeah, I think we'd like to keep the correct `consumer_unacked_size` logging there, but put the calculation into the log block so it's only triggered in debug log mode.

PR #363, opened 2022-06-20
Fix: Don't loop over expire_queue if we have consumers

jage:
For queues with `expire` arguments we tried to loop over expire_queue
when expire TTL had passed and there still was a consumer.


jage:
Need to look at the specs one more time, see if we can trigger on failure.

carlhoerberg (approved):

PR #364, opened 2022-06-29
Bugfix: Ensure receive_or_expire returns bool

jage:
This caused the deliver_loop to stop, introduced in 09c426ad7cb240d9342b6b61b54a277e49bf8451, never released

Set return type to avoid similar issues.


jage:
Think we missed failing specs due to it just hangs, i.e. no GHA failure, just timeout. Will work on improvements in a separate PR.

snichme (approved):

Issue #342, opened 2022-03-24
Management UI: Sort by message rates


johanrhodin:
Can't sort by rates (publish rate, delivery rate, ack rate)
<img width="695" alt="Screen Shot 2022-03-24 at 14 57 56" src="https://user-images.githubusercontent.com/460369/160000076-3f5519d8-557f-4afd-b8c2-08bde5f198e6.png">

Also redelivery rate must be added.

PR #367, opened 2022-07-21
Sort Queues by rates

Christina-84:
This update allows users to sort queues by message rates. 
It also adds Redelivery rate to the queues table, and lets users sort by redelivery rate as well. 

#342 
sort_by_rates

### DIFF
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -44,28 +44,21 @@ module LavinMQ
           return context
         end
         all_items = filter_values(params, iterator.map(&.details_tuple))
-        if sort_by = params.fetch("sort", nil)
+        if sort_by = params.fetch("sort", nil).try &.split(".")
           sorted_items = all_items.to_a
           filtered_count = sorted_items.size
-          case sort_by
-          when /message_stats/
-            sort_key = sort_by.split(".").first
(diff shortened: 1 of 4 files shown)


jage:
"Docker images / main (pull_request)" is failing due to credentials being ignored since you were not a collaborator on this repo when the PR was opened. I've added you and tried to rerun but still fails, so let's just ignore it here as it will (most likely) work next time.

jage:
Did some experiment and found a weird behaviour.

3 queues with some traffic, sorting on delivery rate and expected the paused queue to be at the bottom, but it's in the middle? 😄 
Not sure if it's related to how rates are updated (auto-refresh) or if it's something else but maybe we can try to debug this now that we're looking at sorting. 

Screenshot 2022-08-02 at 16 29 12

carlhoerberg (approved):

carlhoerberg (changes requested):
oh, yeah, it's not as easy as just to add the `data-sort-key` attribute, those attributes have to map to something, and there's no `publish` property on the `item`

Christina-84:
We now sort by rates in queues, but would like to write more dynamic code as it is currently hardcoded to work for message_stats

carlhoerberg:
Fixed the dynamic sorting. Wrote a custom `dig` that always returns a `NamedTuple` except for the last sort key

PR #368, opened 2022-07-25
Show total ready and unacked messages

Christina-84:
Adds total number of ready and unacked messages for all queues to **stats** on **nodes page**

image

### DIFF
--- a/static/js/nodes.js
+++ b/static/js/nodes.js
@@ -11,28 +11,20 @@
   let updateTimer = null
 
   function update (cb) {
-    const statsPromise = lavinmq.http.request('GET', url)
-    Promise.all([
-      statsPromise
-    ]).then(function (responses) {
-      const statsData = responses[0]
-      render(statsData)
+    lavinmq.http.request('GET', url).then(function (response) {
+      render(response)
       if (cb) {
-        cb(statsData)
+        cb(response)
       }
     }).catch(lavinmq.http.standardErrorHandler).catch(stop)
   }
 
-  function render (data, queueData) {
+  function render (data) {
     document.querySelector('#version').innerText = data[0].applications[0].version
 
     for (const node of data) {
       updateDetails(node)
-      if (queueData instanceof Object) {
-        updateStats({ ...node, ...queueData })
-      } else {
-        updateStats(node)
-      }
+      updateStats(node)
     }
   }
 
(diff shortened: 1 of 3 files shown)


jage:
"Docker images / main (pull_request)" is failing due to credentials being ignored since you were not a collaborator on this repo when the PR was opened. I've added you and tried to rerun but still fails, so let's just ignore it here as it will (most likely) work next time.

static/js/nodes.js
```diff
@@ -9,27 +9,38 @@
...
-  if (data === null) {
-    update(render)
-  }
 
   function update (cb) {
-    lavinmq.http.request('GET', url).then(function (response) {
-      data = response
-      render(response)
+    const statsPromise = lavinmq.http.request('GET', url)
+    const queueStatsPromise = lavinmq.http.request('GET', '/api/queues')
```
jage, line 15: We might want to do this sum server side, if we have a lot of queues. Maybe we can do an experiment with 10k queues how long this takes here? :)

Christina-84: Yes true! Ill try putting it server side and re-iterate 👍

static/js/nodes.js
```diff
@@ -9,27 +9,30 @@
...
-  if (data === null) {
-    update(render)
-  }
 
   function update (cb) {
-    lavinmq.http.request('GET', url).then(function (response) {
-      data = response
-      render(response)
+    const statsPromise = lavinmq.http.request('GET', url)
+    Promise.all([
```
carlhoerberg, line 15: `Promise.all` for a single item is necessary.

static/js/nodes.js
```diff
@@ -9,27 +9,30 @@
...
+      statsPromise
+    ]).then(function (responses) {
```
carlhoerberg, line 17: use the `(response) => {}` syntax.

static/js/nodes.js
```diff
@@ -9,27 +9,30 @@
...
+      const statsData = responses[0]
```
carlhoerberg, line 18: are you sure there will always be one item? and why is an array returned if there's only a single item?

static/js/nodes.js
```diff
@@ -9,27 +9,30 @@
...
   }
 
-  function render (data) {
+  function render (data, queueData) {
     document.querySelector('#version').innerText = data[0].applications[0].version
+
     for (const node of data) {
       updateDetails(node)
-      updateStats(node)
+      if (queueData instanceof Object) {
```
carlhoerberg, line 31: when is it not?

carlhoerberg (changes requested):

Christina-84:
Was eager and pushed half ready, now i think the node.js file makes more sense

static/js/nodes.js
```diff
@@ -27,6 +21,7 @@
 
   function render (data) {
     document.querySelector('#version').innerText = data[0].applications[0].version
+
```
carlhoerberg, line 24:
```suggestion
```

carlhoerberg (approved):

jage (approved):
Great!

jage:
Feel free to squash merge, and add a row under unreleased to `Changelog`, either in this branch or directly in `main`.

PR #371, opened 2022-08-08
Collect metrics time

Weistrand:
Measures time it takes to collect metrics

### DIFF
--- a/spec/server_spec.cr
+++ b/spec/server_spec.cr
@@ -973,6 +973,7 @@ describe LavinMQ::Server do
     should_eventually(be_true, 1.seconds) { server.stats_loop_duration_seconds_total > Time::Span.zero }
     server.stats_loop_rates_duration_seconds.should_not eq Time::Span.zero
     server.stats_loop_system_duration_seconds.should_not eq Time::Span.zero
-    LavinMQ::Config.instance.stats_interval = stats_interval
+  ensure
+    LavinMQ::Config.instance.stats_interval = stats_interval if stats_interval
   end
 end
--- a/spec/server_spec.cr
+++ b/spec/server_spec.cr
@@ -968,7 +968,7 @@ describe LavinMQ::Server do
 
   it "should measure time it takes to collect metrics in stats_loop" do
     server = LavinMQ::Server.new("/tmp/spec")
-    should_eventually(be_true, 10.seconds) { server.stats_loop_duration_seconds_total > Time::Span.zero }
+    should_eventually(be_true, 6.seconds) { server.stats_loop_duration_seconds_total > Time::Span.zero }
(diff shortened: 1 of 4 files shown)


spec/server_spec.cr
```diff
@@ -954,4 +954,14 @@ describe LavinMQ::Server do
...
   end
+
+  it "should measure time it takes to collect metrics in stats_loop" do
+    stats_interval = LavinMQ::Config.instance.stats_interval
+    LavinMQ::Config.instance.stats_interval = 100
+    server = LavinMQ::Server.new("/tmp/spec")
+    should_eventually(be_true, 1.seconds) { server.stats_loop_duration_seconds_total > Time::Span.zero }
+    server.stats_loop_rates_duration_seconds.should_not eq Time::Span.zero
+    server.stats_loop_system_duration_seconds.should_not eq Time::Span.zero
+    LavinMQ::Config.instance.stats_interval = stats_interval
```
jage, line 965: Just a minor detail, if the spec crashes the `stats_interval` won't be restored. So we can put this in a `ensure` block.

```suggestion
  ensure
    LavinMQ::Config.instance.stats_interval = stats_interval if stats_interval
```

jage (approved):
Looks great!

Sorry for the delayed approval.

If you want, you can update the Changelog in this PR as well, just add a row under `# Unreleased` with a summary of what's been added.

(Just a minor improvement re: spec but otherwise good to go).

carlhoerberg (approved):

carlhoerberg:
maybe the naming could be better. stats_loop => stats_collection ?

carlhoerberg:
i would probably also take the time to extract  system and rates collections into two seperate methods

carlhoerberg:
rebased and squashed all commits into one, now without conflicts.

src/lavinmq/server.cr
```diff
@@ -254,53 +254,59 @@ module LavinMQ
...
-        until @disk_free_log.size < log_size
-          @disk_free_log.shift
-        end
-        disk_free = fs_stats.available.to_i64
-        @disk_free_log.push disk_free
-        @disk_free = disk_free
-
-        until @disk_total_log.size < log_size
-          @disk_total_log.shift
+          @system_metrics_duration_seconds = Time.measure do
```
carlhoerberg, line 262: extract everything in this block into its own method?

Weistrand: That's what I did, but [this](https://github.com/cloudamqp/lavinmq/compare/6a4e36913473340becf10debe2f4434cb138e32e..1f6b4721e248b4d9310ece34c6ea23429c72ee37) seems to have changed back some things.

PR #372, opened 2022-08-11
VHost limits

carlhoerberg:
https://trello.com/c/HfiVcZPH/100-add-support-for-vhost-limits

### DIFF
--- a/src/lavinmq/http/controller/vhost-limits.cr
+++ b/src/lavinmq/http/controller/vhost-limits.cr
@@ -6,14 +6,15 @@ module LavinMQ
     class VHostLimitsController < Controller
       private def register_routes
         get "/api/vhost-limits" do |context, _params|
-          vhosts = vhosts(user(context)).map { |v| VHostLimitsView.new(v) }
+          vhosts = vhosts(user(context)).compact_map { |v| VHostLimitsView.new(v).self_if_limited }
           page(context, vhosts)
         end
 
         get "/api/vhost-limits/:vhost" do |context, params|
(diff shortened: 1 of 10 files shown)


carlhoerberg:
Screenshot 2022-08-12 at 11-50-13 _ LavinMQ

src/lavinmq/http/controller/vhost-limits.cr
```diff
@@ -0,0 +1,89 @@
...
+        put "/api/vhost-limits/:vhost/:type" do |context, params|
+          u = user(context)
+          refuse_unless_administrator(context, u)
+          context.response.status_code = 400
+          with_vhost(context, params) do |vhost|
+            v = @amqp_server.vhosts[vhost]
+            if body = context.request.body
+              json = JSON.parse(body)
+              if value = json["value"]?.try &.as_i?
+                value = nil if value < 0
```
jage, line 30: Should we reject the request if it's "invalid", now we silently remove the limit.

carlhoerberg: Only 204 on valid values, 400 for all other requests

jage: Yeah, I was thinking we make `-1` invalid. Now it's valid, but setting nil.

carlhoerberg: It's correct behavior, it's the way to delete the limit

carlhoerberg: Same as for rabbitmq

PR #376, opened 2022-08-24
Support nack when queue is closed

Weistrand:
Nack published messages when queue is closed.

### DIFF
--- a/spec/server_spec.cr
+++ b/spec/server_spec.cr
@@ -838,7 +838,7 @@ describe LavinMQ::Server do
     end
   end
 
-  it "supports sends nack if queue is closed" do
+  it "sends nack if queue is closed" do
     with_channel do |ch|
       q = ch.queue("to-be-closed")
       s.vhosts["/"].queues["to-be-closed"].close
@@ -846,7 +846,7 @@ describe LavinMQ::Server do
     end
   end
 
-  it "supports sends nack if queue is closed, direct exchange" do
+  it "sends nack if queue is closed, direct exchange" do
     with_channel do |ch|
       ch.exchange("amq.direct", "direct")
       q1 = ch.queue("not-closed")
@@ -858,7 +858,7 @@ describe LavinMQ::Server do
     end
   end
 
-  it "supports sends nack if queue is closed, topic exchange" do
+  it "sends nack if queue is closed, topic exchange" do
     with_channel do |ch|
       ch.exchange("amq.topic", "topic")
       q1 = ch.queue("not-closed")
@@ -870,7 +870,7 @@ describe LavinMQ::Server do
     end
   end
 
-  it "supports sends nack if queue is closed, fanout exchange" do
+  it "sends nack if queue is closed, fanout exchange" do
     with_channel do |ch|
       ch.exchange("amq.fanout", "fanout")
       q1 = ch.queue("not-closed")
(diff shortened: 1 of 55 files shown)


Weistrand:
I'm a bit unsure of the dead-letter exchange spec, please have an extra look on that one to see if it has the intended behaviour.

spec/server_spec.cr
```diff
@@ -839,6 +839,75 @@ describe LavinMQ::Server do
...
+      q1 = ch.queue("not-closed", args: args)
+      q2 = ch.queue("to-be-closed", args: args)
+      ch.queue_bind(q1.name, "amq.headers", "rk")
+      ch.queue_bind(q2.name, "amq.headers", "rk")
+      s.vhosts["/"].queues["to-be-closed"].close
+      ch.basic_publish_confirm("msg", "amq.headers").should be_false
+    end
+  end
+
+  it "supports sends nack if queue is closed, dead-letter exchange" do
```
jage, line 900: Yeah this will just exercise the direct exchange code path, as it nacks before dead-lettering would have effect.
I'm leaning towards we can just skip the DLQ/DLX part of this for now. So you can just remove this spec and we can merge this.

spec/server_spec.cr
```diff
@@ -839,6 +839,75 @@ describe LavinMQ::Server do
     end
   end
 
+  it "supports sends nack if queue is closed" do
```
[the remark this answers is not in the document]
Weistrand: [8e3c472](https://github.com/cloudamqp/lavinmq/pull/376/commits/8e3c472f2dccf8f35c972e764d5d1d9686157747)

spec/server_spec.cr
Weistrand: [11dd6a3](https://github.com/cloudamqp/lavinmq/pull/376/commits/11dd6a321c23cce9b7cae93a2707a538683417d3)

carlhoerberg:
so if any matching queue is closed we raise and abort the publish, is that the intended behavior? is that how rabbitmq works too?

carlhoerberg:
when are queues closed? just before getting deleted? when the server is shutting down? when a queue is "paused"?

Weistrand:
@jage probably has better answers than me

jage:
>  so if any matching queue is closed we raise and abort the publish, is that the intended behavior? is that how rabbitmq works too?

Yes, nack is the intended behaviour, as per RabbitMQ. If we can't guarantee that we will deliver the message to all destinations (queues) I interpret the docs as we should nack. Though the type of failure is probably a bit different (Erlang process crash + observer restart etc).

> By nack'ing one or more messages, the broker indicates that it was unable to process the messages and refuses responsibility for them

[RabbitMQ nack](https://www.rabbitmq.com/confirms.html#server-sent-nacks).

Since we don't take responsibility anymore, it feels best to abort as early as possible and don't try to save anything.

> when are queues closed? just before getting deleted? when the server is shutting down? when a queue is "paused"?

Yes, delete and shut down, but don't think we accept any new messages then so that's not our focus here.
This was something we started thinking about when we started handling index corruption https://github.com/cloudamqp/lavinmq/pull/318#issuecomment-1064961560

When we discover corruption, we will most likely crash altogether but in the future that might not be true, so I think this makes sense to ensure we keep compatibility.

jage:
IMHO this looks good, but would be good to see if we can provoke this failure in a real running server to see how it behaves.

carlhoerberg (approved):

jage (approved):

PR #378, opened 2022-09-01
added last activity timestamp for vhosts

Christina-84:
Related to this card: https://trello.com/c/r7pGcwWb 
Logs a last activity timestamp for each Vhost (If a vhost has connections, it is active)

### DIFF
--- a/src/lavinmq/server.cr
+++ b/src/lavinmq/server.cr
@@ -243,6 +243,7 @@ module LavinMQ
           connection.channels.each_value(&.update_rates)
         end
         vhost.update_rates
+        vhost.update_last_activity
       end
     end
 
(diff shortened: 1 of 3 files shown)


kickster97:
Needs to store the timestamp persisently, now it only lives when LavinMQ runs.

dentarg:
This will be a feature that RabbitMQ doesn't have, right? Are we keeping a list of differences somewhere? I think it would be good to do.

Also, don't forget to update the API schema with this property. You find it at https://github.com/cloudamqp/lavinmq/blob/f43c8632b2807442a55b95cb826f6dfae25b3a98/openapi/schemas/vhosts.yaml#L6-L11

PR #379, opened 2022-09-08
Allow shared user to download its vhost definition.

tbroden84:
Allow user with policymaker permission be able to download vhost definition. As long as the user also have access to that vhost.

### DIFF
--- a/spec/api/definitions_spec.cr
+++ b/spec/api/definitions_spec.cr
@@ -424,23 +424,6 @@ describe LavinMQ::HTTP::Server do
         s.users.create("guest", "guest", [LavinMQ::Tag::Administrator])
         s.vhosts.each_key { |name| s.users.add_permission("guest", name, /.*/, /.*/, /.*/) }
       end
-
-      it "should refuse user tag access" do
-        s.users.delete("guest")
-        s.users.create("other_name", "guest", [LavinMQ::Tag::Http], save: false) # Will be the new default_user
-        s.vhosts.create("new")
-        s.users.add_permission("other_name", "new", /.*/, /.*/, /.*/)
-        headers = HTTP::Headers{"Authorization" => "Basic b3RoZXJfbmFtZTpndWVzdA=="}
-        response = get("/api/definitions/new", headers: headers)
-        response.status_code.should eq 401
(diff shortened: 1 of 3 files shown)


spec/api/definitions_spec.cr
```diff
@@ -392,6 +392,58 @@ describe LavinMQ::HTTP::Server do
...
+      end
+
+      it "should refuse user tag access" do
+        s.users.delete("guest")
+        # Http tag override refuse_unless_management
+        s.users.create("other_name", "guest", [LavinMQ::Tag::Http], save: false) # Will be the new default_user
+        s.vhosts.create("new")
+        s.users.add_permission("other_name", "new", /.*/, /.*/, /.*/)
+        headers = HTTP::Headers{"Authorization" => "Basic b3RoZXJfbmFtZTpndWVzdA=="}
+        response = get("/api/definitions/new", headers: headers)
```
tbroden84, line 435: New user without tags, have no access to API at all, hence issue with login refused.

Setting tag to Http (to be able to make the API call), will override management and allow exporting vhost definitions. https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/http/controller.cr#L194

spec/api/definitions_spec.cr
```diff
@@ -392,6 +392,57 @@ describe LavinMQ::HTTP::Server do
...
+        s.users.delete("other_name", save: false)
+        s.vhosts.delete("new")
+        s.users.create("guest", "guest", [LavinMQ::Tag::Administrator])
+        s.vhosts.each_key { |name| s.users.add_permission("guest", name, /.*/, /.*/, /.*/) }
+      end
+
+      it "should refuse user tag access" do
+        s.users.delete("guest")
+        # Http tag override refuse_unless_management
+        s.users.create("other_name", "guest", [LavinMQ::Tag::Http], save: false) # Will be the new default_user
```
carlhoerberg, line 431: What's the http tag? There's no such tag in rmq

jage: @snichme do you remember the reasoning here?
Introduced in 0b238da7d6cd5d86d3feeac53fab42ccc4e93080

PR #384, opened 2022-10-19
Index corruption bug

carlhoerberg:
For index files we preallocate files (for use with mmap), but if the
process crashes the file isn't truncated to it's size, so instead we've
used lseek to SEEK_HOLE to find the end, but that rounds up to the
nearest block (4096 bytes), and then we continued writing to writing to
the index file, but that would case a small space with null bytes. When
we next time restored the index we stopped reading when we encountered a
segment position point to segment 0 and position 0. That's causing to
big problems.

a) messages enqueued after a crash would not be restored at the next
starts of the server, not until the queue index was compacted again and
the null bytes disappeared.
b) messages could have been ack:ed, but after the "hole", and thus
messages could reappear after a restart if the SP in the enq was before
a hole and the SP in the enq file after the "hole".


src/lavinmq/queue/durable_queue.cr
```diff
@@ -211,46 +229,47 @@ module LavinMQ
               @ready = ready = ReadyQueue.new Math.pw2ceil(capacity)
             end
             next if acked.try { |a| a.bsearch { |asp| asp >= sp } == sp }
+            # SP refers to a segment that has already been deleted
+            unless @vhost.@segments.has_key? sp.segment
+              lost_msgs += 1
```
carlhoerberg, line 234: are they lost? or are they messages that has already been acked? or both? potentially lost i guess..

PR #387, opened 2022-10-21
WIP: spec to test index corruption at startup after crash

kickster97:
Begining of spec to recreate and test the index corruption bug. 

https://84codes.slack.com/archives/CKPHUKCF6/p1666302566839169
"the message loss happens at the second start after a crash, not at shutdown
a crash needs to happen so that the index files was preallocated, then we used lseek SEEK_HOLE to find the end, but that wasn't the true end, it was rounded up 4096 bytes, filling the rest will null bytes, and then we started to append after that "hole". that meant that at the second start, when restoring the index, we stopped reading when we encountured the first null bytes, and didn't read the ack/enqs that been written during the session after the crash. so those messages were lost, or at least as soon as a GC of segments ran (every 60s)."

### DIFF
--- a/spec/storage_spec.cr
+++ b/spec/storage_spec.cr
@@ -104,6 +104,38 @@ describe LavinMQ::DurableQueue do
     queue.@ready.capacity.should eq Math.pw2ceil(msg_count)
     queue.@ready.size.should eq msg_count
   end
+
+  # TODO: we know this works as expected although logs dont show
+  # next step is to shut down hard in order to try and trigger faulty index recovery
+  it "it should not insert zeros to enq/ack files" do
+    with_channel do |ch|
+      q = ch.queue("corruption_test", durable: true)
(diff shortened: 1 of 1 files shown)


spec/storage_spec.cr
```diff
@@ -104,6 +104,38 @@ describe LavinMQ::DurableQueue do
...
+    with_channel do |ch|
+      q = ch.queue("corruption_test", durable: true)
+      q.publish_confirm "Hello world"
+      # enq_path = queue.@enq.path
+    end
+    queue = s.vhosts["/"].queues["corruption_test"].as(LavinMQ::DurableQueue)
+    queue.@ready.size.should eq 1
+    # close_servers
+    # emulate the file was preallocated after server crash
+    # File.open(enq_path, "r+") { |f| f.truncate(f.size + 24 * 1024**2) }
```
carlhoerberg, line 120: Correct

PR #389, opened 2022-10-25
Sort consumers in queue details

### LINKED ISSUE #386: Lavinmq UI issue disable consumers sorting ?
hi,

last night i cancelled a different consumer  and realized only after few hours as the auto sorting of consumers list is really annoying , as i hover my mouse over cancel , the list must have sorted and i ended up clicking wrong consumer !!!

the issue is "the consumers table in UI must be sorted by a consumer tag ( alphabetically ) "  or any other stable identifier so it doesn't  sort at all or sort  in a predictable manner.

thanks

carlhoerberg:
Fixes #386, but with a cost.

### DIFF
--- a/src/lavinmq/queue/queue.cr
+++ b/src/lavinmq/queue/queue.cr
@@ -958,7 +958,7 @@ module LavinMQ
         end
         builder.field("consumer_details") do
           builder.array do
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Superseded by  #391

PR #391, opened 2022-10-26
Let consumers pull messages from Queues

carlhoerberg:
Let each consumer take care of each delivery, solving the slow/fast consumer problem.


carlhoerberg:
Known issue: the head message in the queue won't be expired until a consumer tries to consume it if the queue got consumers.

carlhoerberg:
And yeah, consumer priorities are not supported either.

carlhoerberg:
> And yeah, consumer priorities are not supported either.

Scratch that, fixed now!

PR #394, opened 2022-11-01
Selects the selected vhost as default

kickster97:
When creating a queue, the default selected vhost is now the overall chosen vhost (right top corner of page), instead of `/`, which some user may not have access to.

### DIFF
--- a/static/js/vhosts.js
+++ b/static/js/vhosts.js
@@ -37,10 +37,8 @@ function addVhostOptions (formId) {
       const opt = document.createElement('option')
       opt.text = vhosts[i].name
       opt.value = vhosts[i].name
+      opt.selected = vhost[i].name === selectedVhost
       select.add(opt)
-    }
-    if (vhosts.map(v => v.name).includes(selectedVhost)) {
-      select.value = selectedVhost
     }
   })
 }
--- a/static/js/vhosts.js
+++ b/static/js/vhosts.js
@@ -32,11 +32,15 @@ function addVhostOptions (formId) {
       return
     }
     // let the selected vhost be put first in the list and '/' last
+    const selectedVhost = window.sessionStorage.getItem('vhost')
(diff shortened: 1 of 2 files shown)


carlhoerberg (changes requested):

static/queues.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
```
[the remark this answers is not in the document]
kickster97: I included it bc my linter corrected the file, but maybe i could do a focused lint job instead some time and comb through everything.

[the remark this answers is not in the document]
carlhoerberg: Correct

static/js/vhosts.js
```diff
@@ -27,15 +27,17 @@ function addVhostOptions (formId) {
...
-      err.textContent = "Error fetching data: Please try to refresh the page!"
+      err.textContent = 'Error fetching data: Please try to refresh the page!'
       select.parentElement.insertAdjacentElement('beforebegin',err)
       return
     }
 
+    const selectedVhost = window.sessionStorage.getItem('vhost')
     for (let i = 0; i < vhosts.length; i++) {
       const opt = document.createElement('option')
       opt.text = vhosts[i].name
```
carlhoerberg, line 38:
```suggestion
```

kickster97: without the text setting, the content of the option tag will be empty

kickster97: I could use the label attribute instead if that is preferred? or is there another way?

carlhoerberg (commented):
Revert the changes in queue.html, and then you can squash merge.

static/js/vhosts.js
carlhoerberg: Ok, docs says that value would be used if no label was set.

carlhoerberg (approved):

PR #395, opened 2022-11-02
Clear the response buffer before writing error

baelter:
So errors while writing paginated responses don't get wrapped like 
```
{"items":[{"error":"internal_server_error","reason":"Internal Server Error"}
```


carlhoerberg:
Flush doesn't clear but writes what's in the output buffer to the socket, no?

PR #396, opened 2022-11-02
Create valid error response

kickster97:
Creates valid error responses even if there is an exception in `array_iterator_to_json` 
The fabricated responses are handled in the Table javascript class

### DIFF
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -43,7 +43,12 @@ module LavinMQ
           {error: "payload_too_large", reason: "Max allowed page_size 10000"}.to_json(context.response)
           return context
         end
-        all_items = filter_values(params, iterator.map(&.details_tuple))
+        iterator = iterator.map do |i|
+          i.details_tuple
+        rescue e
+          {error: e.message}
+        end
+        all_items = filter_values(params, iterator)
         if sort_by = params.fetch("sort", nil).try &.split(".")
           sorted_items = all_items.to_a
           filtered_count = sorted_items.size
@@ -94,7 +99,7 @@ module LavinMQ
           nt = i[keys.first].as?(NamedTuple) || return
           dig(nt, keys[1..])
         else
-          i[keys.first]
+          i[keys.first]? || 0
         end
       end
 
--- a/static/js/table.js
+++ b/static/js/table.js
@@ -14,7 +14,7 @@ function getQueryVariable (variable) {
 function renderTable (id, options = {}, renderRow) {
   let sortKey = getQueryVariable('sort')
   let reverseOrder = strToBool(getQueryVariable('reverseOrder'))
-  const view = window.location.pathname.split("/").pop()
+  const view = window.location.pathname.split('/').pop()
   if (!sortKey || reverseOrder === null) {
     sortKey = window.sessionStorage.getItem(view + '-sortkey')
     reverseOrder = strToBool(window.sessionStorage.getItem(view + '-reverseorder'))
@@ -116,7 +116,7 @@ function renderTable (id, options = {}, renderRow) {
     if (searchTerm) {
       q += `&name=${searchTerm}&use_regex=true`
     }
-    if (sortKey > "") {
+    if (sortKey > '') {
       q += `&sort=${sortKey}&sort_reverse=${reverseOrder}`
     }
     return q
--- a/static/js/table.js
+++ b/static/js/table.js
@@ -128,9 +128,8 @@ function renderTable (id, options = {}, renderRow) {
 
   function fetchAndUpdate () {
     const fullUrl = `${url}?${buildQuery(currentPage)}`
-    const tableError = document.getElementById(id + '-error')
     return HTTP.request('GET', fullUrl).then(function (response) {
-      tableError.textContent = ''
+      toggleDisplayError(false)
       try {
         window.sessionStorage.setItem(fullUrl, JSON.stringify(response))
       } catch (e) {
@@ -139,9 +138,9 @@ function renderTable (id, options = {}, renderRow) {
       updateTable(response)
     }).catch(function (e) {
       if (e.body) {
-        tableError.textContent = 'Error fetching data: ' + e.body
+        toggleDisplayError('Error fetching data: ' + e.body)
       } else {
-        tableError.textContent = "Error fetching data: Can't reach server, please try to refresh the page!"
+        toggleDisplayError("Error fetching data: Can't reach server, please try to refresh the page.")
         console.error(e)
       }
       if (timer) {
@@ -167,26 +166,32 @@ function renderTable (id, options = {}, renderRow) {
     }
 
     let start = 0
+    toggleDisplayError(false)
     for (let i = 0; i < data.length; i++) {
       const item = data[i]
-      const foundIndex = findIndex(t.rows, start, item)
-      if (foundIndex !== -1) {
-        if (foundIndex !== i) {
-          renderRow(t.rows[i], item, true)
-          setKeyAttributes(t.rows[i], item)
-          if (data[foundIndex]) {
-            renderRow(t.rows[foundIndex], data[foundIndex], true)
-            setKeyAttributes(t.rows[foundIndex], data[foundIndex])
+      try {
+        console.log(item)
+        const foundIndex = findIndex(t.rows, start, item)
+        if (foundIndex !== -1) {
+          if (foundIndex !== i) {
+            renderRow(t.rows[i], item, true)
+            setKeyAttributes(t.rows[i], item)
+            if (data[foundIndex]) {
+              renderRow(t.rows[foundIndex], data[foundIndex], true)
+              setKeyAttributes(t.rows[foundIndex], data[foundIndex])
+            }
+          } else {
+            renderRow(t.children[i], item, false)
           }
+          start = Math.min(i + 1, foundIndex)
         } else {
-          renderRow(t.children[i], item, false)
+          const tr = t.insertRow(i)
+          setKeyAttributes(tr, item)
+          renderRow(tr, item, true)
+          start = i + 1
         }
-        start = Math.min(i + 1, foundIndex)
-      } else {
-        const tr = t.insertRow(i)
-        setKeyAttributes(tr, item)
-        renderRow(tr, item, true)
-        start = i + 1
+      } catch (e) {
+        toggleDisplayError(item.error || e.message)
       }
     }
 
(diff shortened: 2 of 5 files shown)


src/lavinmq/http/controller.cr
```diff
@@ -102,11 +102,15 @@ module LavinMQ
         size = 0
         total = 0
         json.array do
-          iterator.each_with_index do |o, i|
-            total += 1
-            next if i < start || start + page_size <= i
-            o.to_json(json)
-            size += 1
+          begin
```
baelter, line 105: Weird that `ameba` doesn't find this but you don't need a begin block, you can rescue in the outer block.

src/lavinmq/http/controller.cr
```diff
@@ -102,11 +102,15 @@ module LavinMQ
...
+            iterator.each_with_index do |o, i|
+              total += 1
+              next if i < start || start + page_size <= i
+              o.to_json(json)
+              size += 1
+            end
+          rescue e
+            json.object { json.field "error", e.message }
```
baelter, line 113: Should we use the `{ error: type, reason: message }` format we use in the `ApiErrorHandler` ?
And set a different status code? 500?

kickster97: Yes, great!

carlhoerberg: i'm not sure about the response code actually... if the result set is very large, and say 900 items have already been written to the response, then a) is it a valid or invalid response? b) the response header with the status code might already have been written to the socket and can't be changed

src/lavinmq/http/controller.cr
```diff
@@ -102,11 +102,15 @@ module LavinMQ
...
-          iterator.each_with_index do |o, i|
-            total += 1
-            next if i < start || start + page_size <= i
-            o.to_json(json)
-            size += 1
+          begin
+            iterator.each_with_index do |o, i|
+              total += 1
+              next if i < start || start + page_size <= i
+              o.to_json(json)
```
carlhoerberg, line 109: only wrap this line in the `begin..rescue` block.

carlhoerberg (changes requested):

src/lavinmq/http/controller.cr
baelter: True, like we fail to generate a response for one of 900 queues, this happens in rabbitmq as well, they just send some null fields back.

static/js/table.js
```diff
@@ -170,7 +170,7 @@ function renderTable (id, options = {}, renderRow) {
     for (let i = 0; i < data.length; i++) {
       const item = data[i]
       const foundIndex = findIndex(t.rows, start, item)
-      if (foundIndex !== -1) {
+      if (foundIndex !== -1 && item.error === undefined) {
```
carlhoerberg, line 173: you want to check for the error property earlier, not when we've already tried to "find" it.

carlhoerberg (changes requested):

src/lavinmq/http/controller.cr
kickster97: So maybe we should just filter out the failure from items, and still render the queues with valid responses? (with a warning that we failed to load some data) ?

src/lavinmq/http/controller.cr
```diff
@@ -108,6 +108,8 @@ module LavinMQ
             o.to_json(json)
             size += 1
           end
+        rescue e
+          json.object { json.field "error", e.message }
```
carlhoerberg, line 112:
```suggestion
            begin
              o.to_json(json)
            rescue e
              json.object { json.field "error", e.message }
            end
            size += 1
          end
```

static/js/table.js
```diff
@@ -178,7 +180,6 @@ function renderTable (id, options = {}, renderRow) {
             renderRow(t.rows[foundIndex], data[foundIndex], true)
             setKeyAttributes(t.rows[foundIndex], data[foundIndex])
           }
-        } else {
```
carlhoerberg, line 181: Are you sure about this? I don't think this line should be remove.

static/js/table.js
```diff
@@ -131,6 +131,8 @@ function renderTable (id, options = {}, renderRow) {
     const tableError = document.getElementById(id + '-error')
     return HTTP.request('GET', fullUrl).then(function (response) {
       tableError.textContent = ''
+      const hasError = response.items.some(i => i.error !== undefined)
+      if (hasError) { throw "item error: invalid item type" }
```
carlhoerberg, line 135: Can't we write something more specific? Like taking the first error message?

carlhoerberg (changes requested):

kickster97:
Rescue when mapping `details_tuple` to the `iterator` because we can have errors here which we will not catch if we wait untill `array_iterator_to_json` to rescue. 

Made error display a little nicer, a complete renovation on msg/error displaying for everything could be done as part of UI-project.

static/main.css
```diff
@@ -381,7 +381,12 @@ footer .logo {
 }
 
 #table-error {
-  color: red;
+  display: none;
```
baelter, line 384: Other use of `#table-error` will be hidden now. We always have to change `display` when adding content to this element now. Maybe write a function in table that does it.

https://github.com/cloudamqp/lavinmq/blob/2bae422dc0ddcebf728fd7cde67f6c797e01a0d3/static/vhosts.html#L74
https://github.com/cloudamqp/lavinmq/blob/83efd006dab9cf727ce3cd44776c9216ce139d43/static/users.html#L89

kickster97: ah yes missed those, thanks!

static/js/table.js
```diff
@@ -437,4 +443,14 @@ function updateQueryState (params) {
   window.history.replaceState(null, '', newurl)
 }
 
-export { renderCell, renderTable, renderHtmlCell }
+function toggleDisplayError (tableError, display, message) {
+  if (display) {
+    tableError.style.display = 'block'
+    tableError.textContent = 'Error fetching data: {' + message + '}'
```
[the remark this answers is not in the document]
kickster97: Ah right, maybe put "Something went wrong: " instead? 
Agree, will fix

static/js/table.js
```diff
@@ -139,9 +138,9 @@ function renderTable (id, options = {}, renderRow) {
       updateTable(response)
     }).catch(function (e) {
       if (e.body) {
-        tableError.textContent = 'Error fetching data: ' + e.body
+        toggleDisplayError(id, true, e.body)
       } else {
-        tableError.textContent = "Error fetching data: Can't reach server, please try to refresh the page!"
+        toggleDisplayError(id, true, "{ error: Can't reach server, please try to refresh the page }")
```
baelter, line 143:
```suggestion
        toggleDisplayError("Error fetching data: Can't reach server, please try to refresh the page.")
```

static/js/table.js
```diff
@@ -139,9 +138,9 @@ function renderTable (id, options = {}, renderRow) {
       updateTable(response)
     }).catch(function (e) {
       if (e.body) {
-        tableError.textContent = 'Error fetching data: ' + e.body
+        toggleDisplayError(id, true, e.body)
```
baelter, line 141:
```suggestion
        toggleDisplayError('Error fetching data: ' + e.body)
```

static/js/table.js
```diff
@@ -437,4 +441,15 @@ function updateQueryState (params) {
   window.history.replaceState(null, '', newurl)
 }
 
-export { renderCell, renderTable, renderHtmlCell }
+function toggleDisplayError (id, display, message) {
+  const tableError = document.getElementById(id + '-error')
```
baelter, line 445:
```suggestion
  const tableError = document.getElementById('table-error')
```

static/users.html
```diff
@@ -86,7 +88,7 @@ <h3>Add user</h3>
           Table.renderCell(tr, 3, hasPassword, 'center')
         })
       }).catch(e => {
-        document.getElementById("table-error").textContent = e.status === 401 ? "Access refused: You need administrator role to see this view" : e.body
+        Table.toggleDisplayError(document.getElementById('table-error'), true, e.status === 401 ? 'Access refused: You need administrator role to see this view' : e.body)
```
baelter, line 91:
```suggestion
        Table.toggleDisplayError(e.status === 401 ? 'Access refused: You need administrator role to see this view' : e.body)
```

static/vhosts.html
```diff
@@ -63,16 +65,15 @@ <h3>Add virtual host</h3>
...
         const urlEncodedVhost = encodeURIComponent(item.name)
         const permissionsUrl = '/api/vhosts/' + urlEncodedVhost + '/permissions'
         const permissionsPromise = new Promise(function (resolve, reject) {
           HTTP.request('GET', permissionsUrl).then(permissions => {
             window.sessionStorage.setItem(permissionsUrl, JSON.stringify(permissions))
             resolve(permissions)
           }).catch(e => {
-              document.getElementById("table-error").textContent = e.status === 401 ? e.body + ": You need administrator role to see this view" : e.body
-              reject(e)
+            Table.toggleDisplayError(document.getElementById('table-error'), true, e.status === 401 ? e.body + ': You need administrator role to see this view' : e.body)
```
baelter, line 75:
```suggestion
            Table.toggleDisplayError(e.status === 401 ? e.body + ': You need administrator role to see this view' : e.body)
```

static/js/table.js
```diff
@@ -128,9 +128,8 @@ function renderTable (id, options = {}, renderRow) {
 
   function fetchAndUpdate () {
     const fullUrl = `${url}?${buildQuery(currentPage)}`
-    const tableError = document.getElementById(id + '-error')
     return HTTP.request('GET', fullUrl).then(function (response) {
-      tableError.textContent = ''
+      toggleDisplayError(id, false, '')
```
baelter, line 132:
```suggestion
      toggleDisplayError(false)
```

static/js/table.js
```diff
@@ -167,26 +166,31 @@ function renderTable (id, options = {}, renderRow) {
...
+          start = i + 1
         }
-        start = Math.min(i + 1, foundIndex)
-      } else {
-        const tr = t.insertRow(i)
-        setKeyAttributes(tr, item)
-        renderRow(tr, item, true)
-        start = i + 1
+      } catch (e) {
+        toggleDisplayError(id, true, e)
```
baelter, line 193:
```suggestion
        toggleDisplayError(item.error || e.message)
```

baelter (commented):
Some refactoring suggestions around the toggleError function

PR #399, opened 2022-11-03
Spec: Keep same Logger for server restarts

jage:
When we run `Log.setup` the second time we'd loose logging in specs.

I'm not sure exactly why logging stops working, but needed a fix to debug some other specs and this worked ...

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -121,7 +121,7 @@ module TestHelpers
   end
 
   def self.create_servers(dir = "/tmp/spec", level = LOG_LEVEL)
-    Log.setup(level)
+    Log.setup(level) unless @@s
     cfg = LavinMQ::Config.instance
     cfg.gc_segments_interval = 1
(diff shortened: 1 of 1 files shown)


tbroden84:
Printed out log level before Log.setup: https://github.com/cloudamqp/lavinmq/blob/main/spec/spec_helper.cr#L124
Full spec run and something change the LOG_LEVEL to error multiple times. 

Without this check of previous created servers, the new Log.setup will override the previous setup (DEBUG or chosen log level). https://crystal-lang.org/api/1.6.1/Log.html#default-logging-configuration

tbroden84:
Suspect the cause coming from all specs containing TestHelpers.setup (e.g. https://github.com/cloudamqp/lavinmq/blob/main/spec/vhost_spec.cr#L33)

And flag check revert LOG_LEVEL back to default=error
https://github.com/cloudamqp/lavinmq/blob/main/spec/spec_helper.cr#L51

PR #400, opened 2022-11-03
Bug fix: user handling

tbroden84:
https://trello.com/c/EIyxewSY

Bug fixes when creating or updating users.

[Users view]
- Adds confirm window when trying to add already existing user (update user instead).
- Auto-complete empty tags when updating existing user.

[User view]
- Adds users tags to input field in update form in user view.

[Both views]
- Block adding multiple pre-defined tags (onclick event).
- Remove length check when setting `hasPassword`, response contains null when password_hash missing. 

[API]
- Filter out uniq tags when create/update user
- Specs

### DIFF
--- a/static/user.html
+++ b/static/user.html
@@ -205,6 +205,7 @@
         HTTP.request('PUT', url, { body })
           .then(() => {
             updateUser()
+            DOM.toast("User updated")
             evt.target.reset()
           }).catch(HTTP.standardErrorHandler)
       })
--- a/static/user.html
+++ b/static/user.html
@@ -124,7 +124,16 @@
             const hasPassword = item.password_hash.length ? '●' : '○'
             document.getElementById('tags').textContent = item.tags
             document.getElementById('hasPassword').textContent = hasPassword
+            tagHelper(item.tags)
           }).catch(HTTP.standardErrorHandler)
+      }
+
+      function tagHelper(tags) {
+        const vals = tags.split(',')
+        vals.forEach((val) =>  {
+          const currentVal = document.querySelector(`[name=tags]`).value
+          document.querySelector(`[name=tags]`).value = currentVal ? currentVal + ', ' + val : val
+        })
       }
       updateUser()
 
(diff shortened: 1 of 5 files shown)


tbroden84:
Not sure how to easy determine when and if to add tags from add user form. As for now, always use tags from the input fields when updating existing user. Left empty, no tags will be assign.

tbroden84:
Added the tags to the update user form in user view. Was previous always empty, regardless if the user already had assigned tags.

tbroden84:
Another annoying thing. It was previous possible to add multiple tags of same type, when using the pre-defined tags in both views.

static/user.html
```diff
@@ -121,11 +121,20 @@ <h3>Danger zone</h3>
...
+            const hasPassword = item.password_hash ? '●' : '○'
             document.getElementById('tags').textContent = item.tags
             document.getElementById('hasPassword').textContent = hasPassword
+            tagHelper(item.tags)
           }).catch(HTTP.standardErrorHandler)
       }
+
+      function tagHelper(tags) {
+        const vals = tags.split(',')
+        vals.forEach((val) =>  {
```
kickster97, line 133: should we have the same check here as in `argumentHelper` in `helpers.js`? 

```js
if (currentVal.includes(val)) { return }
```

kickster97 (commented):
looks good, im just wondering what the difference between `tagHelper` and `argumentHelper` is? It looks like they both recieve tags(?) and do the same logic on those tags, could it be generalised?

static/user.html
tbroden84: Difference is, that `argumentHelper` triggers on onclick event and only contains one tag (the one pressed in the UI). While `tagHelper` format all tags received from the API response, when displaying it in the input field.

tbroden84: The check is used to avoid adding more of the same tag. Perhaps this should be handled by the API too. As for now it's possible to assign more than one of the same tag to a user.

kickster97 (approved):
looks good to me 👍

PR #404, opened 2022-11-08
Include timestamp and TTL in SegmentPosition metadata

carlhoerberg:
Uses 6 more bytes but one less disk lookup is then required when a queue
has a queue TTL (as the timestamp then have to be read from disk). Also
improves the situation for TTL=0 messages which requires special
handling (in the event of or lack of consumers, and on requeue).


carlhoerberg:
Tested to upgrade schema version 2 to 3 locally, I guess we should have some spec for it too. 

java client test result:
`Tests run: 309, Failures: 16, Errors: 22, Skipped: 1`

Throughput performance is about the same, but publishing to a queue with a message ttl policy is about 100% faster.

PR #408, opened 2022-11-15
Update user access controller

tbroden84:
From detected issue regarding uploading definitions from the UI. 

Previous user tag access check also combined lesser access. Instead of just making sure the user have the correct access for any given tag.

E.g. check for administrative rights, also used the policymaker and monitoring check. While the tags it self was repeated in those checks. Enough to just control that the user have the correct tags assigned.

### DIFF
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -204,7 +204,8 @@ module LavinMQ
       end
 
       private def refuse_unless_management(context, user, vhost = nil)
-        unless user.tags.any?
+        unless user.tags.any? { |t| t.administrator? || t.monitoring? ||
+               t.policy_maker? || t.management? }
           @log.warn { "user=#{user.name} does not have management access on vhost=#{vhost}" }
           access_refused(context)
         end
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -196,6 +196,7 @@ module LavinMQ
       end
 
       private def refuse_unless_vhost_access(context, user, vhost)
+        return if user.tags.any? &.administrator?
         unless user.permissions.has_key?(vhost)
           @log.warn { "user=#{user.name} does not have permissions to access vhost=#{vhost}" }
           access_refused(context)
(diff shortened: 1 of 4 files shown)


src/lavinmq/http/controller.cr
```diff
@@ -191,38 +191,35 @@ module LavinMQ
       end
 
       private def refuse_unless_vhost_access(context, user, vhost)
+        return if user.tags.any? &.administrator?
```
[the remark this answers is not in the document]
tbroden84: Grant access to vhost if administrator (not necessary have vhost permission). Otherwise check if user have vhost permission set regardless if it has tags or not. 

Have used it in combination with the other tag checks. E.g. refuse_unless_managment.

src/lavinmq/http/controller.cr
```diff
@@ -191,38 +191,35 @@ module LavinMQ
...
         unless user.permissions.has_key?(vhost)
           @log.warn { "user=#{user.name} does not have permissions to access vhost=#{vhost}" }
           access_refused(context)
         end
       end
 
       private def refuse_unless_management(context, user, vhost = nil)
-        unless user.tags.any? { |t| t.management? || t.administrator? || t.policy_maker? || t.monitoring? }
+        unless user.tags.any?
```
carlhoerberg, line 202: is this correct? what about the `impersonator` tag? that shouldn't give management access?

carlhoerberg (changes requested):

src/lavinmq/http/controller.cr
tbroden84: Looked for the pre-defined tags we have https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/user.cr#L6-L10
initialize and create user are limited to those.

carlhoerberg: Yes, but any tag should be able to be added to a user, not just the predefined.

tbroden84: Use module and constants instead of enum? Then any tag can be set but still keep pre-defined. https://gist.github.com/tbroden84/e1c004dd23462fe76582eaf1792773e7

PR #409, opened 2022-11-15
Vhost and upload definition from overview view

tbroden84:
From detected issue regarding uploading definitions from the UI and monitoring tag was required. 

See https://github.com/cloudamqp/lavinmq/pull/408

Overview view and upload functionality tried to upload definitions for the whole cluster. Regardless of vhost chosen for current logged in user. 

- Add file vhost definition upload endpoint in API.
- No vhost detected in session storage, pick default for user and save to storage.
- Change vhost in upper right menu (`userMenuVhost`), update session storage with vhost.
- Export definition drop down, set to vhost in session storage.
- Import definition, set to vhost in session storage.
	- All, upload defintions to whole cluster (require administrator tag)
	- <vhost>, upload definition to vhost (require poliymaker/administrator tag)
	
Screenshot 2022-11-15 at 08 35 21


carlhoerberg:
I think the vhost selector should be next to the upload button, i think it can easliy be overlooked otherwise. Someone selected a vhost when viewing queues, than switches over to the overview page and upload defintions, thinkig that it is for the whole vhost and then is left wondering why it didn't happen.

tbroden84:
True, would be a lot better.

tbroden84:
Removed fetching vhost from session storage and added separate vhost selector for uploading definitions. Either upload defintions to cluster or only for selected vhost.

Screenshot 2022-11-22 at 09 40 12

carlhoerberg:
very nice!

kickster97 (approved):

PR #410, opened 2022-11-15
A start of spec when disk is full

weistrand:
Tests start of server and publish of message when disk gets full. A small RAM-disk is needed and therefore the spec is put in a different spec folder.

### DIFF
--- a/special_spec/disk_full_spec.cr
+++ b/special_spec/disk_full_spec.cr
@@ -1,0 +1,29 @@
+require "../spec/spec_helper"
+
+# To run these specs you need to create a small RAM-disk, see https://gist.github.com/htr3n/344f06ba2bb20b1056d7d5570fe7f596.
+
+describe "when disk is full" do
+  it "should raise error on start server" do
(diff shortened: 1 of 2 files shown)


special_spec/disk_full_spec.cr
```diff
@@ -0,0 +1,29 @@
...
+        q.publish message * 134_217_728
+      end
+    end
+  end
+
+  it "should raise error on close server" do
+    expect_raises(File::Error) do
+      close_servers
+    end
+  end
```
Weistrand, line 28: Wanted to check what happens when disk gets full when closing, but it is hard to know how much is needed to close.

carlhoerberg:
You can now run the specs with `docker build --target spec .`

Weistrand:
> You can now run the specs with `docker build --target spec .`

🥱
image

carlhoerberg:
Depot is probably the way to go then! I sent you an invite, runs the build
on fast servers in the cloud. Download their cli and just replace docker
with depot in your command.

PR #411, opened 2022-11-16
lavinmqperf: Uniq queue names, allow parallel runs

jage:


src/lavinmqperf.cr
```diff
@@ -350,21 +351,22 @@ class QueueChurn < Perf
   def run
     super
 
+    r = Random.new
```
carlhoerberg, line 354: or use [`Random::DEFAULT`](https://devdocs.io/crystal/random#DEFAULT)?

PR #415, opened 2022-11-17
WIP: Full disk docker compose

jage:
Run with `docker compose build` then `docker compose up`. Specs aren't 100% working but not sure if they where before either.

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -1,0 +1,5 @@
+FROM 84codes/crystal:latest-debian-11
+WORKDIR /src
+COPY shard.* /src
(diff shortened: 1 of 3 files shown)


carlhoerberg:
What about this?
```crystal
system "mount -t tmpfs -o size=10m lavinmq-spec-10m /tmp/spec-10m" || pending! "No root access to mount tmpfs"
begin
  # run specs
ensure
  system "umount /tmp/spec-10m"
end
```
Will work in containers, but not otherwise, and will then just mark it as pending, so can belong among the normal specs. 

ping @Weistrand

PR #424, opened 2022-11-23
build deb packages using debuild

carlhoerberg:
debuild automatically creates a ddeb package with debug symbols


carlhoerberg:
This means `lavinmq-debug` is gone, but instead all binaries are built with `--debug`, which now gives full stack traces. FYI @jage

PR #426, opened 2022-11-29
Upgrade chartjs version

kickster97:
Chart.js upgrade 
- must actively include [luxon.js](https://moment.github.io/luxon/#/) and its [chart.js adapter](https://github.com/chartjs/chartjs-adapter-luxon) , these were chosen as they are ESM compatible 

- Saw the opportunity to remove alot of code by removing the custom legend. Imo it was more messy than it is worth. This changes the position of the legend to the top of the chart, but its possible to move it to any side of the chart. The only thing i think we are missing out by removing the custom legend is the "live" value that was rendered beside each label. However, all values can still be found in the tooltip by hovering the mouse wherever on the chart lines.

### DIFF
--- a/static/channel.html
+++ b/static/channel.html
@@ -88,6 +88,8 @@
     </main>
     <footer>LavinMQ is open source and developed by <a href="https://www.84codes.com" target="_blank"><img class="logo" src="/img/logo-84codes.svg"></a></footer>
     <script src="/js/lib/chart.js"></script>
+    <script src="/js/lib/moment.js"></script>
+    <script src="/js/lib/chartjs-adapter-moment.js"></script>
     <script type="module">
       import * as Table from './js/table.js'
       import * as HTTP from './js/http.js'
--- a/static/channel.html
+++ b/static/channel.html
@@ -87,9 +87,6 @@
       </section>
     </main>
     <footer>LavinMQ is open source and developed by <a href="https://www.84codes.com" target="_blank"><img class="logo" src="/img/logo-84codes.svg"></a></footer>
-    <script src="/js/lib/chart.js"></script>
-    <script src="/js/lib/moment.js"></script>
-    <script src="/js/lib/chartjs-adapter-moment.js"></script>
     <script type="module">
       import * as Table from './js/table.js'
       import * as HTTP from './js/http.js'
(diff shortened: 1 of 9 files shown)


static/channel.html
```diff
@@ -88,6 +88,8 @@ <h3>
     </main>
     <footer>LavinMQ is open source and developed by <a href="https://www.84codes.com" target="_blank"><img class="logo" src="/img/logo-84codes.svg"></a></footer>
     <script src="/js/lib/chart.js"></script>
+    <script src="/js/lib/moment.js"></script>
```
carlhoerberg, line 91: use the ESM variants of these libraries and `import` them in our `chart.js`, that way you don't need to include them in each html file.

carlhoerberg, line 91: use the ESM variants of these libraries and `import` them in our `chart.js`, that way you don't need to include them in each html file.

static/index.html
```diff
@@ -72,6 +72,11 @@ <h3>Export definitions</h3>
     </main>
     <footer>LavinMQ is open source and developed by <a href="https://www.84codes.com" target="_blank"><img class="logo" src="/img/logo-84codes.svg"></a></footer>
     <script src="/js/lib/chart.js"></script>
+    <script src="/js/lib/moment.js"></script>
+    <script src="/js/lib/chartjs-adapter-moment.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chart.js@^3"></script>
+    <script src="https://cdn.jsdelivr.net/npm/moment@^2"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-moment@^1"></script>
```
carlhoerberg, line 79: serve all files locally, no CDN

static/js/chart.js
```diff
@@ -1,11 +1,12 @@
+/* global Chart */
 import * as helpers from './helpers.js'
```
carlhoerberg, line 2: import `moment.js` and its adapter here

carlhoerberg (changes requested):

static/channel.html
kickster97: me & anders tried to solve this but were unsuccessful... might need some help

```javascript
import './lib/moment.js'
import './lib/chartjs-adapter-moment.js'
```
generates error 

```
Uncaught TypeError: Cannot set properties of undefined (setting 'moment')
    at moment.js:12:144
    at moment.js:12:149
```

carlhoerberg (approved):

PR #429, opened 2022-12-05
Decrease queue mem usage

carlhoerberg:
Queues uses too much memory by default, this PR tries to mitigate that.


carlhoerberg:
With 100k queues this branch uses 5.7G RSS, while main branch uses 5.8G.

PR #435, opened 2022-12-08
Build lavinmq with multi threading

carlhoerberg:


spuun:
Old, and needs to be redone when ExecutionContext is available.

Issue #445, opened 2022-12-22
Error at second startup when using delayed message exchange (406 (PRECONDITION-FAILED))


brahima:
Hello,

I'm following the [LavinMQ documentation about delayed message](https://lavinmq.com/documentation/delayed-message) and this [cloudamqp tutorial for rabbitmq with nodejs and amqplib](https://www.cloudamqp.com/blog/rabbitmq_delayed_message_exchange_plugin_with_node_js.html) to have a delayed exchange in nodeJS (amqplib)

**Note** : I'm using the last version of amqplib (0.10.3) with nodejs 18.12.1 and lavinMQ 1.0.0-beta.6 (seal-01) on CloudAMQP.

I have an error with the following code : 

`channel.assertExchange("my-exchange", "x-delayed-message", {autoDelete: false, durable: true,  arguments: {'x-delayed-type':  "fanout"}})`

- The first time (when my channel does not exist yet), that works and my delayed channel is correctly created in lavinmq (see screenshot below) with a type 'fanout' here. I noticed that a special queue (amqp.delay.my-exchange) is also created and deleted with the exchange deletion. So far, so good.

<img width="706" alt="image" src="https://user-images.githubusercontent.com/1575495/209195482-5ec802f6-be98-4963-985f-a62d86e764f0.png">

- Any time i restart my app, i have the following error on the assertExchange call.

**Received Error: Operation failed: ExchangeDeclare; 406 (PRECONDITION-FAILED) with message "PRECONDITION_FAILED - Existing exchange 'my-exchange' declared with other arguments"**

The error seems quite logic to me, and i can't see anything else making the assertion failing. amqplib is expecting a exchange of type 'x-delayed-messaage' but once it is created, its type is fanout ... so the assert_exchange check fails when the exchange already exists.

Also, I think that the python sample provided in the documentation should have the same shortcoming : 
```
channel.exchange_declare(exchange='test-exchange',
                                 exchange_type='x-delayed-message',
                                 arguments={"x-delayed-type": "fanout"})
```
Indeed, the pika client documentation says about _channel.exchange_declare_ :
`This method creates an exchange if it does not already exist, and if the exchange exists, verifies that it is of the correct and expected class.` 

Moreover, from the cloudamqp rabbitmq tutorial i linked above, it seems to me (but not tested) that rabbitmq returns 'x-delayed-message' as the exchange type for the code tutorial to be working as expected. This screenshot from the tutorial page displays the different exchange's types.
[...]

PR #450, opened 2023-01-05
reformat type and arguments when redeclaring exchange via HTTP api

kickster97:
related to #445 
When you declare a delayed message exchange, it is created as an exchange of the specified type. When one then wants to redeclare a delayed message exchange, the match check done to see if it is already declared fails since the reformatting is done AFTER the match method. 
This PR adds specs for redeclaring a delayed message exchange, as well as changes the match() method to compare on the correct values(type and arguments).

### DIFF
--- a/src/lavinmq/exchange/exchange.cr
+++ b/src/lavinmq/exchange/exchange.cr
@@ -89,19 +89,17 @@ module LavinMQ
     end
 
     def match?(frame : AMQP::Frame)
-      type == frame.exchange_type &&
-        @durable == frame.durable &&
-        @auto_delete == frame.auto_delete &&
-        @internal == frame.internal &&
-        @arguments == frame.arguments.to_h
+      match?(frame.exchange_type, frame.durable, frame.auto_delete, frame.internal, frame.arguments)
     end
 
     def match?(type, durable, auto_delete, internal, arguments)
-      self.type == type &&
+      delayed = type == "x-delayed-message"
+      frame_args = arguments.to_h.dup.reject("x-delayed-type").merge({"x-delayed-exchange" => true})
+      self.type == (delayed ? arguments["x-delayed-type"] : type) &&
         @durable == durable &&
         @auto_delete == auto_delete &&
         @internal == internal &&
-        @arguments == arguments.to_h
+        @arguments == (delayed ? frame_args : arguments.to_h)
     end
 
     def in_use?
--- a/spec/api/exchanges_spec.cr
+++ b/spec/api/exchanges_spec.cr
@@ -116,6 +116,23 @@ describe LavinMQ::HTTP::ExchangesController do
       response.status_code.should eq 400
     end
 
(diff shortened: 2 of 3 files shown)


src/lavinmq/exchange/exchange.cr
```diff
@@ -97,6 +97,11 @@ module LavinMQ
     end
 
     def match?(type, durable, auto_delete, internal, arguments)
+      # API-calls that want to re-decalre exchanges needs to reformat type and arguments
+      if type == "x-delayed-message"
+        type = arguments["x-delayed-type"]
+        arguments = { "x-delayed-exchange" => true }
```
spuun, line 103: I think you want 
```suggestion
        type = arguments.delete("x-delayed-type")
        arguments["x-delayed-exchange"] = true
```
else you'll lose any other arguments.

spuun (commented):
You should add specs to verify that everything works, and to prevent regression.

kickster97:
reformatted again, avoiding to modify the frame/exchange parameters since that is done after the match in `vhost.cr`

src/lavinmq/exchange/exchange.cr
```diff
@@ -89,19 +89,23 @@ module LavinMQ
...
     def match?(frame : AMQP::Frame)
-      type == frame.exchange_type &&
+      delayed = frame.exchange_type == "x-delayed-message"
+      frame_args = frame.arguments.to_h.dup.reject("x-delayed-type").merge({"x-delayed-exchange" => true})
+      type == (delayed ? frame.arguments["x-delayed-type"] : frame.exchange_type) &&
         @durable == frame.durable &&
         @auto_delete == frame.auto_delete &&
         @internal == frame.internal &&
-        @arguments == frame.arguments.to_h
+        @arguments == (delayed ? frame_args : frame.arguments.to_h)
```
spuun, line 98: You can probably just forward frame values to the other match method to get rid of some code duplication:

```suggestion
match?(frame.type, frame.durable, frame.auto_delete, frame.internal, frame.arguments)
```

kickster97: nice!

kickster97:
@spuun wdyt, ready for merge? :)

src/lavinmq/exchange/exchange.cr
```diff
@@ -89,19 +89,17 @@ module LavinMQ
     end
 
     def match?(frame : AMQP::Frame)
-      type == frame.exchange_type &&
-        @durable == frame.durable &&
-        @auto_delete == frame.auto_delete &&
-        @internal == frame.internal &&
-        @arguments == frame.arguments.to_h
+      match?(frame.exchange_type, frame.durable, frame.auto_delete, frame.internal, frame.arguments.to_h)
```
spuun, line 92: `.to_h` seems redundant, it's done in the other `match?` method

spuun (approved):

PR #451, opened 2023-01-11
Dont restore SPs for GC:ed messages

spuun:
After hole punching has been performed on a segment it may contain partial/malformed messages, i.e. the hole starts in the middle of a message that has been delivered. This can cause errors in the index migration of the ack file.

If the ack file references the partial message it will fail to parse it, and the migration fails.

The msg file for this new spec contains a malformed message that is referenced by the ack file but not the enq file (i.e. it simulates an acked message that has been partially punched).


carlhoerberg:
As an alternative I guess we could compact the index first, and then migrate? Less code duplication?

carlhoerberg (commented):
As an alternative I guess we could compact the index first, and then migrate? Less code duplication?

carlhoerberg:
Does this take into account that messages that the `ack` file refers to might have been hole punched too?

carlhoerberg:
What do you think of #453 as an alternative solution?

PR #452, opened 2023-01-18
Live log view in manager UI with SSE

viktorerlingsson:
Adds a log view in the manager UI that updates live with SSE, and an /api/livelog endpoint. 
Adds `in_memory_backend` that sends new log entries to any connected clients via channels. 
The in_memory_backend also keeps the latest 128 messages in memory and sends them to newly connected clients.

### DIFF
--- a/src/lavinmq/http/controller/logs.cr
+++ b/src/lavinmq/http/controller/logs.cr
@@ -1,0 +1,55 @@
+require "../controller"
+require "../../config"
+require "../../in_memory_backend"
+
+module LavinMQ
+  module HTTP
+    class LogsController < Controller
+      LogBackend = ::Log::InMemoryBackend.instance
+
+      private def register_routes
+        get "/api/livelog" do |context, _params|
+          channel = LogBackend.add_channel
+          context.response.content_type = "text/event-stream"
+          context.response.upgrade do |io|
+            if last_event_id = context.request.headers["Last-Event-Id"]?
+              last_ts = last_event_id.to_i64
+              LogBackend.entries.each do |entry|
+                next if entry.timestamp.to_unix_ms <= last_ts 
+                print_entry(entry, io)
+              end
+            else
+              LogBackend.entries.each do |entry|
+                print_entry(entry, io)
+              end
+            end
+            io.flush
+            while entry = channel.receive
+              print_entry(entry, io)
+              io.flush
+            end
+          ensure
+            LogBackend.remove_channel(channel)
+          end
+          context
+        end
+
+        get "/api/logs" do |context, _params|
+          context.response.content_type = "text/plain"
+          context.response.headers["Content-Disposition"] = "attachment; filename=logs.txt"
+          LogBackend.entries.each do |e|
+            context.response.puts "#{e.timestamp} [#{e.severity.to_s.upcase}] #{e.source} - #{e.message}"
+          end
+          context
+        end
+      end
+
+      private def print_entry(entry, io)
+        io.print("id: #{entry.timestamp.to_unix_ms}\n")
+        io.print("data: ")
+        {entry.severity, entry.source, entry.message}.to_json(io)
+        io.print("\n\n")
+      end
+    end
+  end
+end
--- a/src/lavinmq/in_memory_backend.cr
+++ b/src/lavinmq/in_memory_backend.cr
@@ -1,0 +1,53 @@
+require "log"
+require "../stdlib/channel"
+
+# A `Log::Backend` for holding last x nr of messages in memory.
+class Log::InMemoryBackend < Log::Backend
+  @entries = Deque(Log::Entry).new
+  @lock = Mutex.new
+  getter channels = Array(Channel(Log::Entry)).new
+  property size = 1024
+
+  @@instance : self = self.new
+
+  def entries
+    @lock.synchronize do
+      @entries.dup
+    end
+  end
+
+  def self.instance
+    @@instance
+  end
+
+  private def initialize
+    super(:direct)
+  end
+
+  def write(entry : Log::Entry) : Nil
+    @lock.synchronize do 
+      if @entries.size >= @size
+        @entries.shift(@entries.size - @size + 1)
+      end
+      @entries << entry
+
+      @channels.each do |ch|
+        ch.try_send(entry)
+      end
+    end
+  end
+
+  def add_channel
+    ch = Channel(Log::Entry).new(128)
+    @lock.synchronize do
+      @channels << ch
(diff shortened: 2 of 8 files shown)


carlhoerberg (changes requested):

src/lavinmq/http/controller/logs.cr
```diff
@@ -0,0 +1,34 @@
...
+            end
+          ensure
+            ::Log::InMemoryBackend.instance.remove_channel(channel)
+          end 
+          context
+        end
+      end
+
+      private def print_entry(entry, io)
+        io.print("data: #{entry.timestamp.to_utc.to_s} #{entry.severity.to_s} #{entry.source} #{entry.message} \n\n")
```
carlhoerberg, line 29:
```suggestion
        io.print("event: #{entry.timestamp.to_unix_ms}")
        io.print("data: #{entry.timestamp.to_utc.to_s} #{entry.severity.to_s} #{entry.source} #{entry.message}\n\n")
```

src/lavinmq/http/controller/logs.cr
```diff
@@ -0,0 +1,34 @@
...
+module LavinMQ
+  module HTTP
+    class LogsController < Controller
+      private def register_routes
+        get "/api/livelog" do |context, _params|
+          channel = ::Log::InMemoryBackend.instance.add_channel
+          context.response.content_type = "text/event-stream"
+          context.response.upgrade do |io|
+            ::Log::InMemoryBackend.instance.entries.each do |entry|
+              print_entry(entry, io)
```
carlhoerberg, line 16:
```suggestion
            if last_event_id = context.request.headers["Last-Event-Id"]?
              last_ts = last_event_id.to_i64
              ::Log::InMemoryBackend.instance.entries.each.skip_while { |e| e.timestamp.to_unix_ms <= last_ts } do
              print_entry(entry, io)
            else              
              ::Log::InMemoryBackend.instance.entries.each do |entry|
                print_entry(entry, io)
              end
```

carlhoerberg (commented):
Support for `Last-Event-Id` by identifying msgs by their timestamp

src/lavinmq/in_memory_backend.cr
```diff
@@ -2,40 +2,40 @@
...
+
+  def add_channel
+    Log.info { "Add channel to InMemoryBackend" }
+    ch = Channel(Log::Entry).new(128)
+    @channels << ch
+    ch
+  end
+
+  def remove_channel(channel)
+    Log.info { "Remove channel from InMemoryBackend" }
```
carlhoerberg, line 38: should be debug logging or none

carlhoerberg (commented):
Did you use `crystal tool format`?

viktorerlingsson:
> Did you use `crystal tool format`?

Yes!

carlhoerberg (approved):
Do you want to add "download logs" in this PR or another?

carlhoerberg:
no, i was thinking of just that. how can we make it bigger?

src/lavinmq/http/controller/logs.cr
```diff
@@ -0,0 +1,42 @@
...
+          ensure
+            ::Log::InMemoryBackend.instance.remove_channel(channel)
+          end
+          context
+        end
+      end
+
+      private def print_entry(entry, io)
+        io.print("id: #{entry.timestamp.to_unix_ms}\n")
+        io.print("data: #{entry.timestamp.to_utc} #{entry.severity} #{entry.source} #{entry.message}\n\n")
```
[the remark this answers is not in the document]
viktorerlingsson: > no, i was thinking of just that. how can we make it bigger?

It depends on setup I guess, but in some cases it could be possible to download the actual log files? If we know where the log files are stored and have access to them. So, maybe not when using stdout/journal for logging, but if you've specified a logfile in config. 

But I agree, we should just implement a download for this solution for now.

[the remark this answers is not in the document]
viktorerlingsson: > should make sure that `entry.message` can't include '\n', or prefix each line in message with `data: `. lines that doesn't start with `data:` will be ignored by sse.

Yes, I'll fix. I think it's best to prefix with `data: ` if `\n` exists in message

viktorerlingsson:
> no, i was thinking of just that. how can we make it bigger?

It depends on setup I guess, but in some cases it could be possible to download the actual log files? If we know where the log files are stored and have access to them. So, maybe not when using stdout/journal for logging, but if you've specified a logfile in config.

But I agree, we should just implement a download for this solution for now.

static/main.css
```diff
@@ -948,3 +948,44 @@ pre.arguments > div {
   text-decoration: none;
   color: #000;
 }
+
+.livelog {
```
carlhoerberg, line 952:
```suggestion
#livelog {
```

static/main.css
```diff
@@ -948,3 +948,44 @@ pre.arguments > div {
...
+  max-height: 85vh;
+  overflow-y: scroll;
+}
+
+.livelog table tbody tr td {
```
carlhoerberg, line 957:
```suggestion
#livelog td {
```

static/main.css
```diff
@@ -948,3 +948,44 @@ pre.arguments > div {
...
+
+#download-logs {
+  margin: 5px 0 5px 0;
+}
+
+.livelog table td, .livelog table th {
+  padding:0;
+}
+
+.livelog table thead .livelog-timestamp {
```
carlhoerberg, line 973: you can use the nth-child psudo selector instead of naming them. 

```suggestion
#livelog th:nth-child(1) {
```

static/main.css
```diff
@@ -948,3 +948,44 @@ pre.arguments > div {
...
+  width:185px;
+}
+
+.livelog table thead .livelog-severity {
```
carlhoerberg, line 977:
```suggestion
#livelog th:nth-child(2) {
```

static/main.css
viktorerlingsson: Yes - but I think the css is slightly easier to read/maintain with named elements rather than using nth-child in cases like this. I don't really mind either way though.

carlhoerberg: Absolut, judgement call. Just an idea.

carlhoerberg (approved):

carlhoerberg:
Would you like to add something else or are we good to squash merge this @viktorerlingsson ?

viktorerlingsson:
> Would you like to add something else or are we good to squash merge this @viktorerlingsson ?

We're good to merge! Just haven't gotten around to it

Issue #455, opened 2023-02-02
Definitions' `null` parsing error


iamolegga:
Don't know if this is an error of definitions export or import but when I'm exporting definitions each `bindings.N.arguments` is `null`. When I'm trying to import definitions again with the UI or ctl I got errors.

- For ctl it's `cast from Nil to Hash(K, V) failed`.
- For UI it's `Upload failed`.

When I replace `null`s with empty objects `{}` in JSON it works

carlhoerberg:
Thank you for reporting! Will fix

PR #457, opened 2023-02-07
Handle empty arguments hash on bindings for definitions import/export

viktorerlingsson:
Export definitions exports empty hash instead of null if there are no arguments for binding.
Import definitions allows import even if arguments on binding is null.

Fixes https://github.com/cloudamqp/lavinmq/issues/455

### DIFF
--- a/src/lavinmq/exchange/exchange.cr
+++ b/src/lavinmq/exchange/exchange.cr
@@ -316,7 +316,7 @@ module LavinMQ
         destination:      @destination.name,
         destination_type: @destination.is_a?(Queue) ? "queue" : "exchange",
         routing_key:      routing_key,
-        arguments:        arguments.nil? ? Hash(String, AMQP::Field).new : arguments,
+        arguments:        arguments || NamedTuple.new,
         properties_key:   BindingDetails.hash_key(@key),
       }
     end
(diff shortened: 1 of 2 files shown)


src/lavinmq/exchange/exchange.cr
```diff
@@ -316,7 +316,7 @@ module LavinMQ
         destination:      @destination.name,
         destination_type: @destination.is_a?(Queue) ? "queue" : "exchange",
         routing_key:      routing_key,
-        arguments:        arguments,
+        arguments:        arguments.nil? ? Hash(String, AMQP::Field).new : arguments,
```
carlhoerberg, line 319: Hashes are allocated on the heap, `NamedTuple`'s on the stack. Both becomes `"{}"` when `.to_json`. 

```suggestion
        arguments:        arguments || NamedTuple.new,
```

```
icr:11> Benchmark.memory { 1000.times { Hash(String, String).new } }
 => 61440
icr:12> Benchmark.memory { 1000.times { NamedTuple.new } }
 => 0
icr:13> Hash(String, String).new.to_json
 => "{}"
icr:14> NamedTuple.new.to_json
 => "{}"
```

Issue #454, opened 2023-01-28
Feature Request: Option to expose the Management UI at configurable URL path other than "/"


nifr:
Hey people, 
first of all thank you so much for your amazing work with lavinmq!

We currently evaluate lavinmq for internal testing and use it in our development environments.
The overall experience, arm64 support, resource-consumption etc. has been superb so far.

We would love to be able to proxy / access the management UI at a custom URL path i.e.  `https://<host>/lavinmq/`.

The current implementation with it's hardcoded asset paths in the HTML pages in folder [static/](https://github.com/cloudamqp/lavinmq/tree/main/static) does not allow for this.

Would it be an option to serve the UI pages as rendered templates and add a custom `path` option for the management UI that prefixes the assets?

Some examples how other vendors implement such feature are:
* OpenSearch Dashboards uses config option [server.basePath](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/config/opensearch_dashboards.yml#L9)
* Redis Insight uses environment [variable](https://docs.redis.com/latest/ri/installing/configurations/#riproxypath) `RIPROXYPATH=/custom-path`
* mailhog uses [option](https://github.com/mailhog/MailHog/blob/master/docs/CONFIG.md#configuring-mailhog) `--web-ui-path=[/custom-path` or environment variable `MH_WEB_UI_PATH=/custom-path`
* dozzle uses [option](https://github.com/amir20/dozzle#environment-variables-and-configuration) `--base=/custom-path` or environment variable `DOZZLE_BASE=/custom-path`
* [..]

An option like this would allow for:
* easier proxy configuration
* TLS certificate re-use
* [..]

Thanks for looking into this. I appreciate any feedback on this even if such feature is - totally understandable - not on your short to mid-term roadmap.

Keep up the good work - you're up to something big with lavinMQ.

Have a great day 😄

carlhoerberg:
Thank you for your kind feedback! Yes, been thinking it might make sense to move to a template system anyway.

carlhoerberg:
We figured we could solve this without templating, just use relative urls which was pretty straight forward. Templating will come as well, but it's not necessary to fix this issue.

carlhoerberg:
No configuration required, will just work.

PR #460, opened 2023-02-20
Relative links in the mgmt interface

carlhoerberg:
So that the mgmt interface can be remapped to a sub directory


carlhoerberg:
Fixes #454

PR #462, opened 2023-02-21
Github Action for Benchmarking

kickster97:
The goal is to continuously benchmark lavinmq/main in order to discover regressions in performance. 
Since both the main and pr branches are built and run on the same GHA runner they should have the same conditions and it is appropriate to compare them as they are presented in the comment made by github-actions bot. However, performance is expected to vary from runner to runner, so it is not appropriate to compare the benchmarking results form one commit to the results of another. 
**co-authered by @spuun**

### DIFF
--- a/.github/workflows/lavinmqperf.yml
+++ b/.github/workflows/lavinmqperf.yml
@@ -8,7 +8,7 @@ on:
       - main
 jobs:
   spec:
-    name: Spec
+    name: benchmark
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


carlhoerberg:
We have tested this before but GitHub CI has notably varied/poor performance. Maybe better luck with depot.dev?

kickster97:
> We have tested this before but GitHub CI has notably varied/poor performance. Maybe better luck with depot.dev?

noted! will try out depot

PR #463, opened 2023-02-21
Responsive UI

viktorerlingsson:
Made some changes to the UI to look better with smaller screens.

Also removed small-menu and replaced it with media queries in css instead

example: 
image


carlhoerberg (approved):
Very nice!

PR #466, opened 2023-03-03
In specs, let OS pick listen ports by default

spuun:
I always forget to set AMQP_PORT etc when i run specs, and I don't want to stop any running LavinMQ/RabbitMQ, so I made this change to always use a free port.


spuun:
This was done in #640

PR #467, opened 2023-03-03
Add Random Exchange Type (x-random)

jdamanalo:
### Additional Context
In rabbit this previously used routing keys.
> It's basically a direct exchange, with the exception that, instead of each consumer bound to that exchange with the same routing key getting a copy of the message, the exchange type randomly selects a queue to route to.

The behavior was changed to not use routing keys in this commit: [Ignore [binding] routing keys](https://github.com/rabbitmq/rabbitmq-server/commit/10115654c8e0ddd970558e89b8390f65e1c3e7ef).

### DIFF
--- a/src/lavinmq/exchange/random.cr
+++ b/src/lavinmq/exchange/random.cr
@@ -1,0 +1,50 @@
+require "./exchange"
+
+module LavinMQ
+  class RandomExchange < Exchange
+    def type : String
+      "x-random"
+    end
+
+    def bind(destination : Queue, routing_key, headers = nil)
+      @queue_bindings[{routing_key, nil}] << destination
+      after_bind(destination, routing_key, headers)
+    end
+
+    def bind(destination : Exchange, routing_key, headers = nil)
+      @exchange_bindings[{routing_key, nil}] << destination
+      after_bind(destination, routing_key, headers)
+    end
+
+    def unbind(destination : Queue, routing_key, headers = nil)
+      @queue_bindings[{routing_key, nil}].delete destination
+      after_unbind(destination, routing_key, headers)
+    end
+
+    def unbind(destination : Exchange, routing_key, headers = nil)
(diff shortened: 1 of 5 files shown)


src/lavinmq/exchange/random.cr
```diff
@@ -0,0 +1,50 @@
...
+
+    def do_queue_matches(routing_key, headers = nil, & : Queue -> _)
+      queues = @queue_bindings.values.map { |s| s.to_a }.flatten
+      without_internal = queues.select { |q| !q.internal? }
+
+      if without_internal.empty?
+        return
+      end
+
+      yield without_internal.sample
```
kickster97, line 37:
```suggestion
yield queues.select { |q| !q.internal? }.sample unless queues.empty?
```

Maintains original logic but with fewer lines of code, which can make it easier to read and maintain.

src/lavinmq/exchange/random.cr
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def do_exchange_matches(routing_key, headers = nil, & : Exchange -> _)
+      exchanges = @exchange_bindings.values.map { |s| s.to_a }.flatten
```
kickster97, line 41:
```suggestion
    exchanges =  @exchange_bindings.values.flat_map { |s| s.to_a }
```

src/lavinmq/exchange/random.cr
```diff
@@ -0,0 +1,50 @@
...
+
+      if exchanges.empty?
+        return
+      end
+
+      yield exchanges.sample
```
kickster97, line 47:
```suggestion
    yield exchanges.sample unless exchanges.empty?
```

kickster97 (changes requested):

kickster97:
This is not in the roadmap for LavinMQ atm, so i will close this PR and we'll open this again if we re-visit this.

PR #471, opened 2023-03-09
Reproducible builds

carlhoerberg:
Ideally two sequential builds should produce the exact same result. https://en.wikipedia.org/wiki/Reproducible_builds


carlhoerberg:
Dropped baked file system in favor of own implementation.

carlhoerberg:
One difference to `baked_file_system` is that this implementation doesn't compress the assets, but also doesn't have to decompress them each time they are served, increasing the binary size with about 3MB.

PR #472, opened 2023-03-13
dont render ui unless logged in

viktorerlingsson:
WIP. 
Checks if user is logged in before rendering UI.

Only implemented for Overview here for proof of concept. 

Needs handling of javascript errors.

### DIFF
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -189,6 +189,28 @@ module LavinMQ
         user
       end
 
+      private def logged_in?(context) : Bool
+        #puts ::HTTP::Cookies.from_client_headers(context.request.headers)["m"].value
+        cookie = ::HTTP::Cookies.from_client_headers(context.request.headers)["m"].value
+        begin
+          base64 = cookie.split("|")[2].split(":")[1].split("%")[0]
+          username, password = Base64.decode_string(base64).split(":")
+          if user = @amqp_server.users[username]?
+            if valid_auth?(user, password)
+              return true
+            end
+          end
+        rescue IndexError; Base64::Error
+        end
+        return false
+      end
+
+      private def valid_auth?(user, password) : Bool
+        return false if user.tags.empty?
+        return false if user.password.to_s.empty?
+        user.password.not_nil!.verify(password)
+      end
+
       def vhosts(user : User)
         @amqp_server.vhosts.each_value.select do |v|
           full_view_vhosts_access = user.tags.any? { |t| t.administrator? || t.monitoring? }
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -190,26 +190,18 @@ module LavinMQ
       end
 
       private def logged_in?(context) : Bool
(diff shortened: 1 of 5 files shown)


carlhoerberg:
doesn't it make more sense to do it in the controller than in the view? https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/http/controller/view_helpers.cr#L29-L49

viktorerlingsson:
> doesn't it make more sense to do it in the controller than in the view? https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/http/controller/view_helpers.cr#L29-L49

Yeah, agree. I'll look into it a bit more, but it might require more changes since we do all the redirects and login handling in javascript. So we still need to 'render' that.

carlhoerberg:
Or do the cookie check before the rendering. No server side check at all.

viktorerlingsson:
> Or do the cookie check before the rendering. No server side check at all.

I don't think that's possible since we use static html for the views - the browser will start to render the HTML regardless of what we do in javascript. I guess we can get around it by for example hiding the html/body element in css and then displaying it again once we have checked the cookie, but that will be a little bit slower, and will show a blank page for a short amount of time for every page load. I don't think that's ideal. 

Maybe we could do a redirect in the backend instead
https://devdocs.io/crystal/http/server/response#redirect%28location%3AString%7CURI%2Cstatus%3AHTTP%3A%3AStatus%3D%3Afound%29-instance-method

carlhoerberg:
Right now we wait for all html to be rendered, then load JS, that can be
changed.

viktorerlingsson:
> Right now we wait for all html to be rendered, then load JS, that can be changed.

Ah, of course. Maybe something like the newest changes then? A little bit of a compromise with a backend check (to actually verify the user/password combination) that does a javascript redirect to /login if not logged in.

spuun:
Imo it doesn't matter that we render views as long as there is no sensitive data in the html.

I think it would be kind of nice to handle all authentication in the `http.js`, so if one is not authenticated we popup a login dialog before http requests can proceed. As soon as the login is done, all requests are done.

It would also be nice to handle 401 and popup a login dialog, then retry the failed request automatically.

spuun:
Have you tested how caching is affecting this? I think it's a risk that the "non rendered" version of a view will be cached by the browser.

But this can be done server-side, so why not do it? Just add an if-statement in the get endpoint generated by `static_view`  macro and do a redirect, as long as an exception is done for `/login` endpoint. Or won't the redirect be possible because of relative paths? hm

viktorerlingsson:
> Have you tested how caching is affecting this? I think it's a risk that the "non rendered" version of a view will be cached by the browser.

>But this can be done server-side, so why not do it? Just add an if-statement in the get endpoint generated by static_view macro and do a redirect, as long as an exception is done for /login endpoint. Or won't the redirect be possible because of relative paths? hm

Yeah you're right - caching messes this up. I'll have a look a server-side redirect instead!

carlhoerberg:
Yes, HTTP caching makes server-side redirects and checks moot. Let's focus on JS checks only, make them earlier and more reliably, no server side changes should be required.

carlhoerberg:
#476 fixes the "flash of layout" issue.

PR #473, opened 2023-03-13
Log exchange

viktorerlingsson:
WIP Log exchange


carlhoerberg (changes requested):
Let's copy more of the behavior from rmq, the exchange is an `internal` topic exchange (which means that publushers can't publish to it, but otherwise it behaves like a normal topic exchange). Also let's call it `amq.lavinmq.log`. And the messages should use headers and not bodies like rmq, so that we don't have to rewrite our cloudamqp log-stream.

src/lavinmq/launcher.cr
```diff
@@ -89,6 +91,26 @@ module LavinMQ
       Log.info &.emit("logger settings", level: @config.log_level.to_s, target: target)
     end
 
+    private def setup_log_exchange
+      queue_name = "logstream"
+      exchange_name = "amq.lavinmq.log"
+      vhost = @amqp_server.vhosts["/"]
+      vhost.declare_exchange(exchange_name, "topic", true, true, true)
+      vhost.declare_queue(queue_name, true, false)
+      vhost.bind_queue(queue_name, exchange_name, "#")
```
carlhoerberg, line 100: No need to declare the queue, that's up to the user to do.

```suggestion
      exchange_name = "amq.lavinmq.log"
      vhost = @amqp_server.vhosts["/"]
      vhost.declare_exchange(exchange_name, "topic", true, true, true)
```

src/lavinmq/launcher.cr
```diff
@@ -89,6 +91,26 @@ module LavinMQ
...
+      log_channel = ::Log::InMemoryBackend.instance.add_channel
+      spawn do
+        while entry = log_channel.receive
+          vhost.publish(msg: Message.new(
+            exchange_name,
+            entry.severity.to_s,
+            "#{entry.timestamp} [#{entry.severity.to_s.upcase}] #{entry.source} - #{entry.message}",
```
carlhoerberg, line 107: The user might want to format the message differently, with the timestamp and the severity, no need to include it in the message body as it's available by other means.
```suggestion
            "#{entry.source} - #{entry.message}",
```

carlhoerberg (changes requested):
Make it optional, through a configuration option. How about log levels? Are all debug messages also sent here? Should that be configurable too?

viktorerlingsson:
> Make it optional, through a configuration option. 

Yes, I'll add a config var to enable/disable the log exchange. 

> How about log levels? Are all debug messages also sent here? Should that be configurable too?

Yeah, it's probably a good idea to be able to configure log level. Maybe we can default to same log level as is set for Lavin?

viktorerlingsson:
>> How about log levels? Are all debug messages also sent here? Should that be configurable too?

> Yeah, it's probably a good idea to be able to configure log level. Maybe we can default to same log level as is set for Lavin?

After some more thought, I'm not sure if it makes sense to be able to use a different log level here. We set the log level of the InMemoryBackend on startup, and that will dictate which log entries are in turn sent to the channel, and then handled here. 

So, we could be more restrictive here (only publishing ERROR and WARN for example), but we won't get more than what is set as log_level in config.

carlhoerberg (approved):

src/lavinmq/launcher.cr
viktorerlingsson: > We don't allow binding to internal exchanges

Hmm, that's only in the mgmt interface though. I can bind to it with a client, so that should be fine.

carlhoerberg: That's a bug. Users shouldnt be able to publish to internal exchanges, but should be able to bind to them

PR #475, opened 2023-03-14
Allow lavinmq to be used as a dependency

baelter:
Shards didn't accept the version format and pinning against a branch, even if it's the default, results in version conflicts for the same dependency is used in the project dependent on lavinmq.

This enables lavinmq to be used as a dependency.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,5 +1,5 @@
 name: lavinmq
-version: 1.0.0-beta.10
+version: 1.0.0-beta10
 
 authors:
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Interesting, where are you using lavinmq as a dependency? Have to check that the way we generate tags, version number for containers, deb and RPM packages are compatible with this format.

carlhoerberg:
When 1.0.0 is released (any day now) this will be resolved, so closing this PR.

PR #476, opened 2023-03-15
Fixes the "click fast and you get logged out" issue

carlhoerberg:
Check if auth cookie exist early and redirect to login if missing, don't check whoami, instead redirect if 401 on xhr requests.

Also no "flash of layout" when cookie is missing/expired.

Replaces #472

### DIFF
--- a/static/js/http.js
+++ b/static/js/http.js
@@ -1,35 +1,17 @@
 import * as Auth from './auth.js'
-import * as Dom from './dom.js'
 
-function testLoggedIn () {
-  const hash = window.location.hash
-  if (hash.startsWith('#/login')) {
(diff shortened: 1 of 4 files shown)


viktorerlingsson:
Nice, was thinking of something similar to what you did in `head.ecr`, but didn't want to duplicate to much code from Auth.js. 

There's still some issues if you log out though, because we don't delete the cookie when doing that. So the flashing is still present for some paths in that case, and some other paths results in error being shown (JSON.parse error, but we should get 401 and therefore redirect?).

Anyways, both problems can be solved by deleting the cookie in `signOut` instead of just clearing separate values.

viktorerlingsson (approved):

PR #477, opened 2023-03-16
Render active menu item server side

spuun:
Render active class on active menu item server-side.


src/lavinmq/http/controller/view_helpers.cr
```diff
@@ -14,6 +15,10 @@ module LavinMQ
         HTML.escape({{value}}, context.response)
       end
 
+      macro active_path?(path)
```
carlhoerberg, line 18: Why would this be a macro and not just a normal method?

carlhoerberg (commented):
👌

src/lavinmq/http/controller/view_helpers.cr
spuun: To not have to know about `route_path` and pass it as an argument too.

carlhoerberg: ok, i get it that macro might be needed here, but can `route_path` be replaced with `context.request.path` or something like that? `context` feels more constant than the `route_path` variable at least.

carlhoerberg: replaced it with 

```
context.request.path == "/#{{{path}}}" || (context.request.path == "/" && {{path}} == :".")
```

because then we could drop the `route_path` variable, and no need for `strip`, which would allocate a new string. the interoplation `"/#{{{path}}"` allocates a new string, but so does `{{path}}.to_s`, which we can't get away with.

PR #480, opened 2023-03-19
dont save password in clear text in cookie

viktorerlingsson:


carlhoerberg (commented):
How about the shovel, do we ask for the password there now or will it just fail?

viktorerlingsson:
> How about the shovel, do we ask for the password there now or will it just fail?

The shovel uses the B64 hashed password, so this _should_ have no consequences.

viktorerlingsson:
> Can remove `password` here too https://github.com/84codes/cloudamqp-api/blob/d3f52eaef62b2acbc3a6a86d7dc19335b6b8720e/lib/services/cluster_mgmt_auth.rb#L76

Yes, thanks! I've removed it from there.

PR #482, opened 2023-03-30
Remove duplicate logging

baelter:
Also logged inside `send_not_implemented`


carlhoerberg:
Thanks! I'm ok with you pushing things like this directly to main

PR #485, opened 2023-04-04
Try login by cookie if Authorization header is missing

viktorerlingsson:
Use cookie to login if Authorization header is not sent (for example for EventSource for livelog and /api/logs for downloading log file)

### DIFF
--- a/src/lavinmq/http/handler/basic_auth.cr
+++ b/src/lavinmq/http/handler/basic_auth.cr
@@ -10,45 +10,71 @@ module LavinMQ
       end
 
       def call(context)
-        if public_path?(context.request.path)
-          return call_next(context)
-        end
-
         if internal_unix_socket?(context)
           context.authenticated_username = UserStore::DIRECT_USER
           return call_next(context)
         end
 
-        if auth = context.request.headers.fetch("Authorization", nil)
-          if auth.starts_with? "Basic "
-            base64 = auth[6..]
-            begin
-              decode = Base64.decode_string(base64)
-              if idx = decode.index(':')
-                username = decode[0...idx]
-                password = decode[idx + 1..]
-                if user = @server.users[username]?
-                  if valid_auth?(user, password) && guest_only_loopback?(context, user)
-                    context.authenticated_username = username
-                    return call_next(context)
-                  end
-                end
-              end
-            rescue Base64::Error
-            end
+        if auth = basic_auth(context) || cookie_auth(context)
+          username, password = auth
+          if valid_auth?(username, password, context.request.remote_address)
+            context.authenticated_username = username
+            return call_next(context)
           end
         end
+
         forbidden(context)
       end
 
-      private def public_path?(request_path) : Bool
-        !request_path.starts_with?("/api/") &&
-          request_path != "/metrics" &&
-          !request_path.starts_with?("/metrics/")
+      private def basic_auth(context)
+        if auth = context.request.headers["Authorization"]?
+          if auth.starts_with? "Basic "
+            base64 = auth[6..]
+            return decode(base64)
+          end
+        end
+      end
+
+      private def cookie_auth(context)
+        if m = context.request.cookies["m"]?
+          if idx = m.value.index(':')
+            base64 = m.value[idx + 1..]
+            base64 = URI.decode(base64)
+            return decode(base64)
+          end
+        end
+      end
+
+      private def decode(base64) : Tuple(String, String)?
+        string = Base64.decode_string(base64)
+        if idx = string.index(':')
+          username = string[0...idx]
+          password = string[idx + 1..]
+          return {username, password}
+        end
+      rescue Base64::Error
+      end
+
+      private def valid_auth?(username, password, remote_address) : Bool
+        return false if password.empty?
+        if user = @server.users[username]?
+          return false if user.tags.empty?
+          if user_password = user.password
+            if user_password.verify(password)
+              if guest_only_loopback?(remote_address, username)
+                return true
+              end
+            end
+          end
+        end
+        false
       end
 
       private def internal_unix_socket?(context) : Bool
-        context.request.remote_address.to_s == HTTP::INTERNAL_UNIX_SOCKET
+        if addr = context.request.remote_address.as?(Socket::UNIXAddress)
+          return addr.to_s == HTTP::INTERNAL_UNIX_SOCKET
+        end
+        false
       end
 
       private def forbidden(context)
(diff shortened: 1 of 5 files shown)


carlhoerberg:
I could download logs both when SSO:ed from customer and admin

viktorerlingsson:
> I don't get this change, what is up with the auth? Very "creative" way of downloading.

> I could download logs both when SSO:ed from customer and admin

Hmm, I couldn't download logs in neither firefox nor chrome earlier, but now I can in firefox. Still doesn't work in chrome though. There's no Authorization header in the request to /api/logs .  

image

viktorerlingsson:
We can do something like this instead, to append username and password to the link href (like we do for the url for livelog):

```
var logLink=document.getElementById('download-logs-link')
const downloadUrl = new URL(window.location.pathname.replace(/\/logs$/, "/api/logs"), window.location.origin)
downloadUrl.username = Auth.getUsername()
downloadUrl.password = Auth.getPassword()
logLink.href=downloadUrl
```

carlhoerberg:
I would like to understand the source of the auth problem first

viktorerlingsson:
> I would like to understand the source of the auth problem first

Yeah, me too!

I can reproduce it in Chrome locally, but not in Firefox anymore. 
If I close chrome (or open an incognito window), I can't download the logs, and we don't send an Authorization header. If I login on any /api/ endpoint once it works, and sometimes it seems to magically start working after a while. 

I'm also getting an error from the EventSource in chrome. 

Hmm, maybe the livelog EventSource is the key here. Because it does requests with https://user:pass@url , which may cause the browser to cache the auth header for later use, for example for downloading logs. 

I'll look into it more tomorrow.

carlhoerberg:
We could also check the cookie server side and authenticate using the
user/pass in there.

viktorerlingsson:
> I would like to understand the source of the auth problem first

At least part of the problem is that Chrome treats http://localhost and http://user:pass@localhost as different sites, which at least caused the EventSource to fail. 

>We could also check the cookie server side and authenticate using the user/pass in there.

Updated the PR to try logging in with cookie in backend instead 👍

static/js/http.js
```diff
@@ -30,17 +30,22 @@ function request (method, path, options = {}) {
...
         }
-        return json
-      }).catch(function (e) {
-        if (!response.ok) {
-          throw new HTTPError(response.status, e.body || e.message || response.statusText)
-        }
-        return e
-      })
+      }
+      if (response.headers.get('Content-Type') == 'application/json') {
```
carlhoerberg, line 44: always `===` in js
```suggestion
      if (response.headers.get('Content-Type') === 'application/json') {
```

static/js/http.js
```diff
@@ -30,17 +30,22 @@ function request (method, path, options = {}) {
...
+        return response.json().catch(function (e) { return e })
+      }else if (response.headers.get('Content-Type') == 'text/plain') {
```
carlhoerberg, line 46:
```suggestion
      } else if (response.headers.get('Content-Type') === 'text/plain') {
```

views/logs.ecr
```diff
@@ -9,7 +9,7 @@
     <% render "header" %>
     <main>
       <section id="livelog" class="card livelog">
-        <button id="download-logs" class="btn-primary"><a href="api/logs">Download logs</a></button>
+        <button id="download-logs" class="btn-primary"><a id="download-logs-link" href="api/logs" target="_blank">Download logs</a></button>
```
carlhoerberg, line 12: `content-disposition: attachment` header should fix this, should be no need for a target in the link
```suggestion
        <button id="download-logs" class="btn-primary"><a id="download-logs-link" href="api/logs">Download logs</a></button>
```

src/lavinmq/http/handler/basic_auth.cr
```diff
@@ -23,24 +23,45 @@ module LavinMQ
...
-                username = decode[0...idx]
-                password = decode[idx + 1..]
-                if user = @server.users[username]?
-                  if valid_auth?(user, password) && guest_only_loopback?(context, user)
-                    context.authenticated_username = username
-                    return call_next(context)
-                  end
-                end
+              username = validate_auth(base64, context)
+              if username
```
carlhoerberg, line 27:
```suggestion
              if username = validate_auth(base64, context)
```

src/lavinmq/http/handler/basic_auth.cr
```diff
@@ -23,24 +23,45 @@ module LavinMQ
...
+                context.authenticated_username = username
+                return call_next(context)
               end
             rescue Base64::Error
             end
           end
         end
+
+        if context.request.cookies["m"]?
```
carlhoerberg, line 36: Extract this to its own method?

src/lavinmq/http/handler/basic_auth.cr
```diff
@@ -23,24 +23,45 @@ module LavinMQ
...
+          base64 = context.request.cookies["m"].value.split("|")[1].split(":")[1]
```
carlhoerberg, line 37: all this splitting is unnecessarily expensive. and is it always correct? always the second element?

src/lavinmq/http/handler/basic_auth.cr
```diff
@@ -23,24 +23,45 @@ module LavinMQ
...
+          base64 = URI.decode(base64)
+          begin
+            username = validate_auth(base64, context)
+            if username
```
carlhoerberg, line 41:
```suggestion
            if username = validate_auth(base64, context)
```

carlhoerberg (changes requested):
Great change!

src/lavinmq/http/handler/basic_auth.cr
viktorerlingsson: Yes, we can make it quicker!
And yes, since we set the cookie ourselves, so we know it always looks the same. 

I did some tests and `base64 = m.value.delete_at(0, (m.value.rindex(":") || 0) + 1)` should be ~2.5x quicker and use a lot less memory.

carlhoerberg: nice! now you're getting into the mindset :) 

or a little bit more readable imho:

```
if idx = m.value.index(':')
  base64 = m.value[idx + 1..]
else
  raise "invalid cookie"
end
```

views/logs.ecr
carlhoerberg: oh, ok! that makes sort of sense. alright, let it stay then!

views/logs.ecr
```diff
@@ -9,7 +9,7 @@
     <% render "header" %>
     <main>
       <section id="livelog" class="card livelog">
-        <button id="download-logs" class="btn-primary"><a href="api/logs">Download logs</a></button>
+        <button id="download-logs" class="btn-primary"><a download href="api/logs">Download logs</a></button>
```
carlhoerberg, line 12: Oh, didn't know about the `download` attribute, good find! https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download

carlhoerberg (approved):

PR #486, opened 2023-04-05
Keep multi selection on table refresh

snichme:
The code around multi selection has been refactored into a separate "module" so all the variables and functions dont mix with the other code, this allows the handler (multiActionHandler) to only expose a small set of methods to work on for the multi select feature.

Part from the refactoring, the main change implemented is that we keep a local hash to keep track of which checkbox was checked so we can marked them as checked again when the table refreshes see method `isSelected`

### DIFF
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -4,6 +4,84 @@ import * as Dom from './dom.js'
 import * as Vhosts from './vhosts.js'
 import * as Table from './table.js'
 
+
+const multiActions = (callback) => {
+  const multiSelectControls = document.getElementById("multiselect-controls")
+  let selectedQueues = {}
+
+  const performAction = (el) => {
+    const action = el.target.dataset.action
+    const elems = document.querySelectorAll("input[data-name]:checked")
+    const totalCount = elems.length
+    let performed = 0
+    elems.forEach(el => {
+      const data = el.dataset;
+      let url;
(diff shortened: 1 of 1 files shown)


spuun:
Ouch! Just realized that this feature is partly broken. If you select a few rows, then wait until table is updated (5 seconds) selection will be gone.

static/js/queues.js
```diff
@@ -4,6 +4,84 @@ import * as Dom from './dom.js'
...
+  }
+
+  const id = (vhost, name) => `${vhost}/${name}`
+
+  const onSelectionChange = () => {
+    const checked = document.querySelectorAll("input[data-name]:checked")
+    selectedQueues = {}
+    checked.forEach(el => {
+      const data = el.dataset
+      selectedQueues[id(data.vhost, data.name)] = true
```
spuun, line 51: `data.vhost` is uri encoded, so `/` is `%2F` here but not down in `isSelected`

spuun (commented):
I think the bug this will fix was introduced by me when i fixed the "edit entities" feature. I'm refactoring tables scripts right now, so I'll make sure not to break it again (preferably checkboxes won't be recreated every refresh, and then this change won't be needed anymore).

snichme:
Ok sounds good. Closing this PR

PR #487, opened 2023-04-05
Respond with 400 instead of 500 for PreconditionFailed errors

snichme:
Considering PreconditionFailed is generated by bad input 400 Bad request seems like a good response code
The body is the Exception message


carlhoerberg (approved):
Great! Are there more places this should be done? Exchanges? Policies?

snichme:
Found this by change so haven't looked into it that much, PreconditionFailed doesn't seem to be raised on policies and exchanges (after a quick search). 
Merge this now and if others arise fix them then?

PR #488, opened 2023-04-05
add exponential backoff for shovel-error connection retries

kickster97:
Adds an exponential backoff function for the reconnection delay on a shovel's runner. 

If you have an persistent error, the error will loop on reconnection_delay which by default is 5 seconds, and that can be a bit annoying. With this PR, the reconnection_delay will after 5 retries exponentially increase up to 5 minutes. 
If you have manually set the reconnection_delay to > 300(5 minutes), it will retry on that delay and not increase.

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -275,12 +275,14 @@ module LavinMQ
       @state = State::Stopped
       @error : String?
       @message_count : UInt64 = 0
+      @retries : Int64 = 0
 
       getter name, vhost
 
       def initialize(@source : AMQPSource, @destination : Destination,
                      @name : String, @vhost : VHost, @reconnect_delay = DEFAULT_RECONNECT_DELAY)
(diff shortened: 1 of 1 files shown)


kickster97:
@spuun this still interesting or should we close?

src/lavinmq/shovel/shovel.cr
```diff
@@ -316,18 +320,28 @@ module LavinMQ
...
       end
 
+      def exponential_reconnect_delay
+        @retries += 1
+        if @retries > 5
+          sleep @delay.seconds
+          @delay = [@delay * 2, 300].min unless @delay > 300
+        else
+          sleep @delay.seconds
+        end
```
spuun, line 342:
```suggestion
        sleep @delay.seconds
        if @retries > 5
          @delay = [@delay * 2, 300].min
        end
```
I guess it can simplified to this. And maybe change `5` and `300` to constants with describing names?

PR #490, opened 2023-04-06
UI table refactor

spuun:
This PR decouples table, pagination, sorting and data retrieval.  This was done to enable more flexibility, e.g. using a custom way of fetching shovel table data and still be able to sort the table (which was broken and removed before 1.0).


static/js/eventemitter.js
```diff
@@ -0,0 +1,20 @@
+class EventEmitter {
```
carlhoerberg, line 1: any reason you're not using an [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget)?

static/js/datasource.js
```diff
@@ -0,0 +1,208 @@
...
+        } catch(e) {
+          console.error(`Failed to load cached query state: ${e}`)
+        }
+      }
+      this._setStateFromQuery()
+      window.addEventListener('hashchange', evt => this._setStateFromQuery())
+    }
+  }
+
+  _setStateFromQuery() {
```
carlhoerberg, line 42: any reason you're not using the private prefix `#`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

static/js/pagination.js
```diff
@@ -0,0 +1,92 @@
+function create(container, dataSource) {
+  dataSource.on('update', update)
+  function makeLink(toPage, label = toPage, opts = {}) {
```
carlhoerberg, line 3: can't these be seperate functions? do they have to be defined inside another function?

static/js/queue.js
```diff
@@ -15,7 +16,19 @@ const resumeQueueForm = document.querySelector('#resumeQueue')
 const messageSnapshotForm = document.querySelector('#messageSnapshot')
 document.title = queue + ' | LavinMQ'
 let consumerListLength = 20
-const consumersTable = Table.renderTable('table', { keyColumns: [], countId: "consumer-count" }, function (tr, item) {
+
+const consumersDataSource = new (class extends DataSource {
```
carlhoerberg, line 20: short but pretty unreadable

static/js/eventemitter.js
spuun: Nope! I've just missed that it exists (I thought it was just an interface)!

static/js/datasource.js
spuun: I did at first, then I saw something related that wasn't fully supported so I changed. Dunno what now.

static/js/pagination.js
spuun: As of now, yes, since they reference dataSource and container. But that could of course be changed. I guess i see `create` as a constructor and `makeLink` and `update` as private methods, like how Table is made. 

I guess going all the way to make it a class would be nice.

PR #496, opened 2023-04-13
Bugfix: json parse error when no body

spuun:


viktorerlingsson:
I don't think we set content-length on /api responses? so `if (!response.headers.has('content-length')` means we always return null instead of `response.json()`

static/js/http.js
```diff
@@ -12,14 +12,16 @@ function request (method, path, options = {}) {
...
-      .then(response => {
-        if (!response.ok)
-          return response.json().then(json => reject(new HTTPError(response.status, json.reason)))
-        return resolv(response.json())
-      })
-  })
+  return window.fetch(path, opts)
+    .then(response => {
+      if (!response.ok) {
+        return response.json().then(json => { throw new HTTPError(response.status, json.reason) })
```
carlhoerberg, line 18: pretty sure i got "uncaught exception" when throwing inside the `then` callback, for some reason, maybe i did it differently, but that's why i introduced the extra `Promise`.. have you tested when eg. 403 is raised? logging in as non admin.

static/js/http.js
```diff
@@ -12,14 +12,16 @@ function request (method, path, options = {}) {
...
+      }
+      if (!response.headers.has('content-length') || parseInt(response.headers.get('content-length')) === 0) {
```
carlhoerberg, line 20: All json responses has transfer-encoding: chunked, and hence no content-length.

carlhoerberg (changes requested):

static/js/http.js
spuun: Lol, bad testing. But yes, i see now. It was the 201 that had a content-length=0. 

Maybe it's safe to assume we never have a body for 201 and 204 then? It was my first idea, but 201 may contain a body (according to specification, if I'm not mistaken).

static/js/http.js
spuun: I think it worked, but maybe not..? Maybe missed a case? I'll verify this.

spuun: It seems to work just fine

carlhoerberg:
I did this: https://github.com/cloudamqp/lavinmq/commit/5c3af4d75626d0bfee80d0fd8252e5f5f0a12beb

PR #498, opened 2023-04-14
do Fiber.yield in deliver_loop

viktorerlingsson:
Do Fiber.yield in deliver_loop to allow lavinmq to do other stuff while consuming

https://trello.com/c/DXlvpqpn/273-ui-doesnt-load-when-only-consuming-with-lavinmqperf

### DIFF
--- a/src/lavinmq/client/channel/consumer.cr
+++ b/src/lavinmq/client/channel/consumer.cr
@@ -62,6 +62,7 @@ module LavinMQ
             queue.consume_get(no_ack) do |env|
               deliver(env.message, env.segment_position, env.redelivered)
             end
+            Fiber.yield
           end
         rescue ex : ClosedError | Queue::ClosedError | Client::Channel::ClosedError | ::Channel::ClosedError
           @log.debug { "deliver loop exiting: #{ex.inspect}" }


viktorerlingsson:
Consume rate is way down with this change, guess because the fiber is being yielded after every consumed message?
Especially very low if also publishing at the same time. 

With change: 
```
Average publish rate: 187128.1 msgs/s
Average consume rate: 67.8 msgs/s
```

Without change: 
```
Average publish rate: 65894.9 msgs/s
Average consume rate: 63875.1 msgs/s
```


When running only one consumer
With change: 
```
Average consume rate: 59335.2 msgs/s
```
Without change: 
```
Average consume rate: 84173.7 msgs/s
```

carlhoerberg:
Yes, see this commit: https://github.com/cloudamqp/lavinmq/commit/258e3247eccd924527d4b39340a1eaa7f0375197

carlhoerberg:
Also, you almost always want to benchmark with --release mode.

viktorerlingsson:
> Yes, see this commit: [258e324](https://github.com/cloudamqp/lavinmq/commit/258e3247eccd924527d4b39340a1eaa7f0375197)

Okay, nice! 👍
I was playing around with something similar before finishing for the day, but wasn't sure how larger messages would affect something like this.

PR #503, opened 2023-04-23
Add docker-compose.yml, environment variables and a script to generate TLS files

brunneis:
- [`Dockerfile`] Define TLS ports using `EXPOSE` instruction
- [`Dockerfile`] Shift the default `lavinmq` arguments to the `CMD` instruction
- [`Dockerfile`] Specify `entrypoint.sh` in the `ENTRYPOINT` instruction
- [`entrypoint.sh`] Implement environment variables for all `lavinmq` argument options
- [`entrypoint.sh`] Give preference to environment variables over `CMD` / Docker command
- [`extras/generate-tls-files.sh`] Include a script for generating the key and certificate required for TLS configuration

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,4 +50,4 @@ WORKDIR /var/lib/lavinmq
 ENV GC_UNMAP_THRESHOLD=1
 HEALTHCHECK CMD ["/usr/bin/lavinmqctl", "status"]
 ENTRYPOINT ["entrypoint.sh"]
-CMD ["-b", "0.0.0.0", "--guest-only-loopback", "false"]
+CMD ["-b", "0.0.0.0", "--guest-only-loopback=false"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,9 +43,11 @@ RUN apt-get update && \
     apt-get install -y libssl3 libevent-2.1-7 libevent-pthreads-2.1-7 ca-certificates && \
     rm -rf /var/lib/apt/lists/* /var/cache/debconf/* /var/log/*
 COPY --from=builder /tmp/bin/* /usr/bin/
-EXPOSE 5672 15672
+COPY entrypoint.sh /usr/bin/
+EXPOSE 5672 15672 5671 15671
 VOLUME /var/lib/lavinmq
 WORKDIR /var/lib/lavinmq
 ENV GC_UNMAP_THRESHOLD=1
 HEALTHCHECK CMD ["/usr/bin/lavinmqctl", "status"]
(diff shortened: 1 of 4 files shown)


colinmollenhour (approved):
These Dockerfile changes are good, definitely using CMD is the more correct way to pass arguments to the entrypoint.sh.

brunneis:
> These Dockerfile changes are good, definitely using CMD is the more correct way to pass arguments to the entrypoint.sh.

Thanks!

kickster97:
Thanks for the pull request, we appreciate your interest and effort! We're currently not accepting external contributions and I'll have to close this, but if you're keen on getting this into LavinMQ, consider opening an issue or discussion. We can revisit it for future planning.

carlhoerberg:
> These Dockerfile changes are good, definitely using CMD is the more correct way to pass arguments to the entrypoint.sh.

No, you probably never want's to skip the default args, CMD will add to ENTRYPOINT, so I don't agree this is improves anything, just easier to make mistakes. 

Why would you duplicate all command line args as ENV vars? Better to do that in lavinmq.cr in that case instead of adding a superfluous script file, that contributers will forget to update when new commandline args are added. And what is the advatage to pass ENV vars to the docker rather than commandline args?

colinmollenhour:
> No, you probably never want's to skip the default args

Then why now make them the default? :)

Stuffing args into ENTRYPOINT makes it clunky to use `docker run` for arbitrary commands. For example, `docker run lavinmq:latest lavinmqctl -p cloudamqp list_queues". This is a really common pattern, see the official [redis entrypoint](https://github.com/docker-library/redis/blob/master/docker-entrypoint.sh) for a great example.

> Why would you duplicate all command line args as ENV vars?

Ideally, the executable supports environment variables and the command line overrides them, but since it doesn't I suppose the pros vs cons are subjective.

carlhoerberg:
thank you @colinmollenhour for your reply! 

> Then why not make them the default (in the executable)? :)
>
Because the defaults (only bind to localhost, only allow guest user on localhost) makes sense in non container environments but not in containers which you (normally) can't access over localhost.
 
> Stuffing args into ENTRYPOINT makes it clunky to use `docker run` for arbitrary commands. For example, `docker run lavinmq:latest lavinmqctl -p cloudamqp list_queues". This is a really common pattern, see the official [redis entrypoint](https://github.com/docker-library/redis/blob/master/docker-entrypoint.sh) for a great example.

Yes, an `ENTRYPOINT` makes it clunky to run any other executable in the image, but the suggested `entrypoint.sh` did `exec lavinmq ...` so that's no different. We could distribute `lavinmqctl` images, which only contains that binary, which would make running that easier.

PR #505, opened 2023-04-25
Iterator segfault bug

carlhoerberg:
Can't reproduce the seg fault anymore with this patch. Please let me know if you can.


viktorerlingsson (approved):
Looks good to me! 
Tested and verified with exchange-exchange binding.

PR #513, opened 2023-05-09
Json argument types

spuun:
The help buttons for adding options to json arguments always added an empty string as "placeholder". This can trick users to enter number values as strings, which results in an error when posting the form.

This PR will make numbers have integer as a placeholder.


viktorerlingsson (approved):

carlhoerberg (approved):
Very nice

PR #515, opened 2023-05-12
Msgs specs

nyior:
The following specs were written:

- Msgs with ttl should expire on requeue
- Msgs should not be requeued on consumer close
- Publish and consume messages larger than 128kb


kickster97 (approved):
lgtm :)

spec/server_spec.cr
```diff
@@ -187,22 +187,56 @@ describe LavinMQ::Server do
...
+
+  it "does not requeue messages on consumer close" do
+    with_channel do |ch|
+      q = ch.queue "msg_q"
+      q.publish_confirm "no requeue"
+      consumed = 0
+      tag = q.subscribe(no_ack: false) { |msg| consumed += 1 }
+      sleep 0.05
+      q.unsubscribe(tag)
+      sleep 0.05
```
carlhoerberg, line 235:
```suggestion
      done = Channel(Nil).new
      tag = q.subscribe(no_ack: false) { |msg| done.send nil }
      done.receive
      q.unsubscribe(tag)
```

carlhoerberg (changes requested):

PR #517, opened 2023-05-16
Add churn rates to /api/overview

viktorerlingsson:
Adds churn_rates to /api/overview 

```
"churn_rates": {
        "connection_created": 3303,
        "connection_created_details": {
            "rate": 489.0
        },
        "connection_closed": 3302,
        "connection_closed_details": {
            "rate": 488.8
        },
        "channel_created": 3303,
        "channel_created_details": {
            "rate": 488.8
        },
        "channel_closed": 3302,
        "channel_closed_details": {
            "rate": 488.8
        },
        "queue_declared": 0,
        "queue_declared_details": {
            "rate": 0.0
        },
        "queue_deleted": 0,
        "queue_deleted_details": {
            "rate": 0.0
        }
    },
```

### DIFF
--- a/src/lavinmq/http/controller/main.cr
+++ b/src/lavinmq/http/controller/main.cr
@@ -27,6 +27,8 @@ module LavinMQ
 
       OVERVIEW_STATS = {"ack", "deliver", "get", "publish", "confirm", "redeliver", "reject"}
       EXCHANGE_TYPES = {"direct", "fanout", "topic", "headers", "x-federation-upstream", "x-consistent-hash"}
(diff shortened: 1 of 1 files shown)


johanrhodin:
Would it be easy to add consumer churn rates?

carlhoerberg (approved):

viktorerlingsson:
> Would it be easy to add consumer churn rates?

We don't keep track of that the same way unfortunately, so it would be more work to add it. I won't add it in this PR, but we can add it to Trello.

PR #518, opened 2023-05-22
dont truncate mfiles on startup

viktorerlingsson:
Disable truncating mfiles on startup

https://trello.com/c/k8uAEipk/310-starting-lavinmq-with-many-empty-queues-after-crash-is-slow

### DIFF
--- a/src/lavinmq/queue/message_store.cr
+++ b/src/lavinmq/queue/message_store.cr
@@ -308,7 +308,7 @@ module LavinMQ
               @size += 1
             end
           rescue ex : IO::EOFError
-            if  mfile.pos < mfile.size
+            if mfile.pos < mfile.size
               Log.warn { "Truncating #{mfile.path} from #{mfile.size} to #{mfile.pos}" }
               mfile.truncate(mfile.pos)
             end
--- a/src/lavinmq/queue/message_store.cr
+++ b/src/lavinmq/queue/message_store.cr
@@ -305,10 +305,6 @@ module LavinMQ
               @size += 1
             end
           rescue ex : IO::EOFError
-            if mfile.pos < mfile.size
-              Log.warn { "Truncating #{mfile.path} from #{mfile.size} to #{mfile.pos}" }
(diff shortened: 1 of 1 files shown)


src/lavinmq/queue/message_store.cr
```diff
@@ -305,10 +305,6 @@ module LavinMQ
               @size += 1
             end
           rescue ex : IO::EOFError
-            if mfile.pos < mfile.size
-              Log.warn { "Truncating #{mfile.path} from #{mfile.size} to #{mfile.pos}" }
-              mfile.truncate(mfile.pos)
```
carlhoerberg, line 310: `@size` still needs to be updated

carlhoerberg (changes requested):

src/lavinmq/queue/message_store.cr
viktorerlingsson: > `@size` still needs to be updated

updated so we do `mfile.resize if ts.zero?` but still truncating if EOF or message doesn't fit

carlhoerberg: gr8!

src/lavinmq/queue/message_store.cr
```diff
@@ -296,7 +296,10 @@ module LavinMQ
             pos = mfile.pos
             raise IO::EOFError.new if pos + BytesMessage::MIN_BYTESIZE >= mfile.size # EOF or a message can't fit, truncate
             ts = IO::ByteFormat::SystemEndian.decode(Int64, mfile.to_slice + pos)
-            raise IO::EOFError.new if ts.zero? # This means that the rest of the file is zero, so truncate it
+            if ts.zero? # This means that the rest of the file is zero, so resize it
+              mfile.resize(mfile.pos)
+              break
+            end
```
carlhoerberg, line 302:
```suggestion
            break mfile.resize(pos) if ts.zero? # This means that the rest of the file is zero, so resize it
```

PR #520, opened 2023-05-29
add configurable values for min and warn for free_disk

viktorerlingsson:
Adds option to set `free_disk_min` and `free_disk_warn` in config to help stop lavinmq from filling up the disk

### DIFF
--- a/src/lavinmq/config.cr
+++ b/src/lavinmq/config.cr
@@ -41,6 +41,8 @@ module LavinMQ
     property? guest_only_loopback : Bool = true
     property max_message_size = 128 * 1024**2
     property? log_exchange : Bool = false
+    property free_disk_min : Int64 = 0  # bytes
+    property free_disk_warn : Int64 = 0 # bytes
     @@instance : Config = self.new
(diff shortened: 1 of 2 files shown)


src/lavinmq/config.cr
```diff
@@ -41,6 +41,8 @@ module LavinMQ
     property? guest_only_loopback : Bool = true
     property max_message_size = 128 * 1024**2
     property? log_exchange : Bool = false
+    property free_disk_min : Int64 = 0  # bytes
+    property free_disk_warn : Int64 = 0 # bytes
```
carlhoerberg, line 45: Should we consider some defaults here?

carlhoerberg (approved):

src/lavinmq/config.cr
viktorerlingsson: Skipping defaults here since lavinmq can be run on a lot of different hardware. I'll add some notes to the documentation about it instead.
We could also add some default values to be set for CloudAMQP deploys.

PR #525, opened 2023-06-12
Stream queues

viktorerlingsson:


src/lavinmq/queue/queue_factory.cr
```diff
@@ -36,5 +39,12 @@ module LavinMQ
         false
       end
     end
+
+    private def self.stream_queue?(frame) : Bool
+      if value = frame.arguments["x-queue-type"]?
+        return value == "stream"
+      end
+      false
```
carlhoerberg, line 47:
```suggestion
      frame.arguments["x-queue-type"]? == "stream"
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,187 @@
...
+        @consumers << consumer
+        notify_consumers_empty(false) if was_empty
+      end
+      @exclusive_consumer = true if consumer.exclusive
+      @has_priority_consumers = true unless consumer.priority.zero?
+      @log.debug { "Adding consumer (now #{@consumers.size})" }
+      notify_observers(:add_consumer, consumer)
+    end
+
+    private def handle_arguments
```
carlhoerberg, line 155: what's different here from the normal Queue implementation? doesnt it need to be overriden?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,187 @@
...
+      get(true) do |env|
+        yield env
+        env.redelivered ? (@redeliver_count += 1) : (@deliver_count += 1)
+      end
+    end
+
+    # yield the next message in the ready queue
+    # returns true if a message was deliviered, false otherwise
+    # if we encouncer an unrecoverable ReadError, close queue
+    private def get(no_ack, & : Envelope -> Nil) : Bool
```
carlhoerberg, line 109: what's different here from the normal implemenation?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,187 @@
...
+
+    def add_offset_header(msg, pos)
+      headers = msg.properties.headers || ::AMQP::Client::Arguments.new
+      headers["x-stream-offset"] = pos.as(AMQ::Protocol::Field)
+      msg.properties.headers = headers
+      msg
+    end
+
+    # save message id / segment position
+    def publish(msg : Message) : Bool
```
carlhoerberg, line 72: i would suggest adding the SegmentPosition as the offset, but do it in the MessageStore, not here

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,187 @@
...
+  class StreamQueue < Queue
+    @durable = true
+    @exclusive_consumer = false
+    @no_ack = true
+
+    private def init_msg_store(data_dir)
+      @msg_store = StreamQueueMessageStore.new(data_dir)
+    end
+
+    class StreamQueueMessageStore < MessageStore
```
carlhoerberg, line 13: You need a read segment/position per consumer, unlike the normal message store.

carlhoerberg: Maybe a MessageStore per consumer? is that possible?

carlhoerberg: that will create duplicate `MFile`s, a bit of waste, but maybe not too bad, can try it out. otherwise somehow store the current Segment and Position per consumer, and with each shift update it.

carlhoerberg: also need a sepearate `@requeue` per consumer.

src/lavinmq/http/controller/queues.cr
```diff
@@ -50,7 +50,9 @@ module LavinMQ
             body = parse_body(context)
             durable = body["durable"]?.try(&.as_bool?) || false
             auto_delete = body["auto_delete"]?.try(&.as_bool?) || false
+            stream_queue = body["stream_queue"]?.try(&.as_bool?) || false
```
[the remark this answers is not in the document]
viktorerlingsson: Hmm, yeah maybe!

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: I don't think we can/should use SegmentPositions as offset, because their position depend on size of message and so on, right? 

For streams in Rabbit, the offset is just an incrementing id, so you always know that the next message after n will have offset n+1. And when you start consuming with an offset=n, you get the message with offset=n, so you need to be able to figure out the offset of the next message.

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: Yeah, not sure what the best solution is yet. I think storing Segment and Position per consumer will be easiest. But we also need a good way to "seek" to an offset I guess? If a new consumer starts consuming at a random offset.

src/lavinmq/client/channel.cr
```diff
@@ -374,10 +378,10 @@ module LavinMQ
           case offset_arg
           when "first", "next" # same as offset = 0
           when "last"
-            offset = -1.as?(UInt64)               # FIX ME!
-          when offset_arg.as?(Int)                # FIX ME!
-            offset_int = offset_arg.as?(Int) || 0 # FIX ME!
-            offset = offset_int.as?(UInt64)       # FIX ME!
+            offset = -1.as?(UInt64) # FIX ME!
```
carlhoerberg, line 381: uint64 can't be negative, should we use Int64 instead?

src/lavinmq/client/channel.cr
```diff
@@ -374,10 +378,10 @@ module LavinMQ
...
+          when offset_arg.as?(Int)  # FIX ME!
+            offset_int = offset_arg.as?(Int) || 0
+            offset = offset_int.to_i.to_u64 # FIX ME!
```
carlhoerberg, line 384:
```suggestion
          when offset_int = offset_arg.as?(Int)
            offset = offset_int.to_u64 # FIX ME!
```

src/lavinmq/client/channel/consumer.cr
```diff
@@ -61,7 +61,7 @@ module LavinMQ
             {% unless flag?(:release) %}
               @log.debug { "Getting a new message" }
             {% end %}
-            queue.consume_get(no_ack, offset) do |env|
+            queue.consume_get(no_ack, offset, self) do |env|
```
carlhoerberg, line 64: only send `self`, everything else can be fetched from it?

src/lavinmq/queue/stream_queue.cr
carlhoerberg: Consumers can't "re-consume specific offsets", then the consumer has to be cancceled and start a new consumer with a certain offset, all other prefetched messages the conusmer might have received have to be redelivered then. So requeueing support is required imho. 

We should not create a new queue per consumer, that won't scale very well. Better that each consumer "remebers" where it is is in the queue.

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,199 @@
...
+      end
+      drop_overflow unless immediate_delivery?
+      true
+    rescue ex : MessageStore::Error
+      @log.error(exception: ex) { "Queue closed due to error" }
+      close
+      raise ex
+    end
+
+    def basic_get(no_ack, force = false, & : Envelope -> Nil) : Bool
```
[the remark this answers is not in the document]
viktorerlingsson: Hmm, basic_get is used when getting messages in the UI right? does it make sense to allow that for stream queues in some way? Either getting with no offset (start of queue) or adding a field for providing offset.

src/lavinmq/queue/stream_queue.cr
```diff
@@ -6,18 +6,61 @@ module LavinMQ
...
+              @bytesize += bytesize
+              @size += 1
+            end
+          rescue ex : IO::EOFError
+            if mfile.pos < mfile.size
+              Log.warn { "Truncating #{mfile.path} from #{mfile.size} to #{mfile.pos}" }
+              mfile.truncate(mfile.pos)
+            end
+            break
+          rescue ex : OverflowError | AMQ::Protocol::Error::FrameDecode
```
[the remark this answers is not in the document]
viktorerlingsson: Just copied it from message_store, haven't looked in to that. I'll check!

src/lavinmq/client/channel/consumer.cr
```diff
@@ -241,6 +241,39 @@ module LavinMQ
...
+
+        def initialize(@channel : Client::Channel, @tag : String,
+                       @queue : Queue, @no_ack : Bool, @exclusive : Bool, @priority : Int32, @offset : Int64?)
+          super(@channel, @tag, @queue, @no_ack, @exclusive, @priority)
+        end
+
+        def update_segment(segment, pos)
+          @segment = segment
+          @pos = pos
+        end
```
carlhoerberg, line 260: why not make `@segment` and `@pos` properties? 

```suggestion
        property segment, pos
```

src/lavinmq/client/channel/consumer.cr
```diff
@@ -241,6 +241,39 @@ module LavinMQ
...
+
+        def update_offset(offset)
+          @offset = offset
+        end
```
carlhoerberg, line 264:
```suggestion
        property offset
```

src/lavinmq/client/channel/consumer.cr
```diff
@@ -241,6 +241,39 @@ module LavinMQ
...
+
+        def reject(unack, requeue)
+          was_full = @unacked == @prefetch_count
+          @unacked -= 1
+          notify_hash_capacity(true) if was_full
```
carlhoerberg, line 269: no need to notify the queue, because the size of the queue hasn't changed, other stream consumers aren't intreseted in this

src/lavinmq/queue/message_store.cr
```diff
@@ -93,7 +93,7 @@ module LavinMQ
         end
       end
 
-      def shift? : Envelope? # ameba:disable Metrics/CyclomaticComplexity
+      def shift?(consumer = Nil) : Envelope? # ameba:disable Metrics/CyclomaticComplexity
```
carlhoerberg, line 96: `Nil != nil`, nil is a value, Nil is a class.

```suggestion
      def shift?(consumer = nil) : Envelope? # ameba:disable Metrics/CyclomaticComplexity
```

src/lavinmq/queue/queue.cr
```diff
@@ -43,6 +43,7 @@ module LavinMQ
     @consumers = Array(Client::Channel::Consumer).new
     @consumers_lock = Mutex.new
     @message_ttl_change = Channel(Nil).new
+    @queue_type = "standard"
```
carlhoerberg, line 46: not needed, check the class instead.

src/lavinmq/queue/queue.cr
```diff
@@ -187,6 +188,10 @@ module LavinMQ
       @exclusive_consumer
     end
 
+    def is_stream_queue?
+      @queue_type == "stream"
```
carlhoerberg, line 192: i don't think this is needed, check `queue.is_a? StreamQueue` instead

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+
+    class StreamQueueMessageStore < MessageStore
+      @last_offset = 0_i64
+
+      def initialize(@data_dir : String, @replicator : Replication::Server?)
+        super
+      end
+
+      # Populate bytesize, size and segment_msg_count
+      private def load_stats_from_segments : Nil
```
carlhoerberg, line 23: is this different from the base class implementation?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+      replicator = @vhost.@replicator
+      @msg_store = StreamQueueMessageStore.new(data_dir, replicator)
+    end
+
+    class StreamQueueMessageStore < MessageStore
+      @last_offset = 0_i64
+
+      def initialize(@data_dir : String, @replicator : Replication::Server?)
+        super
+      end
```
carlhoerberg, line 20: no need to override it if you just do `super` inside
```suggestion
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+          rescue IndexError
+          end
+
+          mfile.pos = 4
+          @segment_msg_count[seg] = count
+        end
+      end
+
+      def shift?(consumer) : Envelope?                              # ameba:disable Metrics/CyclomaticComplexity
+        consumer.update_offset(@last_offset) unless consumer.offset # if no offset provided, use offset of last published message
```
carlhoerberg, line 62: i don't get this, can't the consumer's offset be set when it's starting instead? why do it here in the hot path?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+        consumer_offset = consumer.offset || 0_i64
```
carlhoerberg, line 63: because of the previous line consumer.offset can never be nil, can it?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+        return if @last_offset <= consumer_offset # don't consume unless there's something to consume
+
+        seg = consumer.segment || @segments.first_value
+        pos = consumer.pos || 0_u32
+        requeued = consumer.requeued || Deque(SegmentPosition).new
```
carlhoerberg, line 68: how can `consumer.requeued` be nil? does that even make sense?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+        raise ClosedError.new if @closed
+        if sp = requeued.shift?
+          segment = @segments[sp.segment]
+          begin
+            msg = BytesMessage.from_bytes(segment.to_slice + sp.position)
+            notify_empty(true) if @size.zero?
```
carlhoerberg, line 74: no one is intreseted if this consumers "virtual queue" is empty or not

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+        env = @msg_store_lock.synchronize { @msg_store.shift?(consumer) } || break
+        if has_expired?(env.message) # guarantee to not deliver expired messages
+          expire_msg(env, :expired)
+          next
+        end
+
+        sp = env.segment_position
+        begin
+          yield env # deliver the message
+        rescue ex   # requeue failed delivery
```
carlhoerberg, line 191: if the delivery failed, i can't think of any other reason than that the connection is broken, and thus, requeuing doesn't make sense as the consumer/channel/connection will be closed right after this.

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+    rescue ex : MessageStore::Error
+      @log.error(exception: ex) { "Queue closed due to error" }
+      close
+      raise ClosedError.new(cause: ex)
+    end
+
+    def ack(sp : SegmentPosition) : Nil
+      return if @deleted
+      true
+    end
```
carlhoerberg, line 207:
```suggestion
    def ack(sp : SegmentPosition) : Nil
    end
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,228 @@
...
+
+    # do nothing?
+    def reject(sp : SegmentPosition, requeue : Bool)
+      return if @deleted || @closed
+      true
+    end
+
+    def add_consumer(consumer : Client::Channel::StreamConsumer)
```
carlhoerberg, line 215: how is this different from the base class implementation?

src/lavinmq/client/channel/consumer.cr
```diff
@@ -241,6 +241,39 @@ module LavinMQ
 
         class ClosedError < Error; end
       end
+
+      class StreamConsumer < LavinMQ::Client::Channel::Consumer
```
carlhoerberg, line 245: please move this class to its own file

src/lavinmq/queue/stream_queue.cr
carlhoerberg: ah, one line, updating the `@last_offset`, i think we should figure out a better way of doing that instead of copy pasting 20+ lines.

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: Yes, that's better

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: No, it can't be nil, removed!

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: Yeah, I can't think of any other reason either. I'll remove the rescue block!

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: Hmm, only removed some stuff that shouldn't be possible anyway (regarding exclusive consumer, priority consumers, etc). I think I can safely remove this method!

src/lavinmq/client/channel.cr
```diff
@@ -358,7 +358,13 @@ module LavinMQ
           unless frame.no_wait
             send AMQP::Frame::Basic::ConsumeOk.new(frame.channel, frame.consumer_tag)
           end
-          c = Consumer.new(self, frame.consumer_tag, q, frame.no_ack, frame.exclusive, priority)
+
+          c = if q.is_stream_queue?
```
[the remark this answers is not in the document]
viktorerlingsson: Yeah, that looks better. But class is Queue here for a stream. I guess I need to fix that somewhere else? (Because @queues on vhost is `Hash(String, Queue)` ?)

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: yes, already changed with what we discussed earlier today (but not committed yet) 👍

src/lavinmq/client/channel/stream_consumer.cr
```diff
@@ -0,0 +1,98 @@
...
+module LavinMQ
+  class Client
+    class Channel
+      class StreamConsumer < LavinMQ::Client::Channel::Consumer
+        # @offset : Int64
+        @segment = 1_u32
+        @offset = 0_i64
+        @pos = 4_u32
+        @requeued = Deque(SegmentPosition).new
+        getter requeued
```
carlhoerberg, line 14: can be combined:
```suggestion
        getter requeued = Deque(SegmentPosition).new
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -188,10 +62,7 @@ module LavinMQ
         sp = env.segment_position
         begin
           yield env # deliver the message
-        rescue ex   # requeue failed delivery
-          consumer.requeue(sp)
-          raise ex
-        end
+end
```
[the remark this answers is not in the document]
viktorerlingsson: Hmm, yeah. `crystal tool format` did that for some reason

src/lavinmq/client/channel/consumer.cr
```diff
@@ -7,16 +7,22 @@ module LavinMQ
     class Channel
       class Consumer
         include SortableJSON
-
+        @tag : String
+        @priority : Int32
+        @exclusive : Bool
+        @no_ack : Bool
         getter channel, tag, queue, no_ack, exclusive, priority
```
carlhoerberg, line 14:
```suggestion
        getter tag : String
        getter priority : Int32
        getter exclusive : Bool
        getter no_ack : Bool
        getter channel, queue
```

src/lavinmq/client/channel/consumer.cr
```diff
@@ -7,16 +7,22 @@ module LavinMQ
...
         getter prefetch_count = 0u16
         getter unacked = 0_u32
         getter? closed = false
         @log : Log
         @flow : Bool
 
-        def initialize(@channel : Client::Channel, @tag : String,
-                       @queue : Queue, @no_ack : Bool, @exclusive : Bool, @priority : Int32)
+        def initialize(@channel : Client::Channel, @queue : Queue, @frame : AMQP::Frame::Basic::Consume)
```
carlhoerberg, line 21: does the frame need to be a instance variable? 
```suggestion
        def initialize(@channel : Client::Channel, @queue : Queue, frame : AMQP::Frame::Basic::Consume)
```

src/lavinmq/client/channel/stream_consumer.cr
```diff
@@ -0,0 +1,88 @@
...
+require "../../sortable_json"
+require "../../error"
+
+module LavinMQ
+  class Client
+    class Channel
+      class StreamConsumer < LavinMQ::Client::Channel::Consumer
+        @segment = 1_u32
+        @offset = 0_i64
+        @pos = 4_u32
```
carlhoerberg, line 11:
```suggestion
        property segment = 1_u32
        property offset = 0_i64
        property pos = 4_u32
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -0,0 +1,123 @@
...
+          end
+
+          mfile.pos = 4
+          @segment_msg_count[seg] = count
+        end
+      end
+
+      def empty?(consumer)
+        return true if @size.zero?
+        @last_offset <= consumer.offset
```
carlhoerberg, line 52: Can't the consumer check this them selves? In `StreamConsumer`: `queue.last_offset <= @offset`?

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,102 @@
...
+    @exclusive_consumer = false
+    @no_ack = true
+    getter new_messages : Channel(Bool)
+
+    def initialize(@vhost : VHost, @name : String,
+                   @exclusive = false, @auto_delete = false,
+                   @arguments = Hash(String, AMQP::Field).new)
+      super
+      @new_messages = stream_queue_msg_store.new_messages
+    end
```
carlhoerberg, line 17: no need to assign the channel to an instance variable, is it?
 
```suggestion

    def initialize(@vhost : VHost, @name : String,
                   @exclusive = false, @auto_delete = false,
                   @arguments = Hash(String, AMQP::Field).new)
      super
    end
    
    def new_messages : Channel(Bool)
      stream_queue_msg_store.new_messages
    end
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,102 @@
...
+  class StreamQueue < Queue
+    @durable = true
+    @exclusive_consumer = false
+    @no_ack = true
+    getter new_messages : Channel(Bool)
+
+    def initialize(@vhost : VHost, @name : String,
+                   @exclusive = false, @auto_delete = false,
+                   @arguments = Hash(String, AMQP::Field).new)
+      super
```
carlhoerberg, line 15: should raise an `ArgumentError` if exclusive or auto_delete is true

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,102 @@
...
+
+    # save message id / segment position
+    def publish(msg : Message) : Bool
+      return false if @state.closed?
+      reject_on_overflow(msg)
+      @msg_store_lock.synchronize do
+        @msg_store.push(msg)
+        @publish_count += 1
+      end
+      drop_overflow unless immediate_delivery?
```
carlhoerberg, line 36: `immediate_delivery` doesn't make sense on a stream queue
```suggestion
      drop_overflow
```

src/lavinmq/queue/stream_queue.cr
```diff
@@ -0,0 +1,102 @@
...
+        env.redelivered ? (@redeliver_count += 1) : (@deliver_count += 1)
+      end
+    end
+
+    # yield the next message in the ready queue
+    # returns true if a message was deliviered, false otherwise
+    # if we encouncer an unrecoverable ReadError, close queue
+    private def get(consumer : Client::Channel::StreamConsumer, & : Envelope -> Nil) : Bool
+      raise ClosedError.new if @closed
+      loop do # retry if msg expired or deliver limit hit
```
carlhoerberg, line 67: does delivery limits make sense for stream queues? maybe it does... thinking we maybe can simplify things.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -0,0 +1,123 @@
...
+            end
+            break
+          rescue ex : OverflowError | AMQ::Protocol::Error::FrameDecode
+            raise Error.new(mfile, cause: ex)
+          end
+
+          begin # sets @last_offset to last message offset
+            mfile.seek(mfile.pos - last_bytesize, IO::Seek::Set)
+            msg = BytesMessage.from_bytes(mfile.to_slice + mfile.pos)
+            @last_offset = offset_from_headers(msg.properties.headers)
```
carlhoerberg, line 41: i'm thinking of not overriding this huge method, and instead make a new method that reads the last message in it self, called from the constructor.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -0,0 +1,123 @@
...
+          begin
+            msg = BytesMessage.from_bytes(segment.to_slice + sp.position)
+            return Envelope.new(sp, msg, redelivered: true)
+          rescue ex
+            raise Error.new(segment, cause: ex)
+          end
+        end
+
+        loop do
+          @rfile_id = consumer.segment
```
carlhoerberg, line 70: no point in updating an instance variable here, it will be overwritten by the next consumer
```suggestion
          rfile_id = consumer.segment
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -0,0 +1,123 @@
...
+          rfile = @segments[consumer.segment]
+
+          if consumer.pos == rfile.size # EOF?
+            select_next_read_segment
```
carlhoerberg, line 74:
```suggestion
            select_next_read_segment(consumer)
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -0,0 +1,123 @@
...
+            consumer.segment = @rfile_id
+            consumer.pos = 4_u32
+            rfile = @segments[@rfile_id]
+            next
+          end
+          if deleted?(consumer.segment, consumer.pos)
```
carlhoerberg, line 80: can a message in the middle of a stream queue be deleted? should individual message ttl's be respected?

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: queue_factory will throw PreconditionFailed if exclusive or auto_delete is true for a stream, is that enough?

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: Hmm, yeah I guess. Then we need to overwrite select_next_read_segment from message_store in stream_queue_message_store.

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: > can a message in the middle of a stream queue be deleted? 

Hmm, I'm not sure. It would be kinda weird if id 6 doesn't exist but 5 and 7 do. So I'm thinking it doesn't make sense to delete messages in the middle of a stream queue. 

Same goes for individual message ttl's then I guess.

carlhoerberg: Yes, let's skip that, and everything else mentioned here: https://www.rabbitmq.com/streams.html#feature-matrix

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: fixed

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: I think we cover most things from that matrix now. 


I see that rabbit also has single active consumer for streams in 3.11+. I say we can implement that later? https://www.rabbitmq.com/streams.html#single-active-consumer

src/lavinmq/queue/stream_queue.cr
viktorerlingsson: No, I don't think delivery limits makes sense for stream queues. But that's just an old comment copied from normal queues, it wasn't enforced in stream queues. Removed the comment.

PR #527, opened 2023-06-15
Adds consumer view

kickster97:
Adds a consumer view, with same values as the one found in queue. 

The value with having a separate consumers page is that you can list all consumers across LavinMQ, and not see them one queue at a time. 

Should we remove the consumer form from queues, i don't think so because its nice to have all metrics gathered together. But its worth mentioning that they are more or less the same forms.

### DIFF
--- a/static/js/consumers.js
+++ b/static/js/consumers.js
@@ -17,7 +17,8 @@ const tableOptions = {
   search: true
 }
 
-Table.renderTable('table', tableOptions, function (tr, item, all) {
+Table.renderTable('table', tableOptions, function (tr, item, firstRender) {
+  if (!firstRender) return
   const channelLink = document.createElement('a')
   channelLink.href = 'channel#name=' + encodeURIComponent(item.channel_details.name)
   channelLink.textContent = item.channel_details.name
--- a/static/js/consumers.js
+++ b/static/js/consumers.js
@@ -1,0 +1,57 @@
+import * as Table from './table.js'
+import * as HTTP from './http.js'
+import * as DOM from './dom.js'
+
+const vhost = window.sessionStorage.getItem('vhost')
+let url = 'api/consumers'
+if (vhost && vhost !== '_all') {
+  url = `api/consumers/${encodeURIComponent(vhost)}`
+}
+
+const tableOptions = {
+  url,
+  keyColumns: ['channel_details', 'consumer_tag'],
+  interval: 5000,
+  pagination: true,
+  columnSelector: true,
+  search: true
+}
+
+Table.renderTable('table', tableOptions, function (tr, item, all) {
+  console.log(item)
+  const channelLink = document.createElement('a')
+  channelLink.href = 'channel#name=' + encodeURIComponent(item.channel_details.name)
+  channelLink.textContent = item.channel_details.name
+  const ack = item.ack_required ? '●' : '○'
+  const exclusive = item.exclusive ? '●' : '○'
+  const cancelForm = document.createElement('form')
+  const btn = document.createElement('button')
+  btn.classList.add('btn-primary')
+  btn.type = 'submit'
+  btn.textContent = 'Cancel'
+
+  cancelForm.appendChild(btn)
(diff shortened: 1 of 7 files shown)


static/js/consumers.js
```diff
@@ -0,0 +1,54 @@
...
+  const exclusive = item.exclusive ? '●' : '○'
+  const cancelForm = document.createElement('form')
+  const btn = document.createElement('button')
+  btn.classList.add('btn-primary')
+  btn.type = 'submit'
+  btn.textContent = 'Cancel'
+
+  cancelForm.appendChild(btn)
+  const urlEncodedVhost = encodeURIComponent(vhost)
+  console.log(vhost)
```
carlhoerberg, line 34:
```suggestion
```

views/header.ecr
```diff
@@ -30,6 +30,7 @@
     <% {
       ".":            "Overview",
       "connections":  "Connections",
+      "consumers":  "Consumers",
```
carlhoerberg, line 33:
```suggestion
      "consumers":    "Consumers",
```

carlhoerberg (commented):
Should probably add columns for vhost and queue as well

static/js/consumers.js
```diff
@@ -0,0 +1,54 @@
...
+
+const vhost = window.sessionStorage.getItem('vhost')
+let url = 'api/consumers'
+if (vhost && vhost !== '_all') {
+  url = `api/consumers/${encodeURIComponent(vhost)}`
+}
+
+const tableOptions = {
+  url,
+  keyColumns: ['channel_details'],
```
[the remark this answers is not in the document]
carlhoerberg: Why wouldn't `channel_details` work as key? https://github.com/cloudamqp/lavinmq/blob/fd31b05c7d8193dccfebd48050d4c429fa20a0b6/static/js/table.js#L100

But `consumer_tag` should be added, right? 
```suggestion
  keyColumns: ['channel_details', 'consumer_tag`],
```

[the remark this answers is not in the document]
carlhoerberg: I fixed so that objects (and numbers) can be used a key columns in this PR: https://github.com/cloudamqp/lavinmq/pull/529
but the consumer_tag should still be added.

static/js/consumers.js
```diff
@@ -0,0 +1,57 @@
...
+  url,
+  keyColumns: ['channel_details', 'consumer_tag'],
+  interval: 5000,
+  pagination: true,
+  columnSelector: true,
+  search: true
+}
+
+Table.renderTable('table', tableOptions, function (tr, item, all) {
+  console.log(item)
```
carlhoerberg, line 21: left over logging
```suggestion
```

static/js/consumers.js
```diff
@@ -0,0 +1,57 @@
...
+const tableOptions = {
+  url,
+  keyColumns: ['channel_details', 'consumer_tag'],
+  interval: 5000,
+  pagination: true,
+  columnSelector: true,
+  search: true
+}
+
+Table.renderTable('table', tableOptions, function (tr, item, all) {
```
carlhoerberg, line 20: the third parameter to the callback is weather the row is rendered the first time or is being updated. As no columns are being changed (rows are only added or removed), we don't need to re-render the cells.

```suggestion
Table.renderTable('table', tableOptions, function (tr, item, firstRender) {
  if (!firstRender) return
```

static/js/consumers.js
```diff
@@ -0,0 +1,57 @@
...
+  const exclusive = item.exclusive ? '●' : '○'
+  const cancelForm = document.createElement('form')
+  const btn = document.createElement('button')
+  btn.classList.add('btn-primary')
+  btn.type = 'submit'
+  btn.textContent = 'Cancel'
+
+  cancelForm.appendChild(btn)
+  const urlEncodedVhost = encodeURIComponent(item.queue.vhost)
+  console.log(urlEncodedVhost)
```
carlhoerberg, line 35:
```suggestion
```

carlhoerberg (approved):

PR #530, opened 2023-06-18
Allow importing definitions without vhost

carlhoerberg:
And override vhost regardless if a specific vhost is chosen to import to


carlhoerberg:
When exporting specific vhost definitions, we still include the vhost name, rabbitmq does not..

PR #531, opened 2023-06-18
Don't fsync on confirm but syncfs on transaction

carlhoerberg:
Rabbit is removing fsync on confirm and transactions. Was no-op already. `syncfs` synchronizes all files on the filesystem, which is convinent instead of having to track all possible files that would need to be synced when acking/publishing in a transaction.


carlhoerberg:
`lavinmqperf throughput -c 1` is faster because of the less accounting, but `lavinmqperf throughput -t 1` is slower becasue we actually sync now before ack:ing the tx commit.

PR #534, opened 2023-06-21
reroutes to /login when cookie expires

kickster97:
We currently have two error handlers, one "general", in http.js (standardErrorHandler), and one in datasource.js, for table errors. 
I have added some conditions to make it look smooth, so the datasource error will not trigger if it is a 401.. but im wondering if there is a better way? 
let me know what you think

### DIFF
--- a/static/js/datasource.js
+++ b/static/js/datasource.js
@@ -49,7 +49,8 @@ class DataSource {
     sort_reverse: false,
     name: ''
   }
-  constructor(opts) {
+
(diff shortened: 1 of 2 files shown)


spuun (approved):
Looking at this we can probably rethink and handle stuff better in `HTTP.request` so we don't need that 401 check in `DataSource`, but I'm to busy to figure it out now :)

Let's go with this!

PR #537, opened 2023-06-28
use zstd to compress replication data

carlhoerberg:
Doesn't work as `zstd.cr` currently has no way of flushing data


carlhoerberg:
Replaced with lz4 compression instead

PR #538, opened 2023-06-28
add timer for recieving closeok, force close socket after a while

kickster97:


src/lavinmq/client/client.cr
```diff
@@ -36,6 +36,7 @@ module LavinMQ
     @last_sent_frame = RoughTime.monotonic
     rate_stats({"send_oct", "recv_oct"})
     DEFAULT_EX = "amq.default"
+    @closeok_timer : Int64
```
carlhoerberg, line 39: set the default value here, and you then don't need to do it in the constructor
```suggestion
    @closeok_timer = 0i64
```

src/lavinmq/client/client.cr
```diff
@@ -58,6 +59,7 @@ module LavinMQ
                         end
       @log = Log.for "client[vhost=#{@vhost.name} address=#{@remote_address}#{connection_name}]"
       @connected_at = RoughTime.unix_ms
+      @closeok_timer = 0
```
carlhoerberg, line 62:
```suggestion
```

src/lavinmq/client/client.cr
```diff
@@ -143,6 +145,11 @@ module LavinMQ
           if @running
             process_frame(frame)
           else
+            @closeok_timer += 1
+            if @closeok_timer > 500000
```
carlhoerberg, line 149: we force close after 500 000 frames? i think that's a bit much, and what if they send frames very slowly? what about having a timer instead? say 5-10 seconds or so?

src/lavinmq/client/client.cr
```diff
@@ -143,6 +144,10 @@ module LavinMQ
           if @running
             process_frame(frame)
           else
+            if (RoughTime.monotonic - @closeok_timer) > 5.seconds
```
carlhoerberg, line 147: this only works if the client still sends frames, but imagine a client that just consumes but doesn't properly send the CloseOk frame, eg. because it's busy with other things.. One possibility is to start a new fiber when we close the client, that sleeps 5s, checks if the connection is still open and then force closes it.

carlhoerberg (changes requested):

src/lavinmq/client/client.cr
carlhoerberg: The down side is of course that if we closed 1M connections, and thus spawn 1M new fibers, that can be very memory intensive..

carlhoerberg: or we set a `read_timeout` on the socket when we try to gracefully close the client, so the read_loop will raise an TimeoutError if the client is silent.

kickster97: i think my last commit adds it in the way you're thinking, not sure how to test it easily

PR #539, opened 2023-06-29
Create pull_request_template.md

kickster97:
Adds a pull request template to help organize the commit history

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,8 @@
+### WHY are these changes introduced?
+Please fill me in.
+
+### WHAT is this pull request doing?
+Please fill me in.
+
+### HOW can this pull request be tested?
+Specs? Manual steps? Please fill me in.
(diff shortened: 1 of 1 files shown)


nyior:
> Adds a pull request template to help organize the commit history

Maybe a template for commit messages as well? Maybe something along this line: 
- A one line summary of what the change in the commit is about 
- And a brief description of the reason for the change.

The guys at Zulip have a nice [commit discipline](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-messages)

viktorerlingsson:
For PRs, I don't love having to fill in both **what** and **why**. In many cases I feel like that's basically the same, or that the why just is "to get feature x".  

I'd rather just have the what-section and the test-section.

kickster97:
> For PRs, I don't love having to fill in both **what** and **why**. In many cases I feel like that's basically the same, or that the why just is "to get feature x".
> 
> I'd rather just have the what-section and the test-section.

I very much agree with this, often its the same thing written in different words on those two

PR #540, opened 2023-06-29
Drop queue size-details

carlhoerberg:
Size details was an idea to include min and max sizes of messages/queue but it was removed when the in memory queue index was dropped, so dropping the UI and API now too.

To implement it correctly today we have to scan through the whole message store, doable, but memory intensive, and can ruin the disk cache (useful data evicted, useless swapped in).


carlhoerberg:
Is this very very useful @johanrhodin ?

johanrhodin (approved):
Define "very very useful" :)

Sure go ahead and drop it. We can discuss some alternatives (such as keeping a track of "what is the biggest message that has entered a queue" - could be easy to do while checking max message size) or rmqrecover like functionality.

PR #541, opened 2023-06-29
LZ4 replication

carlhoerberg:
LZ4 compress the data over the replication protocol


carlhoerberg:
Should lz4 be optional? in that case we need to add support for compression negotiation in the replication protocol. (could the one of the bits in the start header)

PR #547, opened 2023-07-12
server settings for replication

### LINKED ISSUE #674: Minimum followers and Max Lag setting for replication
Have a setting `min_followers` for the replication server, where we pause publishes if X amount of followers are not connected. 

And a setting `max_lag`, where we pause publishing if not X number of followers have less than max_lag

kickster97:
Adds properties `min_followers` and `max_lag` to the Config class and the values are parsed from `lavinmq.ini`
The values are then conditioned in `vhost.cr` via the replication servers new `wait_for_max_lag`.  You cannot publish to a queue if you have less followers connected than `min_followers` or if min_followers amount of followers have more lag than `max_lag`.  

Fixes #674

Start LavinMQ and configure the values in your config file, connect too few followers and try to publish.

### DIFF
--- a/spec/replication_spec.cr
+++ b/spec/replication_spec.cr
@@ -5,14 +5,12 @@ describe LavinMQ::Replication::Client do
   data_dir = "/tmp/lavinmq-follower"
 
   before_each do
-    puts "client bfe"
     FileUtils.rm_rf data_dir
     Dir.mkdir_p data_dir
     File.write File.join(data_dir, ".replication_secret"), Server.@replicator.@password, 0o400
   end
 
   after_each do
-    puts "client ae"
     FileUtils.rm_rf data_dir
   end
 
@@ -72,7 +70,6 @@ describe LavinMQ::Replication::Server do
   data_dir = "/tmp/lavinmq-follower"
 
   before_each do
-    puts "before_each"
     FileUtils.rm_rf data_dir
     Dir.mkdir_p data_dir
     File.write File.join(data_dir, ".replication_secret"), Server.@replicator.@password, 0o400
@@ -80,7 +77,6 @@ describe LavinMQ::Replication::Server do
   end
 
   after_each do
-    puts "after_each"
     FileUtils.rm_rf data_dir
   end
 
@@ -160,7 +156,6 @@ describe LavinMQ::Replication::Server do
     # end
 
     it "should not publish when max_lag is reached" do
-      pp "config: #{LavinMQ::Config.instance.max_lag}"
       Server.vhosts["/"].declare_queue("test123", true, false)
       repli = LavinMQ::Replication::Client.new(data_dir)
       done = Channel(Nil).new
@@ -172,37 +167,24 @@ describe LavinMQ::Replication::Server do
 
       client : AMQP::Client::Connection? = nil
       spawn(name: "with_channel") do
-        pp 0
         with_channel do |ch, conn|
-          pp "-1"
           client = conn
-          pp "-2"
-          pp 1
           ch.basic_publish_confirm "hello world", "", "test123"
-          Fiber.list { |f| puts f.inspect }
-          pp 2
-
           ch.basic_publish_confirm "hello world2", "", "test123"
-          pp 3
-          # done.send nil
+          done.send nil
         rescue e
         end
-
       end
-      sleep 1.seconds
+
+      select
+      when done.receive
+        fail "should not receive mssage"
+      when timeout(1.seconds)
+        Server.vhosts["/"].queues["test123"].message_count.should eq 1
+      end
+    ensure
       client.try &.close(no_wait: true)
-      Server.vhosts["/"].queues["test123"].message_count.should eq 1
-
-
-    #   select
-    #   when done.receive
-    #     fail "should not receive mssage"
-    #   when timeout(1.seconds)
-    #     pp "7"
-    #     Server.vhosts["/"].queues["test123"].message_count.should eq 1
-    #   end
-    # ensure
-    #   client.try &.close(no_wait: true)
+      repli.try &.close
     end
   end
 end
--- a/spec/replication_spec.cr
+++ b/spec/replication_spec.cr
@@ -125,7 +125,7 @@ describe LavinMQ::Replication::Server do
       when done.receive
         fail "Should not receive message"
       when timeout(0.1.seconds)
-        #ugly hack to release replicator from waiting for lag
+        # ugly hack to release replicator from waiting for lag
         repli = LavinMQ::Replication::Client.new(data_dir)
         done = Channel(Nil).new
         spawn do
--- a/spec/replication_spec.cr
+++ b/spec/replication_spec.cr
@@ -87,8 +87,6 @@ describe LavinMQ::Replication::Server do
         repli.follow("127.0.0.1", LavinMQ::Config.instance.replication_port)
       end
     end
-
-    # repli.closing
   end
 
   describe "min_followers" do
@@ -128,7 +126,14 @@ describe LavinMQ::Replication::Server do
         fail "Should not receive message"
       when timeout(0.1.seconds)
         client.try &.close(no_wait: true)
-        Server.close
+        # Ugly hack to free us
+        # repli = LavinMQ::Replication::Client.new(data_dir)
+        # done = Channel(Nil).new
+        # spawn do
+        #   repli.follow("127.0.0.1", LavinMQ::Config.instance.replication_port)
+        #   done.send nil
+        # end
+        # done.receive
       end
     end
   end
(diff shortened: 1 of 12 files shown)


src/lavinmq/replication/server.cr
```diff
@@ -24,6 +24,7 @@ module LavinMQ
       @followers = Array(Follower).new
       @password : String
       @files = Hash(String, MFile?).new
+      @min_followers = 5_i64
```
carlhoerberg, line 27: Very high default, I think 0 should be the default. And no way it has to be an `Int64`, I don't think many would need more than 2^31 followers. 
```suggestion
      @min_followers = 0
```

src/lavinmq/replication/server.cr
```diff
@@ -98,6 +99,12 @@ module LavinMQ
         end
       end
 
+      def min_followers : Int64
+        @lock.synchronize do
+          @min_followers.dup # for thread safety
+        end
+      end
```
carlhoerberg, line 106: Reading an integer is always thread safe, as integers are copied.

```suggestion
      getter min_followers
```

src/lavinmq/vhost.cr
```diff
@@ -113,6 +113,9 @@ module LavinMQ
       headers = properties.headers
       find_all_queues(ex, msg.routing_key, headers, visited, found_queues)
       headers.try(&.delete("BCC"))
+      puts @replicator.min_followers
+      puts @replicator.followers.size
+      return false if @replicator.followers.size < @replicator.min_followers
```
carlhoerberg, line 118: Make a method on Replicator, which does the check, so it becomes easier to read, and `@replicationr.followers` duplicates the whole array, which is wasteful if you just want to know its size. 
```suggestion
      return false unless @replicator.has_min_followers?
```

carlhoerberg (changes requested):

src/lavinmq/vhost.cr
```diff
@@ -111,6 +111,7 @@ module LavinMQ
       headers = msg.properties.headers
       find_all_queues(ex, msg.routing_key, headers, visited, found_queues)
       headers.delete("BCC") if headers
+      return false unless @replicator.has_min_followers? || @replicator.has_max_lag?
```
carlhoerberg, line 114: I think that instead of returning we should block and wait

carlhoerberg (changes requested):

src/lavinmq/vhost.cr
kickster97: Did a while/sleep thing instead of returning, maybe there is another way to do it? saw that we use select/when in some places to wait, but might only be good for multiple outcome scenarios.. can't find good docs in the crystal api

kickster97:
@viktorerlingsson this is the solution i worked on with @carlhoerberg yesterday. 

I added the wait_for_followers channel, which works but gets into a bad state if followers connect and disconnects a few times..

src/lavinmq/replication/server.cr
```diff
@@ -98,6 +98,19 @@ module LavinMQ
         end
       end
 
+      def wait_for_max_lag
+        until @followers.size >= Config.instance.min_followers
+          select
+          when @wait_for_followers.receive
+          when timeout(1.seconds)
+          end
```
spuun, line 106: I think you can replace this with, or can `@follower.size` or `Config.instance.min_followers` change value while waiting without sending to `@wait_for_followers`?
```suggestion
@wait_for_followers.receive
```

src/lavinmq/replication/server.cr
```diff
@@ -209,21 +221,41 @@ module LavinMQ
...
         def read_acks(socket = @socket) : Nil
           spawn action_loop, name: "Follower#action_loop"
+          @server.wait_for_followers.try_send nil
           loop do
             len = socket.read_bytes(Int64, IO::ByteFormat::LittleEndian)
             @acked_bytes += len
-            @ack.try_send nil
+            if max_lag = Config.instance.max_lag
+              if lag < max_lag
+                while @ack.try_send lag
```
spuun, line 230:
```suggestion
                while @ack.try_send lag
```
Will `lag` ever change while in this loop? Doesn't this mean that the same `lag` will be sent until the channel is clogged? Just `@ack.send lag` (without looping) will do the same?

src/lavinmq/replication/server.cr
```diff
@@ -209,21 +221,41 @@ module LavinMQ
...
+        def wait_for_max_lag
+          if max_lag = Config.instance.max_lag
+            current_lag = lag
+            until current_lag < max_lag
+              select
+              when timeout(1.seconds)
+              when current_lag = @ack.receive
+              end
+              break if @server.followers.includes?(self) == false
+            end
```
spuun, line 247:
```suggestion
            until current_lag < max_lag
              current_lag = @ack.receive
              break if @server.followers.includes?(self) == false
            end
```
Or may `@server.followers` change without receiving on `@ack` first? (I.e is the timeout needed?)

src/lavinmq/vhost.cr
```diff
@@ -112,6 +112,9 @@ module LavinMQ
       headers = msg.properties.headers
       find_all_queues(ex, msg.routing_key, headers, visited, found_queues)
       headers.delete("BCC") if headers
+
+      @replicator.wait_for_max_lag
```
spuun, line 116: Is it here we should wait? Not anywhere in `Replication::Server`?

src/lavinmq/replication/server.cr
kickster97: Yes, currently @follower.size can change without sending to the channel. We recieve on this channel when a follower connects and starts reading.. but other followers can disconnect when ever

src/lavinmq/replication/server.cr
kickster97: Yes, followers can disconnect whenever, if we remove the timeout a follower can disconnect in the middle and we get stuck on waiting for ack.receive

src/lavinmq/vhost.cr
kickster97: worth discussing.. have not thought about having it wait somewhere else, what are you thinking?

spuun: Dunno :)
Maybe in `action_loop`? I just feel that replication should be responsible for it, not `VHost`. But since replication code is new territory for me, I have no clue really.

src/lavinmq/replication/server.cr
kickster97: .send does not work for some reason.. the publishes do not stop even tough the replication sevrer has to few followers.. but this approach with try_send works as it should, so i'll apply that instead!

src/lavinmq/vhost.cr
kickster97: `action_loop` belongs to the follower, and we want to block publishes on the entire server depending on the perspective of the leader server.. maybe we could put it somewhere in the replicator, but i think it is also pretty clear to put it in the vhosts `publish` method.. but maybe unclear since we do not block publishes per vhost, but on the whole server

kickster97: We now block in the replication server instead: Unfortunately we can't place the block directly in the each_follower function, because the replace function is run when launching, and we will have had no chance to connect any followers resulting in deadlock. Therefore i have placed the block (`wait_for_max_lag`) it in the `append` and `delete_file` functions.

src/lavinmq/replication/server.cr
```diff
@@ -209,21 +223,40 @@ module LavinMQ
...
 
+        def wait_for_max_lag
+          if max_lag = Config.instance.max_lag
+            current_lag = lag
+            until current_lag < max_lag
+              select
+              when timeout(1.seconds)
+              when current_lag = @ack.receive
+              end
+              break if @server.followers.includes?(self) == false
```
spuun, line 247: Maybe we can get rid of the timeout here by closing `@ack` in `read_acks` when the socket i closed?

Something like
```suggestion
              break unless current_lag = @ack.receive
```

src/lavinmq/replication/server.cr
```diff
@@ -173,8 +188,7 @@ module LavinMQ
 
       class Follower
         Log = ::Log.for(self)
-
-        getter ack = Channel(Nil).new
+        getter ack = Channel(Int64).new
```
spuun, line 191: I don't think this getter is used, it seem we always access it via the instance variable.

```suggestion
        @ack = Channel(Int64).new
```

kickster97: you're right!

spuun (approved):
👍

spec/replication_spec.cr
```diff
@@ -65,3 +65,133 @@ describe LavinMQ::Replication::Client do
...
+          q = ch.queue("repli")
+          q.publish_confirm "hello world"
+          done.send nil
+        end
+      end
+      select
+      when done.receive
+        fail "Should not receive message"
+      when timeout(0.1.seconds)
+        #ugly hack to release replicator from waiting for lag
```
kickster97, line 128: Had to work around `spec_helpsers` `after_each()` that calls `server.stop` here.. because `server.stop` only calls `replicator.clear`, which will not finish if we are waiting for lag.

PR #557, opened 2023-08-09
StreamQueue: only unmap unused segments

carlhoerberg:
Previous behavior could unmap segments that other consumers were still delivering from, causing segfaults.


carlhoerberg:
Don't like the 60s sleep.. I think we can do better than that. Maybe a channel that consumers can notify when they switch segment.

viktorerlingsson:
> Don't like the 60s sleep.. I think we can do better than that. Maybe a channel that consumers can notify when they switch segment.

A channel that consumers notify when switching segments sounds good to me. 

Should we unmap on consumer disconnect as well?

carlhoerberg:
Ah, when consumer is disconnecting too, good catch, we don't do that in the normal queues either currently.

PR #558, opened 2023-08-10
quick fix for displaying toasts correctly

kickster97:
This is a quick-fix for not rendering toasts for failed requests when declaring queues. 
However Im going to look closer at http.js and try and get the json parse stuff to work. currently we just always return null for ok requests which feels wierd

### DIFF
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -125,10 +125,12 @@ document.querySelector('#declare').addEventListener('submit', function (evt) {
     arguments: Dom.parseJSON(data.get('arguments'))
   }
   HTTP.request('PUT', url, { body })
-    .then(() => {
+    .then((response) => {
       queuesTable.reload()
       evt.target.reset()
-      Dom.toast('Queue ' + queue + ' created')
+      if (!(typeof response === 'undefined')) {
+        Dom.toast('Queue ' + queue + ' created')
+      }
     })
 })
 queuesTable.on('updated', _ => {
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -126,11 +126,10 @@ document.querySelector('#declare').addEventListener('submit', function (evt) {
   }
   HTTP.request('PUT', url, { body })
     .then((response) => {
+      if (response?.is_error) { return }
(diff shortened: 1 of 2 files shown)


static/js/queues.js
```diff
@@ -125,10 +125,12 @@ document.querySelector('#declare').addEventListener('submit', function (evt) {
     arguments: Dom.parseJSON(data.get('arguments'))
   }
   HTTP.request('PUT', url, { body })
-    .then(() => {
+    .then((response) => {
       queuesTable.reload()
       evt.target.reset()
-      Dom.toast('Queue ' + queue + ' created')
+      if ((typeof response !== 'undefined')) {
```
carlhoerberg, line 131: Would this work? 
```suggestion
      if (response) {
```

kickster97: Yes! 
But I'm thinking that this works because the endpoint currently dosn't return anything. We could add an is_error value to the error response just to be explicit with that it is an error (see my latest commit).

static/js/queues.js
```diff
@@ -125,7 +125,8 @@ document.querySelector('#declare').addEventListener('submit', function (evt) {
     arguments: Dom.parseJSON(data.get('arguments'))
   }
   HTTP.request('PUT', url, { body })
-    .then(() => {
+    .then((response) => {
+      if (response?.is_error) { return }
```
viktorerlingsson, line 129: Could we just check for `response.reason` here instead? It should be set on error, and we wouldn't need to set `is_error` as well.

kickster97: not really, we don't return a reason so the response is undefined (if it is a bad request) or null (if it is ok). (see http.js)

viktorerlingsson: Okay. It looks to me like we always set a reason (either to `response.statusText` or `json.reason`), but maybe statusText doesn't always exist.

kickster97: Oh yes, sorry in `http.js` we set the reason but the json conversion dosn't work for this prticular endpoint because we don't return anything from the endpoint `put api/queues`

PR #560, opened 2023-08-11
Add playwright for frontend specs

spuun:
This will add [playwright](https://playwright.dev/) tests for some of the frontend code. The goal is to verify that all requests are done as expected and that data is handled correct (e.g. rendered). 

Start lavinmq and run

```
# npm install @playwright/test
# npx playwright install --with-deps
# BASE_URL=http://locahost:15672 npx playwright test --reporter=list
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,8 +145,8 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install -y make libevent-2.1-7
+          apt-get update
+          apt-get install -y make libevent-2.1-7
 
       - uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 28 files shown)


.github/workflows/ci.yml
```diff
@@ -134,6 +135,60 @@ jobs:
...
+  path_filter:
+    name: Filters for fronted specs
+    runs-on: ubuntu-latest
+    outputs:
+      views: ${{ steps.changes.outputs.views }}
+      js: ${{ steps.changes.outputs.js }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
```
dentarg, line 147: Seems like a well regarded action 👍  seeing both Sentry and Google Chrome is using it: https://github.com/dorny/paths-filter ([though Sentry forked it](https://github.com/getsentry/paths-filter)) :)

spuun: Yeah, that's what I noticed and why I allowed myself to use it :)

kickster97 (approved):

PR #564, opened 2023-08-16
send dispatch to lavinmq-website on new release

viktorerlingsson:
Sends dispatch action to lavinmq-website when new release is created, so that lavinmq-website automatically can update it's latest stable version info. 

To be used with: https://github.com/84codes/lavinmq-website/pull/48

Needs a token to be able to communicate with lavinmq-website repo


dentarg:
We also want to build new docs (https://github.com/84codes/lavinmq-docs) on a new release, so here's an idea:

Send the dispatch from this repo to https://github.com/84codes/lavinmq-docs and in the docs repo, we send a dispatch to the website repo, to trigger a new build. The website repo can continue to ingest the version file (https://docs.lavinmq.com/latest_release.json) from the docs repo at build time.

WDYT?

viktorerlingsson:
> We also want to build new docs (https://github.com/84codes/lavinmq-docs) on a new release, so here's an idea:
> 
> Send the dispatch from this repo to https://github.com/84codes/lavinmq-docs and in the docs repo, we send a dispatch to the website repo, to trigger a new build. The website repo can continue to ingest the version file (https://docs.lavinmq.com/latest_release.json) from the docs repo at build time.
> 
> WDYT?

Interesting! 
Yeah, we should definitely trigger docs build from here as well, but I'm not sure if I like the website having to depend on the docs. What if docs for some reason is unavailable during website build for example, or docs fail to build.

I think dispatching the version number from here should be more resilient, no?

dentarg:
> I'm not sure if I like the website having to depend on the docs. What if docs for some reason is unavailable during website build for example, or docs fail to build.

I wouldn't worry about that too much, the docs site is hosted on GitHub Pages, like the website is, so they would probably be unavailable at the same time. If the docs can't build, we would likely want to address that ASAP anyway.

> I think dispatching the version number from here should be more resilient, no?

We do get a longer chain with my idea (lavinmq -> docs -> website) but I like it because it means we can continue with the current solution of consuming the versions file at (website) build time, remove the javascript code about it, and avoid the need to make a ([automated commit](https://github.com/84codes/lavinmq-website/pull/48/files#diff-8d4c99794a132146cc00f3b4860b4896f1627711754c64088d1005786ab3f0d9R25)) to the website repo on every new release.

viktorerlingsson:
> We do get a longer chain with my idea (lavinmq -> docs -> website) but I like it because it means we can continue with the current solution of consuming the versions file at (website) build time, remove the javascript code about it, and avoid the need to make a to the website repo on every new release.

Maybe we could get the best of both worlds! If we can keep track of the version number in the action in website, we could trigger a build and use that at build time, to update the version that way. And then remove the need for both JS and the commits. WDYT? I can do a little testing next week.

And do a separate dispatch to docs.

dentarg:
Not sure I follow that latest suggestion, but I still think it is cleaner to dispatch to the docs repo and persist the version number there – it already need to persist the updated docs and that repo is never updated manually by humans, in contrast to the website repo.

If we want to avoid the dispatch to the website repo, the website could perhaps fetch https://docs.lavinmq.com/latest_release.json with JavaScript? (there should be no rate limits on that)

dentarg:
There might be a point in persisting the version number in the website repo, it could be kinda nice have the bumps recorded in the git history along with all other changes. And maybe the docs repo shouldn't be "abused" to host the version file. Added some feedback in https://github.com/84codes/lavinmq-website/pull/48 that should help us get there.

PR #566, opened 2023-08-21
PoC: Add load simulator

viktorerlingsson:

WIP/PoC - Just an idea! 

Adds a load simulator to lavinmqperf. Creates nr of queues, then pubs and consumes from those queues at a specified rate and message size with a specified number of publishers/consumers per queue. Could be helpful for testing what kind of resources are used in specific scenarios. (For example customers wanting to know what plan they should be using)

Mostly copied stuff from throughput. Should definitely be refactored if we continue with this idea. 

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -15,7 +15,7 @@ class Perf
 
   def initialize
     @parser = OptionParser.new
-    @banner = "Usage: #{PROGRAM_NAME} [throughput | bind-churn | queue-churn | connection-churn | connection-count] [arguments]"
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Have you checked out https://github.com/rabbitmq/RabbitTestTool/tree/main/benchmark ? it's pretty cool for testing pretty advanced topoligies.

viktorerlingsson:
> Have you checked out https://github.com/rabbitmq/RabbitTestTool/tree/main/benchmark ? it's pretty cool for testing pretty advanced topoligies.

No, had not seen that before. Interesting! Seems like a competent tool. 

This simpler load testing in lavinmqperf might still be interesting though? As a simpler tool that we will have easily available since it comes with lavin. (It already pretty much works as is, just needs to be polished a bit)

PR #567, opened 2023-08-23
check if disk is full before creating queues and users

viktorerlingsson:
Check if disk is full before creating queues and users, and send `PreconditionFailed` if disk is full (if flow is stopped). 

Also slightly raises the default free disk value to `3 * segment_size` to give Lavin more time to stop flow. 

https://trello.com/c/2iP4j8ca/316-catch-disk-full-errors-instead-of-crashing

Makes sure LavinMQ doesn't fill up disk and crash when creating queues and/or users. 

Fill up disk, try creating queue/user.

### DIFF
--- a/src/lavinmq/server.cr
+++ b/src/lavinmq/server.cr
@@ -385,7 +385,7 @@ module LavinMQ
     getter stats_system_collection_duration_seconds = Time::Span.new
 
     private def control_flow!
-      if is_disk_full?
+      if disk_full?
         if flow?
           @log.info { "Low disk space: #{@disk_free.humanize}B, stopping flow" }
           flow(false)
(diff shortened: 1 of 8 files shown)


src/lavinmq/server.cr
```diff
@@ -385,19 +385,27 @@ module LavinMQ
     getter stats_system_collection_duration_seconds = Time::Span.new
 
     private def control_flow!
-      if @disk_free < 2_i64 * Config.instance.segment_size || @disk_free < Config.instance.free_disk_min
+      if disk_full?
         if flow?
           @log.info { "Low disk space: #{@disk_free.humanize}B, stopping flow" }
           flow(false)
         end
       elsif !flow?
```
kickster97, line 393: Still think the naming is a bit confusing here.. if we don't have flow (flow is stopped because we have a high disk usage?) we want to log "Not low..." ?

viktorerlingsson: I guess we could make it more explicit

the `elsif !flow?` implicitly means `if !disk_full? && !flow?` because of the if above. So if disk is not full but flow is set to false, that means disk usage has been reduced below the limit, and flow should be resumed.

kickster97: makes sense :)

kickster97 (approved):

PR #570, opened 2023-09-01
Dont write all binds to defs

spuun:
definitions.amqp can grow indefinitely because we write all bind and unbind frames to it. This will add guard against that behavior.

Add the same binding multiple times and verify it's not written to definitions.amqp. If log level is debug it will be logged if written to file.

### DIFF
--- a/src/lavinmq/exchange/default.cr
+++ b/src/lavinmq/exchange/default.cr
@@ -6,11 +6,11 @@ module LavinMQ
       "direct"
     end
 
-    def bind(destination, routing_key, headers = nil) : Bool
+    def bind(destination, routing_key, headers = nil)
       raise "Access refused"
     end
 
-    def unbind(destination, routing_key, headers = nil) : Bool
+    def unbind(destination, routing_key, headers = nil)
       raise "Access refused"
     end
 
--- a/src/lavinmq/exchange/default.cr
+++ b/src/lavinmq/exchange/default.cr
@@ -6,19 +6,11 @@ module LavinMQ
       "direct"
     end
 
-    def has_binding?(destination : Queue, routing_key : String, headers : Hash(String, AMQP::Field)?)
-      false
-    end
-
-    def has_binding?(destination : Exchange, routing_key : String, headers : Hash(String, AMQP::Field)?)
-      false
-    end
-
-    def bind(destination, routing_key, headers = nil)
+    def bind(destination, routing_key, headers = nil) : Bool
       raise "Access refused"
     end
 
-    def unbind(destination, routing_key, headers = nil)
+    def unbind(destination, routing_key, headers = nil) : Bool
       raise "Access refused"
     end
 
--- a/src/lavinmq/exchange/default.cr
+++ b/src/lavinmq/exchange/default.cr
@@ -6,6 +6,14 @@ module LavinMQ
       "direct"
     end
 
+    def has_binding?(destination : Queue, routing_key : String, headers : Hash(String, AMQP::Field)?)
+      false
+    end
+
+    def has_binding?(destination : Exchange, routing_key : String, headers : Hash(String, AMQP::Field)?)
+      false
+    end
+
     def bind(destination, routing_key, headers = nil)
       raise "Access refused"
     end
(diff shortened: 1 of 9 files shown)


spuun:
We should probably add some kind of occasional compactation during runtime. Count "deletes" and trigger compaction after X deletions or something, so prevent the file from growing on e.g. high queue churn.

viktorerlingsson:
> We should probably add some kind of occasional compactation during runtime. Count "deletes" and trigger compaction after X deletions or something, so prevent the file from growing on e.g. high queue churn.

Sounds good!

src/lavinmq/exchange/default.cr
```diff
@@ -6,6 +6,14 @@ module LavinMQ
       "direct"
     end
 
+    def has_binding?(destination : Queue, routing_key : String, headers : Hash(String, AMQP::Field)?)
+      false
```
carlhoerberg, line 10:
```suggestion
      routing_key == destination.name
```

src/lavinmq/vhost.cr
```diff
@@ -329,11 +331,13 @@ module LavinMQ
         when AMQP::Frame::Queue::Bind
           x = @exchanges[f.exchange_name]? || return false
           q = @queues[f.queue_name]? || return false
+          return false if x.has_binding?(q, f.routing_key, f.arguments.to_h)
```
carlhoerberg, line 334: Cant Exchange#bind return false if the binding already existed? Now you do f.arguments.to_h twice and introduce a new method that might not be needed

spuun: Yes, that was my initial thought. Dunno why I didn't go with that.

src/lavinmq/exchange/default.cr
spuun: I guess I did return `false` here because `bind` and `unbind` always raises `"Access refused"`, so in it can't have any binding? But maybe it would make more sense to always return true, since no bindings can be added we don't need to write it to any definition file.

However, will change so `bind`/`unbind` returns `true`/`false` instead of having `has_binding?`.

src/lavinmq/exchange/fanout.cr
```diff
@@ -7,23 +7,27 @@ module LavinMQ
     end
 
     def bind(destination : Queue, routing_key, headers = nil)
-      @queue_bindings[{routing_key, nil}] << destination
+      ret = @queue_bindings[{routing_key, nil}].add? destination
       after_bind(destination, routing_key, headers)
+      ret
```
carlhoerberg, line 12: Should `after_bind` be executed even if the binding was already there? 

```suggestion
      if @queue_bindings[{routing_key, nil}].add? destination
        after_bind(destination, routing_key, headers)
        true
      end
```

src/lavinmq/vhost.cr
```diff
@@ -300,12 +300,12 @@ module LavinMQ
         when AMQP::Frame::Exchange::Bind
           src = @exchanges[f.source]? || return false
           dst = @exchanges[f.destination]? || return false
-          src.bind(dst, f.routing_key, f.arguments.to_h)
+          return false unless src.bind(dst, f.routing_key, f.arguments.to_h)
           store_definition(f) if !loading && src.durable && dst.durable
```
carlhoerberg, line 304: Doesn't `false` indicate that the client will get a PreconditionFailed back? as in the case above if the exchange is missing? Should we just not store the defintion instead? 

```suggestion
          if src.bind(dst, f.routing_key, f.arguments.to_h)
            store_definition(f) if !loading && src.durable && dst.durable
          end
```

src/lavinmq/exchange/fanout.cr
spuun: Maybe, I just didn't want to change the behavior. Seems to be federation link that listens for bind/unbind events to forward them to the upstream.

src/lavinmq/vhost.cr
spuun: Maybe :) It wasn't obvious on what the return value meant (or how it's used). At least not used in `Client`. Maybe we should use enums or symbols here and there instead to get more descriptive return values?

spuun: I think i chose to return false because we did so for `Queue::Declare` if the queue already existed.

PR #571, opened 2023-09-05
Compact definitions during runtime

spuun:
With high churns (e.g. tons of short-lived queues) the definitions file can grow indefinitely. This will add compaction every 10000th "delete" frame. (10k was chosen completely arbitrary.)

Run specs, or launch lavinmq and create and delete a queue more than 10k times.

### DIFF
--- a/src/lavinmq/vhost.cr
+++ b/src/lavinmq/vhost.cr
@@ -19,8 +19,6 @@ module LavinMQ
     include SortableJSON
     include Stats
 
-    DEFINITIONS_DIRT_COMPACT_THREASHOLD = 10_000
-
     rate_stats({"channel_closed", "channel_created", "connection_closed", "connection_created",
                 "queue_declared", "queue_deleted", "ack", "deliver", "get", "publish", "confirm",
                 "redeliver", "reject", "consumer_added", "consumer_removed"})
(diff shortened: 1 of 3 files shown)


src/lavinmq/vhost.cr
```diff
@@ -611,15 +614,24 @@ module LavinMQ
...
       bytes = frame.to_slice
       @definitions_file.write bytes
       @replicator.append @definitions_file_path, bytes
       @definitions_file.fsync
+      if dirty
+        @definitions_dirt_counter += 1
+        # By doing equality comparision we'll only start one fiber
+        # without having to keep track if it's started or not
+        if @definitions_dirt_counter == DEFINITIONS_DIRT_COMPACT_THREASHOLD
+          spawn(name: "VHost #{name} definitions compact") { compact! }
```
carlhoerberg, line 632: why in a separate fiber? makes it potentially thread unsafe

spuun: I think one reason was that `store_definition` is called from a  `@definitions_lock.synchronize` context, and that `compact!` also want the lock, and the other was to not block the current fiber with this (even though it's probably fast, and it will block by holding the lock).

PR #573, opened 2023-09-13
Use Log#emit and Log::Metadata

carlhoerberg:
Saves Log instances and uses Metadata instead, somewhat repetitive but the "native" way.

### DIFF
--- a/src/lavinmq/client/amqp_connection.cr
+++ b/src/lavinmq/client/amqp_connection.cr
@@ -7,7 +7,6 @@ module LavinMQ
 
     def self.start(socket, connection_info, vhosts, users) : Client?
       remote_address = connection_info.src
-      Log.context.set(address: remote_address.to_s)
       socket.read_timeout = 15.seconds
       if confirm_header(socket)
         if start_ok = start(socket)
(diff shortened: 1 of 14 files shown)


baelter:
Can't we wrap this with our own `Log`, something like 
```crystal
class LavinMQ::Log < ::Log
  property metadata
  def debug
     msg = yield
     super.debug &.emit msg, metadata
  end
  ...
end

Log = LavinMQ::Log.for(self)
Log.metadata = ...
Log.debug { "yada" }
```
?

spuun (commented):
Maybe the `EntityLog` in #502 is over-engineered with its expand methods and stuff should be more explicit, but i think using a Struct to wrap the Log instance and @metadata is nice so you can use the same syntax?

spuun:
> Can't we wrap this with our own `Log`, something like
> 
> ```crystal
> class LavinMQ::Log < ::Log
>   property metadata
>   def debug
>      msg = yield
>      super.debug &.emit msg, metadata
>   end
>   ...
> end
> 
> Log = LavinMQ::Log.for(self)
> Log.metadata = ...
> Log.debug { "yada" }
> ```
> 
> ?
The EntityLog here does that:
https://github.com/cloudamqp/lavinmq/blob/logging-sources/src/lavinmq/logging.cr

carlhoerberg:
Getting rid of `@log` that we might think be a memory leak. And doesn't do any monkey patching of anything, uses the "native" way of logging metadata with `Log`.

PR #574, opened 2023-09-22
Bugfix corrupt dlx headers

viktorerlingsson:
Fixes a bug where messages that where dead-lettered multiple times would cause memory errors because of an issue in amq-protocol.cr - https://github.com/cloudamqp/amq-protocol.cr/pull/14

Run the spec.

### DIFF
--- a/spec/dlx_spec.cr
+++ b/spec/dlx_spec.cr
@@ -23,7 +23,7 @@ describe "Dead lettering" do
       msg = wait_for { q.get }
 
       x_death = msg.properties.headers.not_nil!["x-death"].as(Array(AMQ::Protocol::Field))
-      x_death.inspect
+      x_death.inspect.should be_a(String) # checks that message and headers can be read
       x_death.size.should eq 2
       x_death[0].as(AMQ::Protocol::Table)["queue"].should eq q_delayed.name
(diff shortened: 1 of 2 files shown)


spec/dlx_spec.cr
```diff
@@ -0,0 +1,32 @@
...
+        {"x-message-ttl" => 1, "x-dead-letter-exchange" => "", "x-dead-letter-routing-key" => q_name}
+      ))
+      q = ch.queue(q_name)
+
+      x = ch.default_exchange
+      x.publish_confirm("ttl", q_delayed_2.name)
+      msg = wait_for { q.get }
+
+      x_death = msg.properties.headers.not_nil!["x-death"].as(Array(AMQ::Protocol::Field))
+      x_death.inspect
```
spuun, line 26:
```suggestion
```

viktorerlingsson: I kinda want that there so that the code would actually crash with amq-protocol 1.1.9. Without it this spec just fails the comparison, and I think it's safer to also know that we can output the whole array.

viktorerlingsson: We could change it to `x_death.inspect.should be_a(String)` so it looks more like a spec I guess :)

PR #575, opened 2023-09-25
Bugfix headers exchange wont match tables

spuun:
Header exchange couldn't match on tables because a mix of `AMQ::Protocol::Table` and `Hash(String, AMQ::Protocol::Field)`. This will change to use `AMQ::Protocol::Table` everywhere.

Run specs.

### Note

Depends on https://github.com/cloudamqp/amq-protocol.cr/pull/15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - A bug in delay exchanges caused messages to be routed to x-dead-letter-exchange instead of bound queues. It also ruined any dead lettering headers.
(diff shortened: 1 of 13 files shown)


spuun:
> Good, but wondering if there can be an issue where `AMQP::Table` is using bytes from disk (`MFile`), eg. from the `definitions.amqp` and then the file is unmapped, causing a segfault.

Ah, right! Will investigate!

spuun:
- `definitions.amqp` is a `File`, so data will be copied
-  frames read from socket will copy data

kickster97 (approved):

viktorerlingsson (approved):
Looks good to me

carlhoerberg (approved):

PR #577, opened 2023-09-26
Yield fiber while purging messages

spuun:
Purging queues with a lot of message takes a while making LavinMQ blocked for other operations. This will make the fiber yield ever 16 384 message.

Run specs


viktorerlingsson (approved):
Looks good to me!

kickster97 (approved):

carlhoerberg:
I would not make this configurable, none except core developers would ever change it. Make a choice eg. 2^14, as in many other places in the code.

PR #578, opened 2023-09-29
Consumer arg support for shovels

viktorerlingsson:
Shovels handle `src-consumer-args` to enable them to handle stream queues. 

Create a shovel for a stream queue via API with x-stream-offset

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -41,7 +41,8 @@ module LavinMQ
 
       def initialize(@name : String, @uri : URI, @queue : String?, @exchange : String? = nil,
                      @exchange_key : String? = nil,
-                     @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH, @ack_mode = DEFAULT_ACK_MODE,
+                     @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH,
+                     @ack_mode = DEFAULT_ACK_MODE, @consumer_args = {} of String => JSON::Any || nil,
                      direct_user : User? = nil)
         @tag = "Shovel[#{@name}]"
         cfg = Config.instance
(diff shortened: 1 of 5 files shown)


src/lavinmq/shovel/shovel.cr
```diff
@@ -41,7 +41,8 @@ module LavinMQ
 
       def initialize(@name : String, @uri : URI, @queue : String?, @exchange : String? = nil,
                      @exchange_key : String? = nil,
-                     @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH, @ack_mode = DEFAULT_ACK_MODE,
+                     @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH,
+                     @ack_mode = DEFAULT_ACK_MODE, @consumer_args = {} of String => JSON::Any || nil,
```
spuun, line 45: Any reason to have `@consumer_args` as a instance variable? Looks like it's only used to populate `@args`?

And the `|| nil` part seems odd. Will it ever do anything? Default value is empty hash. Or do you want `@consumer_args : Hash(String, JSON::Any)? = nil`?

src/lavinmq/shovel/shovel.cr
```diff
@@ -60,6 +61,10 @@ module LavinMQ
         if @queue.nil? && @exchange.nil?
           raise ArgumentError.new("Shovel source requires a queue or an exchange")
         end
+        @args = AMQ::Protocol::Table.new
+        @consumer_args.try &.each do |k, v|
+          @args[k] = v
```
spuun, line 66: How will `JSON::Any` be handled from here? Should it be `v.as_s?` or something?

src/lavinmq/shovel/shovel.cr
viktorerlingsson: Yeah, looks much better that way 👍

src/lavinmq/shovel/shovel.cr
viktorerlingsson: I don't `v.as_s?` is needed for it to work, but it's probably best to be specific

viktorerlingsson:
Moved the definitions upload stuff to a separate PR to keep this PR more focused

src/lavinmq/shovel/shovel.cr
spuun: Oh, `AMQ::Protocol::Table` accepts `JSON::Any` as value. Didn't know that (and I guess it compiled for you 🤦).

carlhoerberg (approved):

PR #583, opened 2023-10-24
Bugfix: DLX doesn't update timestamp on republish

spuun:

When a message was dead letter the timestamp wasn't updated. This caused TTL to misbehave if a message was dead lettered to another queue with TTL:

1. Declare queue `q1` with `x-message-ttl = 100` and `x-dead-letter-routing-key = q2`
2. Declare queue `q2` with `x-message-ttl = 100` and `x-dead-letter-routing-key = q3`
3. Publish message to `q1`.

After 100ms the message is dead lettered to `q2` (via default exchange). Because the message timestamp isn't updated the message will expire immediately in `q2` and dead lettered to `q3`.

This PR will fix this by setting a new timestamp on the message when it's published by the dead lettering.

Run specs


viktorerlingsson (approved):
LGTM!

carlhoerberg (approved):

Issue #596, opened 2023-11-14
Please enable build target Debian 12 (bookworm) in the packagecloud APT repository


nifr:
Hey lavinmq folks 👋 
we're totally amazed by `lavinmq` and have been using it with great joy during development.

Thanks for all the hard work you put into this project and for making it available to the public ❤️ 

As we're currently switching our development containers from Debian 11 (bullseye) to 12 (bookworm) we sadly realized that we can't use your packagecloud APT repository to install `lavinmq` at the moment.

Would it be possible to enable Debian bookworm as a target on packagecloud?

Thank you so much in advance and have a great day 😎 

Note: There is a dependency conflict for `libssl1.1` on `bookworm` if we attempt to use the `bullseye` apt repository instead.

### Details

**Background** Debian 12 (bookworm) was released on `2023-06-10`.

---

**Given** I want to install `lavinmq` via apt on Debian 12 (bullseye) and architecture `arm64`
**When** I follow the installation instructions in [README.md#debianubuntu](https://github.com/cloudamqp/lavinmq/blob/main/README.md#debianubuntu)
**And** I update the list of available packages with `apt-get update`
**Then** the command fails because of a missing `bookworm` release with the following error:

<details>
<summary>APT Error: [..] does not have a Release file</summary>

```
Ign:11 https://packagecloud.io/cloudamqp/lavinmq/debian bookworm InRelease
Err:12 https://packagecloud.io/cloudamqp/lavinmq/debian bookworm Release
  404  Not Found [IP: 54.183.173.4 443]
Reading package lists... Done
E: The repository 'https://packagecloud.io/cloudamqp/lavinmq/debian bookworm Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

</details>

---

**Given** I use the `bullseye` apt repository on packagecloud (instead of `bookworm`)
**When** I update the package list with `apt-get update`
**And** I try to install `lavinmq` with `apt-get install -y lavinmq`
**Then** the installation fails because of broken dependencies with the following error:

<details>
<summary>APT Error: Depends on libssl1.1</summary>

```
Starting pkgProblemResolver with broken count: 1
Starting 2 pkgProblemResolver with broken count: 1
Investigating (0) lavinmq:arm64 < none -> 1.2.5-1 @un puN Ib >
[...]

PR #597, opened 2023-11-14
Build for Debian 12 (bookworm)

dentarg:

Build Debian package for Debian 12, to address https://github.com/cloudamqp/lavinmq/issues/596

Test the package that CI builds :)


dentarg:
Merging this as we got a success report at https://github.com/cloudamqp/lavinmq/issues/596#issuecomment-1810406268 and I've tested it too on amd64 :)

PR #600, opened 2023-11-22
filter out unsupported policies

kickster97:
Adds an array `SUPPORTED_POLICIES` and filters on it `Policy`'s `merge_definitions`

try and add an unsupported policy, eg. HA policy, and view the effective policy definition on the queue

### DIFF
--- a/src/lavinmq/policy.cr
+++ b/src/lavinmq/policy.cr
@@ -10,10 +10,10 @@ module LavinMQ
   end
 
   class Policy
-    SUPPORTED_POLICIES = ["max-length", "max-length-bytes", "message-ttl", "expires", "overflow",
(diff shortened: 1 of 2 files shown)


src/lavinmq/policy.cr
```diff
@@ -10,6 +10,10 @@ module LavinMQ
   end
 
   class Policy
+    SUPPORTED_POLICIES = ["max-length", "max-length-bytes", "message-ttl", "expires", "overflow",
```
carlhoerberg, line 13: Make it a tuple, as it's static

kickster97: Yes, fixed!

spuun (commented):
A spec would be nice!

kickster97:
> A spec would be nice!

@spuun added a spec :)

spuun (approved):

PR #601, opened 2023-11-27
Bugfix: DLX parameters aren't updated when the messages is expired

snichme:

The effect of this bug is that `count` on a DLX message is counted up to 2 but never higher. It's set to 1 when the headers are created, we get the frist count and increase that but instead of updating in place the new count is added as a new field in the table.

AMP::Table doesn't support `Symbol` for it's `#delete` method which results in when using `#merge!` the new pairs are added onto the backing IO for each call.

The result is that after three expires the table looks like this:
```
AMQ::Protocol::Table(@queue="alarms.notifications", @reason="rejected", @exchange="amq.topic", @count=1, @time=2023-11-27 12:46:51.0 UTC, @routing-keys=["alarms.notify"], @count=2, @time=2023-11-27 12:47:01.0 UTC, @routing-keys=["alarms.notify"], @count=2, @time=2023-11-27 12:47:11.0 UTC, @routing-keys=["alarms.notify"])
```

If instead each key is set it overrides the previous value as expected.

Not sure how to write specs for this 

It can be tested manually by publish a message to a queue that is setup to when the message is rejected, it ends up in a DLX queue. Setup a TTL on the DLX queue so that when the TTL expires the messages is published back to the original queue, when that happens, look at the messages headers `x-death`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Don't update user's password hash if given password is the same as current [#586](https://github.com/cloudamqp/lavinmq/pull/586)
 - Remove old segments in the background for stream queues [#608](https://github.com/cloudamqp/lavinmq/pull/608)
+- A bug prevented `count` from being updated correctly in `x-death` headers [#601](https://github.com/cloudamqp/lavinmq/pull/601)
 
 ## [1.2.5] - 2023-11-06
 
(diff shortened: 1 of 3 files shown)


spuun:
I can try to create a spec or two for this.

I've working on fix for `amq-protocol` here: https://github.com/cloudamqp/amq-protocol.cr/pull/16

viktorerlingsson (approved):
LGTM

kickster97 (approved):

PR #603, opened 2023-11-29
Ameba rule updates

spuun:
Update our ameba exceptions to match latest ameba:

- `DocumentationAdmonition` had moved from `Lint` to `Documentation`
- `Naming/BlockParameterName` has been introduced. Added `ch` and `q` as allowed parameter names as I think that makes sense in lavinmq.

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -27,7 +27,7 @@ Lint/NotNil:
   Enabled: true
   Severity: Warning
 
-Lint/DocumentationAdmonition:
+Documentation/DocumentationAdmonition:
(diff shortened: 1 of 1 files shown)


spuun:
This won't remove all complains about block parameter names, but I think some names could be better.

viktorerlingsson (approved):

carlhoerberg:
Fixed in main: https://github.com/cloudamqp/lavinmq/blob/c23a025995e1ece195da2f363cf2e5e9bf7e4d78/.ameba.yml

PR #605, opened 2023-12-05
only log if the permission is changed

kickster97:
Don't log thet the permission is updated if the permission is the same that we already have. 
I tried opting for an early return.. but it looks like we do this on launch, and if we don't have a default user we can't create one. 

Oskar also gave his two cents, and makes the case that we could change these permissions updates in ssh-monitor.. and hat we should log because we do the update. 

try to update permissions without changing them

### DIFF
--- a/src/lavinmq/user_store.cr
+++ b/src/lavinmq/user_store.cr
@@ -45,9 +45,11 @@ module LavinMQ
 
     def add_permission(user, vhost, config, read, write)
       perm = {config: config, read: read, write: write}
+      unless perm == @users[user].permissions[vhost]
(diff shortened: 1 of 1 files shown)


viktorerlingsson (commented):
Looks good to me. 

Maybe we could also move these rows inside the `unless`? If the permissions are the same there shouldn't be a need to update and save them. 
```
@users[user].permissions[vhost] = perm
@users[user].invalidate_acl_caches
save!
```

viktorerlingsson:
> I tried this, but if we don't have a user this wont work.. and its run at launch when we don't have one yet so then it fails :)

Okay!

viktorerlingsson (approved):

PR #608, opened 2023-12-07
Drop overflow in loop for stream_queue to make sure max-age is respected

viktorerlingsson:

Replaces `unmap_unused_segments_loop` with `unmap_and_remove_segments_loop`. The new function also removes segments from stream queues where all messages are past max-age. 

Create a stream queue with max-age (ex. 60s). 
Publish more messages than fits in one segment file. 
Wait the max-age to pass and for the loop to run.
Check that the older segment file has been deleted.

### DIFF
--- a/src/lavinmq/queue/stream_queue.cr
+++ b/src/lavinmq/queue/stream_queue.cr
@@ -8,7 +8,7 @@ module LavinMQ
                    @exclusive = false, @auto_delete = false,
                    @arguments = AMQP::Table.new)
       super
-      spawn unmap_unused_segments_loop, name: "StreamQueue#unmap_unused_segments_loop"
+      spawn unmap_and_remove_segments_loop, name: "StreamQueue#unmap_and_remove_segments_loop"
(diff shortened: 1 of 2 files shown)


src/lavinmq/queue/stream_queue.cr
```diff
@@ -163,6 +163,7 @@ module LavinMQ
       @msg_store_lock.synchronize do
         stream_queue_msg_store.unmap_segments(except: used_segments)
       end
+      stream_queue_msg_store.drop_overflow
```
carlhoerberg, line 166: probably drop segments before trying to unmap them, otherwise double work. And all access to the msg_store has to happen inside its lock. 

```suggestion
        stream_queue_msg_store.drop_overflow
        stream_queue_msg_store.unmap_segments(except: used_segments)
      end
```

carlhoerberg (changes requested):

carlhoerberg (approved):

PR #616, opened 2023-12-18
Use persisted sorting/query information when loading table

### LINKED ISSUE #599: Persist sorting information after deleting queue
While sorting is already persisted via URL, it would be great to persist it via LocalStorage too.

Case:
- Sort queues
- Enter detailed view
- Delete queue
- **Sorting is dropped**

carlhoerberg:

Fixes #599

Open a table view, eg queues, sort on eg. ready msgs, click on the Queue menu item again, the sorting should be the same.

### DIFF
--- a/static/js/datasource.js
+++ b/static/js/datasource.js
@@ -75,9 +75,8 @@ class DataSource {
         } catch (e) {
           console.error(`Failed to load cached query state: ${e}`, cachedState)
         }
-      } else {
-        this._setStateFromHash()
       }
+      this._setStateFromHash()
       window.addEventListener('hashchange', evt => {
         this._setStateFromHash()
         this.reload({ updateState: false })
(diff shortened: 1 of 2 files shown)


spuun (commented):
I know RabbitMQ mgmt works like this but IMO this goes against how "web" works. E.g. if i click "Queues" in the menu I expect to end up with a "fresh view", not a view with some old filters active. With this change there is no way to "reset" the view without fiddling with the url in the url bar.

However, the referenced issue should be fixed.

Maybe we can change or URLs in the menu to contain a default filter? Like, add `#page=1`? Or is it better with some kind of reset filter button?

carlhoerberg:
I don't know if that's what is "expected", if i pull up a google sheet i expect the sheet to be sorted the way when i left it.

carlhoerberg:
I'll merge this as I think it makes the situation better than what it currently is at least

PR #617, opened 2023-12-18
Consumer timeout

carlhoerberg:

Implements support for consumer timeouts.

Spec


src/lavinmq/vhost.cr
```diff
@@ -58,6 +58,18 @@ module LavinMQ
...
+
+    private def check_consumer_timeouts_loop
+      loop do
+        sleep Config.instance.consumer_timeout_loop_interval
+        @connections.each do |c|
+          c.channels.each_value do |ch|
+            ch.check_consumer_timeout
+          end
+        end
+      end
```
spuun, line 72: We should make this fiber close down gracefully when vhost is closed. It's most likely necessary for MT.

PR #619, opened 2023-12-21
Use DirectDispatcher for logging

viktorerlingsson:
Sets up DirectDispatcher for logging so we always log directly, without manually having to yield Fibers. 

Also adds more logging to the startup procedure, making it more verbose.

Relates to the bug that was fixed in https://github.com/cloudamqp/lavinmq/pull/565

https://trello.com/c/FUxNAzYu/402-no-logs-during-startup

Manual for now

### DIFF
--- a/src/lavinmq/launcher.cr
+++ b/src/lavinmq/launcher.cr
@@ -84,7 +84,7 @@ module LavinMQ
       backend = if ENV.has_key?("JOURNAL_STREAM")
                   ::Log::IOBackend.new(io: log_file, formatter: JournalLogFormat, dispatcher: ::Log::DirectDispatcher)
                 else
-                  ::Log::IOBackend.new(io: log_file, formatter: StdoutLogFormat)
+                  ::Log::IOBackend.new(io: log_file, formatter: StdoutLogFormat, dispatcher: ::Log::DirectDispatcher)
                 end
 
       broadcast_backend.append(backend, @config.log_level)
--- a/src/lavinmq/launcher.cr
+++ b/src/lavinmq/launcher.cr
@@ -82,7 +82,7 @@ module LavinMQ
       log_file = (path = @config.log_file) ? File.open(path, "a") : STDOUT
       broadcast_backend = ::Log::BroadcastBackend.new
       backend = if ENV.has_key?("JOURNAL_STREAM")
-                  ::Log::IOBackend.new(io: log_file, formatter: JournalLogFormat)
+                  ::Log::IOBackend.new(io: log_file, formatter: JournalLogFormat, dispatcher: ::Log::DirectDispatcher)
                 else
                   ::Log::IOBackend.new(io: log_file, formatter: StdoutLogFormat)
                 end
--- a/src/lavinmq/queue/message_store.cr
+++ b/src/lavinmq/queue/message_store.cr
@@ -265,6 +265,7 @@ module LavinMQ
       end
 
       private def load_deleted_from_disk
+        count = 0
         Dir.each_child(@data_dir) do |child|
           next unless child.starts_with? "acks."
           seg = child[5, 10].to_u32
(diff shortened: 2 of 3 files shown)


src/lavinmq/queue/message_store.cr
```diff
@@ -282,11 +283,13 @@ module LavinMQ
             end
             @replicator.try &.register_file(file)
           end
+          log_progress_and_yield("Loading acks...") if (count += 1) % 100 == 0
```
spuun: I think it would be nice to include which file it just loaded, and maybe how include the count value as well? Same for all other `log_progress_and_yield` where applicable.

src/lavinmq/queue/message_store.cr
```diff
@@ -383,6 +385,11 @@ module LavinMQ
         end
       end
 
+      private def log_progress_and_yield(message)
+        Log.info { message }
```
spuun, line 388: One "problem" here is that if you set log level to something higher than `info` the message string will still be created for every call to `log_progress_and_yield`. It's probably better to do 
```suggestion
      private def log_progress_and_yield(&block)
        Log.info &block
```
and change all calls to `log_progress_and_yield` to use a block: `log_progress_and_yield { "hello world" }`

src/lavinmq/queue/message_store.cr
```diff
@@ -265,6 +265,8 @@ module LavinMQ
       end
 
       private def load_deleted_from_disk
+        count = 0
+        ack_files = Dir.entries(@data_dir).reject { |f| f.starts_with?("msgs") }.size
```
carlhoerberg, line 269: now you create two arrays, where none is needed.

```suggestion
        ack_files = Dir.each(@data_dir).count { |f| f.starts_with?("acks.") }
```

src/lavinmq/queue/message_store.cr
```diff
@@ -282,6 +284,7 @@ module LavinMQ
             end
             @replicator.try &.register_file(file)
           end
+          log_progress_and_yield("Loading acks (#{count}/#{ack_files})") if (count += 1) % 100 == 0
```
carlhoerberg, line 287: modulus of number that this a power of two is much faster
```suggestion
          log_progress_and_yield("Loading acks (#{count}/#{ack_files})") if (count += 1) % 128 == 0
```

carlhoerberg (changes requested):

src/lavinmq/queue/message_store.cr
viktorerlingsson: I'm not getting any logs when using this approach. Any ideas?

``` 
private def log_progress_and_yield(&block)
  Log.info &block
  Fiber.yield
end
```

called by for example:
```
log_progress_and_yield { "Deleting unused segment #{seg}" }
```

viktorerlingsson: Solved by using `DirectDispatcher` instead, and just logging like normal in this file

carlhoerberg (approved):

src/lavinmq/launcher.cr
```diff
@@ -82,7 +82,7 @@ module LavinMQ
       log_file = (path = @config.log_file) ? File.open(path, "a") : STDOUT
       broadcast_backend = ::Log::BroadcastBackend.new
       backend = if ENV.has_key?("JOURNAL_STREAM")
-                  ::Log::IOBackend.new(io: log_file, formatter: JournalLogFormat)
+                  ::Log::IOBackend.new(io: log_file, formatter: JournalLogFormat, dispatcher: ::Log::DirectDispatcher)
                 else
                   ::Log::IOBackend.new(io: log_file, formatter: StdoutLogFormat)
```
carlhoerberg, line 87: Dont need the direct dispatch here too?

viktorerlingsson: Yeah, you're right,. Added it here as well!

PR #620, opened 2023-12-28
Bugfix: crash during startup because of missing 'x-delayed-type'

spuun:

If a delayed exchange was declared with `type=x-delayed-message` and `arguments={x-delayed-type: '…'}` and LavinMQ crashed before definitions has been compacted, it couldn't start again because of:
`Unhandled exception: Missing required argument 'x-delayed-type' (LavinMQ::Error::ExchangeTypeError)`

In LavinMQ any exchange can be delayed by specifiy `x-delayed-exchange: true` in exchange arguments, and declare frames with `type=x-delayed-message` was converted to `type=<value of x-delayed-type>` with `arguments={x-delayed-exchange: true}`. However, the original frame was stored to the definition file, but with modified arguments causing the crash on startup, i.e. the frame had `type=x-delayed-message` but no `x-delayed-type` in the arguments table.

This change will make sure to clone the arguments table before modifying it.

Specs, or run lavinmq and declare a delayed exchange with `type=x-delayed-message` then kill -9 lavinmq and try start it.

### DIFF
--- a/spec/vhost_spec.cr
+++ b/spec/vhost_spec.cr
@@ -26,6 +26,21 @@ describe LavinMQ::VHost do
     Server.vhosts["test"].exchanges["e"].should_not be_nil
   end
 
+  it "should be able to persist durable delayed exchanges when type = x-delayed-message" do
+    # This spec is to verify a fix where a server couldn't start again after a crash if
+    # an delayed exchange had been declared by specifiying the type as "x-delayed-message".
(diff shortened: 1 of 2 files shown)


kickster97 (approved):

carlhoerberg:
Can definitions with delayed exchanges be exported and then imported in Rabbit?

PR #626, opened 2024-01-30
SystemD.notify_ready when starting as follower

viktorerlingsson:
Sends SystemD.notify_ready when starting LavinMQ as follower, so systemd knows the service has finished activating. 

Start LavinMQ as a service as a follower

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


viktorerlingsson:
Maybe `SystemD.notify_ready` should be before syncing files instead (i.e. after `authenticate`)? Depends on how we want to look at it I guess

carlhoerberg:
Yes, after auth i would say

kickster97 (approved):

carlhoerberg:
Sync can be called multiple times. Put it on top of follow instead

viktorerlingsson:
> Sync can be called multiple times. Put it on top of follow instead

Yeah, that's probably better 👍

PR #627, opened 2024-01-31
Truncate shovel name if too long

viktorerlingsson:
If queue1_name + queue2_name is longer than 255 characters, then moving messages between them will create a shovel with a name longer than 255 characters, which is not allowed. This truncates the shovel name to less then 255 characters. 

Manual - create two queues with long names, move messages between them.

### DIFF
--- a/static/js/queue.js
+++ b/static/js/queue.js
@@ -258,7 +258,7 @@ document.querySelector('#moveMessages').addEventListener('submit', function (evt
   const password = Auth.getPassword()
   const uri = 'amqp://' + encodeURIComponent(username) + ':' + encodeURIComponent(password) + '@localhost/' + urlEncodedVhost
   const dest = document.querySelector('[name=shovel-destination]').value.trim()
-  const name = ('Move ' + queue + ' to ' + dest).substring(0,240) // max length of shovel name is 255
+  const name = ('Move ' + queue + ' to ' + dest).substring(0, 240) // max length of shovel name is 255
   const url = 'api/parameters/shovel/' + urlEncodedVhost + '/' + encodeURIComponent(name)
   const body = {
     name,
--- a/static/js/queue.js
+++ b/static/js/queue.js
@@ -258,7 +258,7 @@ document.querySelector('#moveMessages').addEventListener('submit', function (evt
   const password = Auth.getPassword()
   const uri = 'amqp://' + encodeURIComponent(username) + ':' + encodeURIComponent(password) + '@localhost/' + urlEncodedVhost
   const dest = document.querySelector('[name=shovel-destination]').value.trim()
-  const name = 'Move ' + queue + ' to ' + dest
+  const name = ('Move ' + queue + ' to ' + dest).substring(0,240) // max length of shovel name is 255
   const url = 'api/parameters/shovel/' + urlEncodedVhost + '/' + encodeURIComponent(name)
   const body = {
     name,


viktorerlingsson:
Do we need to handle this somewhere else as well? API?

kickster97 (approved):
nice :)

carlhoerberg:
What is limiting it to 255 chars? I'm pretty sure we should have better error reporting in more places.

viktorerlingsson:
> What is limiting it to 255 chars? 

short string in amq-protocol.cr (rabbit has the same limitation)

> I'm pretty sure we should have better error reporting in more places.

Agreed! I've tried adding some validations here https://github.com/cloudamqp/lavinmq/pull/631 , see this comment for discussion around that! https://github.com/cloudamqp/lavinmq/pull/631#issuecomment-1929259927

carlhoerberg:
Strings in AMQP Tables are always long strings, only keys are limited to 255 chars. 

https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3
https://github.com/cloudamqp/amq-protocol.cr/blob/eddbea6ddb19be0b1750c2760aa5c5ab0eeafed8/src/amq/protocol/table.cr#L297-L299

viktorerlingsson:
I haven't investigated exactly what row caused this to fail, but we use ShortString in frames.cr for a lot of things (queue_name, exchange_name, consumer_tag, routing_key, etc), so I think there's multiple places where any name longer than 255 characters would cause problems?

https://github.com/cloudamqp/amq-protocol.cr/blob/eddbea6ddb19be0b1750c2760aa5c5ab0eeafed8/src/amq/protocol/frames.cr

carlhoerberg:
We should investigate that to understand where the validation should be

carlhoerberg:
The name of the shovel is not used in any of those places you enumerated afaik

viktorerlingsson:
> The name of the shovel is not used in any of those places you enumerated afaik

The shovel name is used as consumer_tag on Consume (and I guess the other methods there as well). 

Stack trace should look something like this (but now I'm just raising an exception in short_string manually when the short string `== "Shovel[my_shovel_name]"`). 


```
  from lib/amq-protocol/src/amq/protocol/short_string.cr:10:11 in 'to_io'
  from /usr/share/crystal/src/io.cr:914:5 in 'write_bytes'
  from lib/amq-protocol/src/amq/protocol/frames.cr:1685:15 in 'to_io'
  from /usr/share/crystal/src/io.cr:914:5 in 'write_bytes'
  from lib/amqp-client/src/amqp-client/connection.cr:166:9 in 'unsafe_write'
  from lib/amqp-client/src/amqp-client/connection.cr:151:9 in 'write'
  from lib/amqp-client/src/amqp-client/channel.cr:685:7 in 'write'
  from lib/amqp-client/src/amqp-client/channel.cr:416:7 in 'basic_consume:no_ack:exclusive:block:args:tag'
  from src/lavinmq/shovel/shovel.cr:127:9 in 'each'
  from src/lavinmq/shovel/shovel.cr:324:11 in 'run'
  from src/lavinmq/shovel/shovel_store.cr:31:13 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:98:34 in '->'
  from ???
```

https://github.com/cloudamqp/amq-protocol.cr/blob/main/src/amq/protocol/frames.cr#L1685

viktorerlingsson:
consumer_tag is set here I guess https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/shovel/shovel.cr#L47

carlhoerberg:
> consumer_tag is set here I guess https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/shovel/shovel.cr#L47

Then it's there we should truncate, not in some javascript

carlhoerberg:
Why even use a custom consumer tag? The channel only have consumer anyway.

viktorerlingsson:
> Then it's there we should truncate, not in some javascript

Sure! Maybe we should even do both? We create the name for this ourselves on the same line of this javascript, and I don't really see why we would create a name that's too long, only for it to be truncated shortly after. `const name = ('Move ' + queue + ' to ' + dest).substring(0, 240)`

I also started this PR https://github.com/cloudamqp/lavinmq/pull/631 to look at names in more places. 


> Why even use a custom consumer tag? The channel only have consumer anyway.

I think it's good to have a consumer tag for the shovel. Makes it easier to identify if there are several shovels or other consumers on a queue IMO

carlhoerberg:
Ok, I think the full name should be there, for the Connection name to be correct. But, for the consumer tag i think "shovel" or something similar is enough.

viktorerlingsson:
> Ok, I think the full name should be there, for the Connection name to be correct. But, for the consumer tag i think "shovel" or something similar is enough.

Yeah, that's probably enough! Maybe consumer_tag can be `shovel-<Random ID>`, just for logging purposes to have different tags.

PR #628, opened 2024-02-01
Fix prometheus endpoints

kickster97:
Fixes the `PrometheusController`, 
Previously the iterator `vhosts` from `target_vhosts()` was misused, resulting in the value being 0 for many metrics. 
Now we let `vhosts` be an array, instead of an iterator. This allows us to iterate over it many times, as we do in `register_routes`. 

Also: Adds a `type` to the metrics that were missing it, so that [datadog is able to scrape everything](https://docs.datadoghq.com/metrics/types/?tab=count#:~:text=Each%20metric%20submitted%20to%20Datadog%20should%20have%20a%20type.). 

accumulate some queues, messages, exchanges etc. and check the endpoint at
http://localhost:15672/metrics

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -8,6 +8,23 @@ describe LavinMQ::HTTP::ConsumersController do
       response.body.lines.any?(&.starts_with? "telemetry_scrape_duration_seconds").should be_true
     end
 
+    it "should perform sanity check on sampled metrics" do
+      vhost = Server.vhosts.create("pmths")
+      vhost.declare_queue("test1", true, false)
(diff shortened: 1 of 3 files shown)


viktorerlingsson (commented):
LGTM 👍 

Agree with @jage , would be great to verify with a spec or two!

kickster97:
Seems like the prometheus parser is not used in any of our projects, has no starts of forks..  @jage is the intention to maintain it? 
A solution could be to just lift in the parser code into lavinmq, as i'm having trouble making it work as a dependency..

viktorerlingsson (approved):
LGTM! 

(but run `crystal tool format`)

PR #629, opened 2024-02-01
Keep the same vhost selected when creating queue

viktorerlingsson:
Keeps the vhost field the same after creating a queue instead of defaulting to /
Makes it a little easier to create multiple queues. 

https://trello.com/c/UmpNY4F5/462-vhost-selection-on-create-queue-resets-when-creating-a-queue

Manual


kickster97:
I tried it out, and when i create a queue it still clears the vhost field. When i reload the page it is re-set

viktorerlingsson:
> I tried it out, and when i create a queue it still clears the vhost field. When i reload the page it is re-set

Hmm, there seems to be an issue with the default vhost for me. It stays selected if I use for example 'cloudamqp', but gets de-selected if I use '/'

kickster97:
lgtm now :)

PR #631, opened 2024-02-01
Make sure name of Queues, Exchanges and Vhosts is not longer than 255 characters

viktorerlingsson:
Stops users from creating queues, vhosts and exchanges with names longer than 255 characters (short string)

### DIFF
--- a/src/lavinmq/http/controller/exchanges.cr
+++ b/src/lavinmq/http/controller/exchanges.cr
@@ -71,7 +71,7 @@ module LavinMQ
               context.response.status_code = 204
             elsif name.starts_with? "amq."
               bad_request(context, "Not allowed to use the amq. prefix")
-            elsif name.size > 255
+            elsif name.bytesize > UInt8::MAX
               bad_request(context, "Exchange name too long, can't exceed 255 characters")
             else
               @amqp_server.vhosts[vhost]
--- a/src/lavinmq/http/controller/exchanges.cr
+++ b/src/lavinmq/http/controller/exchanges.cr
@@ -71,6 +71,8 @@ module LavinMQ
               context.response.status_code = 204
             elsif name.starts_with? "amq."
               bad_request(context, "Not allowed to use the amq. prefix")
+            elsif name.size > 255
+              bad_request(context, "Exchange name too long, can't exceed 255 characters")
             else
               @amqp_server.vhosts[vhost]
                 .declare_exchange(name, type.not_nil!, durable, auto_delete, internal, tbl)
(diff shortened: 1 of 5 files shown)


viktorerlingsson:
Should we do this more 'centralized' instead? Like either in a validator that checks length of all fields with `type=short string`, or propagate the error from amq-protocol or something?

src/lavinmq/http/controller/parameters.cr
```diff
@@ -73,6 +73,9 @@ module LavinMQ
             unless value
               bad_request(context, "Field 'value' is required")
             end
+            if name.size > 255
+              bad_request(context, "#{component} name too long, can't exceed 255 characters")
```
carlhoerberg, line 77: are you sure about parameters not supporing longer names? afaik they are never transfered over the amqp protocol and should not have a limitation, but maybe they are in rabbit?

src/lavinmq/http/controller/exchanges.cr
```diff
@@ -71,6 +71,8 @@ module LavinMQ
               context.response.status_code = 204
             elsif name.starts_with? "amq."
               bad_request(context, "Not allowed to use the amq. prefix")
+            elsif name.size > 255
```
carlhoerberg, line 74: always use `String#bytesize`, it's different from `String#size` when regarding multibyte characters.

src/lavinmq/http/controller/queues.cr
```diff
@@ -64,6 +64,8 @@ module LavinMQ
               context.response.status_code = 204
             elsif name.starts_with? "amq."
               bad_request(context, "Not allowed to use the amq. prefix")
+            elsif name.size > 255
```
carlhoerberg, line 67:
```suggestion
            elsif name.bytesize > UInt8::MAX
```

carlhoerberg (changes requested):

src/lavinmq/http/controller/parameters.cr
viktorerlingsson: No, you're right. Only had issues with long shovel names because we used name as consumer tag as we fixed in another PR.

src/lavinmq/http/controller/vhosts.cr
```diff
@@ -34,6 +34,9 @@ module LavinMQ
           refuse_unless_administrator(context, u)
           name = URI.decode_www_form(params["name"])
           is_update = @amqp_server.vhosts[name]?
+          if name.size > 255
```
carlhoerberg, line 37: Bytesize

carlhoerberg (approved):

PR #634, opened 2024-02-08
Dont use shovel name as consumer tag

viktorerlingsson:
Sets consumer tag for shovels to just `shovel` instead of using the shovel name, since that could be longer than 255 characters. 

Based on discussion in [#627](https://github.com/cloudamqp/lavinmq/pull/627)

Run spec

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -44,7 +44,7 @@ module LavinMQ
                      @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH,
                      @ack_mode = DEFAULT_ACK_MODE, consumer_args : Hash(String, JSON::Any)? = nil,
                      direct_user : User? = nil)
-        @tag = "Shovel-#{Random::Secure.urlsafe_base64(24)}"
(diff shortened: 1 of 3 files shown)


src/lavinmq/shovel/shovel.cr
```diff
@@ -44,7 +44,7 @@ module LavinMQ
                      @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH,
                      @ack_mode = DEFAULT_ACK_MODE, consumer_args : Hash(String, JSON::Any)? = nil,
                      direct_user : User? = nil)
-        @tag = "Shovel[#{@name}]"
+        @tag = "Shovel-#{Random::Secure.urlsafe_base64(24)}"
```
carlhoerberg, line 47: I don't know what this adds, afaik this is never logged anywhere. The name of the receiving queue would make more sense. Else, just `shovel`.

viktorerlingsson: Yeah, you're right. I was thinking we wanted to keep consumer tags unique, but I guess that doesn't matter here since they are always on their own channel.

carlhoerberg (approved):

PR #637, opened 2024-02-28
Bugfix: federation queue link, dont transfer messages unless consumer exists

viktorerlingsson:

Stops federation queue links from transferring messages unless there is an active consumer on the downstream. 
`federate` now publishes directly to vhost instead of using a client. This allows us to use the immediate flag to make sure we don't federate messages unless there is a consumer on the downstream queue. 

Run the specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Follower lag is now based on bytes written to action queue instead of socket [#652](https://github.com/cloudamqp/lavinmq/pull/652)
 
+### Fixed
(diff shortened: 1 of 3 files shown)


spuun:
Just a little remark that came to my mind:

By removing the local `connection` and `channel` it may be harder to understand where messages comes from. Dunno how big of a deal it is, but just want to mention it.

If I can manage to break out `Client`, `Connection`, `Channel` and `Consumer` (which I'd like to do) it would be possible to make these entities without actually having a real TCP connection.

kickster97:
> If I can manage to break out `Client`, `Connection`, `Channel` and `Consumer` (which I'd like to do) it would be possible to make these entities without actually having a real TCP connection.

@spuun, We should remember this now that we have the PR for the abstraction out :)

PR #638, opened 2024-02-29
add nilguard for matching x-delayed-type argument

kickster97:
If you try and redeclare a delayed exchange with no `x-delayed-type` argument, it will crash in `exchanges.cr`: `match?(frame)`

This pull request adds a nilguard for the argument check, and as a result, the `match` method will return false if there is no `x-delayed-type` argument and send a`Preconditioned Failed: Existing exchange 'my_delayed_exchange' declared with other arguments` instead of crashing. 

Declare a delayed exchange "my-delayed-exchange", with argument `{ "x-delayed-type" => "direct" }`, then try and declare the exchange again, but with no `x-delayed-type` argument.

### DIFF
--- a/spec/exchange_spec.cr
+++ b/spec/exchange_spec.cr
@@ -75,11 +75,38 @@ describe LavinMQ::Exchange do
   end
 
   describe "delayed message exchange declaration" do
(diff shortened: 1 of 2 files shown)


viktorerlingsson (approved):
LGTM! 👍

Maybe just add a spec for it as well?

PR #643, opened 2024-03-20
Move specs from node specs to vhost stats spec

spuun:
Moving and refactoring specs.

For some reason these moved specs was located in the file for api nodes controller specs. These specs aren't testing the nodes api, they are testing vhost stats.

Run specs


viktorerlingsson:
Maybe we should keep both spec files? /api/nodes also returns metrics that we use on the nodes page, right? So I think it makes sense to have tests for that as well.

spuun:
> Maybe we should keep both spec files? /api/nodes also returns metrics that we use on the node page, right? So I think it makes sense to have tests for that as well.

True, but I think we can test all that in the one spec that's left. Just add more keys to verify. All "should update metrics" things has nothing to do with the API. It may even be an idea to move API specs to playwright... 

I did this because I had an idea to break out stats from `LavinMQ::Server` to be able to do even more isolated specs and to get smaller classes.

viktorerlingsson:
> True, but I think we can test all that in the one spec that's left. Just add more keys to verify. All "should update metrics" things has nothing to do with the API.

OK, sounds good! 👍

viktorerlingsson (approved):

kickster97 (approved):

PR #644, opened 2024-03-24
log warn instead of preconditioned failed when redeclaring exchange

kickster97:
Logs warning instead of raising preconditioned failure when trying to redeclare exchange with differing arguments, we only raise a preconditioned failed exception when it is an alternate exchange

### DIFF
--- a/src/lavinmq/client/client.cr
+++ b/src/lavinmq/client/client.cr
@@ -533,7 +533,7 @@ module LavinMQ
           send AMQP::Frame::Exchange::DeclareOk.new(frame.channel)
         end
       else
-        send_precondition_failed(frame, "Existing exchange '#{frame.exchange_name}' declared with other arguments")
+        @log.warn { "Existing exchange '#{frame.exchange_name}' declared with other arguments" }
       end
     end
 


baelter:
Not sure this is the way to go, as expected this makes several of the test suites fail. e.g `Bunny::Exchange of type fanout when declared with a different set of attributes raises an exception`. Push on RabbitMQ to fix the "bug"? OR make a special case for the delayed exchange type?

kickster97:
> Not sure this is the way to go, as expected this makes several of the test suites fail. e.g `Bunny::Exchange of type fanout when declared with a different set of attributes raises an exception`. Push on RabbitMQ to fix the "bug"? OR make a special case for the delayed exchange type?

We discussed this solution in the team and thought logging warnings was a good middleground, but I agree it gets messy. I'm leaning towards not merging this, and allowing LavinMQ to follow the AMQP spec. If users are confused by the difference we can help explain where they might be going wrong.

viktorerlingsson:
Maybe we should remove `@arguments == frame_args` from `match?` [(here)](https://github.com/cloudamqp/lavinmq/blob/warn_redeclare_exchange/src/lavinmq/exchange/exchange.cr#L103) instead to achieve the same behavior as RabbitMQ? (Or add some extra logic if we still want to check the arguments in some cases)

PR #645, opened 2024-03-28
Graceful replication shutdown

kickster97:
Wait for followers to be in sync before closing server. 

run specs for #close

### DIFF
--- a/src/lavinmq/replication/follower.cr
+++ b/src/lavinmq/replication/follower.cr
@@ -10,9 +10,13 @@ module LavinMQ
       @acked_bytes = 0_i64
       @sent_bytes = 0_i64
       @actions = Channel(Action).new(4096)
+      getter name
+      getter? closed
 
       def initialize(@socket : TCPSocket, @file_index : FileIndex)
         Log.context.set(address: @socket.remote_address.to_s)
+        @name = @socket.remote_address.to_s
+        @closed = false
         @socket.write_timeout = 5
         @socket.read_timeout = 5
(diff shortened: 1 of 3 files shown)


src/lavinmq/replication/follower.cr
```diff
@@ -137,13 +141,20 @@ module LavinMQ
...
       end
 
-      def close
+      def close(synced_close : Channel({Follower, Bool})? = nil)
+        return if closed?
+        @closed = true
         Log.info { "Disconnected" }
+        if synced_close
+          wait_for_sync
+        end
```
viktorerlingsson, line 150:
```suggestion
        wait_for_sync if synced_close
```

viktorerlingsson:
LGTM 👍 

A couple of questions; 

Do we stop publishers to leader somewhere? Or can this keep going "forever" if you keep publishing? Maybe that is done by the server earlier when initiating shutdown? 

Should there be some kind of timeout on the sync? I don't see anything like that in this code. Maybe we should let that be handled outside of LavinMQ though

spuun:
> Do we stop publishers to leader somewhere? Or can this keep going "forever" if you keep publishing? Maybe that is done by the server earlier when initiating shutdown?

Replication is closed last, so vhost should have closed all client connections before (and shovels/federations).
 
> Should there be some kind of timeout on the sync? I don't see anything like that in this code. Maybe we should let that be handled outside of LavinMQ though

Yes, timeout or similar will probably be part of #547.

viktorerlingsson (approved):

PR #646, opened 2024-03-28
Export replication lag in prometheus metrics

kickster97:
exports the amount of followers and the lag for each follower in prometheus metrics

start lavinmq with a couple of followers and run `laivnmqperf throughput` and check `http://localhost:15672/metrics`

### DIFF
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -252,6 +252,17 @@ module LavinMQ
                       value: @amqp_server.stats_system_collection_duration_seconds.to_f,
                       type:  "gauge",
                       help:  "Time it takes to collect system metrics"})
+        writer.write({name:  "total_connected_followers",
+                      value: @amqp_server.@replicator.followers.size,
+                      type:  "gauge",
+                      help:  "Amount of follower nodes connected"})
+        @amqp_server.@replicator.followers.each_with_index do |f, i|
+        writer.write({name:  "follower_#{i}_lag",
+                      value: f.lag,
+                      type:  "gauge",
+                      help:  "Lag for follower #{i}"})
+  end
+
       end
 
       SERVER_METRICS = {:connection_created, :connection_closed, :channel_created, :channel_closed,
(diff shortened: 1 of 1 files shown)


src/lavinmq/http/controller/prometheus.cr
```diff
@@ -252,6 +252,16 @@ module LavinMQ
                       value: @amqp_server.stats_system_collection_duration_seconds.to_f,
                       type:  "gauge",
                       help:  "Time it takes to collect system metrics"})
+        writer.write({name:  "total_connected_followers",
+                      value: @amqp_server.@replicator.followers.size,
+                      type:  "gauge",
+                      help:  "Amount of follower nodes connected"})
+        @amqp_server.@replicator.followers.each_with_index do |f, i|
+          writer.write({name:  "follower_#{i}_lag",
```
viktorerlingsson, line 260: Could we write something more specific about the follower here so they can be identified? "follower_1" doesn't really mean anything IMO, so would be better with hostname or IP or something.

kickster97: > Could we write something more specific about the follower here so they can be identified? "follower_1" doesn't really mean anything IMO, so would be better with hostname or IP or something.

I added remote ip as an id on the name. Might get a little long, and harder to read? 

```
# TYPE lavinmq_total_connected_followers gauge
# HELP lavinmq_total_connected_followers Amount of follower nodes connected
lavinmq_total_connected_followers 2
# TYPE lavinmq_follower_lag_127.0.0.1:5679 gauge
# HELP lavinmq_follower_lag_127.0.0.1:5679 Lag for follower on address: 127.0.0.1:5679
lavinmq_follower_lag_127.0.0.1:5679 0
# TYPE lavinmq_follower_lag_127.0.0.1:5679 gauge
# HELP lavinmq_follower_lag_127.0.0.1:5679 Lag for follower on address: 127.0.0.1:5679
lavinmq_follower_lag_127.0.0.1:5679 0
```

viktorerlingsson (approved):
LGTM 👍

PR #650, opened 2024-04-02
Use CLA Assistant GitHub action instead of app

kickster97:

[adds CLA github action](https://github.com/contributor-assistant/github-action), allowing us to store json of signatures instead of using centralized storage of signatures with cla assistant app, as well as allows us to use other formats than gist for the CLA document

test by opening, synchronize, closing PR

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -20,8 +20,7 @@ jobs:
         uses: contributor-assistant/github-action@v2.3.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # the below token should have repo scope and must be manually added by you in the repository's secret
-          # This token is required only if you have configured to store the signatures in a remote repository/organization
+          # gives write access to the remote repository
           PERSONAL_ACCESS_TOKEN: ${{ secrets.CLA_SIGNATURES_GITHUB_TOKEN }}
         with:
           path-to-signatures: 'signatures/lavinmq/cla.json'
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -27,7 +27,7 @@ jobs:
           path-to-document: 'https://github.com/cloudamqp/CLA-signatures/blob/main/cla.md'
           # branch should not be protected
           branch: 'main'
-          allowlist: user1,bot*
+          allowlist: 'dependabot[bot]'
 
          # the followings are the optional inputs - If the optional inputs are not given, then default values will be taken
           remote-organization-name: 'cloudamqp'
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -5,7 +5,11 @@ on:
   pull_request_target:
     types: [opened,closed,synchronize]
 
-# explicitly configure permissions, in case your GITHUB_TOKEN workflow permissions are set to read-only in repository settings
+# https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
+# https://github.com/contributor-assistant/github-action needs permissions to
+# - re-run workflows
+# - comment in pull requests
+# - update commit statuses 
 permissions:
   actions: write
   pull-requests: write
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,0 +1,43 @@
+name: "CLA Assistant"
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+
+# explicitly configure permissions, in case your GITHUB_TOKEN workflow permissions are set to read-only in repository settings
+permissions:
+  actions: write
+  contents: write
+  pull-requests: write
+  statuses: write
+
+jobs:
+  CLAAssistant:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/cla.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+
+# explicitly configure permissions, in case your GITHUB_TOKEN workflow permissions are set to read-only in repository settings
+permissions:
+  actions: write
+  contents: write
```
dentarg, line 11: I don't think `contents` should be needed when using a remote repository?

.github/workflows/cla.yml
```diff
@@ -0,0 +1,42 @@
...
+          # This token is required only if you have configured to store the signatures in a remote repository/organization
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: 'signatures/lavinmq/cla.json'
+          path-to-document: 'https://github.com/cloudamqp/CLA-signatures/blob/main/cla.md'
+          # branch should not be protected
+          branch: 'main'
+          allowlist: user1,bot*
+
+         # the followings are the optional inputs - If the optional inputs are not given, then default values will be taken
```
dentarg, line 33: No need for this comment IMHO

.github/workflows/cla.yml
```diff
@@ -0,0 +1,42 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # the below token should have repo scope and must be manually added by you in the repository's secret
+          # This token is required only if you have configured to store the signatures in a remote repository/organization
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: 'signatures/lavinmq/cla.json'
+          path-to-document: 'https://github.com/cloudamqp/CLA-signatures/blob/main/cla.md'
+          # branch should not be protected
+          branch: 'main'
+          allowlist: user1,bot*
```
dentarg, line 31: This is from the README, we should change it to what we actually need, which is Dependabot

```suggestion
          allowlist: 'dependabot[bot]'
```

.github/workflows/cla.yml
kickster97: For line 33?

.github/workflows/cla.yml
```diff
@@ -0,0 +1,42 @@
...
+  CLAAssistant:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "CLA Assistant"
+        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        uses: contributor-assistant/github-action@v2.3.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # the below token should have repo scope and must be manually added by you in the repository's secret
+          # This token is required only if you have configured to store the signatures in a remote repository/organization
```
dentarg, line 24: We can simplify this comment

```suggestion
          # gives write access to the remote repository
```

.github/workflows/cla.yml
```diff
@@ -0,0 +1,32 @@
+name: "CLA Assistant"
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+
+# explicitly configure permissions, in case your GITHUB_TOKEN workflow permissions are set to read-only in repository settings
```
dentarg, line 8: I don't think we need this comment, it's from the contributor-assistant/github-action README, but we can explain why we need the permissions (and link the permissions docs, a nice touch I think)

```suggestion
# https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
# https://github.com/contributor-assistant/github-action needs permissions to
# - re-run workflows
# - comment in pull requests
# - update commit statuses 
```

.github/workflows/cla.yml
```diff
@@ -0,0 +1,32 @@
...
+permissions:
+  actions: write
+  pull-requests: write
+  statuses: write
+
+jobs:
+  CLAAssistant:
+    runs-on: ubuntu-latest
```
dentarg, line 20: I think we want to skip running the CLA Assistant when anyone from our organisation makes PRs

We can do that by making the job conditional (we could have extended the `if` for the step too, but it is already so long)

I think this could work, add it after `runs-on:`

```yaml
if: ${{ github.event.pull_request.author_association != 'MEMBER' || github.event.pull_request.author_association != 'OWNER' }}

Got this from looking at these docs for the event payload https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request (expand `pull_request`)

and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

kickster97: added 
`if: ${{ github.event.pull_request.author_association != 'MEMBER' && github.event.pull_request.author_association != 'OWNER' }}`

dentarg (approved):
Okay let's try this :)

PR #613, opened 2023-12-13
send connection_information to amqp-client

kickster97:
### WHAT is this pull request doing?
Cannot be merged until [#39](https://github.com/cloudamqp/amqp-client.cr/pull/39) is released. 

With this PR we send in connection information 
` product: "LavinMQ", platform: "Crystal", product_version: LavinMQ::VERSION, platform_version: Crystal::VERSION `

to amqp-client, in order for us to be able to have more accurate connection information displayed for federation connections.  

Also: does a little refactor on `start_link`, to dry out some of the code repeated in both `QueueLink` and `ExchangeLink`

### HOW can this pull request be tested?
pull down the amqp-client PR and create an override shards file, then start a federation to another broker and look at the connection information in the UI.

PR #652, opened 2024-04-04
Replication lag improvement

spuun:

Before this we only counted data written to the follower socket as lag, we didn't count the data in the "action queue" of the follower.

This will change so we count data written to the action queue.

Run follower specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Follower lag is now based on bytes written to action queue instead of socket [#652](https://github.com/cloudamqp/lavinmq/pull/652)
+
 ## [1.2.11] - 2024-04-26
 
 ### Fixed
--- a/spec/replication/actions_spec.cr
+++ b/spec/replication/actions_spec.cr
@@ -1,7 +1,7 @@
 require "../spec_helper"
 
-def with_datadir_tempfile(&)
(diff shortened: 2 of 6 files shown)


kickster97:
hmm @dentarg, cla assistant skipped on my pr [here](https://github.com/cloudamqp/lavinmq/pull/613), but not here...

dentarg:
@kickster97 Hmm, I wonder if skipping for `COLLABORATOR` too would help, reading https://stackoverflow.com/questions/63188674/github-actions-detect-author-association it sounds like it. I can experiment some.

dentarg:
I think (hope) the CLA workflow behaves better now going forward. Closing/re-opening didn't really work as I expected, the workflow only ran on the close event: https://github.com/cloudamqp/lavinmq/actions/workflows/cla.yml?query=branch%3Areplication-lag-improvement – I unlocked this PR with the GitHub CLI and also created https://github.com/cloudamqp/lavinmq/pull/654

spuun:
This will also count all data that needs to be sent as lag, not only "message data" (i.e. it counts filename length etc too).

spuun:
> This will also count all data that needs to be sent as lag, not only "message data" (i.e. it counts filename length etc too).

Maybe it should be changed to only count message data.

kickster97 (approved):

kickster97:
> > This will also count all data that needs to be sent as lag, not only "message data" (i.e. it counts filename length etc too).
> 
> Maybe it should be changed to only count message data.

Is it not more accurate if we count all data? Why would we only want to count message data?

PR #654, opened 2024-04-05
CLA workflow: avoid locking PRs after merge

dentarg:
I don't think we want the CLA workflow to lock pull requests if they are closed/merged.

See this issue for context https://github.com/contributor-assistant/github-action/issues/77


kickster97 (approved):
I agree, nice!

PR #655, opened 2024-04-08
Change log source for followers to replication.follower

viktorerlingsson:
Changes the log source for replication followers from "self" to "replication.follower"

Look at logs

### DIFF
--- a/src/lavinmq/replication/follower.cr
+++ b/src/lavinmq/replication/follower.cr
@@ -5,7 +5,7 @@ require "./file_index"
 module LavinMQ
   module Replication
     class Follower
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Do we want to drop the LavinMQ module completely instead? Or hack `Log` to drop it?

spuun:
> Do we want to drop the LavinMQ module completely instead? Or hack `Log` to drop it?

Imo we should keep it (or shorten it to `lmq`). But we need to decide on how we want the logging to work, if we wan't to be able to configure different levels for different sources or not.

carlhoerberg:
fix in another commit

PR #656, opened 2024-04-08
create durable queues and exchanges by default from the UI

viktorerlingsson:
Sets durable to true by default when creating queues and exchanges from the UI

Test in UI


kickster97 (changes requested):
Nice! 
Exchanges also have the durable attribute, so I think we should add these changes to the exchange view also :)

kickster97 (approved):

PR #657, opened 2024-04-10
Refactor and move a MessageStore spec

spuun:
Writing a `MessageStore` spec i found this spec and realized it could be simplified and more isolated to only involve `MessageStore`. I also moved it to the `MessageStore` describe block in `queue_spec`, but that entire block should probably be moved to it's own file at some point.

Run specs


viktorerlingsson (approved):
LGTM 👍

PR #659, opened 2024-04-10
Add and use interface for replicator

spuun:
Adds interface `Replicator` which `Replicator::Server` implements. This will make it possible to stub/mock `Replicator`.

Run specs

### DIFF
--- a/src/lavinmq/replication/replicator.cr
+++ b/src/lavinmq/replication/replicator.cr
@@ -1,0 +1,14 @@
+require "../mfile"
+
+module LavinMQ
+  module Replication
+    module Replicator
+      abstract def register_file(file : File)
+      abstract def register_file(mfile : MFile)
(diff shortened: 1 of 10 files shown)


kickster97 (approved):

viktorerlingsson (approved):

src/lavinmq/replication/replicator.cr
```diff
@@ -0,0 +1,14 @@
+require "../mfile"
+
+module LavinMQ
+  module Replication
+    module Replicator
```
carlhoerberg, line 5: Why not an `abstract class`?

spuun: I like the idea of being able to use multiple interfaces when applicable to reduce the interface you need to mock/stub in specs. `Replication::Server` now implements `Replication::Replicator` and `Replication::FileIndex`.

PR #661, opened 2024-04-17
Automatic offset tracking for stream queues

viktorerlingsson:
Adds broker tracking of consumer offsets in streams if no `x-stream-offset` is provided by the consumer (or if `x-stream-automatic-offset-tracking == true`). Does not track if the consumer tag is generated by the broker. 

Run specs


src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+          tx = @consumer_offsets.to_slice(pos, 8)
+          return IO::ByteFormat::SystemEndian.decode(Int64, tx)
+        end
+      rescue KeyError
+      end
+
+      private def consumer_offset_positions
+        positions = Hash(String, Int64).new
+        slice = @consumer_offsets.to_slice
+        return positions if slice.size.zero?
```
carlhoerberg, line 111:
```suggestion
        return positions if @consumer_offsets.size.zero?
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+        pos = 0
```
carlhoerberg, line 112:
```suggestion
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+
+        loop do
+          ctag_length = IO::ByteFormat::LittleEndian.decode(UInt8, slice[pos, pos + 1])
+          break if ctag_length == 0
+          pos += 1
+          ctag = String.new(slice[pos..pos + ctag_length - 1])
+          pos += ctag_length
+          positions[ctag] = pos
+          pos += 8
+          break if pos >= slice.size
```
carlhoerberg, line 122: might as well use the `IO` interface, nicer and no slower.

```suggestion
          ctag_length = @consumer_offsets.read_byte
          break if ctag_length.zero?
          ctag = @consumer_offsets.read_string(ctag_length)
          positions[ctag] = @consumer_offsets.pos
          @consumer_offsets.skip(8)
        rescue IO::EOFError
          break
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+        end
+        @consumer_offsets.resize(pos) # resize mfile to remove any empty bytes
+        positions
+      end
+
+      def save_offset_by_consumer_tag(consumer_tag, new_offset)
+        pos = 0_i64
```
carlhoerberg, line 129:
```suggestion
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+            buf = uninitialized UInt8[8]
+            IO::ByteFormat::LittleEndian.encode(new_offset.as(Int64), buf.to_slice)
+            @consumer_offsets.write_at(pos, buf.to_slice)
+          end
+        rescue KeyError
+          write_new_ctag_to_file(consumer_tag, new_offset)
+        end
+      end
+
+      # should we write a null byte after each offset?
```
carlhoerberg, line 141:
```suggestion
```
nope

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+          break if pos >= slice.size
+        end
+        @consumer_offsets.resize(pos) # resize mfile to remove any empty bytes
+        positions
+      end
+
+      def save_offset_by_consumer_tag(consumer_tag, new_offset)
+        pos = 0_i64
+        begin
+          if pos = @consumer_offset_positions[consumer_tag]
```
carlhoerberg, line 131:
```suggestion
          if pos = @consumer_offset_positions[consumer_tag]?
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+            buf = uninitialized UInt8[8]
+            IO::ByteFormat::LittleEndian.encode(new_offset.as(Int64), buf.to_slice)
+            @consumer_offsets.write_at(pos, buf.to_slice)
+          end
+        rescue KeyError
+          write_new_ctag_to_file(consumer_tag, new_offset)
+        end
```
carlhoerberg, line 138:
```suggestion
          else
            write_new_ctag_to_file(consumer_tag, new_offset)
          end
        end
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+      end
+
+      # should we write a null byte after each offset?
+      def write_new_ctag_to_file(consumer_tag, new_offset)
```
carlhoerberg, line 142: we are not just writing a ctag

```suggestion
      def store_consumer_offset(consumer_tag, new_offset)
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+          pos += ctag_length
+          positions[ctag] = pos
+          pos += 8
+          break if pos >= slice.size
+        end
+        @consumer_offsets.resize(pos) # resize mfile to remove any empty bytes
+        positions
+      end
+
+      def save_offset_by_consumer_tag(consumer_tag, new_offset)
```
carlhoerberg, line 128:
```suggestion
      def update_consumer_offset(consumer_tag : String, new_offset : Int64)
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +97,103 @@ module LavinMQ
...
+        pos = 0_i64
+        begin
+          if pos = @consumer_offset_positions[consumer_tag]
+            buf = uninitialized UInt8[8]
+            IO::ByteFormat::LittleEndian.encode(new_offset.as(Int64), buf.to_slice)
+            @consumer_offsets.write_at(pos, buf.to_slice)
```
carlhoerberg, line 134:
```suggestion
            @consumer_offsets.pos = pos
            @consumer_offsets.write_bytes new_offset
```
less copying, see https://github.com/crystal-lang/crystal/blob/0aa30372c/src/io/byte_format.cr#L128-L132

carlhoerberg: or else encode directly into the mmap:
```suggestion
            IO::ByteFormat::SystemEndian.encode(new_offset, @consumer_offsets.to_slice(pos, 8))
```

but `MFile#to_slice` returns a read_only Bytes/Slice.

carlhoerberg (changes requested):

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: I can encode directly into the mmap though, but that requires us to add a read_only flag here. Should be OK? 

```
  def to_slice(pos, size, read_only = true)
    raise IO::EOFError.new if pos + size > @size
    Bytes.new(buffer + pos, size, read_only: read_only)
  end
```

carlhoerberg: ah, yes, forgot, `MFile` is append only, the pos is only for reading, which is different from `File`. Use your solution.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+        if pos = @consumer_offset_positions[consumer_tag]?
+          IO::ByteFormat::SystemEndian.encode(new_offset, @consumer_offsets.to_slice(pos, 8, false))
+        else
+          store_consumer_offset(consumer_tag, new_offset)
+        end
+      end
+
+      def store_consumer_offset(consumer_tag : String, new_offset : Int64)
+        slice = consumer_tag.to_slice
+        consumer_tag_length = slice.size.to_u8
```
carlhoerberg, line 136:
```suggestion
        consumer_tag_length = slice.bytesize.to_u8
```
`String#size` != `String#bytesize`

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+        pos = @consumer_offsets.size + slice.size + 1
+
+        length_buffer = uninitialized UInt8[1]
+        IO::ByteFormat::LittleEndian.encode(consumer_tag_length, length_buffer.to_slice)
```
carlhoerberg, line 140: UInt8 are always encoded the same way, no difference between little and big endian. Endianess only applies to types > 1 byte.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+
+        offset_buffer = uninitialized UInt8[8]
+        IO::ByteFormat::LittleEndian.encode(new_offset.as(Int64), offset_buffer.to_slice)
+
+        @consumer_offsets.write(length_buffer.to_slice + slice + offset_buffer.to_slice)
```
carlhoerberg, line 145: why do you merge the buffers here instead of doing indidvual writes?

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+        @consumer_offset_positions[consumer_tag] = pos
+      end
```
carlhoerberg, line 147:
```suggestion
      def store_consumer_offset(consumer_tag : String, new_offset : Int64)
        @consumer_offsets.write_bytes ShortString.new(consumer_tag)
        @consumer_offset_positions[consumer_tag] = @consumer_offsets.pos
        @consumer_offsets.write_bytes @new_offset
      end
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+
+      def cleanup_consumer_offsets
+        offsets_to_save = Hash(String, Int64).new
+        lowest_offset_in_stream, _seg, _pos = offset_at(@segments.first_key, 4u32)
+        @consumer_offset_positions.each do |ctag, _pos|
+          offset = last_offset_by_consumer_tag(ctag)
+          next if !offset || offset < lowest_offset_in_stream
+          offsets_to_save[ctag] = offset
```
carlhoerberg, line 155:
```suggestion
          offsets_to_save[ctag] = offset unless offset && offset > lowest_offset_in_stream
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+        end
+
+        delete_and_reopen_offsets_file
+        @consumer_offset_positions = Hash(String, Int64).new
+        offsets_to_save.each do |ctag, offset|
+          store_consumer_offset(ctag, offset)
+        end
```
carlhoerberg, line 162: This is dangerous, if the machine crashes here all offsets are gone. What you should do is to write the offsets to a new file, then rename it when you're done.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+      end
+
+      def delete_and_reopen_offsets_file
+        @consumer_offsets.close
+        @consumer_offsets.delete
+        @consumer_offsets = MFile.new(@consumer_offset_path, @consumer_offset_capacity)
+      end
```
carlhoerberg, line 185:
```suggestion
      def replace_offsets_file(&)
        old_consumer_offsets = @consumer_offsets
        @consumer_offsets = MFile.new("#{@consumer_offset_path}.tmp", @consumer_offset_capacity)
        yield # fill the new file with correct data in this block
        @consumer_offsets.rename @consumer_offset_path # this will delete the old file on disk
        old_consumer_offsets.close
      end
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +98,92 @@ module LavinMQ
...
+
+        loop do
+          ctag_length = @consumer_offsets.read_byte
+          break if !ctag_length || ctag_length.zero?
+          ctag = @consumer_offsets.read_string(ctag_length)
+          positions[ctag] = @consumer_offsets.pos
+          @consumer_offsets.skip(8)
+        rescue IO::EOFError
+          break
+        end
```
carlhoerberg, line 120:
```suggestion
        loop do
          ctag = ShortString.from_io(@consumer_offsets)
          break if ctag.empty?
          positions[ctag] = @consumer_offsets.pos
          @consumer_offsets.skip(8)
        rescue IO::EOFError
          break
        end
```

carlhoerberg (changes requested):

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: write_bytes does not update pos, but in this case I think we can just use `#size` instead

spec/stream_queue_spec.cr
```diff
@@ -178,4 +210,179 @@ describe LavinMQ::StreamQueue do
...
+      msg = StreamQueueSpecHelpers.consume_one(queue_name, consumer_tag, c_args)
+      StreamQueueSpecHelpers.offset_from_headers(msg.properties.headers).should eq 1
+      sleep 0.1
+
+      # should consume the same message again since tracking was not saved from last consume
+      msg_2 = StreamQueueSpecHelpers.consume_one(queue_name, consumer_tag)
+      StreamQueueSpecHelpers.offset_from_headers(msg_2.properties.headers).should eq 1
+    end
+
+    it "should not use saved offset if x-stream-offset is set" do
```
snichme, line 289: Do we want this behavior?  I'm thinking that if you setup a consumer that says `x-stream-offset=10` and it start consumes to `offset=20` if I restart that consumer then I would expect it to continue from 21, not 10, even though the `x-stream-offset` says 10. 
Otherwise I have to specify `x-stream-offset` first time and then somehow remove the argument if the consumer is restarted and should continue. WDYT?

viktorerlingsson: It's a bit tricky IMO. But in general I think that if you specify x-stream-offset when consuming, that should take precedent over other things. 

Let's say you get messages 10-20, and then decide you want to get them again for whatever reason (or jump to another part of the stream), there would be no way for you to do that if we use our tracking. 

An alternative could be to use an extra argument (x-stream-auto-offset or something), where we think of x-stream-offset as a starting point, and then use our internal tracking for any subsequent requests.

snichme: How this works in Kafka is that offsets are stored based on consumer group name, so if you want to start over using a new offset you need to rename the group. Changing offset when using a group that has already a stored offset on the broker has no effect.  They do have an API call into kafka to reset a consumer group offset so you can tell a running consumer to start reading from somewhere else, but that is just a side note. 

For rabbitmq Streams it looks like the offset it stored based on consumer name (same feature where you can group  consumers on) so if order to "reset" the offset and use another value you need to rename the consumer.  Do we have a name of something for a consumer that we can use in this case?

viktorerlingsson: > For rabbitmq Streams it looks like the offset it stored based on consumer name (same feature where you can group consumers on) so if order to "reset" the offset and use another value you need to rename the consumer. Do we have a name of something for a consumer that we can use in this case?

We use the consumer tag, so you can always "start over" by just setting a new tag.

snichme: Okay, then if we can reset by changing consumer tag I think the broker should use stored offsets if there are any and if not use the one given by the consumer using the argument `x-stream-offset` 

So use stored offset if exists for consumer tags, otherwise use the offset given by consumer

viktorerlingsson: Hmm, I'm not sure I agree - I think that if you provide `x-stream-offset` you explicitly say what you want. That would also  change existing behavior where you always get what you request when providing an `x-stream-offset`. But let's discuss more during tomorrows meeting!

viktorerlingsson: @snichme added `x-stream-use-automatic-offset` so we can have it both ways :)

src/lavinmq/client/channel/stream_consumer.cr
```diff
@@ -34,7 +36,10 @@ module LavinMQ
             raise Error::PreconditionFailed.new("x-priority not supported on stream queues")
           end
           case frame.arguments["x-stream-offset"]?
-          when Nil, Int, Time, "first", "next", "last"
+          when Nil
```
snichme, line 39: RabbitMQ says that if  `x-stream-offset` isn't specified it uses "next" as detault: 
> If it is unspecified the consumer will start reading from the next offset written to the log after the consumer starts

spec/stream_queue_spec.cr
snichme: yay, nice!

src/lavinmq/client/channel/stream_consumer.cr
snichme: Well, that quote is actually from the stream client, as rabbit doesn't support this for the amqp clients. 
So based on that we can do what we want. So if we have one way it works now, leave it at that for now might be good.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,71 @@ module LavinMQ
...
+            store_consumer_offset(ctag, offset)
+          end
+        end
+      end
+
+      def replace_offsets_file(&)
+        old_consumer_offsets = @consumer_offsets
+        @consumer_offsets = MFile.new("#{@consumer_offset_path}.tmp", @consumer_offset_capacity)
+        yield # fill the new file with correct data in this block
+        File.rename "#{@consumer_offset_path}.tmp", @consumer_offset_path
```
carlhoerberg, line 170: should implement `MFile#rename`, so that the internal `@path` is changed too. 
 
```suggestion
        mfile.rename @consumer_offset_path
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,71 @@ module LavinMQ
...
+        old_consumer_offsets.close
```
carlhoerberg, line 171:
```suggestion
        old_consumer_offsets.close(truncate_to_size: false)
```

no need to truncate it, it's being deleted.

src/lavinmq/client/channel/stream_consumer.cr
```diff
@@ -34,7 +36,10 @@ module LavinMQ
...
+            @track_offset = true unless @tag.starts_with?("amq.ctag-")
+          when Int, Time, "first", "next", "last"
+            @track_offset = true if frame.arguments["x-stream-use-automatic-offset"]?
```
carlhoerberg, line 42:
```suggestion
            @track_offset = true if frame.arguments["x-stream-automatic-offset-tracking"]?
```
it is a better name?

carlhoerberg (changes requested):
What happens with `consumer_offset_capacity` is reached? `MFile` doesn't auto expand.

src/lavinmq/client/channel/stream_consumer.cr
viktorerlingsson: Agreed!

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: Implemented [here](https://github.com/cloudamqp/lavinmq/pull/661/commits/e70cc10091ee066d8daa1fce0d19ab0ddecfc590)

src/lavinmq/mfile.cr
```diff
@@ -185,6 +185,14 @@ class MFile < IO
...
   end
 
+  def write_at(pos : Int64, slice : Bytes) : Nil
+    raise ClosedError.new if @closed
+    end_pos = pos + slice.size
+    raise IO::EOFError.new if end_pos > @capacity
+    @size = end_pos if end_pos > @size
+    slice.copy_to(buffer + pos, slice.size)
+  end
+
```
carlhoerberg, line 195: Can't see that this is being used? Am I missing something?

```suggestion
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,71 @@ module LavinMQ
         {msg_offset, segment, pos}
       end
 
+      def last_offset_by_consumer_tag(consumer_tag : String)
+        if pos = @consumer_offset_positions[consumer_tag]?
+          tx = @consumer_offsets.to_slice(pos, 8)
+          return IO::ByteFormat::SystemEndian.decode(Int64, tx)
```
carlhoerberg, line 112: Not that LavinMQ will work on any BigEndian platform that I know of but still, we probably want to set it to a fixed endian so that the data_dir can be moved between machines (also for replication)
 
```suggestion
          return IO::ByteFormat::LittleEndian.decode(Int64, tx)
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,71 @@ module LavinMQ
...
+
+      def update_consumer_offset(consumer_tag : String, new_offset : Int64)
+        if pos = @consumer_offset_positions[consumer_tag]?
+          IO::ByteFormat::SystemEndian.encode(new_offset, @consumer_offsets.to_slice(pos, 8, false))
+        else
+          store_consumer_offset(consumer_tag, new_offset)
+        end
+      end
+
+      def store_consumer_offset(consumer_tag : String, new_offset : Int64)
```
carlhoerberg, line 141: what if the `@consumer_offets` capacity is reached?

viktorerlingsson: Yes, we should expand the file if needed. Working on it!

src/lavinmq/mfile.cr
viktorerlingsson: Yes, you are right. We refactored away from that earlier. Will remove it!

viktorerlingsson:
> What happens with `consumer_offset_capacity` is reached? `MFile` doesn't auto expand.

Fixed [here](implemented [here](https://github.com/cloudamqp/lavinmq/pull/661/commits/367b2e4f292932e71be77069c795fcef8d3578a3))

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -8,11 +8,18 @@ module LavinMQ
       property max_length_bytes : Int64?
       property max_age : Time::Span | Time::MonthSpan | Nil
       getter last_offset : Int64
-      @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
+      @segment_last_ts = Hash(UInt32, Int64).new(0i64)     # used for max-age
+      @consumer_offset_positions = Hash(String, Int64).new # used for consumer offsets
+      @consumer_offsets : MFile
+      @consumer_offset_path : String
+      @consumer_offset_capacity = 32_768
```
carlhoerberg, line 15: Why are these instance variables here? The same information can be accessed with `@consumer_offsets.path` and `@consumer_offsets.capacity`? 
```suggestion
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -49,12 +64,15 @@ module LavinMQ
...
 
-      private def offset_at(seg, pos) : Tuple(Int64, UInt32, UInt32)
+      private def offset_at(seg, pos, retried = false) : Tuple(Int64, UInt32, UInt32)
         return {@last_offset, seg, pos} if @size.zero?
         mfile = @segments[seg]
         msg = BytesMessage.from_bytes(mfile.to_slice + pos)
         offset = offset_from_headers(msg.properties.headers)
         {offset, seg, pos}
+      rescue ex : IndexError # first segment can be empty if message size >= segment size
+        return offset_at(seg + 1, pos, true) unless retried
```
carlhoerberg, line 74: what if many segments has been deleted? this only tired one more segment? and should `pos` be reused? Shouldn't it be in the beginning of the next available segment? 
```suggestion
        return offset_at(@segments.first_key, 4, true) unless retried
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,83 @@ module LavinMQ
...
+        loop do
+          ctag = AMQ::Protocol::ShortString.from_io(@consumer_offsets)
+          break if ctag.empty?
+          positions[ctag] = @consumer_offsets.pos
+          @consumer_offsets.skip(8)
+        rescue IO::EOFError
+          break
+        end
+        @consumer_offsets.pos = 0 if @consumer_offsets.pos == 1
+        @consumer_offsets.resize(@consumer_offsets.pos)
```
carlhoerberg, line 129: so then there's no room left to expand in?

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,83 @@ module LavinMQ
...
+            store_consumer_offset(ctag, offset)
+          end
+        end
+      end
+
+      def replace_offsets_file(&)
+        old_consumer_offsets = @consumer_offsets
+        @consumer_offsets = MFile.new("#{@consumer_offset_path}.tmp", @consumer_offset_capacity)
+        yield # fill the new file with correct data in this block
+        @consumer_offsets.rename(@consumer_offset_path)
```
carlhoerberg, line 182: Use the default size, and use the new `expand_consumer_offset_file` to expand it if needed.

```suggestion
        @consumer_offsets = MFile.new("#{old_consumer_offsets.path}.tmp", 32 * 1024)
        yield # fill the new file with correct data in this block
        @consumer_offsets.rename(old_consumer_offsets.path)
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +106,83 @@ module LavinMQ
...
+
+      def consumer_offset_file_full?(consumer_tag)
+        (@consumer_offsets.size + consumer_tag.bytesize + 8) >= @consumer_offsets.capacity
+      end
+
+      def expand_consumer_offset_file
+        pos = @consumer_offsets.size
+        @consumer_offset_capacity += 32_768
+        @consumer_offsets = MFile.new(@consumer_offset_path, @consumer_offset_capacity)
+        @consumer_offsets.resize(pos)
```
carlhoerberg, line 156: why was `resize` needed? Shouldn't this be enough? 
```suggestion
        @consumer_offsets = MFile.new(@consumer_offsets.path, @consumer_offsets.capacity + 32 * 1024)
```

carlhoerberg (changes requested):

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: Write_bytes writes at the end of the file, and doesn't care about pos, so we need to resize it to pos.

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: Write_bytes writes at the end of the file, so resize is needed.

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: `offset_at()` is always called with `@segments.first_key` (or `@segments.last_key`), so `seg + 1` will always be the second segment (unless we're looking in the last segment, but since it's the last segment, we will never run into the issue of a message spilling over to the next segment in that case). 

And I don't think we ever will have two empty segments in a row? Anytime a message spills over, it will end up in the next segment. The way we expire segments for streams shouldn't allow for any gaps. But maybe getting the next key in the hash is safer? 

`pos `should be set to 4 though 👍

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +104,82 @@ module LavinMQ
...
+
+      def store_consumer_offset(consumer_tag : String, new_offset : Int64)
+        expand_consumer_offset_file if consumer_offset_file_full?(consumer_tag)
+        @consumer_offsets.write_bytes AMQ::Protocol::ShortString.new(consumer_tag)
+        @consumer_offset_positions[consumer_tag] = @consumer_offsets.size
+        @consumer_offsets.write_bytes new_offset
+      end
+
+      def consumer_offset_file_full?(consumer_tag)
+        (@consumer_offsets.size + consumer_tag.bytesize + 8) >= @consumer_offsets.capacity
```
carlhoerberg, line 147: should also include the tag size prefix byte, right? 
 
```suggestion
        (@consumer_offsets.size + 1 + consumer_tag.bytesize + 8) >= @consumer_offsets.capacity
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +104,82 @@ module LavinMQ
...
+          break
+        end
+        @consumer_offsets.pos = 0 if @consumer_offsets.pos == 1
+        @consumer_offsets.resize(@consumer_offsets.pos)
+        positions
+      end
+
+      def update_consumer_offset(consumer_tag : String, new_offset : Int64)
+        if pos = @consumer_offset_positions[consumer_tag]?
+          IO::ByteFormat::SystemEndian.encode(new_offset, @consumer_offsets.to_slice(pos, 8, false))
```
carlhoerberg, line 133: this information is not replicated to followers? i guess `Replicator` doesn't have support for modifying files, so needs to built out.

or rethink, should we do append-only and GC the file instead? Either by assigning each ctag a number, and persist that information too, or by using lz4 compression?

carlhoerberg (changes requested):
Need to think about replication of the consumer offset file.

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: Yes, thanks 👍

src/lavinmq/queue/stream_queue_message_store.cr
viktorerlingsson: > this information is not replicated to followers? i guess Replicator doesn't have support for modifying files, so needs to built out.


Yeah, haven't really thought about replication, but we should obviously do that! 

> or rethink, should we do append-only and GC the file instead? Either by assigning each ctag a number, and persist that information too, or by using lz4 compression?

Hmm, I feel like append-only could cause some issues where the data we need is scattered between potentially large amounts of stale data. But maybe we can handle that by GC'ing (when file is full? or regular intervals) and keep only the latest offset for each ctag? Replacing the file with a new, compacted version. (and expand the file if it's full and there's nothing to GC.) We already keep a hash (`@consumer_offset_positions`) of all tracked ctags and their respective position in the file , so knowing what to keep when compacting should be pretty straight-forward. 

I don't think I understand what you mean by `Either by assigning each ctag a number, and persist that information too, or by using lz4 compression?` though. I guess lz4 would be to use less space on disk in trade-off for a little extra cpu usage? 


And I'm not sure if that's better than building out Replicator to support modifying files? Maybe there are other use cases for supporting modifying files in the future (like log compaction in streams).

viktorerlingsson:
@carlhoerberg Performance seems pretty similar with both solutions, maybe even slightly quicker with append-only (I increased file size to Config.instance.segment_size to make sure we don't have to constantly GC). So I think we can go with that solution 👍

carlhoerberg:
Cool! Havent read all code but looks like the file is still expanded? Is that nessecary? Shouldnt a GC be issued instead? The file size should be a minimum segment size yes, but larger if there are many consumers we are tracking for, so calc the size of the file after GC and the capacity should be a multiple of that, say 1000-10000 larger or so, so that the offset of each consumer could be updated that many times without causing a new GC.

viktorerlingsson:
> Cool! Havent read all code but looks like the file is still expanded? Is that nessecary? Shouldnt a GC be issued instead? The file size should be a minimum segment size yes, but larger if there are many consumers we are tracking for, so calc the size of the file after GC and the capacity should be a multiple of that, say 1000-10000 larger or so, so that the offset of each consumer could be updated that many times without causing a new GC.

Yeah, it will try to compact the file first (only keep the latest offset for each ctag). If the file is still full after that (i.e. there was nothing to compact) it will be expanded. Probably safer to set the capacity to a multiple of tracked ctags as you say though. Will look at that tomorrow!

```
        cleanup_consumer_offsets if consumer_offset_file_full?(consumer_tag)
        expand_consumer_offset_file if consumer_offset_file_full?(consumer_tag)
```

viktorerlingsson:
> Cool! Havent read all code but looks like the file is still expanded? Is that nessecary? Shouldnt a GC be issued instead? The file size should be a minimum segment size yes, but larger if there are many consumers we are tracking for, so calc the size of the file after GC and the capacity should be a multiple of that, say 1000-10000 larger or so, so that the offset of each consumer could be updated that many times without causing a new GC.

@carlhoerberg Added this. Do we need to set some form of max size for the offsets file? capacity could in theory be **very** large if there are many consumers (and/or if ctags are long).

viktorerlingsson:
@carlhoerberg added some replication for the consumer offsets file. Not really sure what the best approach is for when we replace the file during compaction though. Just `replace_file` after we've done the replace? Or do we also want to stream writing to a temp file and then replace the current (and then deleting temp file)? We don't have support for renaming files in replication now I guess

carlhoerberg:
Ah, well spotted, can implement replace file on the client side with write-to-tempfile/rename in another PR

src/lavinmq/mfile.cr
```diff
@@ -231,9 +231,9 @@ class MFile < IO
     Bytes.new(buffer, @size, read_only: true)
   end
 
-  def to_slice(pos, size)
+  def to_slice(pos, size, read_only = true)
     raise IO::EOFError.new if pos + size > @size
-    Bytes.new(buffer + pos, size, read_only: true)
+    Bytes.new(buffer + pos, size, read_only: read_only)
   end
```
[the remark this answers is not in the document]
viktorerlingsson: No, you're right. We removed that when we changed to only appending the consumer offsets file. Removing it!

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -34,7 +36,10 @@ module LavinMQ
           raise LavinMQ::Error::PreconditionFailed.new("x-priority not supported on stream queues")
         end
         case frame.arguments["x-stream-offset"]?
-        when Nil, Int, Time, "first", "next", "last"
+        when Nil
+          @track_offset = true unless @tag.starts_with?("amq.ctag-")
```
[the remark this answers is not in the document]
viktorerlingsson: yes, it's replaced when we get here 👍

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -34,7 +36,10 @@ module LavinMQ
...
+        when Int, Time, "first", "next", "last"
+          @track_offset = true if frame.arguments["x-stream-automatic-offset-tracking"]?
```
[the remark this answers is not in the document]
viktorerlingsson: Ah, correct. `false (Bool)` worked but not `"false" (String)`. Fixed!

PR #662, opened 2024-04-17
set colspan of pagination to nr of columns in the table

viktorerlingsson:
Sets colspan of pagination td to nr of columns in the table to stop it from making the first column in every table very wide if it has enough rows to require pagination. 

before: 
before

after: 
after

Look in UI


spuun (approved):
I wonder how we've missed this. Nice!

PR #663, opened 2024-04-23
export delayed exchanges in a way that is compatible with rabbitmq

kickster97:
rewrites `export_exchanges()` to change the type and arguments of a delayed exchange so it is compatible with RabbitMQ also. 
(we handle the conversion again on import for LavinMQ, so it will also be compatible for LavinMQ) 

add delayed exchange to LavinMQ and export, then import into RabbitMQ, make sure to have the x-delayed-exchange plugin enabled.

### DIFF
--- a/src/lavinmq/http/controller/definitions.cr
+++ b/src/lavinmq/http/controller/definitions.cr
@@ -309,7 +309,7 @@ module LavinMQ
                 {
                   "name":        e.name,
                   "vhost":       e.vhost.name,
-                  "type":        delayed ? type : e.type,
+                  "type":        delayed ? "x-delayed-exchange" : e.type,
                   "durable":     e.durable?,
                   "auto_delete": e.auto_delete?,
                   "internal":    e.internal?,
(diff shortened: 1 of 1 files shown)


src/lavinmq/http/controller/definitions.cr
```diff
@@ -299,14 +299,21 @@ module LavinMQ
...
+                  type = "x-delayed-exchange"
+                  arguments = e.arguments.dup
+                  arguments["x-delayed-type"] = e.type
+                  arguments.delete("x-delayed-exchange")
+                end
                 {
                   "name":        e.name,
                   "vhost":       e.vhost.name,
-                  "type":        e.type,
+                  "type":        delayed ? type : e.type,
```
carlhoerberg, line 312:
```suggestion
                  "type":        delayed ? "x-delayed-exchange" : e.type,
```

src/lavinmq/http/controller/definitions.cr
```diff
@@ -299,14 +299,21 @@ module LavinMQ
           json.array do
             vhosts.each_value do |v|
               v.exchanges.each_value.reject(&.internal?).each do |e|
+                delayed = e.arguments["x-delayed-exchange"]?
+                if delayed
+                  type = "x-delayed-exchange"
```
carlhoerberg, line 304:
```suggestion
```

viktorerlingsson (approved):
LGTM 👍

PR #666, opened 2024-04-25
Abstract client/channel/consumer

spuun:

This will abstract Client/Channel/Consumer so we can implement different types of these entities. The current implementations of each entitiy is moved to the LavinMQ::AMQP namespace.

This abstraction should enable us to implement other protocols.

Run specs


spuun:
I also refactored `Reporter` since it access instance variables. I'll probably break this out to a PR to this PR.

spuun:
I think i'd like to move src/lavinmq/client/amqp/* to src/lavinmq/amqp/* and also include the queue implementations here since they are tied to `AMQP::Consumers`, so make `LavinMQ::AMQP::Queue` and `LavinMQ::AMQP::StreamQueue` and such... Not sure yet.

spuun:
I'll revisit this at some point. Probably starting over.

kickster97:
> I think i'd like to move src/lavinmq/client/amqp/* to src/lavinmq/amqp/* and also include the queue implementations here since they are tied to `AMQP::Consumers`, so make `LavinMQ::AMQP::Queue` and `LavinMQ::AMQP::StreamQueue` and such... Not sure yet.

Makes it easier to introduce other protocols so we don't have to mix e.g. amqp queues and mqtt sessions.. But i'm wondering: should we move **everything AMQP** to under the AMQP namespace? exchanges? etc..

PR #668, opened 2024-04-26
Make Observable/Observer generic

spuun:
This make the event type of Observable/Observer generic. By doing so it's possible to have multiple includes of these modules with different event types and use e.g. enums as events to make it typed.
It also enables different overloads of `Observer#on` for different event types.

Run specs

### DIFF
--- a/src/lavinmq/observable.cr
+++ b/src/lavinmq/observable.cr
@@ -1,19 +1,24 @@
-module Observable
-  @observers = Set(Observer).new
-
-  def register_observer(observer : Observer)
-    @observers.add(observer)
+module LavinMQ
+  module Observer(EventT)
+    abstract def on(event : EventT, data : Object?)
   end
 
-  def unregister_observer(observer : Observer)
-    @observers.delete(observer)
-  end
+  module Observable(EventT)
+    macro included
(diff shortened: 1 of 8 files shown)


carlhoerberg (commented):
use `case ... in ...`

src/lavinmq/observable.cr
```diff
@@ -1,19 +1,24 @@
...
+  module Observer(EventT)
+    abstract def on(event : EventT, data : Object?)
   end
 
-  def unregister_observer(observer : Observer)
-    @observers.delete(observer)
-  end
+  module Observable(EventT)
+    macro included
+      {% ivar = ("@__" + EventT.name.stringify.downcase.gsub(/[^a-z_]+/, "_") + "_observers").id %}
```
carlhoerberg, line 8: maybe give it a more explainable name than the generic `ivar`? 

```suggestion
      {% event_observers = ("@__" + EventT.name.stringify.downcase.gsub(/[^a-z_]+/, "_") + "_observers").id %}
```

carlhoerberg (approved):

src/lavinmq/observable.cr
spuun: Since we're in macro space I want to keep `ivar` in the name ("ivar" seems to be common short name for "instance variable").

Issue #669, opened 2024-04-28
Some of the queues are closed


zzl221000:
lavinmq version: 1.2.5

The status of some queues in the service becomes closed.
I didn't find any documentation about the queues being closed, is there a way to change them to a running status?

viktorerlingsson:
Hi @zzl221000 ,

Could you provide some more information about when this happens? 

What does the log say?

zzl221000:
> What does the log say?

```log
2024-04-26 10:12:31 +00:00 [ERROR] queue[vhost=/ name=icp_inc_queue] - Queue closed due to error
2024-04-26 10:12:31 +00:00 [ERROR] client[vhost=/ address=10.42.0.160:42596] - Unexpected error, while reading: Channel is closed (LavinMQ::Queue::MessageStore::ClosedError)
  from /usr/share/crystal/src/hash.cr:1080:49 in 'delete'
  from /usr/share/crystal/src/mutex.cr:136:7 in 'delete_message'
  from /tmp/src/lavinmq/queue/queue.cr:750:7 in 'ack'
  from /tmp/src/lavinmq/client/channel.cr:483:9 in '??'
  from /tmp/src/lavinmq/client/channel.cr:470:11 in 'basic_ack'
  from /tmp/src/lavinmq/client/client.cr:367:31 in 'process_frame'
  from /tmp/src/lavinmq/client/client.cr:125:13 in 'read_loop'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from ???
2024-04-26 10:12:31 +00:00 [INFO] client[vhost=/ address=10.42.0.160:42596] - Closing, INTERNAL_ERROR - Unexpected error, please report
```
Something seems to have gone wrong.
The service was restarted once before the queue was closed
@viktorerlingsson

viktorerlingsson:
Thanks! 
Could you provide any more information about how to reproduce the issue? Like what is your client trying to do when this happens? 

For getting the queues back up I would suggest trying to restart LavinMQ.

viktorerlingsson:
Could you try upgrading LavinMQ to the latest version (v1.2.11)? There has been some bug fixes since v1.2.5

zzl221000:
I restarted the service when there were more clients consuming (about 200 connections) and a day later some of the queues appeared to be closed.
I checked the consumption code, all using python aiopika library and set up manual ack
[...]

PR #671, opened 2024-05-07
Handle IndexError when reading msg files

viktorerlingsson:
Handles a case where a msg file size is bigger than expected. Tries to move on to the next segment if possible. 

fixes: #669 

Run spec


carlhoerberg (commented):
Let's be more generic than just `BytesMessage::MIN_BYTESIZE`

carlhoerberg (approved):

PR #673, opened 2024-05-10
Fix flaky spec: trigger updates instead of rely on sleeps

spuun:
The `should resume flow when disk is no longer full` flow spec is flaky on macos. This PR try to fix the specs by triggering disk update instead of relying on sleeps and fiber work.

Run specs. For me it failed locally with seed 73372 before this patch.

### DIFF
--- a/spec/flow_spec.cr
+++ b/spec/flow_spec.cr
@@ -37,7 +37,7 @@ describe "Flow" do
 
   it "should stop flow when disk is almost full" do
     LavinMQ::Config.instance.free_disk_min = Int64::MAX
-    Server.system_metrics(nil)
+    Server.update_system_metrics(nil)
     Server.disk_full?.should be_true
   ensure
     LavinMQ::Config.instance.free_disk_min = 0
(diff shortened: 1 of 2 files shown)


carlhoerberg:
Ok, good, but makes me think that `System.system_metrics` should have a better name, reflecting what it's actually doing, eg. `System.update_system_metrics` (and with default parameters).

viktorerlingsson:
👍

I think we should move flow control out of the stats loop  (maybe it's own loop with a shorter sleep? Or actually check disk space every time we open a new file?). I don't think it makes sense to do flow control in the stats loop at least!

kickster97:
> 👍
> 
> I think we should move flow control out of the stats loop (maybe it's own loop with a shorter sleep? Or actually check disk space every time we open a new file?). I don't think it makes sense to do flow control in the stats loop at least!

I agree that it is more clear to separate the flow control from stats loop.. but lets make a separate issue for that :)

viktorerlingsson (approved):

kickster97 (approved):

PR #676, opened 2024-05-14
Refactor Reporter to not know about implementations

spuun:
The `Reporter` class is an odd bird and only used for debugging purposes. It knows too much about classes (even accessing instance variables) which prevents refactoring.

This PR refactors `Reporter` to not know anything about any implementation; instead "reportable" entities must include the `Reportable` module and specify their reportable objects with the `reportables` macro.

The macro may look scary at a first glance, but it will generate a `__report` method that will report all specified "reportables" to the `Reporter` instance. This makes the `Reporter` only responsible for the output, while each entity is responsible for what to report.

Run lavinmq and send `USR1` to observe output.

### DIFF
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -6,12 +6,14 @@ require "../amqp"
 require "../stats"
 require "../sortable_json"
 require "../error"
+require "../reporter"
 
 module LavinMQ
   class Client
     class Channel
       include Stats
       include SortableJSON
+      include Reportable
 
       getter id, name
       property? running = true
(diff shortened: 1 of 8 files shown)


carlhoerberg:
What's the benefit? "Prevents refactoring"? You have to change your `reportables` just the same way as you have to change the `Reporter` today. Now a lot of debugging code is scattered around in individual classes.

spuun:
IMO changing the internals of a class shouldn't affect other classes. I encountered this when I was abstracting client/channel/consumer.

dentarg:
This makes sense. Could be used for general instrumentation and not only debugging?

spuun:
The current approach doesn't work very well if you start to abstract things, like making `Client` an abstract class to introduce amqp clients and mqtt client, since these different client classes may have different internals.

carlhoerberg:
Sure but question is if this is useful at all?

PR #681, opened 2024-05-16
Force 4096 byte frames for websocket connections

carlhoerberg:
To match the frame size of the websocket frames.

The amqp-client.js already defaults to 4096, but for potentially other clients.

### DIFF
--- a/src/lavinmq/client/amqp_connection.cr
+++ b/src/lavinmq/client/amqp_connection.cr
@@ -120,9 +120,10 @@ module LavinMQ
     end
 
     def self.tune(socket)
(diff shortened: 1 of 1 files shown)


src/lavinmq/client/amqp_connection.cr
```diff
@@ -120,9 +120,10 @@ module LavinMQ
     end
 
     def self.tune(socket)
+      frame_max = socket.is_a?(WebSocketIO) ? 4096 : Config.instance.frame_max
```
viktorerlingsson, line 123:
```suggestion
      frame_max = socket.is_a?(WebSocketIO) ? 4096_u32 : Config.instance.frame_max
```

PR #683, opened 2024-05-18
Use Etcd for leader election

carlhoerberg:

Uses etcd for leader election, instances that are not leaders yet will replication from the leader.

* Install `etcd`
* Start etcd or a full cluster:
<details><summary>Etcd cluster</summary>
<p>

```
etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new
```

```
etcd --name infra2 --listen-client-urls http://127.0.0.1:22379 --advertise-client-urls http://127.0.0.1:22379 --listen-peer-urls http://127.0.0.1:22380 --initial-advertise-peer-urls http://127.0.0.1:22380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new 
```

```
etcd --name infra3 --listen-client-urls http://127.0.0.1:32379 --advertise-client-urls http://127.0.0.1:32379 --listen-peer-urls http://127.0.0.1:32380 --initial-advertise-peer-urls http://127.0.0.1:32380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new 
```

</p>
</details> 
* Start two (or more instances of) lavinmq:

```
bin/lavinmq --data-dir=/tmp/amqp1 --clustering --clustering-bind ::
```

```
bin/lavinmq --clustering --data-dir=/tmp/amqp2 --clustering-advertised-uri=tcp://127.2 --clustering-bind=127.2 --bind=127.2
```

* Kill one instance, the other will automatically take over.
* Kill or block/drop different etcd servers and see what happens during netsplit. 

### Todo

- [ ] Specs
- [x] Let follower proxy client connections to leader
  - [x] PROXY protocol
  - [ ] ~Use lz4 compression~
- [x] Maintain a list of in-sync-replicas (followers), in etcd, and don't allow nodes not in that list to become leaders
  - [x] Give each node an unique id, used for ISR (in sync replicas). Let nodes get one from etcd. 
- [x] Make sure followers don't lag behind
- [x] Require at least one follower when replication is enabled
- [x] Don't start replication server when etcd key isn't set (or some other way to enable/disable replication mode)
- [x] Fix bug where leader removes it self from ISR if fails to start
- [x] Don't compact definitions on shutdown to prevent ISR update
- [x] Rename replication to clustering
- [ ] ~Use Txn (etcd transaction) to only allow the leader to change ISR?~
- [x] ISR not updated after failover
- [x] min ISR config
- [x] auto learn new etcd nodes 
- [x] optimize max lag
- [ ] `lavinmqctl reset_isr` that deletes `lavinmq/isr` in etcd?


carlhoerberg:
Problem of doing it to the normal ports is that then we cant do PROXY
protocol header or lz4 compression. Would need new ports for forwarded
traffic.

carlhoerberg:
Forwarding now to the same port on the leader as the port came to, assume that all nodes in the cluster are listening on the same port for the same service. For AMQP connections we now add the PROXY header. On the server side we accept PROXY headers if the source IP is among the followers (if we always accepted users could spoof their IPs).

carlhoerberg:
Rename config parameters from `replication_*` to `clustering_*`? Have a default `etcd_key` and instead have `clustering = true` config setting to enable clustering/replication?

carlhoerberg:
I've updated the instructions on how to run this PR locally.

carlhoerberg:
Also rename `Replication` to `Clustering`? `LavinMQ::Clustering::Client`, `@server.clustering.followers` etc?

carlhoerberg:
Should `min_isr` (mininum in sync replicas) be a config variable? In that case what should the default be 0?

carlhoerberg:
if `min_isr` isn't fulfilled we can't ack queue.declare, publish confirms or consume acks. Those operations will timeout in most clients, is that ok? Or should we send `channel.close` with a proper error message after a ceratin timeout? if there hasn't been a follower for say 30-60 seconds, we start close channels that tries to do stuff?

carlhoerberg:
Currently `clustering_id`, which is used to identify a node in a cluster, if it's in ISR or not, is read from `.clustering_id` in the data dir, or from the config, or by looking at `max_id` in etcd and assigning one from there. But would it be more convenient if an UUID (v4) was generated to `.clustering_id` if that file was missing? And not making it configurable (except editing `.clustering_id` of course). @snichme how was manually configuring the `broker.id` useful in kafka?

snichme:
> if `min_isr` isn't fulfilled we can't ack queue.declare, publish confirms or consume acks. Those operations will timeout in most clients, is that ok? Or should we send `channel.close` with a proper error message after a ceratin timeout? if there hasn't been a follower for say 30-60 seconds, we start close channels that tries to do stuff?

I think it would be helpful if we close the channel and gave the client a helpful message. Timeout errors is difficult to understand and often very non-descriptive. 
But maybe for the first version timeouts is enough and than later we can add better error handling so we dont try to do everything at once.

snichme:
> @snichme how was manually configuring the `broker.id` useful in kafka?

We never really used that in karafka. We just assigned `broker.id` to the number from the server name `server-name-01` would get `broker.id = 1` and so one.  Broker ids in Kafka is a string we any value was possible IIRC.
I don't see any downside in auto-generate an ID as UUIDv4 as long as it's stored in a file that can be moved to a new server if needed.  How about using `hostname` as clustering_id?

src/lavinmq/clustering/controller.cr
```diff
@@ -24,25 +24,13 @@ class LavinMQ::Clustering::Controller
...
-      @config.clustering_id = id
-    end
-    max_id = @etcd.get("#{@config.clustering_etcd_prefix}/max_id") || "-1"
-    if max_id.to_i < id
-      @etcd.put("#{@config.clustering_etcd_prefix}/max_id", id.to_s)
+    id_file_path = File.join(@config.data_dir, ".clustering_id")
+    begin
+      id = File.read(id_file_path).to_i
+    rescue File::NotFoundError
+      id = rand(Int32::MAX)
```
snichme, line 31: I guess there is some risk of collision here,Why not UUID?

carlhoerberg: I would say it's a tolerable risk, only have to be unique within a cluster, of say 3 or 5 nodes, that's a 0,0000002% risk of collision.

carlhoerberg:
Replication performance with different settings of `max_lag`. Two followers, `bin/lavinmqperf throughput -z 10`, on localhost.

| Max lag | Throughput |
|--------|--------|
| 1024 | 714855 |
| 4096 | 772735 |
| 8192 | 799570 |
| 16384 | 800835 |
| 32768 | 785868 | 
| 65536 | 653360 |

With 1024 bytes messages:

| Max lag | Throughput |
|--------|--------|
| 1024 | 409100 |
| 4096 | 423685 |
| 8192 | 437658 |
| 16384 | 445155 |

So let's go with default `max_lag` 8192

PR #684, opened 2024-05-22
Dont run append if @followers is empty

viktorerlingsson:
Replication::Server#append is called from several places, sometimes with `append(path : String, file : MFile, position : Int32, length : Int32)`. That creates a new FileRange even though there are no followers, using unnecessary resources. This returns early from append if the follower array is empty. 

### DIFF
--- a/src/lavinmq/replication/server.cr
+++ b/src/lavinmq/replication/server.cr
@@ -54,12 +54,12 @@ module LavinMQ
       end
 
       def append(path : String, file : MFile, position : Int32, length : Int32)
-        return if @followers.size.zero?
+        return if @followers.empty?
         append path, FileRange.new(file, position, length)
       end
 
       def append(path : String, obj)
-        return if @followers.size.zero?
+        return if @followers.empty?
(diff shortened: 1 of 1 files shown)


spuun (approved):
IMO this shows that we should refactor something; it looks kind of strange to have that if statement in only one of the "action" methods.

carlhoerberg:
this introduced a bug, if a follower is currently syncing the publish is not blocked (as before this PR), and can cause data loss at the follower

carlhoerberg:
> IMO this shows that we should refactor something; it looks kind of strange to have that if statement in only one of the "action" methods.

same could be added to all actions, but `append` is the only one in the hot path.

viktorerlingsson:
> this introduced a bug, if a follower is currently syncing the publish is not blocked (as before this PR), and can cause data loss at the follower

Ah, yes. [Reverted ](https://github.com/cloudamqp/lavinmq/commit/cae46556e30dacf49bb02a511e37ea03041dcaa0)

PR #685, opened 2024-05-22
Handle empty msg files on start

### LINKED ISSUE #624: Outdated schema version 0 for /var/lib/lavinmq/42099b4af021e53fd8fd4e056c2568d7c2e3ffa8/97c1c36cf4b13e7f763f7ca5
hello i'm new to lavinmq

i'm installing lavinmq on my ubuntu laptop for experiment, but i get this error log when starting lavinmq service 
(found from command => journalctl -xeu lavinmq.service)

Outdated schema version 0 for /var/lib/lavinmq/42099b4af021e53fd8fd4e056c2568d7c2e3ffa8/97c1c36cf4b13e7f763f7ca5

after i delete the folder then restart, it works again, why this is happening ?
If my experiment run smoothly, i'm planning to replace rabbitmq with lavinmq on my production server

viktorerlingsson:
Handles empty msg files on startup and sets Schema version to default. 

I think this can happen if a queue is created and lavinmq is killed before anything is written to the file. 

Fixes #624 

Run specs

### DIFF
--- a/spec/schema_version_spec.cr
+++ b/spec/schema_version_spec.cr
@@ -29,7 +29,7 @@ describe LavinMQ::SchemaVersion do
         file.resize(LavinMQ::Config.instance.segment_size)
         # init new message store
         msg_store = LavinMQ::Queue::MessageStore.new(data_dir, nil)
-        msg_store.@segments.first[1].size.should eq 4
+        msg_store.@segments.first_value.size.should eq 4
       end
     end
 
--- a/spec/schema_version_spec.cr
+++ b/spec/schema_version_spec.cr
@@ -3,11 +3,11 @@ require "../src/lavinmq/schema"
 
 describe LavinMQ::SchemaVersion do
   describe "Schema version" do
-    it "Empty file should raise EmptyFile" do
+    it "Empty file should raise IO::EOFError" do
       with_datadir do |data_dir|
         path = File.join(data_dir, "test_schema_version")
         file = MFile.new(path, 12)
-        expect_raises(LavinMQ::EmptyFile) do
+        expect_raises(IO::EOFError) do
           LavinMQ::SchemaVersion.verify(file, :message)
         end
       end
--- a/spec/schema_version_spec.cr
+++ b/spec/schema_version_spec.cr
@@ -1,0 +1,23 @@
+require "./spec_helper"
+require "../src/lavinmq/schema"
+
+describe LavinMQ::SchemaVersion do
+  describe "Schema version" do
+    it "Empty file should raise EmptyFile" do
+      data_dir = "/tmp/lavinmq-spec"
+      path = File.join(data_dir, "test_schema_version")
+      file = MFile.new(path, 12)
+      expect_raises(LavinMQ::EmptyFile) do
+        LavinMQ::SchemaVersion.verify(file, :message)
+      end
+    end
+
+    it "Should verify schema version" do
+      data_dir = "/tmp/lavinmq-spec"
+      path = File.join(data_dir, "test_schema_version")
+      file = MFile.new(path, 12)
+      file.write_bytes LavinMQ::Schema::VERSION
+      LavinMQ::SchemaVersion.verify(file, :message).should eq LavinMQ::SchemaVersion::VERSIONS[:message]
+    end
+  end
+end
--- a/spec/schema_version_spec.cr
+++ b/spec/schema_version_spec.cr
@@ -19,5 +19,27 @@ describe LavinMQ::SchemaVersion do
       file.write_bytes LavinMQ::Schema::VERSION
       LavinMQ::SchemaVersion.verify(file, :message).should eq LavinMQ::SchemaVersion::VERSIONS[:message]
     end
+
+    it "Deletes empty file and creates a new when it is the first file" do
+      data_dir = "/tmp/lavinmq-spec"
+      path = File.join(data_dir, "msgs.0000000001")
+      file = MFile.new(path, LavinMQ::Config.instance.segment_size)
+      file.resize(LavinMQ::Config.instance.segment_size)
+      # init new message store
+      msg_store = LavinMQ::Queue::MessageStore.new(data_dir, nil)
+      msg_store.@segments.first[1].size.should eq 4
+    end
+
+    it "Deletes empty file when it is second file" do
+      v = Server.vhosts["/"].not_nil!
+      v.declare_queue("q", true, false)
+      data_dir = Server.vhosts["/"].queues["q"].@msg_store.@data_dir
+      path = File.join(data_dir, "msgs.0000000002")
+      file = MFile.new(path, LavinMQ::Config.instance.segment_size)
+      file.resize(LavinMQ::Config.instance.segment_size)
+      # init new message store
+      msg_store = LavinMQ::Queue::MessageStore.new(data_dir, nil)
+      msg_store.@segments.size.should eq 1
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


src/lavinmq/schema.cr
```diff
@@ -290,10 +290,15 @@ module LavinMQ
       buf = uninitialized UInt8[4]
       file.read_at(0, buf.to_slice)
       version = IO::ByteFormat::SystemEndian.decode(Int32, buf.to_slice)
+      if version == 0 # if version is 0, read 8 more bytes(ts) and check if that's also 0. If so, the file is empty, set version to default.
+        raise EmptyFile.new if IO::ByteFormat::SystemEndian.decode(Int64, file.to_slice(4, 8)).zero?
```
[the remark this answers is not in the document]
viktorerlingsson: `file.size` is` == segment_size`, so that wouldn't work I think

viktorerlingsson:
> Interesting that this is happening. Is it because an ungraceful shutdown won't truncate the file?

Yes, but I think it has to happen before an fsync as well, because otherwise we would write schema version when opening a segment?

carlhoerberg:
Yes, would be good to understand the root cause of this, although i agree with the fix, if first 16 bytes are 0 most likely the rest is 0 too. But we always write schema version to the `mmap` when a new segment is opened, even if lavinmq crashed exactly after that it would still be written to disk (changes to the `mmap` is persisted on `unmap`/crash), only if there was a powerloss or disk failure would it not be written to disk. Or lavin crashed before writing the schema version, but there's like nanoseconds between the file was created and the schema version is written. https://github.com/cloudamqp/lavinmq/blob/136b743800c2c538340cc34695f5e82ddf0dfe48/src/lavinmq/queue/message_store.cr#L248-L249

src/lavinmq/queue/message_store.cr
```diff
@@ -321,6 +321,14 @@ module LavinMQ
           end
           file.pos = 4
           @segments[seg] = file
+        rescue EmptyFile
+          Log.warn { "Empty file at #{path}, setting schema version to #{Schema::VERSION}" }
+          file = file.not_nil!
+          file.resize(0)
+          file.write_bytes Schema::VERSION
+          @replicator.not_nil!.try &.append path.not_nil!, Schema::VERSION
```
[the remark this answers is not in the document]
viktorerlingsson: This should be solved now when we delete & recreate files, right?

spec/schema_version_spec.cr
```diff
@@ -0,0 +1,53 @@
+require "./spec_helper"
+require "../src/lavinmq/schema"
+
+describe LavinMQ::SchemaVersion do
+  describe "Schema version" do
+    it "Empty file should raise EmptyFile" do
+      with_amqp_server do |_s|
```
carlhoerberg, line 7: do you need an amqp server for this spec?

spec/schema_version_spec.cr
```diff
@@ -0,0 +1,53 @@
...
+        data_dir = "/tmp/lavinmq-spec"
+        path = File.join(data_dir, "test_schema_version")
```
carlhoerberg, line 9: `DATA_DIR` is a constant from spec_helper.cr
```suggestion
        path = File.join(DATA_DIR, "test_schema_version")
```

spec/schema_version_spec.cr
```diff
@@ -0,0 +1,53 @@
...
+
+    it "Deletes empty file and creates a new when it is the first file" do
+      with_amqp_server do |_s|
+        data_dir = "/tmp/lavinmq-spec"
+        path = File.join(data_dir, "msgs.0000000001")
+        file = MFile.new(path, LavinMQ::Config.instance.segment_size)
+        file.resize(LavinMQ::Config.instance.segment_size)
+        # init new message store
+        msg_store = LavinMQ::Queue::MessageStore.new(data_dir, nil)
+        msg_store.@segments.first[1].size.should eq 4
```
carlhoerberg, line 35:
```suggestion
        msg_store.@segments.first_value.size.should eq 4
```

spec/schema_version_spec.cr
```diff
@@ -0,0 +1,53 @@
...
+        data_dir = "/tmp/lavinmq-spec"
+        path = File.join(data_dir, "test_schema_version")
+        file = MFile.new(path, 12)
+        file.write_bytes LavinMQ::Schema::VERSION
+        LavinMQ::SchemaVersion.verify(file, :message).should eq LavinMQ::SchemaVersion::VERSIONS[:message]
+      end
+    end
+
+    it "Deletes empty file and creates a new when it is the first file" do
+      with_amqp_server do |_s|
```
carlhoerberg, line 28:
```suggestion
```

src/lavinmq/schema.cr
```diff
@@ -319,4 +324,7 @@ module LavinMQ
       super "Cannot migrate #{path} from version #{@version}"
     end
   end
+
+  class EmptyFile < Exception
+  end
```
carlhoerberg, line 329:
```suggestion
class EmptyFile < Exception; end
```

src/lavinmq/queue/message_store.cr
```diff
@@ -317,7 +317,22 @@ module LavinMQ
...
           else
-            SchemaVersion.verify(file, :message)
+            begin
+              SchemaVersion.verify(file, :message)
+            rescue EmptyFile
+              # delete empty file, it will be recreated if it's needed
+              Log.warn { "Empty file at #{path}, deleting it" }
+              file.delete.close
+              @replicator.try &.delete_file(path)
+              if idx == 0
```
carlhoerberg, line 327: how come we recreate the first file but not others? Maybe add a comment about that.

spec/schema_version_spec.cr
viktorerlingsson: No, just need a data_dir. Changed to using `with_datadir `instead!

src/lavinmq/queue/message_store.cr
viktorerlingsson: We need at least one segment to exist, otherwise we get errors in a lot of places (for example where we use `first_/last_key` & `first_/last_value`). And for a file that isn't the first I don't think we need to recreate it? Better to let it be recreated when/if needed IMO. 

Maybe could be changed to check if it's the only segment instead? But I don't see how we would end up with an empty file followed by other files. 

Added a comment about it!

src/lavinmq/schema.cr
carlhoerberg, line 294: Want to use `MFile#read_at` to not cause a `mmap` of the whole file. 
```suggestion
        buf = uninitialized UInt8[8]
        file.read_at(0, buf.to_slice)
        raise EmptyFile.new if IO::ByteFormat::SystemEndian.decode(Int64, buf.to_slice).zero?
```

Could also check that it is just zeros this way:

```suggestion
        buf = uninitialized UInt8[8]
        file.read_at(0, buf.to_slice)
        raise EmptyFile.new if buf.all?(&.zero?)
```

And then is a new exception required or could `IO::EOFError` be used? It's not technically true, but in practice.

```suggestion
        buf = uninitialized UInt8[8]
        file.read_at(0, buf.to_slice)
        raise IO::EOFError.new if buf.all?(&.zero?)
```

viktorerlingsson: > And then is a new exception required or could IO::EOFError be used? It's not technically true, but in practice.

No, we can use EOFError, that feels cleaner 👍

carlhoerberg (approved):

PR #687, opened 2024-05-31
User connection endpoint

kickster97:
Adds the endpoint to list connections for a specific username. 

send a get request to the endpoint  "connections/username/:username", should work for the user in question and any user with administrator or monitoring tags.

### DIFF
--- a/spec/api/connections_spec.cr
+++ b/spec/api/connections_spec.cr
@@ -102,11 +102,8 @@ describe LavinMQ::HTTP::ConnectionsController do
   describe "GET /api/connections/username/:username" do
     it "returns connections for a specific user" do
       Server.users.create("arnold", "pw", [LavinMQ::Tag::Administrator])
-      Server.users.add_permission("arnold", "/", /.*/, /.*/, /.*/)
       hdrs = HTTP::Headers{"Authorization" => "Basic YXJub2xkOnB3"}
       with_channel do
-        response = delete("/api/connections/username/arnold", headers: hdrs)
-        response.status_code.should eq 204
         response = get("/api/connections/username/arnold", headers: hdrs)
         body = JSON.parse(response.body)
         body.as_a.empty?.should be_true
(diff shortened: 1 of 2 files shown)


src/lavinmq/http/controller/connections.cr
```diff
@@ -48,6 +48,27 @@ module LavinMQ
...
+          reason = context.request.headers["X-Reason"]? || "Closed via management plugin"
+          connections.each do |c|
+            c.close(reason)
+          end
+          context.response.status_code = 204
+          context
+        end
+      end
+
+      private def get_connections_by_username(context, params)
```
snichme, line 68: Method says `by_username` but you are giving it `params` So I think you should extract username in the caller method and give this method a "username" not params. 
Otherwise the method should be called `get_connections_by_params` but that is really difficult to understand.

src/lavinmq/http/controller/connections.cr
```diff
@@ -48,6 +48,27 @@ module LavinMQ
...
+        username = URI.decode_www_form(params["username"])
+        user = user(context)
+        connections(user).select { |c| c.user.name == username }
```
snichme, line 71: This method should also be allowed for any user that is admin right?  

So an admin can list and close connections for another user

snichme (commented):
I know this is tested via rabbit-hole but I think we should have a unit test in lavin for this endpoint anyway.

src/lavinmq/http/controller/connections.cr
kickster97: `connections(users)` sorts that out, or?

spec/api/connections_spec.cr
```diff
@@ -98,4 +98,34 @@ describe LavinMQ::HTTP::ConnectionsController do
       end
     end
   end
+
+  describe "GET /api/connections/username/:username" do
+    it "returns connections for a specific user" do
```
spuun, line 103: Copy-paste mistake? This spec looks the same as the next one?

spuun (commented):
See inline comment

spec/api/connections_spec.cr
kickster97: hehe yes! just about to push the changes :)

snichme (approved):

PR #688, opened 2024-05-31
Allow array of URIs when creating shovel

kickster97:
Allows for multiple URIs to be sent in when creating a shovel. This input can now be either a string, or an array of strings. If more than one uri string is provided, the shovel will randomly pick one URI from the list until one of the endpoints succeeds.
Needed for [compatibility](https://www.rabbitmq.com/docs/shovel-dynamic#:~:text=Note%20that%20this%20field%20can%20either%20be%20a%20string%2C%20or%20a%20list%20of%20strings.%20If%20more%20than%20one%20string%20is%20provided%2C%20the%20shovel%20will%20randomly%20pick%20one%20URI%20from%20the%20list%20until%20one%20of%20the%20endpoints%20succeeds.) with clients

still needs some small adjustments

Try to create a shovel and send in multiple uris

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -167,6 +167,7 @@ module LavinMQ
 
     class DestinationWrapper < Destination
       @current_dest : Destination?
+
       def initialize(@destinations : Array(Destination))
       end
 
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -39,23 +39,28 @@ module LavinMQ
 
       getter delete_after, last_unacked
 
-      def initialize(@name : String, @uri : URI, @queue : String?, @exchange : String? = nil,
+      def initialize(@name : String, @uris : Array(URI), @queue : String?, @exchange : String? = nil,
                      @exchange_key : String? = nil,
                      @delete_after = DEFAULT_DELETE_AFTER, @prefetch = DEFAULT_PREFETCH,
                      @ack_mode = DEFAULT_ACK_MODE, consumer_args : Hash(String, JSON::Any)? = nil,
                      direct_user : User? = nil)
         @tag = "Shovel"
-        unless @uri.user
-          if direct_user
-            @uri.user = direct_user.name
-            @uri.password = direct_user.plain_text_password
-          else
-            raise ArgumentError.new("direct_user required")
+        cfg = Config.instance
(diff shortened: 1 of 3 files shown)


src/lavinmq/shovel/shovel_store.cr
```diff
@@ -8,6 +8,13 @@ module LavinMQ
 
     forward_missing_to @shovels
 
+    def parse_uris(src_uri) : Array(URI)
```
spuun, line 11: I guess the argument is `JSON::Any`? IMO we can be explicit about since `as_s` and such are JSON stuff. At first I thought the type was `String | Array(String)`.

```suggestion
    def parse_uris(src_uri : JSON::Any) : Array(URI)
```

src/lavinmq/shovel/shovel_store.cr
```diff
@@ -8,6 +8,13 @@ module LavinMQ
...
+      uris = src_uri.as_s? ? [src_uri.as_s] : src_uri.as_a.map(&.as_s)
```
spuun, line 12: This could probably be
```suggestion
      uris = src_uri.as_s? || src_uri.as_a.map(&.as_s)
```

src/lavinmq/shovel/shovel.cr
```diff
@@ -162,6 +165,36 @@ module LavinMQ
       abstract def started? : Bool
     end
 
+    class DestinationWrapper < Destination
```
spuun, line 168: `DestinationWrapper` sounds like an abstract name to me. Maybe we can figure out something more descriptive? `MultiDestination` or something.. hm.

spuun (approved):
Minor comments but nothing that _must_ be fixed.

snichme (approved):

src/lavinmq/shovel/shovel.cr
kickster97: `MultiDestination`, or `UniversalDestination` could work

src/lavinmq/shovel/shovel_store.cr
kickster97: does not work for map, 
`undefined method 'map' for String (compile-time type is (Array(String) | String))`

src/lavinmq/shovel/shovel.cr
kickster97: chose MultiDestinationHandler

PR #689, opened 2024-05-31
Add tags and description to vhost

kickster97:
add tags and description to vhost, and add them to details tuple. 
needed for client [compatibility](https://www.rabbitmq.com/docs/vhosts#:~:text=A%20description,set%20of%20tags)

discuss: do we want this? 

`GET /vhosts` response should have description and tags


snichme (approved):

spuun (commented):
When retrieving a vhost via http api in rabbitmq the root object contains (among others) `description (string)`,  `tags (array)` and `metadata (object)`. The metadata object contains `description (string)` and  `tags (array)`. 

Dunno why it has that metadata object and values in the root object.

Anyway, based on this I'd say we should store tags as an array.

src/lavinmq/http/controller/vhosts.cr
```diff
@@ -33,11 +33,14 @@ module LavinMQ
           u = user(context)
           refuse_unless_administrator(context, u)
           name = URI.decode_www_form(params["name"])
+          body = parse_body(context)
+          tags = body["tags"]?.to_s.split(",") || [] of String
```
spuun, line 37: I think we should remove white spaces
```suggestion
          tags = body["tags"]?.to_s.split(",").map(&:strip) || [] of String
```

Maybe also remove empty strings?

I noticed that in rabbitmq tag string `a,b,, c,d,` results in tag array `["a","b","","c","d",""]`, which I don't think need to mimic.

carlhoerberg:
```suggestion
          tags = body["tags"]?.to_s.split(',').map(&.strip).reject(&.empty?)
```

src/lavinmq/http/controller/vhosts.cr
```diff
@@ -33,11 +33,14 @@ module LavinMQ
...
+          description = body["description"]?.try(&.to_s) || ""
```
carlhoerberg, line 38: `nil.to_s == ""`
```suggestion
          description = body["description"]?.to_s
```

src/lavinmq/vhost.cr
```diff
@@ -42,7 +42,8 @@ module LavinMQ
     @definitions_file_path : String
     @definitions_deletes = 0
 
-    def initialize(@name : String, @server_data_dir : String, @users : UserStore, @replicator : Replication::Replicator)
+    def initialize(@name : String, @tags : Array(String), @description : String,
+                   @server_data_dir : String, @users : UserStore, @replicator : Replication::Replicator)
```
carlhoerberg, line 46:
```suggestion
    def initialize(@name : String, @server_data_dir : String, @users : UserStore, @replicator : Replication::Replicator, @description = "", @tags = Array(String).new(0))
```

spec/vhost_spec.cr
```diff
@@ -99,7 +99,7 @@ describe LavinMQ::VHost do
       username = "test-user"
       user = Server.users.create(username, "password", [LavinMQ::Tag::Administrator])
       vhost = "test-vhost"
-      Server.vhosts.create(vhost, user)
+      Server.vhosts.create(vhost, [""], "", user)
```
carlhoerberg, line 102: make tags and description optional

carlhoerberg:
```suggestion
      Server.vhosts.create(vhost, user)
```

src/lavinmq/vhost_store.cr
```diff
@@ -20,11 +20,11 @@ module LavinMQ
       end
     end
 
-    def create(name : String, user : User = @users.default_user, save : Bool = true)
+    def create(name : String, tags : Array(String) = [] of String, description : String = "", user : User = @users.default_user, save : Bool = true)
```
carlhoerberg, line 23:
```suggestion
    def create(name : String, user : User = @users.default_user, description = "", tags = Array(String).new(0), save : Bool = true)
```

carlhoerberg (approved):

PR #693, opened 2024-06-03
Use Stoplight Elements to render OpenAPI docs

carlhoerberg:

Instead of compiling static html files from `openapi.yaml` using `redoc` the docs are now rendered in the browser using https://github.com/stoplightio/elements

Compile and then browse to http://localhost:15672/docs/


carlhoerberg:
Required (or a variant of it) by #692. Installing nodejs with a recent version in all distros (ubuntu 20.04, debian 11) is a hassle, and just required for building the static html.

static/docs/index.html
```diff
@@ -0,0 +1,20 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <title>LavinMQ HTTP API</title>
+    <script src="https://unpkg.com/@stoplight/elements/web-components.min.js"></script>
+    <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements/styles.min.css">
```
dentarg, line 8: It is probably a good idea lock these to a certain version so they does not break on us unexpectedly. Using Subresource Integrity (SRI) we can also ensure that it is the correct version served from the CDN: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

snichme (approved):
Looks good!

static/docs/index.html
carlhoerberg: or download locally?

carlhoerberg: https://github.com/cloudamqp/lavinmq/pull/693/commits/e91b966906ab4ac3f28329edd91e106fa7579ccc

PR #697, opened 2024-06-12
Refactor: inject data dir to replication classes

spuun:
This is to get rid of some references to `Config.instance` so simplify specs and future refactoring.

Instead of injecting the config instance the data dir is injected. Data dir shouldn't change during runtime, so no need to read data dir from a config instance that may change values on a reload.

`Config.instance` is still referenced from Follower, so maybe we should inject the config instance to `Replication::Server` and `Replication::Follower`, but that can be done in another round.

Run specs


carlhoerberg (approved):
:+1:

snichme (approved):

PR #699, opened 2024-06-14
Bugfix: export definitions changes delayed exchanges

### LINKED ISSUE #698: Definitions export changes delayed exchanges' definition causing 406 on redeclare
Since 1.2.12 you may get a `406 PRECONDITION_FAILED` when re-declaring a delayed exchange. It only happens if a definition export has been done between declaration and re-declaration.

Steps to reproduce:
1. `curl -u guest:guest http://localhost:15672/api/exchanges/%2F/delayed-ex -XPUT -d '{"type":"x-delayed-message", "durable": true, "arguments": {"x-delayed-type": "direct"}}' -H 'Content-Type: application/json'`
2. `curl -u guest:guest http://localhost:15672/api/definitions/%2F`
3. `curl -u guest:guest http://localhost:15672/api/exchanges/%2F/delayed-ex -XPUT -d '{"type":"x-delayed-message", "durable": true, "arguments": {"x-delayed-type": "direct"}}' -H 'Content-Type: application/json'`

spuun:
Fixes #698

Run specs

### DIFF
--- a/src/lavinmq/http/controller/definitions.cr
+++ b/src/lavinmq/http/controller/definitions.cr
@@ -301,7 +301,7 @@ module LavinMQ
               v.exchanges.each_value.reject(&.internal?).each do |e|
                 delayed = e.arguments["x-delayed-exchange"]?
                 if delayed
-                  arguments = e.arguments.dup
+                  arguments = e.arguments.clone
                   arguments["x-delayed-type"] = e.type
                   arguments.delete("x-delayed-exchange")
                 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Bugfix: internal definition of delayed exchanges changed after definitions export [#699](https://github.com/cloudamqp/lavinmq/pull/699)
 
 ## [1.2.12] - 2024-05-24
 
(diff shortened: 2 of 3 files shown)


src/lavinmq/http/controller/definitions.cr
```diff
@@ -301,7 +301,7 @@ module LavinMQ
               v.exchanges.each_value.reject(&.internal?).each do |e|
                 delayed = e.arguments["x-delayed-exchange"]?
                 if delayed
-                  arguments = e.arguments.dup
```
snichme, line 304: What type is `e.arguments` here and what it the difference between `dup` and `clone`? :)

spuun: `e.arguments` is a [AMQ::Protocol::Table](https://github.com/cloudamqp/amq-protocol.cr/blob/main/src/amq/protocol/table.cr).

`dup` will only do a shallow copy and is defined on `Object`. `clone` is a deep copy and defined by the implementation of the class/struct.

In this case a `Table` has a `IO::Memory` and with `dup` the copy will reference the same IO which means that modifying the copy also modifies the original. `Table#clone` will create a new IO and copy the data: https://github.com/cloudamqp/amq-protocol.cr/blob/6a70d2b6857e3bdd2d69b319e57ff3cbefddaddb/src/amq/protocol/table.cr#L28-L32

viktorerlingsson (approved):
LGTM!

carlhoerberg (approved):

PR #704, opened 2024-06-17
Set up specs with before_build.sh for rabbith-hole specs

kickster97:
LavinMQ needs to run before_build with lavinmqctl to set up spec env in order to have the right users created for tests to run and pass. 

run script locally, or look at github actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -358,6 +358,8 @@ jobs:
           cd rabbit-hole
           bash bin/ci/before_build.sh
           go test -v ./...
+        env:
+          RABBITHOLE_LAVINMQCTL: ../bin/lavinmqctl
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -356,6 +356,7 @@ jobs:
       - name: Run specs
         run: |
           cd rabbit-hole
+          bash bin/ci/before_build.sh
           go test -v ./...
```
carlhoerberg, line 360: the path to `lavinmqctl` is wrong, and is `sudo` really required?
```suggestion
          go test -v ./...
        env:
          RABBITHOLE_LAVINMQCTL: ../bin/lavinmqctl
```

carlhoerberg (changes requested):

PR #709, opened 2024-06-20
Show follower id in follower list

spuun:
Now when we have ids on followers we should display it in the follower list.

Run a cluster and look in the follower list.


carlhoerberg (approved):

spuun:
Hm, `Id` or `ID`?

PR #710, opened 2024-06-20
Run frontend specs on ubuntu 24.04

spuun:
Changing to ubuntu 24.04 for frontend specs

Run CI


spuun:
We should wait for playwright to support ubuntu 24.04

PR #711, opened 2024-06-20
clean up trailing '/' in path before routing

kickster97:
clean up any trailing '/' in path before routing, fixes failing rabbit-hole specs

run rabbit-hole specs

### DIFF
--- a/src/lavinmq/http/router.cr
+++ b/src/lavinmq/http/router.cr
@@ -33,7 +33,7 @@ module LavinMQ::HTTP::Router
   {% end %}
 
   def find_route(method, path)
-    search_path = "/#{method}#{path.sub(/\/$/, "")}"
+    search_path = "/#{method}#{path.rstrip('/')}"
     @_routes.each do |r|
       if res = r.pattern.match(search_path)
         return {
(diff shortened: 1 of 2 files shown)


src/lavinmq/http/router.cr
```diff
@@ -33,7 +33,7 @@ module LavinMQ::HTTP::Router
   {% end %}
 
   def find_route(method, path)
-    search_path = "/#{method}#{path}"
+    search_path = "/#{method}#{path.sub(/\/$/, "")}"
```
spuun, line 36: There is a method for this :)
```suggestion
    search_path = "/#{method}#{path.rstrip('/')}"
```

spuun (commented):
It's even simpler!

spuun:
Oh, actually, we should only rstrip if it ends with slash!

spuun:
> Oh, actually, we should only rstrip if it ends with slash!

Uhm, and if path is longer than 1. Maybe just do 
```
    search_path = "/#{method}/#{path.strip('/')}"
```

kickster97:
> search_path = "/#{method}/#{path.strip('/')}"

ah, assuming this is for handling the `get "/"` cases?

spuun:
> > search_path = "/#{method}/#{path.strip('/')}"
> 
> ah, assuming this is for handling the `get "/"` cases?

Yes!

spuun (approved):
Let's add a spec for the "/" case, then merge!

spuun (approved):

PR #712, opened 2024-06-22
Show information about unacked messages

### LINKED ISSUE #686: Show information about unacked messages
Add views/API endpoints that shows information about unacked messages in a queue. See comment below for more information https://github.com/cloudamqp/lavinmq/issues/686#issuecomment-2135565645

viktorerlingsson:
Adds an api endpoint and a view to display information about unacked messages in a queue. 

Fixes #686 

- Update docs

Run specs

### DIFF
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -384,6 +384,9 @@ module LavinMQ
             @client.vhost.event_tick(EventType::ClientGet)
             ok = q.basic_get(frame.no_ack) do |env|
               delivery_tag = next_delivery_tag(q, env.segment_position, frame.no_ack, nil)
+              unless frame.no_ack # track unacked messages
+                q.basic_get_unacked << UnackedMessage.new(delivery_tag, "", @client.channels.first_value.name, RoughTime.monotonic)
+              end
               get_ok = AMQP::Frame::Basic::GetOk.new(frame.channel, delivery_tag,
                 env.redelivered, env.message.exchange_name,
                 env.message.routing_key, q.message_count)
@@ -482,6 +485,7 @@ module LavinMQ
           c.ack(unack.sp)
         end
         unack.queue.ack(unack.sp)
+        unack.queue.basic_get_unacked.reject! { |u| u.message_tag == unack.tag }
         @client.vhost.event_tick(EventType::ClientAck)
         @ack_count += 1
       end
@@ -560,6 +564,7 @@ module LavinMQ
           c.reject(unack.sp, requeue)
         end
         unack.queue.reject(unack.sp, requeue)
+        unack.queue.basic_get_unacked.reject! { |u| u.message_tag == unack.tag }
         @reject_count += 1
         @client.vhost.event_tick(EventType::ClientReject)
       end
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -31,7 +31,6 @@ module LavinMQ
       @next_msg_size = 0_u64
       @next_msg_props : AMQP::Properties?
       @delivery_tag = 0_u64
-      # @unacked = Deque(Unack).new
       @unack_lock = Mutex.new(:checked)
       @next_msg_body_file : File?
       @direct_reply_consumer : String?
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -31,7 +31,7 @@ module LavinMQ
       @next_msg_size = 0_u64
       @next_msg_props : AMQP::Properties?
       @delivery_tag = 0_u64
-      #@unacked = Deque(Unack).new
+      # @unacked = Deque(Unack).new
       @unack_lock = Mutex.new(:checked)
       @next_msg_body_file : File?
       @direct_reply_consumer : String?
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -21,6 +21,7 @@ module LavinMQ
       getter prefetch_count = 0_u16
       getter global_prefetch_count = 0_u16
       getter has_capacity = ::Channel(Nil).new
+      getter unacked = Deque(Unack).new
       @confirm = false
       @confirm_total = 0_u64
       @next_publish_mandatory = false
@@ -30,7 +31,7 @@ module LavinMQ
       @next_msg_size = 0_u64
       @next_msg_props : AMQP::Properties?
       @delivery_tag = 0_u64
-      @unacked = Deque(Unack).new
+      #@unacked = Deque(Unack).new
       @unack_lock = Mutex.new(:checked)
       @next_msg_body_file : File?
       @direct_reply_consumer : String?
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -385,7 +385,7 @@ module LavinMQ
             ok = q.basic_get(frame.no_ack) do |env|
               delivery_tag = next_delivery_tag(q, env.segment_position, frame.no_ack, nil)
               unless frame.no_ack # track unacked messages
-                q.basic_get_unacked << UnackedMessage.new(delivery_tag, "", @client.channels.first_value.name, RoughTime.monotonic)
+                q.basic_get_unacked << UnackedMessage.new(self, delivery_tag, RoughTime.monotonic)
               end
               get_ok = AMQP::Frame::Basic::GetOk.new(frame.channel, delivery_tag,
                 env.redelivered, env.message.exchange_name,
@@ -485,7 +485,7 @@ module LavinMQ
           c.ack(unack.sp)
         end
         unack.queue.ack(unack.sp)
-        unack.queue.basic_get_unacked.reject! { |u| u.message_tag == unack.tag }
+        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
         @client.vhost.event_tick(EventType::ClientAck)
         @ack_count += 1
       end
@@ -564,7 +564,7 @@ module LavinMQ
           c.reject(unack.sp, requeue)
         end
         unack.queue.reject(unack.sp, requeue)
-        unack.queue.basic_get_unacked.reject! { |u| u.message_tag == unack.tag }
+        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
         @reject_count += 1
         @client.vhost.event_tick(EventType::ClientReject)
       end
@@ -637,6 +637,7 @@ module LavinMQ
           @unacked.each do |unack|
             @log.debug { "Requeing unacked msg #{unack.sp}" }
             unack.queue.reject(unack.sp, true)
+            unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
           end
           @unacked.clear
         end
--- a/src/lavinmq/client/channel.cr
+++ b/src/lavinmq/client/channel.cr
@@ -485,7 +485,7 @@ module LavinMQ
           c.ack(unack.sp)
         end
         unack.queue.ack(unack.sp)
-        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
+        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.delivery_tag == unack.tag }
         @client.vhost.event_tick(EventType::ClientAck)
         @ack_count += 1
       end
@@ -564,7 +564,7 @@ module LavinMQ
           c.reject(unack.sp, requeue)
         end
         unack.queue.reject(unack.sp, requeue)
-        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
+        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.delivery_tag == unack.tag }
         @reject_count += 1
         @client.vhost.event_tick(EventType::ClientReject)
       end
@@ -637,7 +637,7 @@ module LavinMQ
           @unacked.each do |unack|
             @log.debug { "Requeing unacked msg #{unack.sp}" }
             unack.queue.reject(unack.sp, true)
-            unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
+            unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.delivery_tag == unack.tag }
           end
           @unacked.clear
         end
--- a/src/lavinmq/http/controller/queues.cr
+++ b/src/lavinmq/http/controller/queues.cr
@@ -45,12 +45,12 @@ module LavinMQ
         get "/api/queues/:vhost/:name/unacked" do |context, params|
           with_vhost(context, params) do |vhost|
             refuse_unless_management(context, user(context), vhost)
-            unacked_messages = Array(UnackedMessage).new
-            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
-              c.unacked_messages.each do |u|
-                next unless u.queue.name == params["name"]
+            consumers = @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each
+            unacked_messages = consumers.flat_map do |c|
+              c.unacked_messages.each.compact_map do |u|
+                  next unless u.queue.name == params["name"]
                 if consumer = u.consumer
-                  unacked_messages << UnackedMessage.new(
+                  UnackedMessage.new(
                     u.tag, consumer.tag,
                     (RoughTime.monotonic - u.delivered_at).to_i,
                     c.channel_name
@@ -58,7 +58,10 @@ module LavinMQ
                 end
               end
             end
-            page(context, unacked_messages.each)
+            unacked_messages = unacked_messages.chain(@amqp_server.vhosts[vhost].queues[params["name"]].@basic_get_unacked.each.flat_map do |b|
+              UnackedMessage.new(0, "basic_get", 0, "basic_get")
+            end)
+            page(context, unacked_messages)
           end
         end
 
--- a/src/lavinmq/http/controller/queues.cr
+++ b/src/lavinmq/http/controller/queues.cr
@@ -47,7 +47,7 @@ module LavinMQ
             refuse_unless_management(context, user(context), vhost)
             unacked_messages = Hash(UInt64, UnackedMessage).new
             @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
-              c.@channel.@unacked.reject{ |u| u.queue.name != params["name"] }.each do |u|
+              c.@channel.@unacked.reject { |u| u.queue.name != params["name"] }.each do |u|
                 if consumer = u.consumer
                   unacked_messages[u.tag] = UnackedMessage.new(u.tag, consumer.tag, u.delivered_at.to_s)
                 end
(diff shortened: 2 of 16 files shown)


src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,21 @@ module LavinMQ
           end
         end
 
+        get "/api/queues/:vhost/:name/unacked" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            unacked_messages = Hash(UInt64, UnackedMessage).new
+            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
+              c.@channel.@unacked.reject { |u| u.queue.name != params["name"] }.each do |u|
```
carlhoerberg, line 50: `reject` generates a new array, but `each.reject` will "stream" the result, without generating a new array. 

```suggestion
              c.@channel.@unacked.each.select { |u| u.queue.name == params["name"] } do |u|
```

carlhoerberg: and accessing instance variables with `@` in a class should be avoided. Instead we can expose a `unacked` method on `Consumer` that does this.

src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,21 @@ module LavinMQ
...
+                if consumer = u.consumer
+                  unacked_messages[u.tag] = UnackedMessage.new(u.tag, consumer.tag, u.delivered_at.to_s)
+                end
+              end
+            end
+            page(context, unacked_messages.each_value)
```
carlhoerberg, line 56: if we only use the values (`each_value`), why was a hash created to begin with and not an array?

src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,21 @@ module LavinMQ
...
         end
 
+        get "/api/queues/:vhost/:name/unacked" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            unacked_messages = Hash(UInt64, UnackedMessage).new
+            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
+              c.@channel.@unacked.reject { |u| u.queue.name != params["name"] }.each do |u|
+                if consumer = u.consumer
+                  unacked_messages[u.tag] = UnackedMessage.new(u.tag, consumer.tag, u.delivered_at.to_s)
```
carlhoerberg, line 52: consumer tag is only unique within a Channel, should we keep a reference to the Channel?

carlhoerberg (commented):
Cool!

src/lavinmq/http/controller/queues.cr
carlhoerberg: `deliviered_at` is `Time.monotonic`, so the value doesn't mean anything in it self. What you probably want to do is something like `(RoughTime.monotonic - u.delivered_at).total_seconds`, to get how long it has been unacked for.

src/lavinmq/unacked_message.cr
```diff
@@ -0,0 +1,22 @@
+module LavinMQ
+  class UnackedMessage
```
carlhoerberg, line 2: can this be a struct?

viktorerlingsson: Yes!

src/lavinmq/http/controller/queues.cr
viktorerlingsson: Oops, remnant of something I tried earlier. Changed it.

src/lavinmq/http/controller/queues.cr
viktorerlingsson: > consumer tag is only unique within a Channel, should we keep a reference to the Channel?

Yes, that's probably a good idea! Added channel.name and a link to the channel on the unacked-page

viktorerlingsson: > deliviered_at is a Time::Span, so the value doesn't mean much in it self. What you probably want to do is something like (RoughTime.monotonic - u.delivered_at).total_seconds, to get how long it has been unacked for.

Yes 👍 Should we do total_seconds as float or just to_i? Not sure if we need to be more exact than whole seconds

views/unacked.ecr
spuun, line 1: We should add playwright tests for this view! :)

viktorerlingsson: Yes, added a very simple one 👍 
Not sure if we should do any more frontend testing on this? There's not much functionality on this page and it re-uses most frontend stuff from `Table`/`DataSource`

spuun: Just testing that the ajax request is performed is probably enough 👍

static/js/unacked.js
```diff
@@ -0,0 +1,32 @@
...
+const vhost = search.get('vhost')
+const url = `api/queues/${encodeURIComponent(vhost)}/${encodeURIComponent(queue)}/unacked`
+const channelBaseUrl = 'channel#name='
+document.getElementById('queue-link').href = `/queue#vhost=${encodeURIComponent(vhost)}&name=${encodeURIComponent(queue)}`
+document.getElementById('queue-link').innerHTML = queue
+document.querySelector('#pagename-label').textContent = queue + ' in virtual host ' + vhost
+
+const tableOptions = {
+  dataSource: new UrlDataSource(url, { useQueryState: false }),
+  keyColumns: ['message_tag', 'consumer_tag', 'channel_name', 'unacked_for_seconds'],
```
spuun, line 15: Not sure, but `message_tag` and `channel_name` should be enough as key?

```suggestion
  keyColumns: ['message_tag', 'channel_name'],
```

static/js/unacked.js
```diff
@@ -0,0 +1,32 @@
...
+  search: false,
+  countId: 'unacked-count'
+}
+
+Table.renderTable('table', tableOptions, function (tr, item, firstRender) {
+  if (!firstRender) return
+  Table.renderCell(tr, 0, item.message_tag)
+  Table.renderCell(tr, 1, item.consumer_tag)
+  const channel = document.createElement('a')
+  channel.href = channelBaseUrl + item.channel_name
```
spuun, line 28: No need for uri encoding `channel_name` here?

views/unacked.ecr
```diff
@@ -0,0 +1,35 @@
...
+        </h3>
+        <div class="table-wrapper">
+          <div id="table-error"></div>
+          <table id="table" class="table">
+            <thead>
+              <tr>
+                <th class="left">Message tag</th>
+                <th class="left">Consumer tag</th>
+                <th class="left">Channel</th>
+                <th class="left">Seconds unacked</th>
```
spuun, line 24: Maybe we should decide on how to align different type of columns? My thinking is that integers should be right-aligned?

spuun (approved):
LGTM

static/js/unacked.js
viktorerlingsson: Hmm, it works for me without, but probably safest to encode it.

static/js/unacked.js
spuun: Ah! That's because this is done:
https://github.com/cloudamqp/lavinmq/blob/b4b9045f7af679714308bd1c1fb673e1ee412c92/static/js/unacked.js#L23-L33

Try this instead:
```javascript
Table.renderTable('table', tableOptions, function (tr, item, firstRender) {
  if (firstRender) {
    Table.renderCell(tr, 0, item.message_tag)
    Table.renderCell(tr, 1, item.consumer_tag)
    const channel = document.createElement('a')
    channel.href = channelBaseUrl + encodeURIComponent(item.channel_name)
    channel.textContent = item.channel_name
    Table.renderCell(tr, 2, channel)
  }
  Table.renderCell(tr, 3, item.unacked_for_seconds, 'right')
  Table.renderCell(tr, 4, item.prefetch_count, 'right')
})
```

viktorerlingsson: Aha! 👍

src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,25 @@ module LavinMQ
           end
         end
 
+        get "/api/queues/:vhost/:name/unacked" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            unacked_messages = Array(UnackedMessage).new
+            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
+              c.unacked_messages.select { |u| u.queue.name == params["name"] }.each do |u|
+                if consumer = u.consumer
```
carlhoerberg, line 51: what about messages that are unacked via BasicGet, we don't want display them?

src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,25 @@ module LavinMQ
           end
         end
 
+        get "/api/queues/:vhost/:name/unacked" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            unacked_messages = Array(UnackedMessage).new
+            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
+              c.unacked_messages.select { |u| u.queue.name == params["name"] }.each do |u|
```
carlhoerberg, line 50: take care to not create arrays if we don't need to: 

```suggestion
              c.unacked_messages.each.select { |u| u.queue.name == params["name"] } do |u|
```

src/lavinmq/unacked_message.cr
```diff
@@ -0,0 +1,25 @@
...
+    getter consumer_tag : String
+    getter unacked_for_seconds : Int64
+    getter channel_name : String
+
+    def initialize(message_tag, consumer_tag, unacked_for_seconds, channel_name)
+      @message_tag = message_tag
+      @consumer_tag = consumer_tag
+      @unacked_for_seconds = unacked_for_seconds
+      @channel_name = channel_name
+    end
```
carlhoerberg, line 14:
```suggestion
    getter message_tag, consumer_tag, unacked_for_seconds, channel_name

    def initialize(@message_tag : UInt64, @consumer_tag : String, @unacked_for_seconds : Int64, @channel_name : String)
    end
```

static/js/unacked.js
```diff
@@ -0,0 +1,33 @@
+import * as Table from './table.js'
+import { UrlDataSource } from './datasource.js'
+
+const search = new URLSearchParams(window.location.hash.substring(1))
+const queue = search.get('name')
+const vhost = search.get('vhost')
+const url = `api/queues/${encodeURIComponent(vhost)}/${encodeURIComponent(queue)}/unacked`
+const channelBaseUrl = 'channel#name='
+document.getElementById('queue-link').href = `/queue#vhost=${encodeURIComponent(vhost)}&name=${encodeURIComponent(queue)}`
```
carlhoerberg, line 9: let it work when /api is submapped into another path, so don't use absolute urls. 

```suggestion
document.getElementById('queue-link').href = `queue#vhost=${encodeURIComponent(vhost)}&name=${encodeURIComponent(queue)}`
```

static/js/unacked.js
```diff
@@ -0,0 +1,33 @@
...
+document.getElementById('queue-link').innerHTML = queue
+document.querySelector('#pagename-label').textContent = queue + ' in virtual host ' + vhost
```
carlhoerberg, line 11: prefer string interpolation
```suggestion
document.querySelector('#pagename-label').textContent = `${queue} in virtual host ${vhost}`
```

carlhoerberg (commented):
i don't know how, but i guess we should show the messsages unacked via BasicGet too

src/lavinmq/http/controller/queues.cr
viktorerlingsson: Wait, that doesn't work, we need the each at the end for the loop? I removed the select and added a next in the loop instead. 
```
c.unacked_messages.each do |u|
  next unless u.queue.name == params["name"]
```

src/lavinmq/http/controller/queues.cr
viktorerlingsson: Hmm, yeah we probably do! Not sure how we would do that though, we only really save them as increasing/decreasing `unacked_count` and `unacked_bytesize` now right? So we don't really have enough information. 
Maybe we could add an Array(UnackedMessage) to queues where we save any unacked via BasicGet? 

Something like this maybe? (but actually working 😄 )
```
queue_unacked = Array(UnackedMessage.new

private def mark_unacked(sp, &)
      @log.debug { "Counting as unacked: #{sp}" }
      @unacked_lock.synchronize do
        @unacked_count += 1
        @unacked_bytesize += sp.bytesize
        queue_unacked.push UnackedMessage.new(..) # track unacked messages:
```

and do something like `queue_unacked.delete(msg)` in `basic_ack `& `basic_reject` and in the rescue in `mark_unacked`

WDYT?

src/lavinmq/http/controller/queues.cr
spuun: I guess we can get rid of `unacked_messages = Array(UnackedMessage).new` too by making sure everything is iterators?

src/lavinmq/http/controller/queues.cr
```diff
@@ -41,6 +42,26 @@ module LavinMQ
           end
         end
 
+        get "/api/queues/:vhost/:name/unacked" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            unacked_messages = Array(UnackedMessage).new
+            @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each do |c|
+              c.unacked_messages.each do |u|
```
spuun, line 50: This will remove the array:
```suggestion
            consumers = @amqp_server.vhosts[vhost].queues[params["name"]].@consumers.each
            unacked_messages = consumers.flat_map do |c|
              c.unacked_messages.each.compact_map do |u|
```

spuun: … and pass `unacked_messages` to `page`, not `unacked_messages.each`.

viktorerlingsson: Thanks! Will give it a try tonight or tomorrow 👍

src/lavinmq/http/controller/queues.cr
carlhoerberg: yes, let's store basicget messages in a `Deque` in `Queue`, shouldn't be that many

carlhoerberg: no need to store the unacked messages deliviered to consumers.

viktorerlingsson: I have something that kinda works in my branch now, but not really sure what we want to store/display for the basic_get unacked messages. They don't really have a consumer_tag, channel_name or delivery_tag. Guess we can have the time for how long it's been unacked at least.

carlhoerberg:
Basic get messages should have a channel name (they belong to a channel),
and has a delivery tag (they can be ack/rejected)

src/lavinmq/queue/queue.cr
```diff
@@ -32,6 +32,7 @@ module LavinMQ
     @consumers = Array(Client::Channel::Consumer).new
     @consumers_lock = Mutex.new
     @message_ttl_change = Channel(Nil).new
+    @basic_get_unacked = Deque(Envelope).new
```
carlhoerberg, line 35: why storing the whole envelope? Only need the `SegmentPosition`, `Channel` and `delivery_tag`, no?

src/lavinmq/queue/queue.cr
```diff
@@ -651,6 +652,7 @@ module LavinMQ
       @queue_expiration_ttl_change.try_send? nil
       @get_count += 1
       get(no_ack) do |env|
+        @basic_get_unacked << env unless no_ack # track unacked messages
```
carlhoerberg, line 655: this shoudl proably be done in another place, where we have the channel and deliverytag

src/lavinmq/queue/queue.cr
viktorerlingsson: Yeah, was thinking we might need some info from the msg, but since we won't then we don't!

src/lavinmq/queue/queue.cr
viktorerlingsson: updated to do this in `channel.basic_get` instead

src/lavinmq/client/channel.cr
```diff
@@ -384,6 +384,9 @@ module LavinMQ
             @client.vhost.event_tick(EventType::ClientGet)
             ok = q.basic_get(frame.no_ack) do |env|
               delivery_tag = next_delivery_tag(q, env.segment_position, frame.no_ack, nil)
+              unless frame.no_ack # track unacked messages
+                q.basic_get_unacked << UnackedMessage.new(delivery_tag, "", @client.channels.first_value.name, RoughTime.monotonic)
```
carlhoerberg, line 388: Why not?
```suggestion
                q.basic_get_unacked << UnackedMessage.new(delivery_tag, "", @name, RoughTime.monotonic)
```

but in addition to that i think `UnackedMessage` should include a reference to the actual `Channel`. 
```suggestion
                q.basic_get_unacked << UnackedMessage.new(delivery_tag, "", self, RoughTime.monotonic)
```
And no basic-get message ever has a consumer tag so: 
```suggestion
                q.basic_get_unacked << UnackedMessage.new(self, delivery_tag RoughTime.monotonic)
```

src/lavinmq/client/channel.cr
```diff
@@ -482,6 +485,7 @@ module LavinMQ
           c.ack(unack.sp)
         end
         unack.queue.ack(unack.sp)
+        unack.queue.basic_get_unacked.reject! { |u| u.message_tag == unack.tag }
```
carlhoerberg, line 488: message tag is not unique, you have to refernce the channel too. 

```suggestion
        unack.queue.basic_get_unacked.reject! { |u| u.channel == self && u.message_tag == unack.tag }
```

src/lavinmq/queue/queue.cr
```diff
@@ -33,6 +33,7 @@ module LavinMQ
     @consumers_lock = Mutex.new
     @message_ttl_change = Channel(Nil).new
 
+    getter basic_get_unacked = Deque(UnackedMessage).new
```
carlhoerberg, line 36: are you removing the messages from this Dequeue when a channel closes?

carlhoerberg (changes requested):

src/lavinmq/client/channel.cr
viktorerlingsson: changed!

src/lavinmq/queue/queue.cr
viktorerlingsson: no, added removal to `close `and a spec for it now

src/lavinmq/queue/queue.cr
```diff
@@ -687,7 +688,7 @@ module LavinMQ
           end
           delete_message(sp)
         else
-          mark_unacked(sp) do
+          mark_unacked(env) do
```
carlhoerberg, line 691: This change is not needed anymore is it?

src/lavinmq/unacked_message.cr
```diff
@@ -0,0 +1,18 @@
...
+  struct UnackedMessage
+    include SortableJSON
+    getter message_tag, consumer_tag, channel, delivered_at
+
+    def initialize(@channel : LavinMQ::Client::Channel, @message_tag : UInt64, @delivered_at : Time::Span, @consumer_tag : String? = nil)
+    end
+
+    def details_tuple
+      {
+        message_tag:         @message_tag,
```
carlhoerberg, line 11: message_tag should be delivery_tag no? I dont think we use the name message_tag anywhere else.

viktorerlingsson:
> The openapi yaml doesnt need an update now when message_tag renamed to delivery_tag?

This change should be enough, no? 
https://github.com/cloudamqp/lavinmq/pull/712/commits/393aecf62e7040aaab1420b864940e0fec655ee2#diff-0fa9d75567bde2a04b564335b08d2f6dc6324077d0a612102b9d29d56d2bfd9d

> Do we display in the ui that messages without a consumer tag are retrieved by BasicGet?

Added so that will be displayed now!

carlhoerberg (approved):
Nice!

PR #713, opened 2024-06-23
Add a Logger to handle Log metadata

spuun:
I think we need some syntax sugar around all emit metadata calls

This will add a `Logger` to wrap all emit metadata Log stuff.

Look at code, test run.


carlhoerberg:
Looks good!

kickster97 (approved):
nice!

carlhoerberg (approved):

PR #719, opened 2024-06-30
Cluster proxy unix

carlhoerberg:

Clustering follower can forward AMQP/HTTP from UNIX socket if configured.  

```
bin/lavinmq --clustering --data-dir=/tmp/amqp2 --clustering-advertised-uri=tcp://127.2 --clustering-bind=127.2 --bind=127.2 --amqp-unix-path=/tmp/amqp --
http-unix-path=/tmp/http
```

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -54,7 +54,7 @@ module LavinMQ
           spawn http_proxy.forward_to(host, @config.http_port), name: "HTTP proxy"
         end
         if unix_amqp_proxy = @unix_amqp_proxy
-          spawn unix_amqp_proxy.forward_to(host, @config.amqp_port, true), name: "AMQP proxy"
+          spawn unix_amqp_proxy.forward_to(host, @config.amqp_port), name: "AMQP proxy"
         end
         if unix_http_proxy = @unix_http_proxy
           spawn unix_http_proxy.forward_to(host, @config.http_port), name: "HTTP proxy"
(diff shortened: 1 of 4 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -49,6 +53,12 @@ module LavinMQ
         if http_proxy = @http_proxy
           spawn http_proxy.forward_to(host, @config.http_port), name: "HTTP proxy"
         end
+        if unix_amqp_proxy = @unix_amqp_proxy
+          spawn unix_amqp_proxy.forward_to(host, @config.amqp_port, true), name: "AMQP proxy"
```
spuun, line 57: This will set `@proxy_header = true` in `Proxy`, but `@server` is a `UNIXServer` so all clients will be `UNIXSocket`.

However, `@proxy_header == true` only affects `TCPSocket` clients in `#handle_client` so something is weird.

Or am I missing something?

carlhoerberg: ah, yes, copy paste error. makes no sense to forward the local/remote address of the UNIX socket to the leader.

src/lavinmq/clustering/proxy.cr
```diff
@@ -8,13 +8,20 @@ module LavinMQ
...
       def initialize(bind_host, bind_port)
-        @server = TCPServer.new(bind_host, bind_port)
+        @server = s = TCPServer.new(bind_host, bind_port)
+        @local_address = s.local_address.to_s
+      end
+
+      def initialize(path : String)
+        @server = s = UNIXServer.new(path)
+        @local_address = s.local_address.to_s
       end
```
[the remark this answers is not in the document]
carlhoerberg: maybe, but if injecting any `Socket+` the `@local_socket` will be complicated/ugly, not all Sockets got a `#local_address`. 

Yes, was thinking of subclassing, but got a lot more lines of code for very marginal benefit.

Issue #721, opened 2024-07-01
LavinMQ in clustering mode crashes if etcd is restarted


carlhoerberg:
It should not crash, not until the TTL has ran out and it still hasn't been able to connect to any etcd node.

PR #722, opened 2024-07-01
Clustering etcd restart

carlhoerberg:

Stop listen for leader changes after a node has become leader. `elect_observe` doesn't handle that a etcd node is restarted, so this is a short term fix for #721  

No specs yet.


carlhoerberg:
So far only fixes the leader, a follower will still crash if etcd is restarted.

carlhoerberg:
Should probably put all the refactoring in its own PR

PR #723, opened 2024-07-03
don't strip debug symbols in deb packages

carlhoerberg:

With `dh_strip --keep-debug` debug symbols was placed in external files, but for some reason it doesn't seem to work. Now skipping stripping symbols completely. 

Install deb, call something that raises an exception.


oskgu360 (approved):
Looking good!

Gave it a spin on a ubuntu 22.04 machine, and seems to be kept now!

carlhoerberg:
Thank you for trying it out @oskgu360!

PR #726, opened 2024-07-05
make keyColumns look for unique combinations on bindings tables

### LINKED ISSUE #568: Unstable API result of /api/exchanges/<vhost>/<exchange>/bindings/source
The `/api/exchanges/<vhost>/<exchange>/bindings/source` results (visible on `/exchange#vhost=<vhost>&name=<exchange>` UI path) are random and this is visible on every UI refresh. The shown bindings change position on every refresh. It looks like the API call result is unstable and I wonder how pagination could work in this case.

I tested this on LavinMQ 1.1.7 and 1.2.2.

kickster97:
bindings tables were rendering in random because it did not have a unique column combination for keeping track of rows. 
this adds a unique combination of columns keys so they don't re-render out of order each update. 

fixes #568 

create queues exchanges and bindings, and look at UI to see that the bindings don't jump around in the table.

### DIFF
--- a/static/js/exchange.js
+++ b/static/js/exchange.js
@@ -49,7 +49,7 @@ updateExchange()
 const tableOptions = {
   dataSource: new UrlDataSource(exchangeUrl + '/bindings/source', { useQueryState: false }),
   pagination: true,
-  keyColumns: ['properties_key']
+  keyColumns: ['properties_key', 'destination', 'routing_key']
 }
 const bindingsTable = Table.renderTable('bindings-table', tableOptions, function (tr, item, all) {
   if (!all) return
--- a/static/js/exchange.js
+++ b/static/js/exchange.js
@@ -49,7 +49,7 @@ updateExchange()
 const tableOptions = {
   dataSource: new UrlDataSource(exchangeUrl + '/bindings/source', { useQueryState: false }),
   pagination: true,
-  keyColumns: ['properties_key', 'destination', 'routing_key']
+  keyColumns: ['destination', 'properties_key']
 }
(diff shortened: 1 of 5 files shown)


static/js/exchange.js
```diff
@@ -49,7 +49,7 @@ updateExchange()
 const tableOptions = {
   dataSource: new UrlDataSource(exchangeUrl + '/bindings/source', { useQueryState: false }),
   pagination: true,
-  keyColumns: ['properties_key']
+  keyColumns: ['properties_key', 'destination', 'routing_key']
```
carlhoerberg, line 52: Are you sure `routing_key` is required? `properties_key` should already include the routing key: https://github.com/cloudamqp/lavinmq/blob/9c2ae1ce4089630479e9e67b919ee84d8d0ae593/src/lavinmq/exchange/exchange.cr#L233-L244 

```suggestion
  keyColumns: ['destination', 'properties_key']
```

static/js/queue.js
```diff
@@ -144,7 +144,7 @@ setInterval(updateQueue, 5000)
 
 const tableOptions = {
   dataSource: new UrlDataSource(queueUrl + '/bindings', { useQueryState: false }),
-  keyColumns: ['properties_key'],
+  keyColumns: ['properties_key', 'source', 'routing_key'],
```
carlhoerberg, line 147:
```suggestion
  keyColumns: ['source', 'properties_key'],
```

static/js/exchange.js
kickster97: yes you're right! 
why put properties_key last in the array?

carlhoerberg (approved):

PR #732, opened 2024-07-17
Immix GC

carlhoerberg:

Replaces the [Boehm GC](https://github.com/ivmai/bdwgc/) with [Immix GC](https://github.com/ysbaddaden/gc) by @ysbaddaden

Build with `-Dgc_none` to activate

`rm bin/lavinmq && make bin/lavinmq`

run benchmarks such as `bin/lavinmqperf queue-churn`

Initial benchmarks shows about 67% faster queue churn rate.

But also that is somewhat buggy, crashes on the `lavinmqperf bind-churn` benchmark.


ysbaddaden:
Oh my :flushed: 

> 67% faster queue churn rate

That's surprising. I remember it was a little faster than Boehm overall, but collections are much slower (not parallel, yet).

Maybe having objects allocated at the same time usually being allocated next to each others (immix) rather than objects of the same size being allocated together (boehm) is helping with data locality?

> that is somewhat buggy

Damn.

PR #735, opened 2024-07-22
only consider MIN_ISR requirements if listeners are started

### LINKED ISSUE #727: Be able to start a new cluster with min_isr > 0
Currently can't start a new cluster with min_isr > 0 because the writes requires nodes, but the server hasn't started its listeners yet so no followers can connect, catch 22.

viktorerlingsson:
Ignores MIN_ISR requirements until listener has been started. Fixes #727 

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -202,7 +202,6 @@ module LavinMQ
           Log.warn { "ISR requirement not met (#{isr_count}/#{@config.clustering_min_isr})" }
           isr_count = @has_followers.receive? || return
         end
-        end
         @lock.synchronize do
           update_isr if @dirty_isr
           @followers.each do |f|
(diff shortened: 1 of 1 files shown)


viktorerlingsson:
This seems to work for starting with `min_isr > 0`. WDYT @carlhoerberg ?

carlhoerberg:
I'm just wondering if we should have min_isr at all,  Postgres doesnt, Kafka only doesnt ack publishes when min isr isnt fulfilled. Maybe we should do the same? Only not confirm publishes? 

But at least this doesnt brakes the "guarantee", if another server with the exact same data dir starts they would end up with the same data anyway.

viktorerlingsson:
> I'm just wondering if we should have min_isr at all, Postgres doesnt, Kafka only doesnt ack publishes when min isr isnt fulfilled. Maybe we should do the same? Only not confirm publishes?

Hmm, yeah, I think that makes sense! 

> But at least this doesnt brakes the "guarantee", if another server with the exact same data dir starts they would end up with the same data anyway.

👍

carlhoerberg:
But probably have to be in combination with storing metadata/definitions in etcd.

carlhoerberg:
Fixed in #797

PR #736, opened 2024-07-22
add signal traps for INT and TERM for followers

viktorerlingsson:
Adds signal traps for INT and TERM to properly shut down follower and close sockets. 

Without this I end up with extra unix & http sockets when closing a follower with ctrl+c. 

Manual

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -41,7 +41,7 @@ module LavinMQ
         Signal::TERM.trap { close_and_exit }
       end
 
-      def close_and_exit
+      private def close_and_exit
         Log.info { "Received termination signal, shutting down..." }
         close
         exit 0
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -33,6 +33,18 @@ module LavinMQ
           @unix_amqp_proxy = Proxy.new(@config.unix_path) unless @config.unix_path.empty?
           @unix_http_proxy = Proxy.new(@config.http_unix_path) unless @config.http_unix_path.empty?
         end
+        setup_signal_traps
+      end
+
+      private def setup_signal_traps
+        Signal::INT.trap { close_and_exit }
+        Signal::TERM.trap { close_and_exit }
+      end
+
+      def close_and_exit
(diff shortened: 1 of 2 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -33,6 +33,18 @@ module LavinMQ
           @unix_amqp_proxy = Proxy.new(@config.unix_path) unless @config.unix_path.empty?
           @unix_http_proxy = Proxy.new(@config.http_unix_path) unless @config.http_unix_path.empty?
         end
+        setup_signal_traps
+      end
+
+      private def setup_signal_traps
+        Signal::INT.trap { close_and_exit }
+        Signal::TERM.trap { close_and_exit }
+      end
```
carlhoerberg, line 42:
```suggestion
        Signal::INT.trap { close_and_exit }
        Signal::TERM.trap { close_and_exit }
      end
```

src/lavinmq/clustering/client.cr
```diff
@@ -33,6 +33,18 @@ module LavinMQ
...
+
+      def close_and_exit
```
carlhoerberg, line 44:
```suggestion
      private def close_and_exit
```

carlhoerberg (commented):
How come sockets was "left over"? If the process exits all sockets opened by that process is closed as well?

viktorerlingsson:
> How come sockets was "left over"? If the process exits all sockets opened by that process is closed as well?

Yeah, I still had the "files" for `lavinmq.sock` and `lavinmq-http.sock` after closing a follower with ctrl+c, and could not restart the follower unless I removed them. Would they remain if the sockets weren't properly unbound or something?

src/lavinmq/clustering/client.cr
viktorerlingsson: Sure, I just copied from launcher :)

carlhoerberg:
I see, you need to try to delete the file before just like in `Launcher`: https://github.com/cloudamqp/lavinmq/blob/6bb3de850c37854272af5a49d31c5382d07b5e1d/src/lavinmq/server.cr#L182

Can happen if the clustering client was killed (eg. OOM), even with this patch.

kickster97 (approved):

PR #737, opened 2024-07-24
Perf refactor

carlhoerberg:
use `WaitGroup` instead of `Channel(Nil)` and `puts` instead of many `print`

`lavinmqperf throughput -C 1000 -D 1000`


viktorerlingsson (approved):
LGTM 👍

PR #742, opened 2024-07-30
use correct chart config for axis label

kickster97:
use the correct syntax for chart.js config when adding y-axis labels, so now the label will actually render.

start LavinMQ and look at graphs


carlhoerberg:
Looks good!

PR #747, opened 2024-08-07
Abstract AMQP client, consumer, channel

kickster97:
Abstract the client, consumer, channel, amqp_connection classes into the AMQP namespace. This is to prepare our codebase for introducing multiple protocols (MQTT support, etc..) 

run specs


spuun (approved):

carlhoerberg:
We now lost all git blame history on these files :face_with_diagonal_mouth:

PR #748, opened 2024-08-08
Always use x-dead-letter-exchange when deliver a delayed message

spuun:
When an exchange is delayed it will create a queue and use message expiration and dead lettering for the delay. When a message is published to the exchange it will put the message in the "delay queue", and when the message is expired it will be republished to the message's exchange and the message is routed.

However, it's possible to publish messages to this queue via the default exchange and then the exchange field of the message is an empty string. This caused the message to be republished to the default exchange instead of the delayed exchange, resulting in the message looping between the queue and default exchange causing lavinmq to freeze.

This PR will make sure that the `x-dead-letter-exchange` argument for the queue is used when the messages is republished due to expiration.

Run specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wait for followers to synchronize on shutdown of leader
 - Shovel the exact number of messages available on start if `delete-after=queue-length`, not more
 - Prevet a queue that's overflowing to consume too much resources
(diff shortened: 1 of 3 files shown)


kickster97 (approved):
nice :)

carlhoerberg:
Shouldnt we prevent publishing directly to that internal queue instead? That the default exchange cant route to the queue?

carlhoerberg:
This will cause a performance hit as the headers have to be parsed for each message twice. Not critical for this exchange type, but we should be aware of it.

spuun:
Yes, this is a "quick fix" and then we need to discuss how to limit internal queues.

spuun:
Btw, which header parsing are you referring too? `arguments`?

carlhoerberg:
That's much better!

PR #749, opened 2024-08-09
Bugfix. federation binds to default exchange

spuun:
If the upstream configuration is created without an exchange it's stored as an empty string. Currently only a nil check is done if the downstream exchange should be used, meaning that the empty string is used and the link tries to bind to the upstream default exchange.

This will add a check for empty string.

Run specs

### DIFF
--- a/src/lavinmq/federation/upstream.cr
+++ b/src/lavinmq/federation/upstream.cr
@@ -63,7 +63,10 @@ module LavinMQ
         if link = @ex_links[federated_exchange.name]?
           return link
         end
-        upstream_exchange = @exchange ||= federated_exchange.name
+        upstream_exchange = @exchange || federated_exchange.name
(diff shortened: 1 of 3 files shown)


src/lavinmq/federation/upstream.cr
```diff
@@ -63,7 +63,10 @@ module LavinMQ
         if link = @ex_links[federated_exchange.name]?
           return link
         end
-        upstream_exchange = @exchange ||= federated_exchange.name
+        upstream_exchange = @exchange || federated_exchange.name
+        if upstream_exchange.nil? || upstream_exchange.empty?
```
kickster97, line 67: if we get true for `.nil?` here, we will reassign it `nil`? since if `@exchange` is `nil`, we will have assigned `federated_exchange.name `
Do we really need to do the check for `upstream_exchange.nil?` then?

kickster97 (approved):

PR #750, opened 2024-08-09
Cleanup upstream exchange

kickster97:
If a federation gets deleted, we do a cleanup process where we are supposed to delete the internal queue that the federation created on the upstream server. 
this PR adds that we also delete the internal exchange. 

specs coming

### DIFF
--- a/src/lavinmq/federation/link.cr
+++ b/src/lavinmq/federation/link.cr
@@ -324,7 +324,6 @@ module LavinMQ
             ch = c.channel
             ch.queue_delete(@upstream_q)
             ch.exchange_delete(@upstream_q)
-            ch.exchange_delete(@upstream_exchange)
           end
         rescue e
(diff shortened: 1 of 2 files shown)


src/lavinmq/federation/link.cr
```diff
@@ -323,6 +323,8 @@ module LavinMQ
           ::AMQP::Client.start(upstream_uri) do |c|
             ch = c.channel
             ch.queue_delete(@upstream_q)
+            ch.exchange_delete(@upstream_q)
+            ch.exchange_delete(@upstream_exchange)
```
spuun, line 327: I'm not sure that we can delete this exchange unconditionally. It's not sure that we're creating it. We should probably not delete it no matter what...

kickster97: Maybe not the regular one, i added that as a guess, but the internal exchange should be deleted at cleanup

PR #754, opened 2024-08-16
Bugfix: make sure AMQPSource reconnects

spuun:
The shovel Runner will restart Source and Destionation when an exception occurs in any of them. The Source was however not restarted because it was in a "started" state because it didn't change state on all exceptions because exception was rescued in the wrong place.

This PR will rescue all exceptions and stop the source before re-raising the exception to the Runner.

Specs


snichme (approved):
Good findings, not always obvious where to put the `rescue` in these scenarios.

PR #772, opened 2024-08-28
Fix .deb build workflow

spuun:
When preparing to build the deb packet we explicit `COPY extras/lavinmq.service debian/`. Then when `debuild` runs it will exec `make install` which will copy `lavinmq.service` to `debian/lavinmq/lib/systemd/system/lavinmq.service`.
For some reason this results in an error when trying to install to deb packet in ubuntu 24.04:

```
dpkg: error processing archive /var/cache/apt/archives/lavinmq_2.0.0-rc.3-1_amd64.deb (--unpack):
 unable to open '/usr/lib/systemd/system/lavinmq.service.dpkg-new': No such file or directory
No apport report written because the error message indicates an issue on the local system
```

Check GH actions?


walro (approved):
Installs fine on 24.04 at least, haven't tested 22.04.

spuun:
22.04 works

PR #776, opened 2024-09-03
Dont log handled exception as error with backtrace

spuun:
This will change log level and remove backtrace when we log errors in the shovel. Since the errors are handled and the shovel should reconnect warn should be enough.

This PR needs cloudamqp/amqp-client.cr#47 to get rid of all scary logging.

Run a shovel and close the connection from mgmt ui.

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -10,7 +10,7 @@ shards:
 
   amqp-client:
     git: https://github.com/cloudamqp/amqp-client.cr.git
-    version: 1.2.6
+    version: 1.2.7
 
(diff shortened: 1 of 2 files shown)


spuun:
Note that this won't take fully effect until we've released a new version of [amqp-client.cr](https://github.com/cloudamqp/amqp-client.cr).

viktorerlingsson (approved):

snichme (approved):

kickster97 (approved):

PR #777, opened 2024-09-04
Add timeout for shovels batch ack

spuun:
To increase performance of shovels ack is done in "batches", i.e. the shovel will ack for every half prefetch message. This may however cause a lot of unacked messages in the source queue because the queue is drained and the number of messages consumed isn't a multiple of half prefetch.

This PR will introduce a timeout for when we'll ack even though the batch threshold isn't reached.

Run specs

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -102,7 +102,7 @@ module LavinMQ
           return if ch.closed?
 
           # We batch ack for faster shovel
-          batch_full = delivery_tag % (@prefetch / 2).ceil.to_i == 0
+          batch_full = delivery_tag % ack_batch_size == 0
           if !batch || batch_full || at_end?(delivery_tag)
             @last_unacked = nil
             ch.basic_ack(delivery_tag, multiple: true)
@@ -115,6 +115,10 @@ module LavinMQ
 
       def started? : Bool
         !@q.nil? && !@conn.try &.closed?
+      end
+
+      private def ack_batch_size
+        (@prefetch / 2).ceil.to_i
       end
 
       private def open_channel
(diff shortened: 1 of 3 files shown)


spuun:
We may need a lock in `AMQPSource#ack` when MT is introduced.

src/lavinmq/shovel/shovel.cr
```diff
@@ -134,6 +139,36 @@ module LavinMQ
...
+          break if ch.closed?
+
+          # We have nothing in memory
+          next if last_unacked.nil?
+
+          # @last_unacked is nil after an ack has been sent, i.e if nil
+          # there is nothing to ack
+          next if @last_unacked.nil?
+
+          # Our memory is the same is the current @last_unacked which means
```
snichme, line 165:
```suggestion
          # Our memory is the same as the current @last_unacked which means
```

typo

snichme (approved):
Looks good, I guess we will have to wait and see if `3.seconds` is a good default or this that something you can configure when setting the shovel as well?

carlhoerberg:
Or you can use a sized Channel which you push processed delivery tags to, and as long as you can pop from it you do nothing but when it blocks you ack the last delivery tag.

spuun:
> Or you can use a sized Channel which you push processed delivery tags to, and as long as you can pop from it you do nothing but when it blocks you ack the last delivery tag.

Yes, I was thinking about that approach too. But if we can control that fibers runs in the same thread, I guess we don't have to think about locks at all.

spuun:
> Looks good, I guess we will have to wait and see if `3.seconds` is a good default or this that something you can configure when setting the shovel as well?

Yeah, dunno. It can probably be lowered to 1 second since as long as message rate is higher than 1 msg/s acks will be batched. What's important is that we ack "eventually" so we're not forcing the source broker to have half prefetch messages in memory "forever" because they are unacked.

spuun:
It's not as simple as one could think to just create an "ack loop fiber", so I'll go with this for now.

PR #781, opened 2024-09-09
Run frontend specs on ubuntu 24.04

spuun:
Run frontend specs on ubuntu 24.04

Run frontend specs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,7 @@ jobs:
           path: bin/
 
   path_filter:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -157,7 +157,7 @@ jobs:
 
   spec_frontend:
     name: Fronted specs
```
viktorerlingsson, line 159: Maybe also fix this typo 😄 (and same in `path_filter` above)

PR #783, opened 2024-09-13
use decoded id for prometheus metrics

viktorerlingsson:
use decoded id for prometheus metris

Manual


kickster97 (approved):

carlhoerberg:
Thanks!

(the terminology is that it is now _encoded_ as a base 32 string. you can't really decode an integer, it is what it is, it can only be encoded (represented) in different ways, as a base 10 string (default for `Int#to_s`) or now as a base 32 string.)

PR #784, opened 2024-09-16
Pass Time::Span to sleep (to please upcoming deprecation)

spuun:
`sleep(Number)` is deprecated in crystal 1.14, see crystal-lang/crystal#14962

This will make sure we pass `Time::Span` everywhere. I was lazy and just added `.seconds`, but it would probably be more suitable to use `.miliseconds` in a lot of places.

Runs specs


kickster97 (approved):
the threshold is low to change to milliseconds now that you already have a PR so i would vote for doing it right the first time, but still approve :)

carlhoerberg (approved):

PR #785, opened 2024-09-18
lavinmqctl on follower nodes

viktorerlingsson:
Adds a HTTP server on followers that bind to `internal_unix_socket` and returns a 503. This 503 is handled by lavinmqctl which will exit with code `2` and show an error message stating that it's a follower. 

Start a follower, run lavinmqctl against it.

### DIFF
--- a/src/lavinmq/http/http_server.cr
+++ b/src/lavinmq/http/http_server.cr
@@ -14,8 +14,6 @@ module LavinMQ
     Log = ::Log.for "http"
 
     class Server
-      getter http
-
       def initialize(@amqp_server : LavinMQ::Server)
         handlers = [
           StrictTransportSecurity.new,
@@ -66,9 +64,9 @@ module LavinMQ
         addr
       end
 
-      def self.bind_internal_unix(http)
+      def bind_internal_unix
         File.delete?(INTERNAL_UNIX_SOCKET)
-        addr = http.bind_unix(INTERNAL_UNIX_SOCKET)
+        addr = @http.bind_unix(INTERNAL_UNIX_SOCKET)
         File.chmod(INTERNAL_UNIX_SOCKET, 0o660)
         Log.info { "Bound to #{addr}" }
         addr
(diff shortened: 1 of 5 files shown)


src/lavinmq/http/http_server.cr
```diff
@@ -80,6 +82,20 @@ module LavinMQ
...
       end
+
+      # Starts a HTTP server that binds to the internal UNIX socket used by lavinmqctl.
+      # The server returns 503 to signal that the node is a follower and can not handle the request.
+      def self.follower_internal_socket_http_server
+        http_server = ::HTTP::Server.new do |context|
+          context.response.status_code = 503
+          context.response.status_message = "This node is a follower and does not handle HTTP requests."
+        end
+        self.bind_internal_unix(http_server)
```
snichme, line 93: it makes sense to reuse `bind_internal_unix` but it also leaves the other method somewhat strange, for example `listen`  uses `@http` but that isn't the "correct" server when we are in follower mode. 

I think this works but makes it harder to track which "http server" we are talking about. So I would suggest copy and inline the lines from `bind_internal_unix` into the follower method.  That way we dont need to change the signature of the `bind_internal_unix` either

viktorerlingsson: Yeah, I was trying not to duplicate code, but maybe you're right. It feels a little messy mixing as it is now.

src/lavinmqctl.cr
```diff
@@ -250,6 +250,11 @@ class LavinMQCtl
 
   private def handle_response(resp, *ok)
     return if ok.includes? resp.status_code
+    if resp.status_code == 503
+      puts "[ERROR] This node is a follower and does not handle lavinmqctl commands."
```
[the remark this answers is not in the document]
viktorerlingsson: Yes, we can do that. Wanted to be me more explicit in the message you get from lavinmqctl, but we can do that in the response body instead 👍

snichme (approved):

PR #786, opened 2024-09-18
Publish to exchange

snichme:

Move the logic of publishing a message to an exchange from Vhost to the exchange. 

The idea came because of the need for the exchange to handle the messages differently for MQTT support 
but it has additional benefits like `@queue_bindings` and `@exchange_bindings` can have different types and as you may see I've used only one "bindings" for both queue and exchange. for example for FanoutExchange it's just a `@bindings : Set(Destination)` 

There is probably some more things that should move down to each subclass to make the code clearer and to support the different types that `@bindings` has now. 

specs

### DIFF
--- a/spec/message_routing_spec.cr
+++ b/spec/message_routing_spec.cr
@@ -5,8 +5,10 @@ module LavinMQ
     # Monkey patch for backward compability and easier testing
     def matches(routing_key, headers = nil) : Set(Queue | Exchange)
       s = Set(Queue | Exchange).new
-      queue_matches(routing_key, headers) { |q| s << q }
-      exchange_matches(routing_key, headers) { |x| s << x }
+      qs = Set(Queue).new
+      es = Set(Exchange).new
+      find_queues(routing_key, headers, qs, es)
+      qs.each { |q| s << q }
       s
     end
   end
@@ -19,7 +21,9 @@ describe LavinMQ::DirectExchange do
       q1 = LavinMQ::Queue.new(vhost, "q1")
       x = LavinMQ::DirectExchange.new(vhost, "")
       x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
-      x.matches("q1").should eq(Set{q1})
+      found_queues = Set(LavinMQ::Queue).new
+      x.find_queues("q1", nil, found_queues, Set(LavinMQ::Exchange).new)
+      found_queues.should eq(Set{q1})
     end
   end
 
@@ -27,7 +31,10 @@ describe LavinMQ::DirectExchange do
     with_amqp_server do |s|
       vhost = s.vhosts.create("x")
       x = LavinMQ::DirectExchange.new(vhost, "")
-      x.matches("q1").should be_empty
+
+      found_queues = Set(LavinMQ::Queue).new
+      x.find_queues("q1", nil, found_queues, Set(LavinMQ::Exchange).new)
+      found_queues.should be_empty
     end
   end
 end
@@ -39,7 +46,10 @@ describe LavinMQ::FanoutExchange do
       q1 = LavinMQ::Queue.new(vhost, "q1")
       x = LavinMQ::FanoutExchange.new(vhost, "")
       x.bind(q1, "")
-      x.matches("any").should eq(Set{q1})
+
+      found_queues = Set(LavinMQ::Queue).new
+      x.find_queues("q1", nil, found_queues, Set(LavinMQ::Exchange).new)
+      found_queues.should eq(Set{q1})
     end
   end
 
@@ -47,7 +57,9 @@ describe LavinMQ::FanoutExchange do
     with_amqp_server do |s|
       vhost = s.vhosts.create("x")
       x = LavinMQ::FanoutExchange.new(vhost, "")
-      x.matches("q1").should be_empty
+      found_queues = Set(LavinMQ::Queue).new
+      x.find_queues("q1", nil, found_queues, Set(LavinMQ::Exchange).new)
+      found_queues.should be_empty
     end
   end
 end
@@ -304,6 +316,108 @@ describe LavinMQ::HeadersExchange do
         x.bind(q13, "", nil)
         x.matches("", nil).size.should eq 1
       end
+    end
+  end
+end
+
+describe LavinMQ::Exchange do
+  it "should handle CC in header" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["CC"] = ["q2"]
+      x.find_queues("q1", headers, found_queues)
+      found_queues.should eq(Set{q1, q2})
+    end
+  end
+  it "should raise if CC header isn't array" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["CC"] = "q2"
+      expect_raises(LavinMQ::Error::PreconditionFailed) do
+        x.find_queues("q1", headers, found_queues)
+      end
+    end
+  end
+
+  it "should handle BCC in header" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["BCC"] = ["q2"]
+      x.find_queues("q1", headers, found_queues)
+      found_queues.should eq(Set{q1, q2})
+    end
+  end
+
+  it "should raise if BCC header isn't array" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["BCC"] = "q2"
+      expect_raises(LavinMQ::Error::PreconditionFailed) do
+        x.find_queues("q1", headers, found_queues)
+      end
+    end
+  end
+
+  it "should drop BCC from header" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["BCC"] = ["q2"]
+      x.find_queues("q1", headers, found_queues)
+      headers["BCC"]?.should be_nil
+    end
+  end
+
+  it "should read both CC and BCC" do
+    with_amqp_server do |s|
+      vhost = s.vhosts.create("x")
+      q1 = LavinMQ::Queue.new(vhost, "q1")
+      q2 = LavinMQ::Queue.new(vhost, "q2")
+      q3 = LavinMQ::Queue.new(vhost, "q3")
+      x = LavinMQ::DirectExchange.new(vhost, "")
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      x.bind(q3, "q3", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["CC"] = ["q2"]
+      headers["BCC"] = ["q3"]
+      x.find_queues("q1", headers, found_queues)
+      found_queues.should eq(Set{q1, q2, q3})
     end
   end
 end
--- a/src/lavinmq/exchange/consistent_hash.cr
+++ b/src/lavinmq/exchange/consistent_hash.cr
@@ -4,9 +4,50 @@ require "../consistent_hasher.cr"
 module LavinMQ
   class ConsistentHashExchange < Exchange
     @hasher = ConsistentHasher(Destination).new
+    @bindings = Hash(Destination, String).new
 
     def type : String
       "x-consistent-hash"
+    end
+
+    def bindings_details : Iterator(BindingDetails)
+      @bindings.each.map do |d, w|
+        binding_key = BindingKey.new(w, @arguments)
+        BindingDetails.new(name, vhost.name, binding_key, d)
+      end
+    end
+
+    def bind(destination : Destination, routing_key : String, headers : AMQP::Table?)
+      return false if @bindings.has_key? destination
+      @bindings[destination] = routing_key
+      w = weight(routing_key)
+      @hasher.add(destination.name, w, destination)
+      binding_key = BindingKey.new(routing_key, @arguments)
+      data = BindingDetails.new(name, vhost.name, binding_key, destination)
+      notify_observers(ExchangeEvent::Bind, data)
+      true
+    end
+
+    def unbind(destination : Destination, routing_key : String, headers : AMQP::Table?)
+      return false unless @bindings.delete destination
+      w = weight(routing_key)
+      @hasher.remove(destination.name, w)
+
+      binding_key = BindingKey.new(routing_key, @arguments)
+      data = BindingDetails.new(name, vhost.name, binding_key, destination)
+      notify_observers(ExchangeEvent::Unbind, data)
+
+      delete if @auto_delete && @bindings.empty?
+      true
+    end
+
+    protected def bindings(routing_key, headers) : Iterator(Destination)
+      key = hash_key(routing_key, headers)
+      if d = @hasher.get(key)
+        {d}.each
+      else
+        Iterator(Destination).empty
+      end
     end
 
     private def weight(routing_key : String) : UInt32
@@ -21,47 +62,6 @@ module LavinMQ
       when String then value.as(String)
       when Nil    then ""
       else             raise Error::PreconditionFailed.new("Routing header must be string")
-      end
-    end
-
-    def bind(destination : Destination, routing_key : String, headers : AMQP::Table?)
-      w = weight(routing_key)
-      @hasher.add(destination.name, w, destination)
-      ret = case destination
-            when Queue
-              @queue_bindings[{routing_key, headers}].add? destination
-            when Exchange
-              @exchange_bindings[{routing_key, headers}].add? destination
-            end
-      after_bind(destination, routing_key, headers)
-      ret
-    end
-
-    def unbind(destination : Destination, routing_key : String, headers : AMQP::Table?)
-      w = weight(routing_key)
-      ret = case destination
-            when Queue
-              @queue_bindings[{routing_key, headers}].delete destination
-            when Exchange
-              @exchange_bindings[{routing_key, headers}].delete destination
-            end
-      @hasher.remove(destination.name, w)
-      ret
-    end
-
-    def do_queue_matches(routing_key : String, headers : AMQP::Table?, & : Queue -> _)
-      key = hash_key(routing_key, headers)
-      case dest = @hasher.get(key)
-      when Queue
-        yield dest.as(Queue)
-      end
-    end
-
-    def do_exchange_matches(routing_key : String, headers : AMQP::Table?, & : Exchange -> _)
-      key = hash_key(routing_key, headers)
-      case dest = @hasher.get(key)
-      when Exchange
-        yield dest.as(Exchange)
       end
     end
   end
--- a/src/lavinmq/exchange/default.cr
+++ b/src/lavinmq/exchange/default.cr
@@ -6,22 +6,24 @@ module LavinMQ
       "direct"
     end
 
-    def bind(destination, routing_key, headers = nil)
-      raise "Access refused"
+    def bindings_details : Iterator(BindingDetails)
+      Iterator(BindingDetails).empty
     end
 
-    def unbind(destination, routing_key, headers = nil)
-      raise "Access refused"
-    end
-
-    def do_queue_matches(routing_key, headers = nil, & : Queue -> _)
+    protected def bindings(routing_key, headers) : Iterator(Destination)
       if q = @vhost.queues[routing_key]?
-        yield q
+        Tuple(Destination).new(q).each
+      else
+        Iterator(Destination).empty
       end
     end
 
-    def do_exchange_matches(routing_key, headers = nil, & : Exchange -> _)
-      # noop
+    def bind(destination, routing_key, headers = nil)
+      raise LavinMQ::Exchange::AccessRefused.new(self)
+    end
+
+    def unbind(destination, routing_key, headers = nil)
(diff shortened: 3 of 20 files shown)


src/lavinmq/exchange/headers.cr
```diff
@@ -6,53 +6,61 @@ module LavinMQ
...
+    @bindings = Hash(AMQP::Table, Set(Destination)).new do |h, k|
+      h[k] = Set(Destination).new
+    end
+
+    def bindings_details : Iterator(BindingDetails)
+      @bindings.each.map do |args, ds|
+        ds.map do |d|
+          BindingDetails.new(name, vhost.name, "", args, d)
+        end
+      end.flatten
```
spuun, line 18: Should be possible to use `Iterator#flat_map` to get rid of allocations. You can probably do `ds.each.map` too?

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,30 @@ module LavinMQ
...
+      ([] of Destination).each
     end
 
-    def do_queue_matches(routing_key, headers = nil, & : Queue -> _)
+    protected def bindings(routing_key, headers) : Iterator(Destination)
       if q = @vhost.queues[routing_key]?
-        yield q
+        a = Array(Destination).new
+        a << q
+        a.each
```
carlhoerberg, line 21: Use a Tuple to avoid allocations 
```suggestion
        {q}.each
```

src/lavinmq/exchange/direct.cr
```diff
@@ -2,40 +2,53 @@ require "./exchange"
...
+        true
+      else
+        false
+      end
     end
 
-    def do_queue_matches(routing_key, headers = nil, & : Queue -> _)
-      @queue_bindings[{routing_key, nil}].each { |q| yield q }
+    protected def bindings : Iterator(Destination)
+      @bindings.each.flat_map { |_, v| v.to_a }
```
carlhoerberg, line 47:
```suggestion
      @bindings.values.each.flat_map &.to_a
```

src/lavinmq/exchange/headers.cr
```diff
@@ -63,27 +71,19 @@ module LavinMQ
       end
     end
 
-    # ameba:disable Metrics/CyclomaticComplexity
-    private def matches(bindings, routing_key, headers, & : Queue | Exchange ->)
-      bindings.each do |bt, dst|
-        args = bt[1] || next
+    private def matches(headers) : Iterator(Destination) # , & : Destination ->)
+      @bindings.each.select do |args, _|
```
carlhoerberg, line 75:
```suggestion
      @bindings.each_key.select do |args|
```

spec/tx_spec.cr
```diff
@@ -34,19 +34,19 @@ describe "Transactions" do
           q2.bind("amq.fanout", "")
           x = ch.exchange("amq.fanout", "fanout")
           2.times do |i|
-            x.publish i.to_s * 200_000, ""
+            x.publish i.to_s * 2, ""
```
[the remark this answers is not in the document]
snichme: The test wasn't passing and when debugging it printed the body to stdout, so while testing I just reduce the number for now. Will revert back when green

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,30 @@ module LavinMQ
...
-    def bind(destination, routing_key, headers = nil)
-      raise "Access refused"
+    def bindings_details : Iterator(BindingDetails)
+      ([] of BindingDetails).each
     end
 
-    def unbind(destination, routing_key, headers = nil)
-      raise "Access refused"
+    def bindings : Iterator(Destination)
+      ([] of Destination).each
```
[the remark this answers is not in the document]
snichme: Nice, thanks

src/lavinmq/exchange/default.cr
snichme: `{q}.each` gives a type error 

but `Slice(Destination).empty.each.chain({q}.each)` works

src/lavinmq/exchange/headers.cr
snichme: `each_key` gives a iterator of just they keys right? Because I need the values from the hash after the `.select`

src/lavinmq/exchange/default.cr
spuun: What if you do `Tuple(Destination).new(q).each`?
Another way could be `Iterator(Destination).of(q).first(1)`, but this is probably two allocation while Tuple will be one. A third way is `Iterator(Destination).chain({ {q}.each })`, but also two allocations I think.

src/lavinmq/exchange/default.cr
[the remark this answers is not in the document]
spuun: Maybe add an `struct EmptyIterator(T)` to get rid of the allocations completely.

I think it could be an improvement to stdlib; add `Iterator(T).empty` that returns an `struct EmptyIterator(T)`.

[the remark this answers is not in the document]
spuun: I created an issue for that https://github.com/crystal-lang/crystal/issues/15038

src/lavinmq/exchange/topic.cr
```diff
@@ -2,75 +2,77 @@ require "./exchange"
...
+        true
+      else
+        false
       end
     end
 
-    def do_exchange_matches(routing_key, headers = nil, & : Exchange -> _)
-      matches(@exchange_binding_keys, routing_key, headers) { |e| yield e.as(Exchange) }
+    protected def bindings : Iterator(Destination)
+      @bindings.each.flat_map { |_, v| v.to_a }
```
carlhoerberg, line 51:
```suggestion
      @bindings.each_value.flat_map &.each
```

src/lavinmq/exchange/topic.cr
```diff
@@ -127,8 +129,8 @@ module LavinMQ
           break unless ok
           i += 1
         end
-        dst.each { |d| yield d } if ok
-      end
+        ok
+      end.flat_map { |_, v| v.to_a }
```
carlhoerberg, line 133:
```suggestion
      end.flat_map { |_, v| v.each }
```

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,30 @@ module LavinMQ
...
+        a.each
+      else
+        ([] of Destination).each
       end
     end
 
-    def do_exchange_matches(routing_key, headers = nil, & : Exchange -> _)
-      # noop
+    def bind(destination, routing_key, headers = nil)
+      raise "Access refused"
```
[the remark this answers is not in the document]
snichme: Yes, before it was the client that checked this, and also the exchange, changed this in https://github.com/cloudamqp/lavinmq/pull/786/commits/dc93030d354062e88503cbfab7590d64a85dc6f1 so the client catches the error from the change and sends a "access_refused" response

snichme:
@spuun @carlhoerberg any thoughts about always including "default" exchange when getting bindings for a queue heere: https://github.com/cloudamqp/lavinmq/blob/publish-to-exchange/src/lavinmq/vhost.cr#L286 
instead of adding the default in the http controllers

carlhoerberg:
Im fine with that

src/lavinmq/exchange/consistent_hash.cr
```diff
@@ -4,64 +4,71 @@ require "../consistent_hasher.cr"
...
-      w = weight(routing_key)
-      ret = case destination
-            when Queue
-              @queue_bindings[{routing_key, headers}].delete destination
-            when Exchange
-              @exchange_bindings[{routing_key, headers}].delete destination
-            end
-      @hasher.remove(destination.name, w)
-      ret
+      if @bindings.delete destination
```
spuun, line 32: I'd probably go with early return instead of if-else
```suggestion
      return false unless @bindings.delete destination
```

src/lavinmq/exchange/consistent_hash.cr
```diff
@@ -4,64 +4,71 @@ require "../consistent_hasher.cr"
...
+
+        binding_key = BindingKey.new(routing_key, @arguments)
+        data = BindingDetails.new(name, vhost.name, binding_key, destination)
+        notify_observers(ExchangeEvent::Unbind, data)
+
+        delete if @auto_delete && @bindings.empty?
+        true
+      else
+        false
+      end
```
spuun, line 44: Together with prev comment
```suggestion
```

src/lavinmq/exchange/consistent_hash.cr
```diff
@@ -4,64 +4,71 @@ require "../consistent_hasher.cr"
...
     end
 
-    def do_queue_matches(routing_key : String, headers : AMQP::Table?, & : Queue -> _)
+    protected def bindings : Iterator(Destination)
+      @bindings.keys.each
```
spuun, line 48:
```suggestion
      @bindings.each_key
```

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,28 @@ module LavinMQ
       "direct"
     end
 
-    def bind(destination, routing_key, headers = nil)
-      raise "Access refused"
+    def bindings_details : Iterator(BindingDetails)
+      Slice(BindingDetails).empty.each
```
spuun, line 10: Monkey patch our code with https://github.com/crystal-lang/crystal/pull/15039 and uses that to get rid of an allocation?
```suggestion
   Iterator(BindingDetails).empty
```

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,28 @@ module LavinMQ
...
     end
 
-    def unbind(destination, routing_key, headers = nil)
-      raise "Access refused"
+    def bindings : Iterator(Destination)
+      Slice(Destination).empty.each
```
spuun, line 14: ditto

src/lavinmq/exchange/default.cr
```diff
@@ -6,22 +6,28 @@ module LavinMQ
...
     end
 
-    def do_queue_matches(routing_key, headers = nil, & : Queue -> _)
+    protected def bindings(routing_key, headers) : Iterator(Destination)
       if q = @vhost.queues[routing_key]?
-        yield q
+        Tuple(Destination).new(q).each
+      else
+        Slice(Destination).empty.each
```
spuun, line 21: If suggested monkey patch:
```suggestion
        Iterator(Destionation).empty
```

src/lavinmq/exchange/direct.cr
```diff
@@ -2,40 +2,56 @@ require "./exchange"
...
       "direct"
     end
 
-    def bind(destination : Queue, routing_key, headers = nil)
-      ret = @queue_bindings[{routing_key, nil}].add? destination
-      after_bind(destination, routing_key, headers)
-      ret
+    def bindings_details : Iterator(BindingDetails)
+      @bindings.each.flat_map do |key, ds|
+        ds.map do |d|
```
spuun, line 15:
```suggestion
        ds.each.map do |d|
```

src/lavinmq/exchange/direct.cr
```diff
@@ -2,40 +2,56 @@ require "./exchange"
...
     end
 
-    def unbind(destination : Queue, routing_key, headers = nil)
-      ret = @queue_bindings[{routing_key, nil}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
-    end
+    def unbind(destination : Destination, routing_key, headers = nil) : Bool
+      rk_bindings = @bindings[routing_key]
+      if rk_bindings.delete destination
```
spuun, line 35:
```suggestion
      return false unless rk_bindings.delete destination
```

src/lavinmq/exchange/direct.cr
```diff
@@ -2,40 +2,56 @@ require "./exchange"
...
 
-    def unbind(destination : Exchange, routing_key, headers = nil)
-      ret = @exchange_bindings[{routing_key, nil}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
+        delete if @auto_delete && @bindings.each_value.all?(&.empty?)
+        true
+      else
+        false
+      end
```
spuun, line 46:
```suggestion
```

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
 
-    def bind(destination : Exchange, routing_key, headers)
-      validate!(headers)
-      args = headers ? @arguments.clone.merge!(headers) : @arguments
-      ret = @exchange_bindings[{routing_key, args}].add? destination
-      after_bind(destination, routing_key, headers)
-      ret
+    def bindings_details : Iterator(BindingDetails)
+      @bindings.each.flat_map do |args, ds|
+        ds.map do |d|
```
spuun, line 22:
```suggestion
        ds.each.map do |d|
```

src/lavinmq/exchange/fanout.cr
```diff
@@ -2,40 +2,48 @@ require "./exchange"
...
+        BindingDetails.new(name, vhost.name, binding_key, d)
+      end
     end
 
-    def unbind(destination : Queue, routing_key, headers = nil)
-      ret = @queue_bindings[{routing_key, nil}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
+    def bind(destination : Destination, routing_key, headers = nil)
+      if @bindings.add? destination
```
spuun, line 19:
```suggestion
      return false unless @bindings.add? destination
```

src/lavinmq/exchange/fanout.cr
```diff
@@ -2,40 +2,48 @@ require "./exchange"
...
+        binding_key = BindingKey.new("")
+        data = BindingDetails.new(name, vhost.name, binding_key, destination)
+        notify_observers(ExchangeEvent::Bind, data)
+        true
+      else
+        false
+      end
```
spuun, line 26:
```suggestion
```

src/lavinmq/exchange/fanout.cr
```diff
@@ -2,40 +2,48 @@ require "./exchange"
...
     end
 
-    def unbind(destination : Exchange, routing_key, headers = nil)
-      ret = @exchange_bindings[{routing_key, nil}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
+    def unbind(destination : Destination, routing_key, headers = nil)
+      if @bindings.delete destination
```
spuun, line 30:
```suggestion
      return false unless @bindings.delete destination
```

src/lavinmq/exchange/fanout.cr
```diff
@@ -2,40 +2,48 @@ require "./exchange"
...
+        binding_key = BindingKey.new("")
+        data = BindingDetails.new(name, vhost.name, binding_key, destination)
+        notify_observers(ExchangeEvent::Unbind, data)
+        delete if @auto_delete && @bindings.empty?
+        true
+      else
+        false
+      end
```
spuun, line 38:
```suggestion
```

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
     end
 
-    def unbind(destination : Queue, routing_key, headers)
+    def bind(destination : Destination, routing_key, headers)
+      validate!(headers)
       args = headers ? @arguments.clone.merge!(headers) : @arguments
-      ret = @queue_bindings[{routing_key, args}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
+      if @bindings[args].add? destination
```
spuun, line 32:
```suggestion
      return false unless @bindings[args].add? destination
```

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
+        binding_key = BindingKey.new(routing_key, args)
+        data = BindingDetails.new(name, vhost.name, binding_key, destination)
+        notify_observers(ExchangeEvent::Bind, data)
+        true
+      else
+        false
+      end
```
spuun, line 39:
```suggestion
```

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
 
-    def unbind(destination : Exchange, routing_key, headers)
+    def unbind(destination : Destination, routing_key, headers)
       args = headers ? @arguments.clone.merge!(headers) : @arguments
-      ret = @exchange_bindings[{routing_key, args}].delete destination
-      after_unbind(destination, routing_key, headers)
-      ret
-    end
+      bds = @bindings[args]
+      if bds.delete(destination)
```
spuun, line 45:
```suggestion
      return false unless bds.delete(destination)
```

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
+        data = BindingDetails.new(name, vhost.name, binding_key, destination)
+        notify_observers(ExchangeEvent::Unbind, data)
 
-    def do_queue_matches(routing_key, headers = nil, & : Queue ->)
-      matches(@queue_bindings, routing_key, headers) do |destination|
-        yield destination.as(Queue)
+        delete if @auto_delete && @bindings.each_value.all?(&.empty?)
+        true
+      else
+        false
```
spuun, line 55:
```suggestion
```
(also remove `end` on next line, but couldn't get it in the suggestion because it's not part of a change?)

src/lavinmq/exchange/headers.cr
```diff
@@ -13,44 +13,55 @@ module LavinMQ
...
       end
     end
 
-    def do_exchange_matches(routing_key, headers = nil, & : Exchange ->)
-      matches(@exchange_bindings, routing_key, headers) { |e| yield e.as(Exchange) }
+    protected def bindings : Iterator(Destination)
+      @bindings.values.each.flat_map(&.each)
```
spuun, line 60:
```suggestion
      @bindings.each_value.flat_map(&.each)
```

src/lavinmq/exchange/topic.cr
```diff
@@ -2,75 +2,77 @@ require "./exchange"
...
+      select_matches(routing_key).each
     end
 
     # ameba:disable Metrics/CyclomaticComplexity
-    private def matches(binding_keys, routing_key, headers = nil, & : Queue | Exchange -> _)
-      return if binding_keys.empty?
+    private def select_matches(routing_key) : Iterator(Destination)
+      binding_keys = @bindings
+
+      return ([] of Destination).each if binding_keys.empty?
```
spuun, line 62: If we monkey patch 
```suggestion
      return Iterator(Destination).empty if binding_keys.empty?
```

spuun (commented):
I ran out of stamina to suggest a few other early returns... I'm not gonna fight for early returns, but i think it's nice.

snichme:
@spuun agree, early return is nicer. Fixed all now I thinkg

src/lavinmq/exchange.cr
```diff
@@ -1 +1,11 @@
...
+
+module LavinMQ
+  struct EmptyIterator(T)
+    include Iterator(T)
+
+    def next
+      stop
+    end
+  end
+end
```
spuun, line 11: I was thinking that we could monkey patch stdlib iterator to have the empty method. If my PR is accepted we can just remove that patch and the method from stdlib will be used.

snichme: Ok, I will try that

snichme: Done https://github.com/cloudamqp/lavinmq/pull/786/commits/af69bfdb80182c1e1781f52655d3c466f17ff67b

snichme: and https://github.com/cloudamqp/lavinmq/pull/786/commits/f55e03a8f9d2dd685314affc0d00d2aab66fca22 :)

spuun: Coolio!

snichme:
Specs fails because of this: https://github.com/cloudamqp/amqp-client.cr/pull/48

kickster97 (approved):
has been working well in the mqtt branch!

spec/message_routing_spec.cr
```diff
@@ -307,3 +319,105 @@ describe LavinMQ::HeadersExchange do
...
+      x.bind(q1, "q1", LavinMQ::AMQP::Table.new)
+      x.bind(q2, "q2", LavinMQ::AMQP::Table.new)
+      found_queues = Set(LavinMQ::Queue).new
+      headers = LavinMQ::AMQP::Table.new
+      headers["CC"] = ["q2"]
+      x.find_queues("q1", headers, found_queues)
+      found_queues.should eq(Set{q1, q2})
+    end
+  end
+  it "should raise unless CC header isn't array" do
```
viktorerlingsson, line 339: double negative here hurts my brain 🙂

src/lavinmq/exchange/exchange.cr
```diff
@@ -167,18 +144,83 @@ module LavinMQ
...
+    end
+
+    def find_queues(routing_key : String, headers : AMQP::Table?,
+                    queues : Set(Queue),
+                    exchanges : Set(Exchange) = Set(Exchange).new) : Nil
+      return unless exchanges.add? self
+      bindings(routing_key, headers).each do |d|
+        case d
+        when Queue
+          queues.add(d) unless d.closed?
```
carlhoerberg, line 189: We also check if the queue is closed when doing the publish

src/lavinmq/exchange/exchange.cr
```diff
@@ -167,18 +144,83 @@ module LavinMQ
...
+
+    private def find_cc_queues(headers, key, queues)
+      return unless cc = headers[key]?
+      cc = cc.as?(Array(AMQP::Field))
+
+      err = Error::PreconditionFailed.new("#{key} header not a string array")
+      raise err unless cc
+
+      hdrs = headers.clone
+      hdrs.delete "CC"
```
carlhoerberg, line 216: Is this right? What about BCC then? And what about the next line?

src/lavinmq/exchange/exchange.cr
```diff
@@ -167,18 +144,83 @@ module LavinMQ
...
+          ae.find_queues(routing_key, headers, queues, exchanges)
+        end
+      end
+    end
+
+    private def find_cc_queues(headers, key, queues)
+      return unless cc = headers[key]?
+      cc = cc.as?(Array(AMQP::Field))
+
+      err = Error::PreconditionFailed.new("#{key} header not a string array")
```
carlhoerberg, line 212: This means that we are always allocating, even when never raising

src/lavinmq/exchange/headers.cr
```diff
@@ -2,8 +2,8 @@ require "./exchange"
 
 module LavinMQ
   class HeadersExchange < Exchange
-    def type : String
-      "headers"
+    @bindings = Hash(AMQP::Table, Set(Destination)).new do |h, k|
```
carlhoerberg, line 5: Maybe not in this PR but maybe we should do a performance comparison when the key is a Hash instead of a amqp::table

src/lavinmq/exchange.cr
```diff
@@ -1 +1,11 @@
 require "./exchange/*"
+
+module LavinMQ
+  struct EmptyIterator(T)
```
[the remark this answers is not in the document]
snichme: No, but I moved it to `/stdlib` not sure why you got an older commit to look at. 
It's here: https://github.com/cloudamqp/lavinmq/blob/publish-to-exchange/src/stdlib/iterator.cr now

src/lavinmq/exchange/exchange.cr
carlhoerberg: it's only cheap if the pointer is in the CPU cache, otherwise it's a pointer deference. `queues` is a `Set` so there's no dup check. That a Queue is in the bindings hash but closed should be a very unlikely condition, like only when the server is shutting down or something.

src/lavinmq/exchange/exchange.cr
carlhoerberg: Extract it to a method (2 extra rows), or duplicate it (1 extra row). Allocations in the hot path is bad.

src/lavinmq/exchange/exchange.cr
snichme: But if the queue is closed, we dont try to insert it so no need for a  check in the `Set` , do you think its faster to try to insert in the `Set` than deref `d` and check if it's closed?

src/lavinmq/exchange/exchange.cr
snichme: Will just copy the line

carlhoerberg (approved):

src/lavinmq/exchange/exchange.cr
carlhoerberg: it doesn't reduce the number of inserts, as it's a very exceptional case that it is closed but still in bindings. so let's only verify that it is closed or not when we really need it, when doing the publish to the queue.

src/lavinmq/exchange/exchange.cr
```diff
@@ -105,22 +108,12 @@ module LavinMQ
...
-      return true if @vhost.exchanges.any? { |_, x| x.exchange_bindings.any? { |_, exs| exs.includes? self } }
-      false
-    end
-
-    def bindings_details
-      Iterator.chain({@queue_bindings.each, @exchange_bindings.each}).flat_map do |(key, destinations)|
-        destinations.map { |destination| binding_details(key, destination) }
+      return true unless bindings_details.empty?
+      @vhost.exchanges.any? do |_, x|
+        x.bindings_details.each.any? { |bd| bd.destination == self }
```
carlhoerberg, line 113:
```suggestion
        x.bindings_details.any? { |bd| bd.destination == self }
```

carlhoerberg: `.any?` does only iterate over the hash, doesn't generate a new array or anything, so the extra `.each` isn't necessary.

src/lavinmq/exchange/exchange.cr
carlhoerberg:
```suggestion
          queues.add(d)
```

src/lavinmq/exchange/exchange.cr
```diff
@@ -167,18 +144,87 @@ module LavinMQ
...
+                    queues : Set(Queue),
+                    exchanges : Set(Exchange) = Set(Exchange).new) : Nil
+      return unless exchanges.add? self
+      bindings(routing_key, headers).each do |d|
+        case d
+        when Queue
+          queues.add(d) unless d.closed?
+        when Exchange
+          d.find_queues(routing_key, headers, queues, exchanges)
+        end
```
carlhoerberg, line 197:
```suggestion
        in Queue
          queues.add(d) unless d.closed?
        in Exchange
          d.find_queues(routing_key, headers, queues, exchanges)
        end
```

PR #787, opened 2024-09-19
Lavinmqctl cluster status

viktorerlingsson:
Adds the `cluster_status` cmd to lavinmqctl, which shows information about the running cluster.

Run `lavinmqctl cluster_status`


snichme (approved):
Nice! 

One thing (probably outside the scope of this PR) have we talked about keeping the output from the ctl stable? If something starts scripting with the ctl they probably want the same output even if they get a new version. 
This is hard (and I dont think we should commit to that) but if we don't do it maybe add support for outputing to different formats like json as well

viktorerlingsson:
> One thing (probably outside the scope of this PR) have we talked about keeping the output from the ctl stable? If something starts scripting with the ctl they probably want the same output even if they get a new version. This is hard (and I dont think we should commit to that) but if we don't do it maybe add support for outputing to different formats like json as well

I don't remember us talking about that, but I think it's a good idea to support at least json output as well!

PR #788, opened 2024-09-20
Federation queue link should reconnect

viktorerlingsson:
Makes sure federation links reconnects after the upstream restarts

Run the spec

### DIFF
--- a/src/lavinmq/federation/link.cr
+++ b/src/lavinmq/federation/link.cr
@@ -232,7 +232,11 @@ module LavinMQ
             state(State::Running)
             unless @federated_q.immediate_delivery?
               @log.debug { "Waiting for consumers" }
-              @consumer_available.receive?
+              select
+              when @consumer_available.receive?
+              when timeout(1.second)
+                return if @upstream_connection.try &.closed?
+              end
             end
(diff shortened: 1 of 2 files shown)


src/lavinmq/federation/link.cr
```diff
@@ -232,7 +232,11 @@ module LavinMQ
             state(State::Running)
             unless @federated_q.immediate_delivery?
               @log.debug { "Waiting for consumers" }
-              @consumer_available.receive?
+              select
+              when @consumer_available.receive?
+              when timeout(1.second)
+                return if @upstream_connection.try &.closed?
+              end
```
spuun, line 239: The timeout will cause a reconnect loop if no consumers are available. You probably need a loop around this select and break when consumer_available.

spuun (changes requested):
See inline comment.

src/lavinmq/federation/link.cr
viktorerlingsson: Ah, yes, you're right 👍

spuun (approved):
👍

kickster97 (approved):

PR #789, opened 2024-09-24
Remove min_isr

viktorerlingsson:
Removes settings and code related to `min_isr` since we're not happy with how it works at the moment. 

"Backup" saved in https://github.com/cloudamqp/lavinmq/tree/min_isr_backup

Run specs

### DIFF
--- a/spec/clustering_spec.cr
+++ b/spec/clustering_spec.cr
@@ -40,7 +40,6 @@ describe LavinMQ::Clustering::Client do
   end
 
   it "can stream changes" do
-    LavinMQ::Config.instance.clustering_min_isr = 1
     replicator = LavinMQ::Clustering::Server.new(LavinMQ::Config.instance, LavinMQ::Etcd.new, 0)
     tcp_server = TCPServer.new("localhost", 0)
     spawn(replicator.listen(tcp_server), name: "repli server spec")
--- a/spec/clustering_spec.cr
+++ b/spec/clustering_spec.cr
@@ -50,7 +50,7 @@ describe LavinMQ::Clustering::Client do
       repli.follow("localhost", tcp_server.local_address.port)
       done.send nil
     end
-
+    wait_for { replicator.followers.size == 1 }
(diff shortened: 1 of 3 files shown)


carlhoerberg (changes requested):
The commit needs a lot more explanation

viktorerlingsson:
> The commit needs a lot more explanation

Something like this? Or are you looking for something else? 

Setting minimum in sync replicas (`min_isr`) > 0 with the current implementation will block LavinMQ from starting. The reason for this is that LavinMQ does some file actions on startup (cleanup, truncate, etc - for example the actions taken [here](https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/queue/message_store.cr#L31-L34)). With `min_isr` set, these actions needs to be replicated to at least n (`min_isr`) number of nodes before LavinMQ can continue. However, the listener for the replication server has not been started yet when these actions are taken, which makes it impossible for any followers to connect to the leader, and therefore makes it impossible to replicate the actions to the required number of nodes. This leaves the leader node stuck on waiting for actions to be replicated to followers who are themselves waiting for the leader to accept connections. 

Several workarounds for this where taken into consideration, but we deemed none of them to be good enough. Which is why we decided to remove `min_isr` for now, until we have a better idea of how to implement it.

dentarg:
What can happen without the `min_isr` functionality? I think that should be spelled out

viktorerlingsson:
> What can happen without a properly working `min_isr` functionality? I think that should be spelled out

Sure, added this to the comment: 
"Using clustering without `min_isr` will work the same way as if `min_isr` would be set to 0: The leader will not wait for followers to connect/reconnect before accepting connections and data. If followers are connected and in-sync, they will be kept in-sync for as long as possible. But if a follower disconnects, the leader node will not wait for it to come back online before accepting new connections and data. This means that it is possible to have data that isn't synced to any follower."

kickster97 (approved):

PR #790, opened 2024-09-24
Output formats (json) for lavinmqctl

viktorerlingsson:
Adds an option to specify output format (json) for lavinmqctl cmds. Also changes all methods to use the same `output` method. 

Example usage: 
```
$ bin/lavinmqctl list_queues -f json
[{"name":"amq.delayed.delay-test","messages":"0"},{"name":"perf-test","messages":"0"},{"name":"delay-q-test","messages":"0"},{"name":"delay2","messages":"0"},{"name":"q1","messages":"0"},{"name":"q2","messages":"0"},{"name":"federation: ex1 -> viktor-lenovo:/:ex1","messages":"0"}]
```

Manual

### DIFF
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -264,25 +264,15 @@ class LavinMQCtl
       puts data.to_json
     else
       case data
-      when String
-        output_string(data)
       when Hash
-        output_hash(data)
+        data.keys.each do |k|
+          puts "#{k}: #{data[k]}"
+        end
       when Array
         output_array(data, columns)
       else
         puts data
       end
-    end
-  end
-
-  private def output_string(data : String)
-    puts data
-  end
-
-  private def output_hash(data : Hash)
-    data.keys.each do |k|
-      puts "#{k}: #{data[k]}"
     end
   end
 
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -289,7 +289,7 @@ class LavinMQCtl
 
   private def output_array(data : Array, columns : Array(String)?)
     if columns
-      puts columns.join("\t")
+      puts columns.join(STDOUT, "\t")
     else
       if first = data.first?
         first = first.as_h if first.is_a?(JSON::Any)
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -424,23 +424,6 @@ class LavinMQCtl
     end
   end
 
-  private def print_erlang_terms(h : Hash)
-    str = "["
-    last_index = h.size - 1
-    h.each_with_index do |(key, value), i|
-      print "{\"#{key}\","
-      case value.raw
-      when Hash   then str += print_erlang_terms(value.as_h)
-      when String then str += "\"#{value}\""
-      else             str += value.to_s
-      end
-      str += "}"
-      str += "," unless i == last_index
-    end
-    str += "]"
-    return str
-  end
-
   private def close_connection
     name = ARGV.shift?
     abort @banner unless name
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -234,22 +234,12 @@ class LavinMQCtl
       # Only used by tests in cloudamqp/rabbitmq-java-client
       @options["node"] = v
     end
-    @parser.on("-q", "--quiet", "suppress informational messages") do
-      @options["quiet"] = "yes"
-    end
-    @parser.on("-s", "--silent", "suppress informational messages and table formatting") do
-      @options["silent"] = "yes"
-    end
-    @parser.on("-f format", "--format=format", "Output format (text, json)") do |v|
+    @parser.on("-f format", "--format=format", "Format output (text, json)") do |v|
       if v != "text" && v != "json"
         abort "Invalid format: #{v}"
       end
       @options["format"] = v
     end
-  end
-
-  private def quiet?
-    @options["quiet"]? || @options["silent"]?
   end
 
   private def entity_arg
@@ -276,8 +266,6 @@ class LavinMQCtl
       case data
       when String
         output_string(data)
-      when NamedTuple
-        output_tuple(data)
       when Hash
         output_hash(data)
       when Array
@@ -287,15 +275,9 @@ class LavinMQCtl
       end
     end
   end
-  
+
   private def output_string(data : String)
     puts data
-  end
-
-  private def output_tuple(data : NamedTuple)
-    data.each do |k,v|
-      puts "#{k}: #{v}"
-    end
   end
 
   private def output_hash(data : Hash)
@@ -312,7 +294,7 @@ class LavinMQCtl
       values = [] of String
       if item.is_a?(Hash)
         item.keys.each do |k|
-          values << item[k]
+          values << item[k].to_s
         end
       elsif item.is_a?(JSON::Any)
         if hash = item.as_h
@@ -357,14 +339,13 @@ class LavinMQCtl
   private def list_users
     resp = http.get "/api/users", @headers
     handle_response(resp, 200)
-    user_array = [] of Hash(String, String)
     if users = JSON.parse(resp.body).as_a?
-      users.each do |u|
+      uu = users.map do |u|
         next unless user = u.as_h?
-        user_array << { "name" => user["name"].to_s, "tags"=> user["tags"].to_s }
+        {"name" => user["name"].to_s, "tags" => user["tags"].to_s}
       end
+      output uu
     end
-    output user_array
   end
 
   private def add_user
@@ -401,15 +382,13 @@ class LavinMQCtl
   private def list_queues
     vhost = @options["vhost"]? || "/"
     resp = http.get "/api/queues/#{URI.encode_www_form(vhost)}", @headers
-    queues_array = [] of Hash(String, String)
     handle_response(resp, 200)
     if queues = JSON.parse(resp.body).as_a?
-      queues.each do |u|
+      qq = queues.map do |u|
         next unless q = u.as_h?
-        queues_array << { "name" => q["name"].to_s, "messages" => q["messages"].to_s }
+        {"name" => q["name"].to_s, "messages" => q["messages"].to_s}
       end
-      output queues_array, ["name", "messages"]
-      true
+      output qq, ["name", "messages"]
     else
       abort "invalid data"
     end
@@ -445,11 +424,17 @@ class LavinMQCtl
     resp = http.get "/api/connections", @headers
     handle_response(resp, 200)
     if conns = JSON.parse(resp.body).as_a?
-      conn_arr = conns.map do |u|
+      cc = conns.map do |u|
         next unless conn = u.as_h?
-        columns.map { |c| conn[c] }
+        {
+          # columns.map { |c| conn[c] }
+          "user"      => conn["user"].to_s,
+          "peer_host" => conn["peer_host"].to_s,
+          "peer_port" => conn["peer_port"].to_s,
+          "state"     => conn["state"].to_s,
+        }
       end
-      output conn_arr, columns
+      output cc, columns
     else
       abort "invalid data"
     end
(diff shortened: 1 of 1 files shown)


src/lavinmqctl.cr
```diff
@@ -234,18 +234,14 @@ class LavinMQCtl
       # Only used by tests in cloudamqp/rabbitmq-java-client
       @options["node"] = v
     end
-    @parser.on("-q", "--quiet", "suppress informational messages") do
-      @options["quiet"] = "yes"
-    end
-    @parser.on("-s", "--silent", "suppress informational messages and table formatting") do
```
kickster97, line 240: When I use it as a human, I like it to be super clear on what I am getting back.. do we remove that completely now? 
Maybe we could still have the informational messages, but they are suppressed when we format the output to json? 
like, `-f json` also sets `@options["quiet"] = "yes"`

viktorerlingsson: To be clear, the only thing that is removed are these messages:
```
puts "Listing queues for vhost #{vhost} ..." unless quiet?
puts "Listing connections ..." unless quiet?
puts "Closing connection #{name} ..." unless quiet?
puts "Listing policies for vhost #{vhost} ..." unless quiet?
puts "Listing exchanges for vhost #{vhost} ..." unless quiet?
```

carlhoerberg: They are there for compatibility with rabbitmqctl

src/lavinmqctl.cr
```diff
@@ -255,14 +251,54 @@ class LavinMQCtl
...
-    puts "#{resp.status_code} - #{resp.status}"
-    puts resp.body if resp.body? && !resp.headers["Content-Type"]?.try(&.starts_with?("text/html"))
+    output "#{resp.status_code} - #{resp.status}"
+    output resp.body if resp.body? && !resp.headers["Content-Type"]?.try(&.starts_with?("text/html"))
     exit 1
   end
 
+  private def output(data, columns = nil)
+    if @options["format"]? == "json"
+      puts data.to_json
```
carlhoerberg, line 264: data.to_json(STDOUT)

src/lavinmqctl.cr
```diff
@@ -255,14 +251,54 @@ class LavinMQCtl
...
+  private def output_array(data : Array, columns : Array(String)?)
+    if columns
+      puts columns.join("\t")
+    end
+    data.each do |item|
+      values = [] of String
+      if item.is_a?(Hash)
+        item.keys.each do |k|
+          values << item[k].to_s
+        end
```
carlhoerberg, line 288: Isnt this equal to:

```suggestion
        values = item.values.map &.to_s
```
?

src/lavinmqctl.cr
```diff
@@ -601,35 +613,40 @@ class LavinMQCtl
...
-      puts "Followers:"
-      puts "ID \t| Address \t\t| Lag"
-      followers.each do |f|
-        lag = f.dig("sent_bytes").as_i64 - f.dig("acked_bytes").as_i64
-        puts "#{f.dig("id")} \t| #{f.dig("remote_address")} \t| #{lag}"
-      end
-    else
-      puts "This node has no followers"
+      followers.map do |f|
+        {
```
carlhoerberg, line 638: Why arent you using NamedTuples? Saves allocations.

src/lavinmqctl.cr
```diff
@@ -501,17 +505,23 @@ class LavinMQCtl
...
         next unless p = u.as_h?
-        puts "#{p["vhost"]}\t#{p["name"]}\t#{p["pattern"]}\t#{p["apply-to"]}\t#{p["definition"]}\t#{p["priority"]}"
+        {
+          "vhost"      => p["vhost"].to_s,
+          "name"       => p["name"].to_s,
+          "pattern"    => p["pattern"].to_s,
+          "apply-to"   => p["apply-to"].to_s,
+          "definition" => p["definition"].as_h,
+          "priority"   => p["priority"].to_s,
+        }
```
carlhoerberg, line 519: Is p including more properties or why the remapping?

src/lavinmqctl.cr
```diff
@@ -380,50 +412,21 @@ class LavinMQCtl
...
+        next unless conn = u.as_h?
+        conn.select { |k, v| columns.includes? k }
       end
+      output cc, columns
     else
       abort "invalid data"
     end
   end
 
-  private def print_erlang_terms(h : Hash)
```
carlhoerberg, line 407: We are dropping rabbitmqctl compatible output?

carlhoerberg (changes requested):

src/lavinmqctl.cr
viktorerlingsson: No, I'll add it back. Didn't see a reason to keep it, but didn't think about compatibility.

src/lavinmqctl.cr
viktorerlingsson: added back the options and messages, java client tests now passes again 👍

kickster97 (approved):

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+        output_array(data, columns)
+      else
+        puts data
+      end
+    end
+  end
+
+  private def output_array(data : Array, columns : Array(String)?)
+    if columns
+      puts columns.join("\t")
```
carlhoerberg, line 292:
```suggestion
      puts columns.join(STDOUT, "\t")
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+    else
+      if first = data.first?
+        first = first.as_h if first.is_a?(JSON::Any)
+        puts first.keys.join("\t")
```
carlhoerberg, line 296:
```suggestion
        puts first.each_key.join(STDOUT, "\t")
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+      end
+    end
+    data.each do |item|
+      values = [] of String
+      if item.is_a?(Hash) || item.is_a?(NamedTuple)
+        values = item.values.map &.to_s
```
carlhoerberg, line 302:
```suggestion
        puts item.each_value.join(STDOUT, "\t")
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+      elsif item.is_a?(JSON::Any)
+        if hash = item.as_h
+          values = hash.values.map &.to_s
```
carlhoerberg, line 305:
```suggestion
          puts hash.each_value.join(STDOUT, "\t")
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+    if columns
+      puts columns.join("\t")
+    else
+      if first = data.first?
+        first = first.as_h if first.is_a?(JSON::Any)
+        puts first.keys.join("\t")
+      end
+    end
+    data.each do |item|
+      values = [] of String
```
carlhoerberg, line 300:
```suggestion
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+      if item.is_a?(Hash) || item.is_a?(NamedTuple)
+        values = item.values.map &.to_s
+      elsif item.is_a?(JSON::Any)
+        if hash = item.as_h
+          values = hash.values.map &.to_s
+        end
+      else
+        values << item.to_s
+      end
+      puts values.join("\t")
```
carlhoerberg, line 310:
```suggestion
```

src/lavinmqctl.cr
```diff
@@ -255,14 +261,56 @@ class LavinMQCtl
...
+    data.each do |item|
+      values = [] of String
+      if item.is_a?(Hash) || item.is_a?(NamedTuple)
+        values = item.values.map &.to_s
+      elsif item.is_a?(JSON::Any)
+        if hash = item.as_h
+          values = hash.values.map &.to_s
+        end
+      else
+        values << item.to_s
```
carlhoerberg, line 308:
```suggestion
        puts item
```

src/lavinmqctl.cr
```diff
@@ -383,20 +429,27 @@ class LavinMQCtl
...
-            else
-              print conn[c]
+      if @options["format"]? == "json"
+        cc = conns.map do |u|
+          next unless conn = u.as_h?
+          conn.select { |k, _v| columns.includes? k }
+        end
+        output cc
+      else
+        puts columns.join("\t")
```
carlhoerberg, line 439:
```suggestion
        puts columns.join(STDOUT, "\t")
```

src/lavinmqctl.cr
```diff
@@ -601,35 +655,40 @@ class LavinMQCtl
...
+      followers.map do |f|
+        {
+          id:      f.dig("id"),
+          address: f.dig("remote_address"),
+          lag:     f.dig("sent_bytes").as_i64 - f.dig("acked_bytes").as_i64,
+        }
+      end
+      cluster_status_obj = {
+        this_node: body[0].dig("name"),
+        version:   body[0].dig("applications")[0].dig("version"),
```
carlhoerberg, line 688:
```suggestion
        this_node: body.dig(0, "name"),
        version:   body.dig(0, "applications", 0, "version"),
```

src/lavinmqctl.cr
viktorerlingsson: `each_key` is expected to be invoked with a block, so this throws an error. 
Maybe just `puts first.keys.join(STDOUT, "\t")` instead?

src/lavinmqctl.cr
viktorerlingsson: `each_value` expects to be invoked with a block, so this throws an error. Maybe just print each value instead of the join? 
```
    data.each do |item|
      if item.is_a?(Hash) || item.is_a?(NamedTuple)
        item.each_value { |v| print "#{v}\t" }
      elsif item.is_a?(JSON::Any)
        if hash = item.as_h
          hash.each_value { |v| print "#{v}\t" }
        end
      else
        print item.to_s
      end
      puts
    end
```

src/lavinmqctl.cr
viktorerlingsson: same as https://github.com/cloudamqp/lavinmq/pull/790#discussion_r1776927873

src/lavinmqctl.cr
spuun: What types may `data` contain? `Hash#each_key` returns an iterator (if not invoked with a block), but `NamedTuple#each_key` only has one overload that requires a block.

src/lavinmqctl.cr
spuun: You can do like this:
```crystal
data.each do |item|
  case item
  when Hash
    item.each_value.join(STDOUT, "\t")
  when JSON::Any
    item.as_h.each_value.join(STDOUT, "\t")
  when NamedTuple
    item.keys.each.join(STDOUT, "\t")
  else
    item.to_s(STDOUT)
  end
  puts
end```

`NamedTuple#keys` returns a tuple, so no allocation.

File: src/lavinmqctl.cr (hunk 8 above)
Comment (viktorerlingsson) [reply src/lavinmqctl.cr:296]: `data.first` will either be a `NamedTuple` or a `JSON::Any`

File: src/lavinmqctl.cr (hunk 9 above)
Comment (viktorerlingsson) [reply src/lavinmqctl.cr:302]: nice 👍 (but we want the values from the `NamedTuple` so we should use `NamedTuple#values` instead I guess. Also returns a tuple)

File: src/lavinmqctl.cr (hunk 8 above)
Comment (viktorerlingsson) [reply src/lavinmqctl.cr:296]: so maybe something like this? 
```
        if first.is_a?(NamedTuple)
          puts first.keys.join(STDOUT, "\t")
        elsif first.is_a?(JSON::Any)
          puts first.as_h.each_key.join(STDOUT, "\t")
        end
      end
```

src/lavinmqctl.cr
spuun: Oh, right! 😁

spuun: Btw, maybe you can change `item.keys.each.join(STDOUT, "\t")` to just `item.keys.join(STDOUT, "\t")`?

PR #791, opened 2024-09-25
Remove some tuple usage

spuun:
It's very annoying when tuples are nested in tuples and you need to do `value[0][1]`. This will change `BindingKey` and `QueueBinding` from `Tuple` to `struct`.

Interesting is that

```crystal
alias QueueBinding = Tuple(BindingKey, Exchange)
```
didn't force the order of the values based on type, see

https://github.com/cloudamqp/lavinmq/blob/0603aa921a37e5bb41ba803b78a77082f4375602/src/lavinmq/vhost.cr#L365-L372

where the orders in line 369 are swapped compared to `QueueBinding`

One problem with this refactoring is that `Hash#[](key)` isn't type, which means we don't get any compile time error when we access our binding hashes with the old value, so there's a risk i've missed change the type in one place or two.

Run specs

### DIFF
--- a/spec/api/bindings_spec.cr
+++ b/spec/api/bindings_spec.cr
@@ -65,7 +65,7 @@ describe LavinMQ::HTTP::BindingsController do
         response = http.post("/api/bindings/%2f/e/be1/q/bindings_q1", body: body)
         response.status_code.should eq 201
         response.headers["Location"].should eq "bindings_q1/rk"
-        s.vhosts["/"].exchanges["be1"].queue_bindings.last_key.first.should eq "rk"
+        s.vhosts["/"].exchanges["be1"].queue_bindings.last_key.routing_key.should eq "rk"
       end
     end
 
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -59,7 +59,7 @@ describe LavinMQ::Queue do
           x.publish_confirm "test message", q.name
           q.get(no_ack: true).try(&.body_io.to_s).should eq("test message")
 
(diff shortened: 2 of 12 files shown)


spuun:
IMO we should avoid `Tuple` and `NamedTuple` everywhere.

spuun:
I monkey patched `Hash` locally and didn't find anything related to this change.

snichme:
This will conflict with https://github.com/cloudamqp/lavinmq/pull/786 but I like the approach of using a struct instead of an alias. How about I introduced the struct in mr PR instead?

carlhoerberg (approved):

carlhoerberg:
> IMO we should avoid `Tuple` and `NamedTuple` everywhere.

Totally do not agree. In this case it makes sense, but there are definitely cases where Tuple or NamedTuples is the better option.

spuun:
> > IMO we should avoid `Tuple` and `NamedTuple` everywhere.
> 
> Totally do not agree. In this case it makes sense, but there are definitely cases where Tuple or NamedTuples is the better option.

Well, I'm generalizing a bit. But I think that many uses of NamedTuple can be a struct instead.

spuun:
> This will conflict with #786 but I like the approach of using a struct instead of an alias. How about I introduced the struct in mr PR instead?

@kickster97 and I wanted this in the mqtt branch, but I guess it's no rush. I think this can be done even better by changing some interfaces to require a `BindingKey` instead of constructing it everywhere like I did now.

spuun:
Superseded by #786

spuun:
Superseded by #786

PR #792, opened 2024-09-26
add log for unsupported queue types

kickster97:
This pull request adds a private method check_unsupported_queue_type to QueueFactory. 
If an unsupported queue type is sent in the arguments, we now let user know that that queue type will be ignored. 

run lavin and try to declare a quorum queue, observe logs

### DIFF
--- a/src/lavinmq/queue/queue_factory.cr
+++ b/src/lavinmq/queue/queue_factory.cr
@@ -57,7 +57,7 @@ module LavinMQ
 
     private def self.warn_if_unsupported_queue_type(frame)
       if frame.arguments["x-queue-type"]?
-        Log.warn { "The queue type #{frame.arguments["x-queue-type"]} is not supported by LavinMQ and will revert to default" }
+        Log.warn { "The queue type #{frame.arguments["x-queue-type"]} is not supported by LavinMQ and will be changed to the default queue type" }
       end
     end
   end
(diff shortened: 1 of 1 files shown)


src/lavinmq/queue/queue_factory.cr
```diff
@@ -52,5 +54,11 @@ module LavinMQ
     private def self.stream_queue?(frame) : Bool
       frame.arguments["x-queue-type"]? == "stream"
     end
+
+    private def self.check_unsupported_queue_type(frame)
```
viktorerlingsson, line 58: Could the name of the method be `warn_if_unsupported_queue_type` instead? I think it's easier to understand what it does then
```suggestion
    private def self.warn_if_unsupported_queue_type(frame)
```

kickster97: sure!

src/lavinmq/queue/queue_factory.cr
```diff
@@ -52,5 +54,11 @@ module LavinMQ
...
+      if frame.arguments["x-queue-type"]?
+        Log.warn { "The queue type #{frame.arguments["x-queue-type"]} is not supported by LavinMQ and will revert to default" }
```
viktorerlingsson, line 60: Maybe this is a little more clear? Just a suggestion though!
```suggestion
        Log.warn { "The queue type #{frame.arguments["x-queue-type"]} is not supported by LavinMQ and will be changed to the default queue type" }
```

kickster97: nice!

viktorerlingsson (approved):

PR #795, opened 2024-09-30
Download original java-client code and apply patch

snichme:

Same as being done to rabbit-hole tests, clone original repo and apply patch to disable tests that LavinMQ doesn't support or makes no sense to LavinMQ

Check GHA

### DIFF
--- a/.github/rabbitmq-java-client.patch
+++ b/.github/rabbitmq-java-client.patch
@@ -1,0 +1,567 @@
+From 6a96bec6d484bbe156b3cd4ee326ae4cadd81e5c Mon Sep 17 00:00:00 2001
+From: Magnus Landerblom <magnus@landerblom.se>
+Date: Mon, 30 Sep 2024 22:44:56 +0200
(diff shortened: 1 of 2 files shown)


kickster97 (approved):
nice!

snichme:
The check fails now because the `curl` command to download the patch points to `main` but the branch isn't merged. 
So it will work when merged

dentarg:
> The check fails now because the `curl` command to download the patch points to `main` but the branch isn't merged.
> So it will work when merged

Why not apply the patch from the current branch? I guess this step isn't cloning the lavinmq repo and instead uses artifact from other build steps, but I think cloning the repo to get the patch is better, then you can patch the patch in PRs and see it working?

dentarg:
Here's how you can checkout just a single file: https://github.com/84codes/actions/blob/78e025f017394020a4a7be07e53546ba33100235/.github/workflows/ruby-ci.yml#L93-L102

snichme:
@dentarg great tips, fixed that now

PR #798, opened 2024-10-03
Add tailwind complier

baelter:
Adding support for tailwind

`make css` and `make tw-watch`


makefiles/tailwindcss.mk
```diff
@@ -0,0 +1,57 @@
+# Variables
+TW_VERSION := 3.4.13
+TW_BASE_URL := https://github.com/tailwindlabs/tailwindcss/releases/download/v$(TW_VERSION)
+TW_CHECKSUM_FILE := sha256sums.txt
```
dentarg, line 4: I think we should commit this file in the repo, otherwise we can not be sure of its integrity (unless the tailwind project would sign it and we could verify its integrity that way). It should be updated when we bump the version here.

carlhoerberg:
wow, a 123MB binary required to compile css styles? Pretty far from the hand crafted CSS we today. Tailwind better be good.. I would much rather see us continue and improve the current CSS.

baelter:
> wow, a 123MB binary required to compile css styles? Pretty far from the hand crafted CSS we got today. Tailwind better be good and save us oceans of time.. I would much rather see us continue and improve the current CSS.

"Only" ~40mb :) But I agree it doesn't feel necessary with a css compiler. However if the design team really want to use tailwind (they already built the new UI in lavin and the cloudamqp console in it) I can live with it. At least the final css is small. 

Another option would be to just commit the compiled css to lavin and let developers worry about the compiler.

Anyway, I learned a lot about makefiles :)

baelter:
Talked to @bengtmagnus and we will build our own css (based on tailwind) without need for any compiler so this won't be needed right now.

carlhoerberg:
:heart: 

> "Only" ~40mb :) But I agree it doesn't feel necessary with a css compiler. However if the design team really want to use tailwind (they already built the new UI in lavin and the cloudamqp console in it) I can live with it. At least the final css is small.

You're right 3.x is 40MB, I see now that I looked at [v4.0.0-alpha.26](https://github.com/tailwindlabs/tailwindcss/releases/tag/v4.0.0-alpha.26) where the binary had ballooned 3x for some reason :grimacing:

PR #800, opened 2024-10-08
Uniform log sources

spuun:
Create all `Log` instances from `LavinMQ::Log`. This will make all log sources prefixed with "lmq" which in turn makes it easy to see if a log comes from a lib (e.g. amqp-client) or from lavinmq itself.

Run lavin with `--debug` and look at log output.


snichme (approved):

viktorerlingsson (approved):

carlhoerberg:
Should probably `require "./lavinmq/logger"` in each file that needs it, rather on a top level. If we want to compile any part independently. Pretty sure vim will complain otherwise, as it won't be able to compile it to check formatting, linting etc.

carlhoerberg (commented):
Should probably require "./lavinmq/logger" in each file that needs it, rather on a top level. If we want to compile any part independently. Pretty sure vim will complain otherwise, as it won't be able to compile it to check formatting, linting etc.

carlhoerberg (approved):

PR #802, opened 2024-10-09
Watch, recompile and reload view changes

baelter:
To enable fast iteration on UI development this PR adds a couple of things:

1. make targets and associated files to compile ECR views separately and serve them from the static/ folder.
2. make target to watch ECR files and recompile them when they change.
3. livereload client and server to refresh the browser when views a recompiled.

Run two terminals with
`bin/lavinmq`
and
`make dev-ui`
Open a browser and make some changes to ECR files and behold.


views/vhost.ecr
```diff
@@ -110,6 +110,6 @@
         </form>
       </section>
     </main>
-    <% render "footer" %>
+    <% render "partials/footer" %>
   </body>
-</html>
+</etml>
```
carlhoerberg, line 115: Mistake?

views/vhost.ecr
```diff
@@ -1,12 +1,12 @@
 <%- pagename = "Virtual host" -%>
-<!DOCTYPE html>
+<!DOCTYPE >
```
carlhoerberg, line 2: ?

views/partials/head.ecr
```diff
@@ -10,3 +10,6 @@
 <link rel="apple-touch-icon" sizes="180x180" href="img/favicon.png">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <meta name="google" content="notranslate">
+<%- {% unless flag?(:release) %} -%>
+  <script src="http://localhost:35729/livereload.js"></script>
```
carlhoerberg, line 14: Can the liveview server be started and integrated in Lavinmqs http server?

shard.lock
```diff
@@ -12,6 +12,14 @@ shards:
...
     version: 1.2.7
 
+  inotify:
+    git: https://github.com/petoem/inotify.cr.git
+    version: 1.0.3
+
+  live_reload:
+    git: https://github.com/bcardiff/live_reload.cr.git
+    version: 0.1.0
+
```
carlhoerberg, line 22: Could be development dependencies?

baelter: They are?

views/partials/head.ecr
baelter: Yes

baelter: ... but, I don't think it's worth the effort? You could perhaps set another http server, setup the websocket handler, keep tabs of connection and handle the js yourself. https://github.com/bcardiff/live_reload.cr/blob/master/src/live_reload.cr#L140

But running a separate server like this seems to the standard thing to do https://github.com/livereload/livereload-js?tab=readme-ov-file#what-is-livereload

spuun:
```
 6 | watcher = Inotify.watch dir do |event|
                       ^----
Error: undefined method 'watch' for Inotify:Module
```

Inotify only supports linux:
https://github.com/petoem/inotify.cr/blob/master/src/inotify.cr#L3

spuun (approved):

viktorerlingsson:
Shouldn't live_reload be a `development_dependency`?

baelter:
> Shouldn't live_reload be a `development_dependency`?

No, I removed the dependency in running live_reload in crystal. https://github.com/cloudamqp/lavinmq/pull/802#discussion_r1795464183

PR #805, opened 2024-10-11
add clustering_max_lag to /metrics

viktorerlingsson:
Adds `clustering_max_lag` to prometheus metrics 

`curl http://guest:guest@localhost:15672/metrics`

### DIFF
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -264,6 +264,10 @@ module LavinMQ
                       value: @amqp_server.followers.size,
                       type:  "gauge",
                       help:  "Amount of follower nodes connected"})
+        writer.write({name:  "clustering_max_lag",
+                      value: Config.instance.clustering_max_lag,
+                      type:  "gauge",
+                      help:  "Max unsynced replicated messages"})
         @amqp_server.followers.each do |f|
           writer.write({name:   "follower_lag",
                         labels: {id: f.id.to_s(36)},
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -267,7 +267,7 @@ module LavinMQ
         writer.write({name:  "clustering_max_lag",
                       value: Config.instance.clustering_max_lag,
                       type:  "gauge",
-                      help:  "Max unsynced replicated messages"})
+                      help:  "Maximum allowed unsynced replicated messages"})
         @amqp_server.followers.each do |f|
           writer.write({name:   "follower_lag",
                         labels: {id: f.id.to_s(36)},


src/lavinmq/http/controller/prometheus.cr
```diff
@@ -264,6 +264,10 @@ module LavinMQ
                       value: @amqp_server.followers.size,
                       type:  "gauge",
                       help:  "Amount of follower nodes connected"})
+        writer.write({name:  "clustering_max_lag",
+                      value: Config.instance.clustering_max_lag,
+                      type:  "gauge",
+                      help:  "Max unsynced replicated messages"})
```
dentarg, line 270: WDYT?

```suggestion
                      help:  "Maximum allowed unsynced replicated messages"})
```

Also, here it says "actions" https://github.com/cloudamqp/lavinmq/blob/a05fb6b28332776b66d2129f8f9ce7750e2a1532/src/lavinmq/config.cr#L58, not sure if we want to be consistent :)

viktorerlingsson: I just copied the text from here https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/config.cr#L136 🙈 
But yeah, we should be consistent!

dentarg: Maybe easiest to update the comment then, hehe :)

viktorerlingsson: Yeah :) 
Hmm, this all might be a little confusing. `clustering_max_lag` is the maximum number of actions that a follower can lag behind, but the `lag` we show in GUI/metrics is how many actual bytes the follower is lagging behind. So the numbers don't really have much to do with each other... 🤔

snichme (approved):

src/lavinmq/http/controller/prometheus.cr
kickster97: Agree that it can be confusing. Could we name it something like max_allowed_cluster_actions, and let lag be lag_bytes?

viktorerlingsson: Yeah, it might be reasonable to rename at least one of them. 
Maybe rename 
`lag` -> `lag_in_bytes`, 
`max_lag` -> `max_lag_in_actions`, 
and add `lag_in_actions` 🤔

viktorerlingsson: Here's PRs for the changes mentioned in this thread
https://github.com/cloudamqp/lavinmq/pull/810
https://github.com/cloudamqp/lavinmq/pull/811

kickster97 (approved):

viktorerlingsson:
Closing this because we decided to not implement metrics for `lag_in_actions`, so this doesn't make much sense to implement.

PR #811, opened 2024-10-16
Adds lag_in_actions to metrics

viktorerlingsson:
Adds number of unsynced actions as `lag_in_actions` to metrics, by checking the size of the `@actions` channel's internal Dequeue. 

Manual

### DIFF
--- a/src/lavinmq/clustering/follower.cr
+++ b/src/lavinmq/clustering/follower.cr
@@ -177,6 +177,7 @@ module LavinMQ
           sent_bytes:         @sent_bytes,
           acked_bytes:        @acked_bytes,
           lag_in_bytes:       lag_in_bytes,
+          lag_in_actions:     lag_in_actions,
           compression_ratio:  @lz4.compression_ratio,
           uncompressed_bytes: @lz4.uncompressed_bytes,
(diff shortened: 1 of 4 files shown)


kickster97 (approved):

carlhoerberg:
But it's not the whole story, right? Actions could still be in the tcp send buffer, or being processed at the follower.

viktorerlingsson:
> But it's not the whole story, right? Actions could still be in the tcp send buffer, or being processed at the follower.

Hmm, yeah that's true. And I guess there's no good way for us to know how many actions are there. 🤔

src/lavinmq/clustering/follower.cr
```diff
@@ -177,6 +177,7 @@ module LavinMQ
           sent_bytes:         @sent_bytes,
           acked_bytes:        @acked_bytes,
           lag_in_bytes:       lag_in_bytes,
+          lag_in_actions:     lag_in_actions,
```
snichme, line 180: We talked about calling it `clustering_actions_queue_size` instead

viktorerlingsson:
Closing this since it doesn't really show useful/correct data, since actions can be in the tcp send buffer, or being processed at the follower.

PR #812, opened 2024-10-18
Clustering - Write to tempfile first when getting full file

### LINKED ISSUE #806: Clustering - add action for renaming replicated files
Add a RenameAction to clustering.

Needed for https://github.com/cloudamqp/lavinmq/pull/661

We probably need to add a more explicit way to select what action to be taken. 
Now it's determined by an Int64 that also holds the length of the content to sync: `delete` file if it's 0, `append` to file if it's negative and `add` file if it's positive. 

So maybe add an Int8 for action selection.

viktorerlingsson:
Writes full file to a tempfile first and then renames it to replace the old file when getting full file on followers. This ensures that followers always has a complete file ready. 

The `Clustering::AddAction` is only ever used by `replace_file` as far as I can tell, meaning we shouldn't have to look for existing files. 

Fixes #806 

Run spec.

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -228,11 +228,12 @@ module LavinMQ
             else
               File.delete? File.join(@data_dir, filename)
             end
-          when .positive? # full file is coming
-            Log.debug { "Getting full file #{filename} (#{len} bytes)" }
-            f = @files[filename]
-            f.truncate
+          when .positive? # replace file
+            Log.debug { "Replacing file #{filename} (#{len} bytes)" }
+            f = @files["#{filename}.tmp"]
             IO.copy(lz4, f, len) == len || raise IO::EOFError.new("Full file not received")
+            f.rename f.path[0..-5]
+            @files.delete("#{filename}.tmp")
           end
           ack_bytes = len.abs + sizeof(Int64) + filename_len + sizeof(Int32)
           acks.send(ack_bytes)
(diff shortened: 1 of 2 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -228,11 +228,12 @@ module LavinMQ
...
-          when .positive? # full file is coming
-            Log.debug { "Getting full file #{filename} (#{len} bytes)" }
-            f = @files[filename]
-            f.truncate
+          when .positive? # replace file
+            Log.debug { "Replacing file #{filename} (#{len} bytes)" }
+            f = @files["#{filename}.tmp"]
             IO.copy(lz4, f, len) == len || raise IO::EOFError.new("Full file not received")
+            f.rename f.path[0..-5]
+            @files.delete("#{filename}.tmp")
```
spuun, line 236: I don't know if it's necessary to add the file back to the hash with it's new name, it would still be opened next time it's accessed.

```suggestion
            @files[filename] = @files.delete("#{filename}.tmp")
```

If not adding it back we can probably close it right away instead of letting it happen during GC.

viktorerlingsson: > I don't know if it's necessary to add the file back to the hash with it's new name, it would still be opened next time it's accessed.

Yeah, shouldn't be necessary to add it back I think.

> If not adding it back we can probably close it right away instead of letting it happen during GC.

Hmm, yeah. Not sure if I understand how that suggested change accomplishes that though?

carlhoerberg:
```suggestion
            @files.delete("#{filename}.tmp").close
```

carlhoerberg (approved):

PR #813, opened 2024-10-21
Consumer default prefetch

viktorerlingsson:
Adds a configurable `default_consumer_prefetch` for all consumers. Defaults to UInt16::MAX (65535), and can be set either in config (`default_consumer_prefetch`) or cli (`--default-consumer-prefetch=`). 

Manual

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -103,7 +103,7 @@ class Throughput < Perf
     @parser.on("-p", "--persistent", "Persistent messages (default false)") do
       @persistent = true
     end
-    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default 0, unlimited)") do |v|
+    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default @prefetch)") do |v|
       @prefetch = v.to_u16
     end
     @parser.on("-g", "--poll", "Poll with basic_get instead of consuming") do
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -44,7 +44,7 @@ class Throughput < Perf
   @consume_rate = 0
   @max_unconfirm = 0
   @persistent = false
-  @prefetch = 0_u32
+  @prefetch = UInt16::MAX
   @quiet = false
   @poll = false
   @json_output = false
@@ -104,7 +104,7 @@ class Throughput < Perf
       @persistent = true
     end
     @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default 0, unlimited)") do |v|
-      @prefetch = v.to_u32
+      @prefetch = v.to_u16
     end
     @parser.on("-g", "--poll", "Poll with basic_get instead of consuming") do
       @poll = true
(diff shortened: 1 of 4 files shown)


carlhoerberg:
Just came to think about it, what does the spec got to say about the initial prefetch? Is it anything in the spec that prevents is from doing this change?

spuun (commented):
Dunno what implications it may have to all of a sudden not treat `0` as unlimited. Maybe `0` and `UInt16::MAX` both should be treated as unlimited?

IMO having `0` as unlimited makes sense (since `0` prefetch doesn't make sense, I think?) while `UInt16::MAX` maybe shouldn't mean unlimited. Unlimited can still go beyond `UInt16::MAX`, right? (Probably bad in every use case though..)

carlhoerberg:
Can still make a unlimited prefetch by calling qos with 0? It's just that a new channel has a limited prefetch by default? We should not make uint16::max to mean unlimited, that doesnt makes sense.

spuun:
> Can still make a unlimited prefetch by calling qos with 0? It's just that a new channel has a limited prefetch by default? We should not make uint16::max to mean unlimited, that doesnt makes sense.

Yes, true. I see now that I misread some changes... 🤦

viktorerlingsson:
> Just came to think about it, what does the spec got to say about the initial prefetch? Is it anything in the spec that prevents is from doing this change?

No, there's nothing in the spec that would prevent this as far as I can tell. The only thing about prefetch I see is that the server should never send MORE messages than specified by prefetch, but it's OK to send less.

src/lavinmqperf.cr
```diff
@@ -104,7 +104,7 @@ class Throughput < Perf
       @persistent = true
     end
     @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default 0, unlimited)") do |v|
```
carlhoerberg, line 106:
```suggestion
    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default @prefetch)") do |v|
```

src/lavinmqperf.cr
```diff
@@ -103,8 +103,8 @@ class Throughput < Perf
     @parser.on("-p", "--persistent", "Persistent messages (default false)") do
       @persistent = true
     end
-    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default 0, unlimited)") do |v|
-      @prefetch = v.to_u32
+    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default @prefetch)") do |v|
```
carlhoerberg, line 106:
```suggestion
    @parser.on("-P", "--prefetch=number", "Number of messages to prefetch (default #{@prefetch})") do |v|
```

src/lavinmqperf.cr
```diff
@@ -264,7 +264,7 @@ class Throughput < Perf
         ch.queue(@queue, passive: true)
       end
       ch.tx_select if @ack_in_transaction > 0
-      ch.prefetch @prefetch unless @prefetch.zero?
+      ch.prefetch @prefetch unless @prefetch == UInt16::MAX
```
carlhoerberg, line 267: Maybe skip this as it would yield different prefetch values for lavin and rabbit
```suggestion
```

src/lavinmqperf.cr
```diff
@@ -303,7 +303,7 @@ class Throughput < Perf
         ch = a.channel
         ch.queue(@queue, passive: true)
       end
-      ch.prefetch @prefetch unless @prefetch.zero?
+      ch.prefetch @prefetch unless @prefetch == UInt16::MAX
```
carlhoerberg, line 306:
```suggestion
```

carlhoerberg (approved):

PR #816, opened 2024-10-25
Improve broker initiated client disconnects

spuun:
Writing to client sockets may block a fiber because the buffers are full. When this happens this fiber will hold the write lock in `Client` until it manage to write the data. Because if this it's not possible to force close a connection from the management UI, and the http request will timeout. It will also affect lavinmq shutdown because `VHost#close` will be blocked from closing all connections effectively.

This PR will: 
- add a timeout argument to `#close` that will set a `write_timeout` on the socket.
- spawn fibers to perform "async" disconnects in parallel on shutdown

Manually with a "bad" client, e.g. unlimited prefetch and no acking. Let it consume until LavinMQ stops because of tcp back pressue. Try kill the connection.

### DIFF
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -417,7 +417,7 @@ module LavinMQ
         @log.debug { "#{ex.inspect} when closing socket" }
       end
 
-      def close(reason = nil, timeout force_close_timeout : Time::Span = 60.seconds)
+      def close(reason = nil, timeout : Time::Span = 60.seconds)
         reason ||= "Connection closed"
         @log.info { "Closing: #{reason}" }
 
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -262,7 +262,7 @@ module LavinMQ
         end
         true
       rescue ex : IO::Error | OpenSSL::SSL::Error
-        @log.debug { "Lost connection, while sending (#{ex.inspect})" }
+        @log.debug { "Lost connection, while sending message (#{ex.inspect})" } unless closed?
         close_socket
         Fiber.yield
         false
@@ -410,23 +410,57 @@ module LavinMQ
         @vhost.rm_connection(self)
       end
 
-      private def close_socket
+      private def close_socket(force : Bool = false)
         @running = false
-        @socket.close
-        @log.debug { "Socket closed" }
+        if force && (socket = @socket.as?(Socket))
+          # Socket#close_write will close on "kernel level" which will ensure
+          # we abort any blocking writes. Socket#close tries to flush the socket
+          # etc, meaning that we may block for a long time (or forever for broken
+          # clients)
+          socket.close_write
+          socket.close
+        else
+          @socket.close
+        end
       rescue ex
         @log.debug { "#{ex.inspect} when closing socket" }
       end
 
-      def close(reason = nil)
+      def close(reason = nil, timeout force_close_timeout : Time::Span = 10.seconds)
         reason ||= "Connection closed"
-        @log.info { "Closing, #{reason}" }
-        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
-        @running = false
+        @log.info { "Closing: #{reason}" }
+
+        # We spawn fibers to make #close non-blocking. Required for e.g. http API calls.
+        # It's probably good for VHost#close too which will close all it's connections.
+        # If we run into problems with too many fibers we can probably make one timeout fiber
+        # handling all fibers we're about to close.
+
+        ch = ::Channel(Nil).new
+        # send may be block by @write_lock, therefore we spawn to send
+        spawn(name: "Client send Close") do
+          send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
+          @running = false
+          ch.send nil rescue ::Channel::ClosedError
+        rescue IO::Error
+        end
+
+        # The @write_lock may be locked "forever" (consumer not consuming and buffers filled up),
+        # so we may never be able to send Close. We therefore spawn this timeout fiber
+        spawn(name: "Client send Close tiemout") do
+          select
+          when ch.receive
+            @log.info { "Close frame sent, now force closing socket" }
+          when timeout force_close_timeout
+            @log.info { "Timeout sending close, force closing socket" }
+          end
+          ch.close
+          # We always force close... Maybe wrong?
+          force_close
+        end
       end
 
       def force_close
-        close_socket
+        close_socket(force: true)
       end
 
       def closed?
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -417,7 +417,7 @@ module LavinMQ
         @log.debug { "#{ex.inspect} when closing socket" }
       end
 
-      def close(reason = nil, timeout : Time::Span = 60.seconds)
+      def close(reason = nil, timeout : Time::Span = 5.seconds)
         reason ||= "Connection closed"
         @log.info { "Closing: #{reason}" }
 
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -421,13 +421,14 @@ module LavinMQ
         reason ||= "Connection closed"
         @log.info { "Closing: #{reason}" }
 
-        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
-        @running = false
-
         socket = @socket
         if socket.responds_to?(:"write_timeout=")
           socket.write_timeout = timeout
+          socket.read_timeout = timeout
         end
+
+        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
+        @running = false
 
         socket.close
       end
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -410,57 +410,27 @@ module LavinMQ
         @vhost.rm_connection(self)
       end
 
-      private def close_socket(force : Bool = false)
+      private def close_socket
         @running = false
-        return @socket.close unless force
-
-        case @socket
-        when Socket
-          @socket.close_write
-        when OpenSSL::Socket
-          @socket.@bio.io.close_write
-        end
-
         @socket.close
       rescue ex
         @log.debug { "#{ex.inspect} when closing socket" }
       end
 
-      def close(reason = nil, timeout force_close_timeout : Time::Span = 10.seconds)
+      def close(reason = nil, timeout force_close_timeout : Time::Span = 60.seconds)
         reason ||= "Connection closed"
         @log.info { "Closing: #{reason}" }
 
-        # We spawn fibers to make #close non-blocking. Required for e.g. http API calls.
-        # It's probably good for VHost#close too which will close all it's connections.
-        # If we run into problems with too many fibers we can probably make one timeout fiber
-        # handling all fibers we're about to close.
-
-        ch = ::Channel(Nil).new
-        # send may be block by @write_lock, therefore we spawn to send
-        spawn(name: "Client send Close") do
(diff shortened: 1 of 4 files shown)


src/lavinmq/amqp/client.cr
```diff
@@ -410,23 +410,57 @@ module LavinMQ
         @vhost.rm_connection(self)
       end
 
-      private def close_socket
+      private def close_socket(force : Bool = false)
         @running = false
-        @socket.close
-        @log.debug { "Socket closed" }
+        if force && (socket = @socket.as?(Socket))
```
carlhoerberg, line 415: for `OpenSSL::SSL::Socket` you can do `@socket.@bio.io.close_write`

src/lavinmq/amqp/client.cr
```diff
@@ -410,23 +410,57 @@ module LavinMQ
...
         reason ||= "Connection closed"
-        @log.info { "Closing, #{reason}" }
-        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
-        @running = false
+        @log.info { "Closing: #{reason}" }
+
+        # We spawn fibers to make #close non-blocking. Required for e.g. http API calls.
+        # It's probably good for VHost#close too which will close all it's connections.
+        # If we run into problems with too many fibers we can probably make one timeout fiber
+        # handling all fibers we're about to close.
```
carlhoerberg, line 436: Could we avoid fibers by using `Socket#write_timeout` and `Socket#read_timeout`? Then on `IO::TimeoutError` here and in the read_loop we force close the socket?

src/lavinmq/amqp/client.cr
carlhoerberg: or monkey patch `OpenSSL::SSL::Socket`

spuun: Oh, I actually forgot about that case!

src/lavinmq/amqp/client.cr
spuun: Yeah, maybe that could be an option. Will give it a try!

spuun: Seems to work!

src/lavinmq/amqp/client.cr
```diff
@@ -413,16 +413,20 @@ module LavinMQ
       private def close_socket
         @running = false
         @socket.close
-        @log.debug { "Socket closed" }
       rescue ex
         @log.debug { "#{ex.inspect} when closing socket" }
       end
 
-      def close(reason = nil)
+      def close(reason = nil, timeout : Time::Span = 60.seconds)
```
carlhoerberg, line 420: isn't 60s very long?

carlhoerberg: ok, i think we can be more agressive, 5 or 10s?

carlhoerberg (approved):

src/lavinmq/amqp/client.cr
```diff
@@ -413,16 +413,20 @@ module LavinMQ
...
         reason ||= "Connection closed"
-        @log.info { "Closing, #{reason}" }
-        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
-        @running = false
+        @log.info { "Closing: #{reason}" }
+
+        socket = @socket
+        if socket.responds_to?(:"write_timeout=")
+          socket.write_timeout = force_close_timeout
```
carlhoerberg, line 426:
```suggestion
          socket.write_timeout = timeout
```

src/lavinmq/amqp/client.cr
```diff
@@ -413,16 +413,20 @@ module LavinMQ
...
-        send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
-        @running = false
+        @log.info { "Closing: #{reason}" }
+
+        socket = @socket
+        if socket.responds_to?(:"write_timeout=")
+          socket.write_timeout = timeout
+        end
+
+        socket.close
```
carlhoerberg, line 429: we're not sending a Close::Connection frame anymore?

carlhoerberg (changes requested):

src/lavinmq/amqp/client.cr
spuun: We are locally on my computer. I guess i forgot to push that commit 🙃

src/lavinmq/amqp/client.cr
```diff
@@ -413,16 +413,23 @@ module LavinMQ
...
-        @log.info { "Closing, #{reason}" }
+        @log.info { "Closing: #{reason}" }
+
         send AMQP::Frame::Connection::Close.new(320_u16, "CONNECTION_FORCED - #{reason}", 0_u16, 0_u16)
         @running = false
+
+        socket = @socket
+        if socket.responds_to?(:"write_timeout=")
+          socket.write_timeout = timeout
+        end
```
carlhoerberg, line 430: shouldn't the write timeout be before the actual send? 

How about also setting a `read_timeout`? So that the read_loop will close the connection even if it doesn't receive a frame (because the client is overloaded or something).

carlhoerberg (changes requested):

src/lavinmq/amqp/client.cr
spuun: > How about also setting a `read_timeout`? So that the read_loop will close the connection even if it doesn't receive a frame (because the client is overloaded or something).
Yeah, probably. I didn't because I think we always set a read timeout.

carlhoerberg:
Yes, but im thinking we can have that logic elsewhere, like push all connections to a Channel and then let a (limited) amount of fibers to work it off until it's empty.

spuun:
Updated the title and description. I've added the solution with fibers and a channel to disconnects clients "async" on shutdown.

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,68 @@ module LavinMQ
...
+        wg.add
+        spawn(name: "vhost close connection #{fiber_id}") do
+          loop do
+            select
+            when client = ch.receive
+              client.close(reason)
+            when timeout 50.milliseconds
+              break
+            end
+          rescue
```
carlhoerberg, line 389: log something

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,68 @@ module LavinMQ
...
+          end
```
carlhoerberg, line 390: close the channel after sending all instead, eliminated timing issues. 
```suggestion
          while client = ch.receive?
            client.close(reason)
          end
```

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,68 @@ module LavinMQ
...
+            break
+          when timeout 200.milliseconds
+            # Spawn more fibers if all current are too busy closing "zombies"
+            spawn_close_fiber.call
+          end
+        end
+      end
+
+      wg.wait
+      ch.close
```
carlhoerberg, line 413: close the channel first, so that the fibers can stop, then wait for all fibers to finish

```suggestion
      ch.close
      wg.wait
```

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,68 @@ module LavinMQ
       upstreams.stop_all
     end
 
+    private def close_connections(reason, close_done : Channel(Nil))
+      ch = Channel(Client).new
```
carlhoerberg, line 373: use a buffered `Channel` instead? And add fibers if `wg.wait` times out? would it result in less code? 

```suggestion
      ch = Channel(Client).new(@connections.size)
```

src/lavinmq/http/controller/connections.cr
```diff
@@ -38,7 +38,7 @@ module LavinMQ
         delete "/api/connections/:name" do |context, params|
           with_connection(context, params) do |c|
             reason = context.request.headers["X-Reason"]? || "Closed via management plugin"
-            c.close(reason)
+            c.close(reason, timeout: 3.seconds)
```
carlhoerberg, line 41: the default 5s is not good enough here?

src/lavinmq/vhost.cr
spuun: Hm, not that easy. No timeout in `WaitGroup#wait`.

The waitgroup is make sure we're not returning from `VHost#close` before all `Client#close` is done.
The timeout to send is to know that we're not stuck with closing.

I think a change would require something like

```crystal
@connections.times do 
  loop do
    select
    when connection_closed.recieve?
      break
    when timout 100.milliseconds
      spawn_close_fiber.call
    end
  end
end
```

together with a `connection_closed.send` after `Client#close`.

Or am I missing something? 🤔

carlhoerberg: ok

src/lavinmq/vhost.cr
```diff
@@ -379,15 +379,7 @@ module LavinMQ
...
-          loop do
-            select
-            when client = ch.receive
-              client.close(reason)
-            when timeout 50.milliseconds
-              break
-            end
-          rescue
-          end
+          loop { client.close ch.receive? }
```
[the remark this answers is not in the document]
carlhoerberg: ah, fixed in latest commit :)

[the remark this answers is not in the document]
spuun: Don't ask me what I did there, haha. I think i mixed that up with another thing I was doing.

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,54 @@ module LavinMQ
...
+        @log.debug { "Close sent to all connections" }
+        close_done.close
       end
+
+      select
+      when close_done.receive?
+        @log.info { "All connections closed gracefully" }
+      when timeout 5.seconds
+        @log.warn { "Timeout waiting for connections to close. #{@connections.size} left." }
+      end
```
carlhoerberg, line 417: Ah, my bad, for some reason I was thinking this was in a loop, that we logged every 5s but didn't timeout until after like 30-60s. 

```suggestion
      6.times do |i|
        select
        when close_done.receive?
          @log.info { "All connections closed gracefully" }
          return
        when timeout 5.seconds
          @log.warn { "Waiting for #{@connections.size} connections to close gracefully, #{(5-i)*5} more seconds" }
        end
      end
```

Or we just resort to 10-15s timeout. 5s feels a bit too short when dealing with huge amount of connections.

carlhoerberg (approved):

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,54 @@ module LavinMQ
...
+            else
+              fiber_id = fiber_count &+= 1
+              @log.trace { "spawning close conn fiber #{fiber_id} " }
+              wg.spawn do
+                while client_to_close = to_close.receive?
+                  client_to_close.close(reason)
+                end
+                @log.trace { "exiting close conn fiber #{fiber_id} " }
+              end
+              Fiber.yield
```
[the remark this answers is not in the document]
spuun: I'm not sure why, but it did spawn one extra fiber. Not a big deal, but...

src/lavinmq/vhost.cr
```diff
@@ -369,18 +369,54 @@ module LavinMQ
...
+        fiber_count = 0
+        @connections.each do |client|
+          loop do
+            select
+            when to_close.send client
+              break
+            else
+              fiber_id = fiber_count &+= 1
+              @log.trace { "spawning close conn fiber #{fiber_id} " }
+              wg.spawn do
```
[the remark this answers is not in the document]
spuun: Agree.

src/lavinmq/vhost.cr
spuun: Yes, I think one longer timeout is fine.

src/lavinmq/vhost_store.cr
```diff
@@ -43,7 +43,11 @@ module LavinMQ
     end
 
     def close
-      @vhosts.each_value &.close
+      WaitGroup.wait do |wg|
+        @vhosts.each_value do |vhost|
+          wg.spawn &->vhost.close
+        end
+      end
```
spuun, line 50: @carlhoerberg wdyt about this? If we don't have something like this it make take a very long time to shutdown a broker with many vhosts.

carlhoerberg: Perfect!

src/lavinmq/vhost.cr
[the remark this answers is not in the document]
spuun: PR sent.

PR #817, opened 2024-10-25
Streams - keep track of first offset and timestamp per segment

viktorerlingsson:
Saves a hash with the first offset for each segment, greatly reducing the time (and memory used) to find a message by offset.
Does the same for timestamps. 

Tests done by getting a message by a random offset 10 times in a stream with 10M messages. 

main: 
```
Getting 10 messages by random offsets...
Average time: 660.8 milliseconds
Maximum resident set size (kbytes): 824424
```

this branch: 
```
Getting 10 messages by random offsets...
Average time: 52.9 milliseconds
Maximum resident set size (kbytes): 23692
```

Existing specs in `stream_queue_spec.cr` should cover this pretty well.

### DIFF
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -67,7 +67,6 @@ module LavinMQ::AMQP
         segment = find_segment_by_offset(offset)
         pos = 4u32
         msg_offset = 0i64
-
         loop do
           rfile = @segments[segment]?
           if rfile.nil? || pos == rfile.size
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -9,14 +9,13 @@ module LavinMQ::AMQP
       property max_age : Time::Span | Time::MonthSpan | Nil
       getter last_offset : Int64
       @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
-      @offset_index : Hash(UInt32, Int64)              # segment_id => offset of first msg
-      @timestamp_index : Hash(UInt32, Int64)           # segment_id => ts of first msg
+      @offset_index = Hash(UInt32, Int64).new          # segment_id => offset of first msg
+      @timestamp_index = Hash(UInt32, Int64).new       # segment_id => ts of first msg
 
       def initialize(*args, **kwargs)
         super
         @last_offset = get_last_offset
-        @offset_index = build_segment_offset_index
-        @timestamp_index = build_segment_offset_index_ts
+        build_segment_indexes
         drop_overflow
       end
 
@@ -235,21 +234,14 @@ module LavinMQ::AMQP
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
       end
 
-      private def build_segment_offset_index : Hash(UInt32, Int64)
-        @segments.transform_values do |mfile|
+      private def build_segment_indexes
+        @segments.each do |seg_id, mfile|
           msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
-          offset_from_headers(msg.properties.headers)
+          @offset_index[seg_id] = offset_from_headers(msg.properties.headers)
+          @timestamp_index[seg_id] = msg.timestamp
         rescue IndexError
-          @last_offset
-        end
-      end
-
-      private def build_segment_offset_index_ts : Hash(UInt32, Int64)
-        @segments.transform_values do |mfile|
-          msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
-          msg.timestamp
-        rescue IndexError
-          RoughTime.unix_ms
+          @offset_index[seg_id] = @last_offset
+          @timestamp_index[seg_id] = RoughTime.unix_ms
         end
       end
     end
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -9,11 +9,13 @@ module LavinMQ::AMQP
       property max_age : Time::Span | Time::MonthSpan | Nil
       getter last_offset : Int64
       @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
+      @first_offset_per_segment = Hash(UInt32, Int64).new
 
       def initialize(*args, **kwargs)
         super
         @last_offset = get_last_offset
         drop_overflow
+        find_first_offset_per_segment
       end
 
       private def get_last_offset : Int64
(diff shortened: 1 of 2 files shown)


spuun:
Cool!
Maybe a sorted array could perform even better since that would enable binary search. Store tuples of `{UInt32, Int64}`.
Which makes me wonder... Will the keys in the Hash always be sorted? I.e. are they always inserted sorted?

viktorerlingsson:
> Maybe a sorted array could perform even better since that would enable binary search. Store tuples of `{UInt32, Int64}`. Which makes me wonder... Will the keys in the Hash always be sorted? I.e. are they always inserted sorted?

Yeah, probably! I don't think it will make a huge difference, since the number of segments won't be super big. But can give it a try 👍 

Yes, it should always be sorted. We only add larger offsets and delete the smallest.

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -211,6 +225,14 @@ module LavinMQ
       private def offset_from_headers(headers) : Int64
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
       end
+
+      private def find_first_offset_per_segment
```
carlhoerberg, line 229:
```suggestion
      private def build_segment_offset_index
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -9,11 +9,13 @@ module LavinMQ
       property max_age : Time::Span | Time::MonthSpan | Nil
       getter last_offset : Int64
       @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
+      @first_offset_per_segment = Hash(UInt32, Int64).new
```
carlhoerberg, line 12: better naming?
```suggestion
      @offset_index = Hash(UInt32, Int64).new # segment_id => offset of first msg
```
or 
```suggestion
      @segment_first_index = Hash(UInt32, Int64).new
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -211,6 +225,14 @@ module LavinMQ
       private def offset_from_headers(headers) : Int64
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
       end
+
+      private def find_first_offset_per_segment
+        @segments.each do |seg_id, mfile|
+          msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
+          @first_offset_per_segment[seg_id] = offset_from_headers(msg.properties.headers)
+        rescue IndexError
+        end
```
carlhoerberg, line 234: I would create the new hash here and return it in the constuctor:

```suggestion
        @segments.transform_values do |seg_id, mfile|
          msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
          offset_from_headers(msg.properties.headers)
        end
```

src/lavinmq/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +90,16 @@ module LavinMQ
         {msg_offset, segment, pos}
       end
 
+      private def find_segment_by_offset(offset) : UInt32
```
carlhoerberg, line 93:
```suggestion
      private def offset_index_lookup(offset) : UInt32
```

carlhoerberg (commented):
Good feature!

Could theoretically build an index over timestamps too, right?

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -9,10 +9,14 @@ module LavinMQ::AMQP
       property max_age : Time::Span | Time::MonthSpan | Nil
       getter last_offset : Int64
       @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
+      @offset_index : Hash(UInt32, Int64)              # segment_id => offset of first msg
+      @offset_index_ts : Hash(UInt32, Int64)           # segment_id => ts of first msg
```
carlhoerberg, line 13: better name that describes the property better? 
```suggestion
      @timestamp_index : Hash(UInt32, Int64)           # segment_id => ts of first msg
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +92,23 @@ module LavinMQ::AMQP
         {msg_offset, segment, pos}
       end
 
+      private def offset_index_lookup(offset) : UInt32
+        seg = @segments.first_key
+        case offset
+        when Int
+          @offset_index.each do |seg_id, first_seg_offset|
+            break if first_seg_offset >= offset
```
carlhoerberg, line 100: if `first_seg_offset` matches the searched for offset we should give that segment, not the full previous segment, right? 
```suggestion
            break if first_seg_offset > offset
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -88,6 +92,23 @@ module LavinMQ::AMQP
...
+            seg = seg_id
+          end
+        when Time
+          @offset_index_ts.each do |seg_id, first_seg_ts|
+            break if Time.unix_ms(first_seg_ts) >= offset
```
carlhoerberg, line 105:
```suggestion
            break if Time.unix_ms(first_seg_ts) > offset
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -147,6 +168,7 @@ module LavinMQ::AMQP
       private def open_new_segment(next_msg_size = 0) : MFile
         super.tap do
           drop_overflow
+          @offset_index[@segments.last_key] = @last_offset
```
carlhoerberg, line 171: is `@last_offset` really correct here? or is it `@last_offset + 1` that will be written as the first offset to the new segment? 
```suggestion
          @offset_index[@segments.last_key] = @last_offset
```

carlhoerberg: what about `@timestamp_index`? It isn't updated?

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -178,6 +200,7 @@ module LavinMQ::AMQP
           msg_count = @segment_msg_count.delete(seg_id)
           @size -= msg_count if msg_count
           @segment_last_ts.delete(seg_id)
+          @offset_index.delete(seg_id)
```
carlhoerberg, line 203:
```suggestion
          @offset_index.delete(seg_id)
          @timestamp_index.delete(seg_id)
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -211,6 +234,24 @@ module LavinMQ::AMQP
...
+      end
+
+      private def build_segment_offset_index_ts : Hash(UInt32, Int64)
+        @segments.transform_values do |mfile|
+          msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
+          msg.timestamp
+        rescue IndexError
+          RoughTime.unix_ms
+        end
+      end
```
carlhoerberg, line 254: Build both indexes at the same time? Now we have to read all segments twice.

carlhoerberg (changes requested):

PR #832, opened 2024-11-08
Add both script and Makefile target for building rpm

riffraff169:
This PR adds a Makefile target for building rpms using the spec file, and adds a script that builds and also installs rpmdevtools to setup the rpmbuild tree. The Makefile target assumes rpmdevtools is already installed.
This also edits lavinmq.spec to reflect the current version.

The Makefile target can be tested just by running `make rpm`.
The script can be tested by running it. It works from either inside the extras directory, or in the base dir. The script is build_rpm.sh, so from either lavinmq directory: `./extras/build_rpm.sh`, or from the extras directory as `./build_rpm.sh`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,12 @@ test: lib
 format:
 	crystal tool format --check
 
+rpm: clean
+	rpmdev-setuptree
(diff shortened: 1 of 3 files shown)


riffraff169:

Issue #829, opened 2024-11-07
LavinMQ crashing during startup because of "Lease lost <number>"


viktorerlingsson:
**Describe the bug**
LavinMQ failed to start on a single node cluster with a "Lost lease" error from `lease_keepalive`. 

```
Nov 07 10:35:48 systemd Starting LavinMQ...
...
<launching, loading data...>
...
Nov 07 10:35:55 lavinmq lmq.etcd Lost leadership of lavinmq/leader: Lost lease 3008003254974389534
....
Nov 07 10:35:58 lavinmq lmq.launcher Lost leadership lease
Nov 07 10:35:58 lavinmq lmq.launcher Stopping
```

This unexpected shutdown during the startup then caused other errors in the startup, and left the LavinMQ service in an inactive (dead) state. 

Logs from after "Stopping": 
```
Nov 07 10:35:58 lavinmq Unhandled exception in spawn(name: HTTP listener): Can't re-start closed server (Exception)
Nov 07 10:35:58 lavinmq   from usr/share/crystal/src/http/server.cr:457:5 in 'listen'
Nov 07 10:35:58 lavinmq   from usr/share/crystal/src/fiber.cr:143:11 in 'run'
Nov 07 10:35:58 lavinmq   from ???
Nov 07 10:35:58 systemd Started LavinMQ.
```

Service status
```
lavinmq.service - LavinMQ
     Loaded: loaded (/lib/systemd/system/lavinmq.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2024-11-07 09:03:28 UTC; 4min 58s ago
       Docs: https://github.com/cloudamqp/lavinmq
    Process: 988772 ExecStart=/usr/bin/lavinmq -c /etc/lavinmq/lavinmq.ini (code=exited, status=0/SUCCESS)
   Main PID: 988772 (code=exited, status=0/SUCCESS)
```

**Describe your setup**
One single node running ubuntu 22.04.
LavinMQ v2.0.0 and etcd 3.5.16 running on the same machine. 

**How to reproduce**
Have not been able to reproduce it again. 

**Expected behavior**
LavinMQ starts, gets lease from etcd and keeps running.

viktorerlingsson:
The problem is probably that the lease TTL expires during startup if the startup takes more than 5s, leading to there being no lease when LavinMQ has actually started.

viktorerlingsson:
Fixed by #834

PR #834, opened 2024-11-08
make sure etcd lease does not expire during slow startup

viktorerlingsson:
Adds Fiber.yield to `load_deleted_from_disk`, `load_segments_from_disk`, `load_stats_from_segments`, and `delete_unused_segments`. 

When LavinMQ starts as a leader it creates an etcd lease with a TTL of 5s. If LavinMQ takes more than 5s to startup that lease can expire before LavinMQ starts because lease_keepalive can not run during startup because the fiber reading data from disk does not yield. This fixes that by adding yield to all functions reading data from disk during startup, allowing the lease_keepalive function to refresh the etcd lease. 

Solves #829 

Create a lot data, start LavinMQ with clustering enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- etcd lease will no longer expire during slow startup [#834](https://github.com/cloudamqp/lavinmq/pull/834)
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


viktorerlingsson:
This does make startup slightly slower. Maybe we don't have to yield for every file all the time?

spuun:
> This does make startup slightly slower. Maybe we don't have to yield for every file all the time?

Most likely not. Dunno if we "better safe than sorry", or if we should add some logic to it. With MT we may remove the yields again.

PR #835, opened 2024-11-08
wait for follower to ack all actions before gracefully shutting down

### LINKED ISSUE #827: Leader should fully sync to all followers on graceful shutdown
When a leader is shut down (either by systemctl stop or ctrl-c) while there is high throughput of data, the leader does not seem to wait for all data to be synced to all followers before shutting down. This leads to risk of data loss when a follower is promoted to leader. 

The leader should close client connections and then wait until all data has been synced to all followers before shutting down. There should maybe be a timeout of some time, after which it forcefully shuts down even if it hasn't been able to fully sync.

viktorerlingsson:
Before this change a leader would just wait for all actions to be sent before gracefully shutting down, meaning there's a risk that a few actions could be waiting to be processed by a follower. This changes that so a leader will now wait for followers to ack all actions before gracefully shutting down. 

Fixes #827 

Run spec

### DIFF
--- a/src/lavinmq/clustering/follower.cr
+++ b/src/lavinmq/clustering/follower.cr
@@ -75,7 +75,7 @@ module LavinMQ
       private def read_ack(socket = @socket) : Int64
         len = socket.read_bytes(Int64, IO::ByteFormat::LittleEndian)
         @acked_bytes += len
-        if @closed && lag_in_bytes == 0
+        if @closed && lag_in_bytes.zero?
           @closed_change.close
         end
         len
--- a/src/lavinmq/clustering/follower.cr
+++ b/src/lavinmq/clustering/follower.cr
@@ -11,6 +11,7 @@ module LavinMQ
       @acked_bytes = 0_i64
       @sent_bytes = 0_i64
       @actions = Channel(Action).new(Config.instance.clustering_max_unsynced_actions)
+      @closed = false
       getter id = -1
       getter remote_address
 
@@ -63,7 +64,6 @@ module LavinMQ
         rescue IO::Error
           # ignore connection errors while closing
         end
-        @closed.close
       end
 
       private def sync(bytes, socket = @socket) : Nil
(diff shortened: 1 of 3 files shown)


kickster97 (approved):

src/lavinmq/clustering/follower.cr
```diff
@@ -75,6 +75,9 @@ module LavinMQ
       private def read_ack(socket = @socket) : Int64
         len = socket.read_bytes(Int64, IO::ByteFormat::LittleEndian)
         @acked_bytes += len
+        if @closed && lag_in_bytes == 0
```
carlhoerberg, line 78:
```suggestion
        if @closed && lag_in_bytes.zero?
```

src/lavinmq/clustering/follower.cr
```diff
@@ -164,11 +167,14 @@ module LavinMQ
...
-      @closed = Channel(Nil).new
+      @closed_change = Channel(Nil).new
 
       def close
+        @closed = true
         @actions.close
-        @closed.receive?
+        if lag_in_bytes > 0
+          @closed_change.receive?
+        end
```
carlhoerberg, line 177:
```suggestion
        @closed_change.receive? if lag_in_bytes > 0
```

src/lavinmq/clustering/follower.cr
```diff
@@ -164,11 +167,14 @@ module LavinMQ
         lag_size
       end
 
-      @closed = Channel(Nil).new
+      @closed_change = Channel(Nil).new
```
carlhoerberg, line 170: use a better name? `in_sync`, `closed_and_in_sync`, something else? 

```suggestion
      @closed_and_in_sync = Channel(Nil).new
```

viktorerlingsson: Yeah, I think `closed_and_in_sync` conveys what it does pretty well 👍

carlhoerberg (approved):

PR #837, opened 2024-11-09
Make detail tuples typed-ish

spuun:
"detail tuples" are `NamedTuple`s which causes a lot of limitations and struggle when refactoring.

One approach to mitigate this would have been to declare a lot of structs for every case we need, but that would require a lot of work and make the code less flexible.

This PR will introduce `Metadata` which is a wrapper around `NamedTuple`. This makes it possible to add helper methods on the details instead of doing a lot of "workarounds" where the data is accessed. Looking at the diff for `src/lavinmq/http/controller.cr` one will see that the code is simplified.

Values is wrapped in `Metadata::Value` which is for `Metadata` what `JSON::Any` is to `JSON`.

One thing that maybe needs to be thought through and fixed is `Value#<=>`. It will probably almost always compare to values of the same type and thereby compare correct, but other edge cases maybe needs to be changed.

Run specs.

### DIFF
--- a/src/lavinmq/data_dir_lock.cr
+++ b/src/lavinmq/data_dir_lock.cr
@@ -14,6 +14,7 @@ module LavinMQ
     def acquire
       begin
         @lock.flock_exclusive(blocking: false)
+        Log.debug { "Data directory lock acquired" }
       rescue
         Log.info { "Data directory locked by '#{@lock.gets_to_end}'" }
         Log.info { "Waiting for file lock to be released" }
--- a/src/lavinmq/data_dir_lock.cr
+++ b/src/lavinmq/data_dir_lock.cr
@@ -14,7 +14,6 @@ module LavinMQ
     def acquire
       begin
         @lock.flock_exclusive(blocking: false)
-        Log.debug { "Data directory lock acquired" }
       rescue
         Log.info { "Data directory locked by '#{@lock.gets_to_end}'" }
         Log.info { "Waiting for file lock to be released" }
--- a/src/lavinmq/metadata.cr
+++ b/src/lavinmq/metadata.cr
@@ -45,7 +45,7 @@ module LavinMQ
           end
         {% end %}
 
-        return 0
+        0
       end
 
       delegate to_json, to_s, to: @value
--- a/src/lavinmq/metadata.cr
+++ b/src/lavinmq/metadata.cr
@@ -24,13 +24,13 @@ module LavinMQ
         @value
       end
 
-      def <=>(other : Value(U)) forall U
+      def <=>(other : Value)
         return 0 if @value.nil? && other.@value.nil?
         return -1 if @value.nil?
         return 1 if other.@value.nil?
 
-        {% if T <= Number && U <= Number %}
-          if other_value = other.@value
+        {% if T <= Number %}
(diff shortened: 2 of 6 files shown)


kickster97 (approved):

src/lavinmq/metadata.cr
```diff
@@ -0,0 +1,113 @@
...
+      end
+
+      def <=>(other : Value)
+        return 0 if @value.nil? && other.@value.nil?
+        return -1 if @value.nil?
+        return 1 if other.@value.nil?
+
+        if self.type != other.type
+          return @value.to_s <=> other.@value.to_s
+        end
```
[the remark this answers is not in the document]
spuun: Yeah i was thinking that we maybe needed more cases here. But dunno if we would ever return different tuples. I think it isn't possible. We're always having `Client#metadata` or `Channel#metadata` etc.

spuun:
> can you show how this will simplify something in a controller?

It simplifies `Controller#page` method.

carlhoerberg:
> > can you show how this will simplify something in a controller?
> 
> It simplifies `Controller#page` method.

ok, i see it now. it's less code in `Controller#page` but does it have other benefits, like more correct, doesn't have to load all data to sort etc?

carlhoerberg:
It's not like `Metadata::Value#fetch` is easily digestible, so in a sense one can say that the complexity has just been pushed around and we've added more indirection. That is if there are no other benefits that I don't see.

spuun:
> ok, i see it now. it's less code in `Controller#page` but does it have other benefits, like more correct, doesn't have to load all data to sort etc?

Not yet, but it enables other refactoring to improve it.

One purpose of this is to simplify the abstraction of different entities. I've struggled with `detail_tuples` and all of the `Iterators` when trying to abstract first `Client` and now `Channel`. I'm just trying to decouple in smaller steps.

> It's not like `Metadata::Value#fetch` is easily digestible, so in a sense one can say that the complexity has just been pushed around and we've added more indirection. That is if there are no other benefits that I don't see.

The "tree traversing" is a little bit of a mess because of macro language limitations, but the generated code should be straight forward. 

Moving the logic also makes it easier to test it.

src/lavinmq/data_dir_lock.cr
```diff
@@ -14,6 +14,7 @@ module LavinMQ
     def acquire
       begin
         @lock.flock_exclusive(blocking: false)
+        Log.debug { "Data directory lock acquired" }
```
carlhoerberg, line 17: not really related to this PR?

src/lavinmq/metadata.cr
```diff
@@ -27,27 +29,37 @@ module LavinMQ
         @value
       end
 
-      def <=>(other : Value)
+      def <=>(other : Value(T2)) forall T2
```
carlhoerberg, line 32: nitpicking, but isn't `U` normally used for the second type? 
```suggestion
      def <=>(other : Value(U)) forall U
```

src/lavinmq/data_dir_lock.cr
spuun: 🤦 
I'm doing to much in parallel.

spuun:
We should go with structs instead one day.

PR #839, opened 2024-11-12
Fix Prometheus node/cluster labels

jage:

* Add lavinmq_node label, hostname to differentiate nodes in cluster
* Fix lavinmq_cluster to be a common cluster name for all nodes, this will allow Prometheus grouping by cluster

Cluster name must be set on all nodes, to the same value, via the `clustering_name` variable.

Start LavinMQ with `--clustering-name=my-cluster-name`, scrape `metrics`.

```
$ curl -s http://guest:guest@localhost:15672/metrics | grep cluster
lavinmq_identity_info{lavinmq_version="2.0.0-9-g558bb98f", lavinmq_node="8E84C28F.local", lavinmq_cluster="my-cluster-name"} 1
```

Example from Prometheus TSDB:

Screenshot 2024-11-12 at 10 26 37

### DIFF
--- a/src/lavinmq/config.cr
+++ b/src/lavinmq/config.cr
@@ -123,7 +123,7 @@ module LavinMQ
         p.on("--clustering", "Enable clustering") do
           @clustering = true
         end
-        p.on("--clustering-name=NAME", "Cluster name (default: \"\")") do |v|
+        p.on("--clustering-name=NAME", "Cluster name, currently only used to group Prometheus metrics (default: \"\")") do |v|
           @clustering_name = v
         end
         p.on("--clustering-advertised-uri=URI", "Advertised URI for the clustering server") do |v|


kickster97 (approved):
looks good :) 
Minor: is `cluster_name` better than `clustering_name`?

kickster97:
> > looks good :) Minor: is `cluster_name` better than `clustering_name`?
> 
> 
> 
> Definitely more aesthetically pleasing but ... 🙂
> 
> ... all other options are grouped under `clustering`, e.g. in the configuration file it's a section with `[clustering]` so now it's just `name = ...`. 

Ah.. makes sense, oh well!

src/lavinmq/config.cr
```diff
@@ -122,6 +123,9 @@ module LavinMQ
         p.on("--clustering", "Enable clustering") do
           @clustering = true
         end
+        p.on("--clustering-name=NAME", "Cluster name (default: \"\")") do |v|
```
carlhoerberg, line 126: Add some more documentation around it? 

```suggestion
        p.on("--clustering-name=NAME", "Cluster name, currently only used to group Prometheus metrics (default: \"\")") do |v|
```

carlhoerberg:
How can you group on it unless you add the label to all metrics? Doing some kind of transformation somewhere? If that's case, isn't it possible to deduct the (CloudAMQP) cluster name by just dropping the last 3 chars from the node name?

carlhoerberg:
> Relevant questions here:
> 
>  * Is the intent to have a `lavinmq_cluster` label? I assumed yes since it was there., and I think LavinMQ should have have this label to avoid relabelling during scrape.
> 
>  * Do you have any intent on ever having an entitiy "cluster name" for anything else then metrics? If not this could be implemented in other ways (`--extra-metrics-labels` or something like that)

Yeah, nothing comes to mind now how we could use this for anything else, hence my hesitation. The label makes sense, but `--extra-metrics-label lavinmq_cluster=red-panda --extra-metrics-label foo=bar` sounded like a more generic idea!

jage:
Stale PR, closing.

PR #842, opened 2024-11-13
Dont call stop if already stopping

spuun:
When stopping the leader it will release its leadership lease, which in turn triggers "lost leadership" which calls stop again. This PR adds a state variable to prevent double stop.

Start lavinmq in clustering mode, then stop it. No more double logging of `lmq.launcher Stopping`.

### DIFF
--- a/src/lavinmq/launcher.cr
+++ b/src/lavinmq/launcher.cr
@@ -13,8 +13,8 @@ module LavinMQ
     @tls_context : OpenSSL::SSL::Context::Server?
     @first_shutdown_attempt = true
     @data_dir_lock : DataDirLock?
-    @closed = false
     @lease : Channel(Nil)?
+    @running = false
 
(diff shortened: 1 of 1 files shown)


viktorerlingsson (approved):

src/lavinmq/launcher.cr
```diff
@@ -56,10 +57,11 @@ module LavinMQ
           GC.collect
         end
       end
-      stop
+      stop if @running
```
carlhoerberg, line 60: I think it's nicer if the metod `stop` returns early if it's not running

carlhoerberg (approved):

snichme (approved):

carlhoerberg:
This solves the same problem plus exits with 1 https://github.com/cloudamqp/lavinmq/pull/846 if loosing leadership

PR #843, opened 2024-11-14
deliver packet not msg from session

snichme:
`Session` delivers `MQTT:Publish` instead of `Envelope` to consumer

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -37,8 +37,9 @@ module LavinMQ
             Channel.receive_first(@msg_store.empty_change, @consumers_empty_change)
             next
           end
-          get(false) do |env|
-            consumers.first.deliver(env.message, env.segment_position, env.redelivered)
+          consumer = consumers.first.as(MQTT::Consumer)
+          get_packet(false) do |pub_packet|
+            consumer.deliver(pub_packet)
           end
           Fiber.yield if (i &+= 1) % 32768 == 0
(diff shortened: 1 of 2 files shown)


src/lavinmq/mqtt/session.cr
```diff
@@ -131,6 +133,20 @@ module LavinMQ
...
+        msg = env.message
+        retained = msg.properties.try &.headers.try &.["x-mqtt-retain"]? == true
+        qos = msg.properties.delivery_mode || 0u8
+        MQTT::Publish.new(
+          packet_id: packet_id,
+          payload: msg.body,
+          dup: env.redelivered,
+          qos: qos,
+          retain: retained,
+          topic: msg.routing_key.tr(".", "/")
```
snichme, line 146: @spuun is this `.tr`  really correct?

spuun: I don't think you need to between topic and routing key here. I think we're storing it as topic. 
But if it's a routing key you want to convert `.` to `/`.

kickster97: yep thats a rest that i forgot to remove

snichme: removing the `tr` then. 

WDYT of the change otherwise?

kickster97 (approved):

src/lavinmq/mqtt/session.cr
kickster97: looks good :)

PR #850, opened 2024-11-16
Add WaitGroup to ensure consumer has sent all before removing itself

spuun:
I'm not sure we've run into this potential bug, but this is a solution to make sure the consumer has written all to socket before it's removing itself from a queue. The bug may trigger if it's the last consumer because then the queue will unmap its segments, but the consumer may still have references to a mapped segment that it tries to write to socket, resulting in a crash.

Uhm... yolo? Should try to add specs for this.


snichme (approved):

viktorerlingsson:
Does this impact performance?

spuun:
> Does this impact performance?

Can't say it's noticeable. I guess benchmarks are in margin of error. This fix is what's needed for the "flush fiber" I've been playing around with, and that gives a 10-15% increased throughput.

viktorerlingsson:
> Can't say it's noticeable. I guess benchmarks are in margin of error. This fix is what's needed for the "flush fiber" I've been playing around with, and that gives a 10-15% increased throughput.

Okay, nice! 👍

viktorerlingsson (approved):

PR #851, opened 2024-11-16
A deleted queue shouldn't allow publish

spuun:
Not sure how/if we can end up with this, but maybe we've seen that a message has been published to a queue that's partially deleted (message store deleted, but vhost still references the queue).

This adds a deleted guard to `#publish`

No sure how yet.


snichme (approved):
Makes sense to check if it's deleted, but I'm curious if you ran into some issue regarding this or just found this by looking at the code?

spuun:
> Makes sense to check if it's deleted, but I'm curious if you ran into some issue regarding this or just found this by looking at the code?

It looks like we did, but dunno how to reproduce it.

PR #854, opened 2024-11-19
Message deduplication

### LINKED ISSUE #833: Message deduplication
Add message deduplication support

Optional feature on exchange and/or queue level. Look at https://github.com/noxdafox/rabbitmq-message-deduplication for inspiration.

snichme:

Implementation for message deduplication on exchanges. 
It works for any exchange type and is enabled by setting the argument `x-message-deduplication = true` on any exchange. 

Features
- [X] Deduplication on Exchanges 
- [X] Deduplication cache storage in memory 
- [X] TTL for each message
- [X] Default TTL
- [x] Deduplication on queues 

Fixes #833 
 
Specs.

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -13,6 +13,7 @@ require "./state"
 require "./event"
 require "./message_store"
 require "../../unacked_message"
+require "../../deduplication"
 
 module LavinMQ::AMQP
   class Queue < LavinMQ::Queue
@@ -111,8 +112,8 @@ module LavinMQ::AMQP
 
     # Creates @[x]_count and @[x]_rate and @[y]_log
     rate_stats(
-      {"ack", "deliver", "deliver_get", "confirm", "get", "get_no_ack", "publish", "redeliver", "reject", "return_unroutable"},
-      {"message_count", "unacked_count", "dedup"})
+      {"ack", "deliver", "deliver_get", "confirm", "get", "get_no_ack", "publish", "redeliver", "reject", "return_unroutable", "dedup"},
+      {"message_count", "unacked_count"})
 
     getter name, arguments, vhost, consumers, last_get_time
     getter? auto_delete, exclusive
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -112,7 +112,7 @@ module LavinMQ::AMQP
     # Creates @[x]_count and @[x]_rate and @[y]_log
     rate_stats(
       {"ack", "deliver", "deliver_get", "confirm", "get", "get_no_ack", "publish", "redeliver", "reject", "return_unroutable"},
-      {"message_count", "unacked_count"})
+      {"message_count", "unacked_count", "dedup"})
 
     getter name, arguments, vhost, consumers, last_get_time
     getter? auto_delete, exclusive
@@ -127,6 +127,7 @@ module LavinMQ::AMQP
     @data_dir : String
     Log = LavinMQ::Log.for "queue"
     @metadata : ::Log::Metadata
+    @deduper : Deduplication::Deduper?
 
     def initialize(@vhost : VHost, @name : String,
                    @exclusive = false, @auto_delete = false,
@@ -271,6 +272,14 @@ module LavinMQ::AMQP
       @single_active_consumer_queue = parse_header("x-single-active-consumer", Bool) == true
       @consumer_timeout = parse_header("x-consumer-timeout", Int).try &.to_u64
       validate_positive("x-consumer-timeout", @consumer_timeout)
+      if parse_header("x-message-deduplication", Bool)
+        size = parse_header("x-cache-size", Int).try(&.to_u32)
+        raise LavinMQ::Error::PreconditionFailed.new("Invalid x-cache-size for message deduplication") unless size
+        ttl = parse_header("x-cache-ttl", Int).try(&.to_u32)
+        header_key = parse_header("x-deduplication-header", String)
+        cache = Deduplication::MemoryCache(AMQ::Protocol::Field).new(size)
+        @deduper = Deduplication::Deduper.new(cache, ttl, header_key)
+      end
     end
 
     private macro parse_header(header, type)
@@ -421,6 +430,13 @@ module LavinMQ::AMQP
 
     def publish(msg : Message) : Bool
       return false if @deleted || @state.closed?
+      if d = @deduper
+        if d.duplicate?(msg)
+          @dedup_count += 1
+          return false
+        end
+        d.add(msg)
+      end
       reject_on_overflow(msg)
       @msg_store_lock.synchronize do
         @msg_store.push(msg)
--- a/src/lavinmq/deduplication.cr
+++ b/src/lavinmq/deduplication.cr
@@ -1,0 +1,66 @@
+module LavinMQ
+  module Deduplication
+    abstract class Cache(T)
+      abstract def contains?(key : T) : Bool
+      abstract def insert(key : T, ttl : UInt32?)
+    end
+
+    class MemoryCache(T) < Cache(T)
+      def initialize(@size : UInt32)
+        @store = Hash(T, Int64?).new
+      end
+
+      def contains?(key : T) : Bool
+        return false unless @store.has_key?(key)
+        ttd = @store[key]
+        return true unless ttd
+        return true if ttd > RoughTime.unix_ms
+        @store.delete(key)
+        false
+      end
+
+      def insert(key : T, ttl : UInt32? = nil)
+        @store.shift if @store.size >= @size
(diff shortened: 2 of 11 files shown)


snichme:
Documentation for the website: https://gist.github.com/snichme/40a33df7ea9a02c9a7d77a5c81d1f52d

src/lavinmq/exchange/dedup_ext.cr
```diff
@@ -0,0 +1,67 @@
...
+        return true unless ttd
+        return true if ttd > RoughTime.unix_ms
+        @store.delete(key)
+        false
+      end
+
+      def insert(key : T, ttl : UInt32? = nil)
+        @store.delete(@store.first_key) if @store.size >= @size
+        ttd = RoughTime.unix_ms + ttl if ttl
+        @store[key] = ttd
```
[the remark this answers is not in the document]
kickster97: maybe not necessary, this is just me exercising  😅

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -414,6 +423,13 @@ module LavinMQ::AMQP
 
     def publish(msg : Message) : Bool
       return false if @deleted || @state.closed?
+      if d = @deduper
+        if d.duplicate?(msg)
+          @dedup_count += 1
+          return false
+        end
+        d.add(msg)
+      end
```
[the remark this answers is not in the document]
snichme: No, we need to `return` from the method. Have some additional lines is good if it helps the reader IMO.

src/lavinmq/exchange/dedup_ext.cr
```diff
@@ -0,0 +1,67 @@
...
+
+      private def dedup_ttl(msg) : UInt32?
+        headers = msg.properties.headers
+        def_ttl = @default_ttl
+        return def_ttl unless headers
+        value = headers["x-cache-ttl"]?
+        return def_ttl unless value
+        value = value.try(&.as?(Int32))
+        return def_ttl unless value
+        value.to_u32 || def_ttl
```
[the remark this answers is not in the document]
snichme: I do like to keep it as one-lines if possible. But not sure if it's good here since having multple lines makes it easier to read and since it's isolated in it's own method it doesn't conflict with the rest of the code.

src/lavinmq/amqp/queue/queue.cr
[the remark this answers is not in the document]
kickster97: don't understand.. my version also returns? 
yeah i guess.. i'm just allergic to nested if statements 😅

src/lavinmq/exchange/dedup_ext.cr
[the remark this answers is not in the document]
kickster97: Yeah, since its isolated maybe its better to have a few extra lines.. and my suggestion has way to many try blocks xd  mostly me just doing exercise 
But.. i do feel like its easier to read if we can avoid changing the value of `value` so many times

src/lavinmq/amqp/queue/queue.cr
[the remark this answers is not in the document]
snichme: No it doesn't :) the statement will be either false of Nil 
so if you do `my_value = d.duplicate?(msg) ? (@dedup_count += 1; false) : d.add(msg)` it will be `false` or `nil` but in order to return from the method you need `return false` after the `@dedup_count` 

we could extract it to a separate method like so: 

```
    # returns true if message is duplicated
    def handle_duplicated(msg : Message) : Bool
      return false unless d = @deduper
      if d.duplicate?(msg)
        @dedup_count += 1
        return true
      end
      d.add(msg)
      false
    end

      return false if @deleted || @state.closed?
      return false if duplicated_message(msg) 
     ....
```

but the method does so much different stuff, checking dups, adding to cache and increment counter so it's also not the best

kickster97 (approved):

src/lavinmq/exchange/dedup_ext.cr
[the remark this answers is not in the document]
viktorerlingsson: > Thats why I called it _ext for extension. 

Okay 👍 

> There are mixed files in different places (like in queue/ there are are queues, stores and events) so that is why I did like this. 

Hmm, yeah, you're right. Maybe we should give our overall file structure some more thought! 🙂 

>It's also used for both queues and exchanges so have it somewhere else makes sense. directly under /src/lavinmq maybe?

Yeah, I think directly under `/src/lavinmq` makes more sense! And if we add more "extensions" in the future we could move those to a separate extensions dir?

src/lavinmq/deduplication.cr
```diff
@@ -0,0 +1,66 @@
...
+  module Deduplication
+    abstract class Cache(T)
+      abstract def contains?(key : T) : Bool
+      abstract def insert(key : T, ttl : UInt32?)
+    end
+
+    class MemoryCache(T) < Cache(T)
+      def initialize(@size : UInt32)
+        @store = Hash(T, Int64?).new
+      end
```
carlhoerberg, line 11: isn't the purpose of the TTL is to keep the size of the cache reasonable small? So I think you should have a fiber that checks the TTL of each key and evict the ones that has expired. Not check the TTL in the `contains?` method. 

monotonic is a better alternative when you only are interested in the timespan, not the exact point of time. 

```suggestion
    class MemoryCache(T) < Cache(T)
      def initialize(@size : UInt32)
        @store = Hash(T, Int64?).new
        spawn(name: "memorycache expiry") do
          loop do
            sleep 60.seconds
            @store.reject! { |_, ttd| tdd.nil? || RoughTime.monotonic > ttd }
          end
        end
```

src/lavinmq/deduplication.cr
```diff
@@ -0,0 +1,66 @@
...
+
+      def contains?(key : T) : Bool
+        return false unless @store.has_key?(key)
+        ttd = @store[key]
+        return true unless ttd
+        return true if ttd > RoughTime.unix_ms
+        @store.delete(key)
+        false
```
carlhoerberg, line 19:
```suggestion
        @store.has_key?(key)
```

src/lavinmq/deduplication.cr
```diff
@@ -0,0 +1,66 @@
...
+      end
+
+      def insert(key : T, ttl : UInt32? = nil)
+        @store.shift if @store.size >= @size
+        @store[key] = ttl ? RoughTime.unix_ms + ttl : nil
```
carlhoerberg, line 24:
```suggestion
        @store[key] = ttl ? RoughTime.monotonic + ttl : nil
```

src/lavinmq/deduplication.cr
carlhoerberg: thinking more about it, maybe i'm wrong about the purpose of the TTL here.

src/lavinmq/deduplication.cr
```diff
@@ -0,0 +1,66 @@
+module LavinMQ
+  module Deduplication
+    abstract class Cache(T)
+      abstract def contains?(key : T) : Bool
+      abstract def insert(key : T, ttl : UInt32?)
+    end
+
+    class MemoryCache(T) < Cache(T)
+      def initialize(@size : UInt32)
+        @store = Hash(T, Int64?).new
```
carlhoerberg, line 10:
```suggestion
        @store = Hash(T, Int64?).new(initial_size: @size)
```

spuun:
I think the cache needs to be thread-safe?

snichme:
> I think the cache needs to be thread-safe?

@spuun For it to work properly now or to support MT later on?

src/lavinmq/deduplication.cr
```diff
@@ -0,0 +1,71 @@
...
+        return true unless ttd
+        return true if ttd > RoughTime.monotonic
+        @store.delete(key)
+        false
+      end
+
+      def insert(key : T, ttl : UInt32? = nil)
+        @store.shift if @store.size >= @size
+        val = if ttl
+                RoughTime.monotonic + Time::Span.new(nanoseconds: ttl * 1_000_000)
```
carlhoerberg, line 25:
```suggestion
                RoughTime.monotonic + ttl.seconds.total_nanoseconds
```

snichme: Did this `val = ttl.try { |v| RoughTime.monotonic + v.milliseconds }`  instead

src/lavinmq/deduplication.cr
snichme: TTL is primarily used, as of now, to only dedup a message for a certain period. 
Lets say you want to ignore copies of the same message for 1 second, that is what TTL is used for. 

I think we can add a fiber that checks all outdated entries and removes them but the cache is limited in size since we shift out older message if the cache is full when new entries arrive. 
Even though a fiber is cheap their is still more things to run for each queue and exchange that enables this.

snichme: Suggestion: Go without the fiber for now and add that later on if we see the need.

carlhoerberg (approved):

PR #861, opened 2024-11-26
Oauth2 support

kickster97:
ConnectionFactory uses a chain of responsibility with authentication handlers to work through different authentication backends. 

The idéa is that `config` will state the authentication backends one wishes to use. 
`ConnectionFactory` will then initialize authentication handlers for those backends. 
Those handlers will then chain the responsibility of authenticating and returning a user, or nil if no authentication method was successful 

here is an example where we configure two authentication methods, basic auth and Oauth2

authhanldersexample drawio

run specs

### DIFF
--- a/spec/auth_sepc.cr
+++ b/spec/auth_sepc.cr
@@ -1,0 +1,6 @@
+require "./spec_helper"
+
+describe LavinMQ::AuthHandler do
+
+
+end
--- a/spec/cache_spec.cr
+++ b/spec/cache_spec.cr
@@ -1,36 +1,36 @@
-require "./spec_helper"
+# require "./spec_helper"
 
-describe LavinMQ::Cache do
-  cache = LavinMQ::Cache(String, String).new(1.seconds)
+# describe LavinMQ::Cache do
+#   cache = LavinMQ::Cache(String, String).new(1.seconds)
 
-  it "set key" do
-    cache.set("key1", "allow").should eq "allow"
-  end
+#   it "set key" do
+#     cache.set("key1", "allow").should eq "allow"
+#   end
 
-  it "get key" do
-    cache.set("keyget", "deny")
-    cache.get?("keyget").should eq "deny"
-  end
+#   it "get key" do
+#     cache.set("keyget", "deny")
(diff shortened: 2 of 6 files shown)


JadeKharats:
Looking at how RabbitMQ works and with Carl's answer, I understood that you could configure a list of authentication/authorization services and the calls were made in sequence.

I had written this
`
module LavinMQ
  class AuthenticationChain
    @first_service : AuthenticationService?

    def initialize
      @first_service = nil
    end

    def add_service(service : AuthenticationService)
      if first = @first_service
        current = first
        while next_service = current.next_service
          current = next_service
        end
        current.then(service)
      else
        @first_service = service
      end
      self
    end

    def authorize?(username : String, password : String)
      if service = @first_service
        service.authorize?(username, password)
      end
    end
  end
end
`

with AuthenticationService which looks like what you did with the Handler
`
module LavinMQ
  abstract class AuthenticationService
    property next_service : AuthenticationService?

    abstract def authorize?(username : String, password : String)

    def then(service : AuthenticationService) : AuthenticationService
      @next_service = service
      service
    end

    protected def try_next(username : String, password : String)
      if next_service = @next_service
        next_service.authorize?(username, password)
      end
    end
  end
end
`

JadeKharats:
On the config side, I also followed Carl's advice by adding an auth section to have the order of the auth services and their config

`
    private def parse_auth(settings)
      settings.each do |config, v|
        case config
        when /^auth_backends\.(\d+)$/  then @auth_backends[$1.to_i] = v
        when "auth_http.http_method"   then @auth_http_method = v
        when "auth_http.user_path"     then @auth_http_user_path = v
        when "auth_http.vhost_path"    then @auth_http_vhost_path = v
        when "auth_http.resource_path" then @auth_http_resource_path = v
        when "auth_http.topic_path"    then @auth_http_topic_path = v
        else
          STDERR.puts "WARNING: Unrecognized configuration 'auth/#{config}'"
        end
      end
    end
`

JadeKharats:
In my opinion, the current user store should be the default if no config.
In the Rabbit configuration, the built-in user store is called 'internal', for compatibility reasons, I think we should keep this name.

PR #863, opened 2024-11-27
yield by each x delivered bytes instead of each y msg/frame

viktorerlingsson:
Changes `read_loop` and `deliver_loop` to Fiber.yield based on delivered bytes instead of number of messages/frames. This should make it so LavinMQ does not freeze when handling large volumes of large messages.  

Performance wise it seems pretty similar to main with small messages, and much more stable with larger messages. 

this branch
```
lavinmqperf throughput
Average publish rate: 802665.3 msgs/s
Average consume rate: 798787.9 msgs/s

lavinmqperf throughput -s 100000
Average publish rate: 7346.7 msgs/s
Average consume rate: 5862.9 msgs/s

lavinmqperf throughput -s 1000000
Average publish rate: 738.6 msgs/s
Average consume rate: 738.3 msgs/s
```

main
```
lavinmqperf throughput
Average publish rate: 792352.1 msgs/s
Average consume rate: 793866.8 msgs/s

lavinmqperf throughput -s 100000
Average publish rate: 6883.7 msgs/s
Average consume rate: 131.7 msgs/s

lavinmqperf throughput -s 1000000
Average publish rate: 542.9 msgs/s
Average consume rate: 4.3 msgs/s
```

Existing specs should cover this pretty well.

### DIFF
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -105,7 +105,7 @@ module LavinMQ
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
             if (received_bytes &+= frame.bytesize) > Config.instance.yield_each_received_bytes
-              received_bytes = 0
+              received_bytes = 0_u32
               Fiber.yield
             end
             frame_size_ok?(frame) || return
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -97,15 +97,15 @@ module LavinMQ
       end
 
       private def read_loop
-        i = 0
+        delivered_bytes = 0_i32
         socket = @socket
         loop do
           AMQP::Frame.from_io(socket) do |frame|
             {% unless flag?(:release) %}
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
-            if (i += 1) == 8192
-              i = 0
+            if (delivered_bytes &+= frame.bytesize) > 131_072 # 8192 * 16
+              delivered_bytes = 0
               Fiber.yield
             end
             frame_size_ok?(frame) || return
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -104,7 +104,7 @@ module LavinMQ
             {% unless flag?(:release) %}
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
-            if (received_bytes &+= frame.bytesize) > 131_072 # 8192 * 16
+            if (received_bytes &+= frame.bytesize) > Config.instance.yield_each_received_bytes
               received_bytes = 0
               Fiber.yield
             end
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -97,15 +97,15 @@ module LavinMQ
       end
 
       private def read_loop
-        delivered_bytes = 0_i32
+        received_bytes = 0_i32
         socket = @socket
         loop do
           AMQP::Frame.from_io(socket) do |frame|
             {% unless flag?(:release) %}
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
-            if (delivered_bytes &+= frame.bytesize) > 131_072 # 8192 * 16
-              delivered_bytes = 0
+            if (received_bytes &+= frame.bytesize) > 131_072 # 8192 * 16
+              received_bytes = 0
               Fiber.yield
             end
             frame_size_ok?(frame) || return
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -97,7 +97,7 @@ module LavinMQ
       end
 
       private def read_loop
-        received_bytes = 0_i32
+        received_bytes = 0_u32
         socket = @socket
         loop do
           AMQP::Frame.from_io(socket) do |frame|
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -75,7 +75,7 @@ module LavinMQ
             deliver(env.message, env.segment_position, env.redelivered)
             delivered_bytes &+= env.message.bytesize
           end
-          if delivered_bytes > 1_048_576 # 32768 * 32
+          if delivered_bytes > Config.instance.yield_each_delivered_bytes
             delivered_bytes = 0
             Fiber.yield
           end
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -56,7 +56,7 @@ module LavinMQ
       private def deliver_loop
         wait_for_single_active_consumer
         queue = @queue
-        i = 0
+        delivered_bytes = 0_i32
         loop do
           wait_for_capacity
           loop do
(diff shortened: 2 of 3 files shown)


src/lavinmq/amqp/client.cr
```diff
@@ -97,15 +97,15 @@ module LavinMQ
...
+        delivered_bytes = 0_i32
         socket = @socket
         loop do
           AMQP::Frame.from_io(socket) do |frame|
             {% unless flag?(:release) %}
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
-            if (i += 1) == 8192
-              i = 0
+            if (delivered_bytes &+= frame.bytesize) > 131_072 # 8192 * 16
```
walro, line 107: Just curious, would Crystal's compiler do the `8192 * 16` for you? if so I guess you could leave it as that and skip it as a comment :) I assume you wanted to avoid the multiplication for each iteration

viktorerlingsson: oh, I assume the compiler would do that multiplication just once anyway I think 🤔 
But I wanted it to be as clear as possible as to after how many bytes we yield. The comment is more of a clarification of why I selected the number 131_072 (because the smaller frames seem to be around 14-15 bytes, so 16 * 8192 makes some sense)

walro: Ah roger, thanks!

viktorerlingsson: Just because I got curious 🤓 
Seems very similar performance wise! 

```
def mult_each()
  i = 0
  loop do
    break if (i &+= 1) > 8192 * 16
  end
end

def consta()
  i = 0
  loop do
    break if (i &+= 1) > 131_072
  end
end
```

```
viktor@viktor-lenovo:~/code/_crystal-test$ crystal run bench.cr --release
mult_each 911.09M (  1.10ns) (± 4.20%)  0.0B/op        fastest
   consta 908.26M (  1.10ns) (± 5.03%)  0.0B/op   1.00× slower
```

src/lavinmq/amqp/client.cr
```diff
@@ -97,15 +97,15 @@ module LavinMQ
       end
 
       private def read_loop
-        i = 0
+        delivered_bytes = 0_i32
```
spuun, line 100:
```suggestion
        received_bytes = 0_i32
```
to not confuse anyone

src/lavinmq/amqp/client.cr
spuun, line 107: Make `131_072` an advanced config option

src/lavinmq/amqp/consumer.cr
```diff
@@ -73,8 +73,12 @@ module LavinMQ
           {% end %}
           queue.consume_get(self) do |env|
             deliver(env.message, env.segment_position, env.redelivered)
+            delivered_bytes &+= env.message.bytesize
+          end
+          if delivered_bytes > 1_048_576 # 32768 * 32
```
spuun, line 78: Make `1_048_576` an advanced config option

src/lavinmq/amqp/consumer.cr
```diff
@@ -73,8 +73,12 @@ module LavinMQ
           {% end %}
           queue.consume_get(self) do |env|
             deliver(env.message, env.segment_position, env.redelivered)
+            delivered_bytes &+= env.message.bytesize
+          end
+          if delivered_bytes > Config.instance.yield_each_delivered_bytes
```
carlhoerberg, line 78: does not have a performance impact to do the double variable dereference in a tight loop?

viktorerlingsson: Doesn't seem to make any noticeable impact when benchmarking at least!

src/lavinmq/config.cr
```diff
@@ -60,6 +60,8 @@ module LavinMQ
     property consumer_timeout : UInt64? = nil
     property consumer_timeout_loop_interval = 60 # seconds
     property default_consumer_prefetch = UInt16::MAX
+    property yield_each_received_bytes = 131_072    # after how many received bytes the read_loop fiber yields
```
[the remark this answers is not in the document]
viktorerlingsson: Yeah, need to figure out how to explain it in a way that's easy to understand though... I'll give it some more thought! 
Maybe something along the lines of "a higher value generally increases throughput on a single consumer, but could cause issues if consuming slowly"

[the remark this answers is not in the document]
viktorerlingsson: Updated to: 
```
    property yield_each_received_bytes = 131_072 # number of bytes received from a publisher before yielding to other fibers
    property yield_each_delivered_bytes = 1_048_576 # number of bytes sent to a consumer before yielding to other fibers
```
is that better? 

It's hard to describe it much better in a short sentence IMO, and I don't want to add a long description here. I think it would be better if we had some more thorough documentation of "advanced" config vars in the docs!

src/lavinmq/amqp/client.cr
```diff
@@ -97,15 +97,15 @@ module LavinMQ
       end
 
       private def read_loop
-        i = 0
+        received_bytes = 0_i32
```
carlhoerberg, line 100: definitely unsigned, but maybe even u64
```suggestion
        received_bytes = 0_u32
```

carlhoerberg (approved):

src/lavinmq/amqp/client.cr
viktorerlingsson: I think u32 should be enough? would mean we can handle message sizes up to 4gb

src/lavinmq/amqp/client.cr
```diff
@@ -97,15 +97,15 @@ module LavinMQ
...
         socket = @socket
         loop do
           AMQP::Frame.from_io(socket) do |frame|
             {% unless flag?(:release) %}
               @log.trace { "Received #{frame.inspect}" }
             {% end %}
-            if (i += 1) == 8192
-              i = 0
+            if (received_bytes &+= frame.bytesize) > Config.instance.yield_each_received_bytes
+              received_bytes = 0
```
carlhoerberg, line 108: you have to change this one too, otherwise you're changing type of the variable. 
```suggestion
              received_bytes = 0u32
```

carlhoerberg:
Or skip the type change, it would only become a problem if someone sent a frame larger than 2 GiB..

src/lavinmq/config.cr
```diff
@@ -60,6 +60,8 @@ module LavinMQ
     property consumer_timeout : UInt64? = nil
     property consumer_timeout_loop_interval = 60 # seconds
     property default_consumer_prefetch = UInt16::MAX
+    property yield_each_received_bytes = 131_072 # number of bytes received from a publisher before yielding to other fibers
+    property yield_each_delivered_bytes = 1_048_576 # number of bytes sent to a consumer before yielding to other fibers
```
carlhoerberg, line 64:
```suggestion
    property yield_each_received_bytes = 131_072 # max number of bytes to read from a client connection without letting other tasks in the server do any work
    property yield_each_delivered_bytes = 1_048_576 # max number of bytes sent to a client without tending to other tasks in the server
```

carlhoerberg (commented):
Trying to write good docs for these settings makes me think that this probably shouldn't be user configurable settings at all. We only want it to be configurable to be able to easy test out new settings internally. The long term solution is either a fixed value, not something we expect users to change, or better yet, waiting for the better MT implementation in crystal that force "yield" fibers after 10ms or something.

spuun:
> > Trying to write good docs for these settings makes me think that this probably shouldn't be user configurable settings at all. We only want it to be configurable to be able to easy test out new settings internally. The long term solution is either a fixed value, not something we expect users to change, or better yet, waiting for the better MT implementation in crystal that force "yield" fibers after 10ms or something.
> 
> Yeah, agreed for the most part. In most cases these values shouldn't need to be changed. There might be a case for changing them in some extreme cases, but it can also cause some "harm" if users set the values without understanding them.
> 
> @spuun since you suggested making the values configurable, WDYT?

Yeah, hm. I think it could be nice to have them until MT to be able to tweak individual instances. "Hide" them as ENVs instead? Or just document? Put them in an `[experimental]` section?

I think it would be nice to avoid having to do releases just to tweak values such this even more.

carlhoerberg:
"experimental" section sounds good

PR #865, opened 2024-12-03
rescue if lavinmq starts with faulty msg_store

kickster97:
today if we get an error when initialising the message store in `load_segments_from_disk` and `load_stats_from_segments` we will raise an exception and LavinMQ will not be able to start. With these changes we will instead just close the queue and let LavinMQ start up with closed queues. 

try to start up with a faulty msg_store.

### DIFF
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -22,6 +22,7 @@ module LavinMQ
       @segment_msg_count = Hash(UInt32, UInt32).new(0u32)
       @requeued = Deque(SegmentPosition).new
       @closed = false
+      getter closed
       getter bytesize = 0u64
       getter size = 0u32
       getter empty_change = Channel(Bool).new
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -348,7 +348,7 @@ module LavinMQ
                 next
               end
             rescue ex
-              @log.error { "Closing message store: invalid SchemaVersion in #{path}" }
+              @log.error { "Could not initialize segment, closing message store: #{ex.message}" }
               close
             end
           end
@@ -380,7 +380,7 @@ module LavinMQ
           rescue ex : IO::EOFError
             break
           rescue ex : OverflowError | AMQ::Protocol::Error::FrameDecode
-            @log.error { "Closing message store: Failed to read segment #{seg} at pos #{mfile.pos}, #{ex}" }
+            @log.error { "Could not initialize segment, closing message store: Failed to read segment #{seg} at pos #{mfile.pos}. #{ex}" }
             close
           end
           mfile.pos = 4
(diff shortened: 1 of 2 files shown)


src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -347,6 +347,9 @@ module LavinMQ
                 @segments.delete seg
                 next
               end
+            rescue ex
+              Log.error { "Closing message store: invalid SchemaVersion in #{path}" }
```
viktorerlingsson, line 351: I think the log message could be even more descriptive, something like "Could not initialize segment, closing message store: Invalid SchemaVersion in #{path}"

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -377,7 +380,8 @@ module LavinMQ
           rescue ex : IO::EOFError
             break
           rescue ex : OverflowError | AMQ::Protocol::Error::FrameDecode
-            raise Error.new(mfile, cause: ex)
+            Log.error { "Closing message store: Failed to read segment #{seg} at pos #{mfile.pos}, #{ex}" }
```
viktorerlingsson, line 383: I think the log message could be even more descriptive, something like "Could not initialize segment, closing message store: Failed to read segment #{seg} at pos #{mfile.pos}. #{ex}"

viktorerlingsson (approved):
Added a couple of comments about log message, otherwise it looks good to me!

src/lavinmq/amqp/queue/message_store.cr
spuun, line 351:
```suggestion
              @log.error { "Closing message store: invalid SchemaVersion in #{path}" }
```

src/lavinmq/amqp/queue/message_store.cr
spuun, line 383:
```suggestion
            @log.error { "Closing message store: Failed to read segment #{seg} at pos #{mfile.pos}, #{ex}" }
```

spuun (approved):

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -347,6 +347,9 @@ module LavinMQ
                 @segments.delete seg
                 next
               end
+            rescue ex
+              @log.error { "Closing message store: invalid SchemaVersion in #{path}" }
```
carlhoerberg, line 351: how do we know that it is a "invalid SchemaVersion" error if we catch all errors? aren't we raising a specific exception then? Catch only that then, or log `ex.message`.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -802,6 +802,7 @@ module LavinMQ::AMQP
         expire_msg(sp, :rejected)
       end
     rescue ex : MessageStore::Error
+      @log.error(ex) { "Queue closed due to error" }
       close
       raise ex
```
carlhoerberg, line 807: where will this be reraised to? Should we log there instead?

kickster97: I put it here so it would be more centralized, so we dont have to change the log string in multiple places if we make changes.. but maybe we want to have different logs at different places where we do `reject`?

PR #867, opened 2024-12-04
Bugfix: use next instead of return to continue listen for events

spuun:
When restarting a leader it could end up in a idling state, not turning into a follower. This because if it starts up fast enough it hasn't lost its leadership according to etcd, and the first election event will return the node as leader, and it can't be follower to itself.

Chaning the `return` to `next` will make it will wait for the next event that should be another node being promoted.

Failed to create specs for this so testing manually for now.

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -45,12 +45,14 @@ class LavinMQ::Clustering::Controller
   # Listens for leader change events
   private def follow_leader
     repli_client = nil
+    Log.info { "Listening for leader changes..." }
     @etcd.elect_listen("#{@config.clustering_etcd_prefix}/leader") do |uri|
+      Log.debug { "Leader event: #{uri}" }
       next if repli_client.try &.follows?(uri) # if lost connection to etcd we continue follow the leader as is
       repli_client.try &.close
       if uri == @advertised_uri # if this instance has become leader
         Log.debug { "Is leader, don't replicate from self" }
-        return
+        next
       end
       Log.info { "Leader: #{uri}" }
       key = "#{@config.clustering_etcd_prefix}/clustering_secret"


carlhoerberg (approved):

carlhoerberg:
But wait, I think the thinking was that if this node is the leader it should crash/close when it looses its leadership? Are we not doing that? A leader restarts, resumes its leadership, it should need to listen for other leadership changes because when it looses its leadership it should exit the process?

spuun:
> But wait, I think the thinking was that if this node is the leader it should crash/close when it looses its leadership? Are we not doing that? A leader restarts, resumes its leadership, it should need to listen for other leadership changes because when it looses its leadership it should exit the process?

Yeah, well, it's stopping for some reason.  I think I may have found another bug now which is the real issue!

spuun:
Hm, no. Because it returns from `follow_leader` it get stuck on leader election, and it's the follower (I'm testing with two nodes) that's elected. And since the old leader has returned from `follow_leader` it will never follow.

And because it's stuck in leader election without following, it means message loss if we publish to the new leader and it dies. No data has been synced and the "idling" non-follower will be leader but with old data.

carlhoerberg:
> And because it's stuck in leader election without following, it means message loss if we publish to the new leader and it dies. No data has been synced and the "idling" non-follower will be leader but with old data.

how can the idling non-followr become leader? it should check the ISR before trying to become leader

spuun:
> > And because it's stuck in leader election without following, it means message loss if we publish to the new leader and it dies. No data has been synced and the "idling" non-follower will be leader but with old data.
> 
> how can the idling non-followr become leader? it should check the ISR before trying to become leader

It's still in the ISR set, so it will just continue to leader election.

spuun:
I think this is the flow:
1. Node 1 (leader) stops gracefully, but doesn't revoke leadership (fixed by #869)
1. Node 1 instantly starts again, before a follower has been elected leader (because of lease ttl)
1. Node 1 notice that it can't follow itself and returns from `follow_leader`
1. Node 1 is still in ISR and can continue to leader election
1. Lease TTL happens and node 2 is elected leader
1. Node 1 won't follow becaues it returned from `follow_leader`
1. Publish data to node 2 which won't be replicated to node 1
1. Kill node 2
1. New lease TTL, node 1 is elected leader
1. Data loss because node 1 never synced

spuun:
Maybe we should add a check to verify the node is in ISR when it's elected, else revoke lease and start over?

spuun:
I'll wait for #871 to be merged and test together with that.

PR #869, opened 2024-12-06
Bugfix: revoke leadership on graceful shutdown

spuun:
On a graceful shutdown lavin exits before the etcd keepalive fiber has had a chance to revoke the leadership. This will yield after the lease has been closed to give the keepalive fiber a chance to revoke the lease.

No specs, yet. Needs to be tested manually.

### DIFF
--- a/src/lavinmq/launcher.cr
+++ b/src/lavinmq/launcher.cr
@@ -45,7 +45,10 @@ module LavinMQ
         if lease = @lease
           select
           when lease.receive?
-            break unless @running
+            unless @running
+              Fiber.yield
+              break
+            end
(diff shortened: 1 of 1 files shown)


spuun:
This is a quickfix. I don't like to rely on just a yield, it would have been nicer to use real synchronization with a channel or a waitgroup.

carlhoerberg:
`Etcd#elect` returns a `Channel` that got two purposes
1) you can wait for it to close which means we lost leadership
2) if you close it we will eventually revoke the leadership. 

If we instead return a `Leadership` class that got two methods
1) `wait` which waits until we loose the leadership
2) `revoke` that synchronous revokes the leadership

PR #871, opened 2024-12-09
Sync release leadership on stop

carlhoerberg:
Synchronous release the release leadership when shutting down gracefully.

Specs

Fixes #869


spuun:
> Instead of `wait` loops self maybe it should return `false` when timeout and `true` when lost?

Yes, i think that would be more clear.

spuun (approved):

PR #874, opened 2024-12-11
Dont wait on a disconnected follower

spuun:
While testing clustering stuff I noticed that the leader sometimes wait for a disconnected follower to be fully in sync. This PR prevents at least some of these cases.

Manually run a cluster and stop follower in different stage. No "Waiting for follower to be in sync" should be logged.

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -160,12 +160,16 @@ module LavinMQ
         end
       rescue ex : AuthenticationError
         Log.warn { "Follower negotiation error" }
+        socket.close
       rescue ex : InvalidStartHeaderError
         Log.warn { ex.message }
+        socket.close
       rescue ex : IO::EOFError
         Log.info { "Follower disconnected" }
+        socket.close
       rescue ex : IO::Error
         Log.warn { "Follower disonnected: #{ex.message}" }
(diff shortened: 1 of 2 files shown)


src/lavinmq/clustering/server.cr
```diff
@@ -160,12 +160,16 @@ module LavinMQ
...
         Log.warn { ex.message }
+        socket.close
       rescue ex : IO::EOFError
         Log.info { "Follower disconnected" }
+        socket.close
       rescue ex : IO::Error
         Log.warn { "Follower disonnected: #{ex.message}" }
+        socket.close
       ensure
         follower.try &.close
```
carlhoerberg, line 174: shouldn't `Follower#close` close the socket?

spuun: It will be done by the action loop once it's drained. But we should probably close the socket there is no lag or if we get a timeout waiting for the follower to sync. I guess we could add an ensure statement and always close the socket in `#close`.

carlhoerberg (approved):

spuun:
I'm not sure this is needed anymore after other things has been refactored. I reproduce what this is fixing...

spuun:
This is not an issue anymore.

PR #876, opened 2024-12-11
use a waitgroup to wait for reserve holders of MFiles before unmapping

carlhoerberg:
Please fill me in.

### DIFF
--- a/src/lavinmq/clustering/actions.cr
+++ b/src/lavinmq/clustering/actions.cr
@@ -14,6 +14,7 @@ module LavinMQ
 
       abstract def lag_size : Int64
       abstract def send(socket : IO, log = Log) : Int64
+      abstract def done
 
       getter filename
 
@@ -25,6 +26,7 @@ module LavinMQ
 
     struct AddAction < Action
       def initialize(@data_dir : String, @filename : String, @mfile : MFile? = nil)
+        @mfile.try &.reserve
       end
 
       def lag_size : Int64
@@ -53,11 +55,22 @@ module LavinMQ
             size
           end
         end
+      ensure
+        done
+      end
+
+      def done
+        if mfile = @mfile
+          mfile.unreserve
+        end
       end
     end
 
     struct AppendAction < Action
       def initialize(@data_dir : String, @filename : String, @obj : Bytes | FileRange | UInt32 | Int32)
+        if range = @obj.as?(FileRange)
+          range.mfile.reserve
+        end
       end
 
       def lag_size : Int64
@@ -92,6 +105,14 @@ module LavinMQ
         end
         log.debug { "Append #{len} bytes to #{@filename}" }
         len
+      ensure
+        done
+      end
+
+      def done
+        if fr = @obj.as?(FileRange)
+          fr.mfile.unreserve
+        end
       end
     end
 
@@ -108,6 +129,9 @@ module LavinMQ
         send_filename(socket)
         socket.write_bytes 0i64
         0i64
+      end
+
+      def done
       end
     end
   end
--- a/src/lavinmq/mfile.cr
+++ b/src/lavinmq/mfile.cr
@@ -118,15 +118,7 @@ class MFile < IO
       raise File::Error.from_errno("Error truncating file", file: @path) if code < 0
     end
 
-    # unmap if non has reserved the file, race condition prone?
-    if @wg.@counter.get(:acquire).zero?
-      unmap
-    else
-      spawn(name: "munmap #{@path}") do
-        @wg.wait
-        unmap
-      end
-    end
+    unmap
   ensure
     unless @fd == -1
       code = LibC.close(@fd)
@@ -150,6 +142,18 @@ class MFile < IO
 
   # unload the memory mapping, will be remapped on demand
   def unmap : Nil
+    # unmap if non has reserved the file, race condition prone?
+    if @wg.@counter.get(:acquire).zero?
+      unsafe_unmap
+    else
+      spawn(name: "munmap #{@path}") do
+        @wg.wait
+        unsafe_unmap
+      end
+    end
+  end
+
+  private def unsafe_unmap : Nil
     b = @buffer
     c = @capacity
     @buffer = Pointer(UInt8).null
--- a/src/lavinmq/mfile.cr
+++ b/src/lavinmq/mfile.cr
@@ -50,9 +50,8 @@ class MFile < IO
     end
   end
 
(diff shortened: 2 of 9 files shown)


src/lavinmq/mfile.cr
```diff
@@ -100,13 +103,30 @@ class MFile < IO
...
   # The file will be truncated to the current position unless readonly or deleted
   def close(truncate_to_size = true)
-    # unmap occurs on finalize
     if truncate_to_size && !@readonly && !@deleted && @fd > 0
       code = LibC.ftruncate(@fd, @size)
       raise File::Error.from_errno("Error truncating file", file: @path) if code < 0
     end
+
+    # unmap if non has reserved the file, race condition prone?
+    if @wg.@counter.get(:acquire).zero?
```
spuun, line 122: Since we're in `#close` I guess we can be sure here that counter won't be increased, only decreased? So no risk that counter is zero in this check but non-zero when unmap is called?

What about always spawning instead of this if-statement? It would be nice to not have to access an instance variable. Or maybe monkey patch WaitGroup with a `done?` method.

src/lavinmq/clustering/actions.cr
```diff
@@ -14,6 +14,7 @@ module LavinMQ
 
       abstract def lag_size : Int64
       abstract def send(socket : IO, log = Log) : Int64
+      abstract def abort
```
spuun, line 17: What about naming it `done` and call it from the action loop when an `Action` has been sent, instead of assuming that `send` will be called only once for an `Action` and call `unreserve` from inside `send`?

src/lavinmq/mfile.cr
carlhoerberg: "risk" that it is zero? the optimization here was for that fact that it will always be zero for single node clusters and for transient queues etc, and then we don't need to spawn a fiber. 

could monkey patch though.

src/lavinmq/clustering/actions.cr
carlhoerberg: makes sense, thanks!

carlhoerberg: put `done` in `ensure` in the `Action#send` method instead

carlhoerberg: yes, the assumption is that `Action#send` is only called once

src/lavinmq/mfile.cr
spuun: I mean if it's zero when we do the check, but it has changed when we call unmap so we're unmapping even though it's reserved. However, I think we can assume that the counter will never be increased at this point since we're in `#close` so the `MFile` shouldn't be used anymore?

src/lavinmq/clustering/actions.cr
spuun: Instead of assuming that, which would be an internal implementation detail the caller may need to know about, I think it would make sense to do 

```crystal
action.send(lz4, Log)
action.done
```

in the action loop.

carlhoerberg: It gets ugly because you have to have a begin/recue block, and then you have two loops, so two blocks on top of that

spuun: True

spuun:
Two things that we probably must remove

- `unmap` from `Queue#rm_consumer`
- `unmap` from USR2 handler

carlhoerberg:
> Two things that we probably must remove
>     * `unmap` from `Queue#rm_consumer`
>     * `unmap` from USR2 handler

I don't know, both call `MessageStore#unmap_segments`, which doesn't unmap the last segment that is still being written to.

https://github.com/cloudamqp/lavinmq/blob/c8ca36a93bfefbbf6f3b9f2d8cb6c4faa39013b9/src/lavinmq/amqp/queue/message_store.cr#L231-L237

spuun:
> I don't know, both call `MessageStore#unmap_segments`, which doesn't unmap the last segment that is still being written to.

Isn't it possible that a segment that isn't yet fully replicated is unmapped? If it's a large message I think the unmap may occur in the middle of the lz4 compression.

Will these unmaps be necessary if we handle unmapping properly, which this PR hopefully solves?

carlhoerberg:
> > I don't know, both call `MessageStore#unmap_segments`, which doesn't unmap the last segment that is still being written to.
> 
> Isn't it possible that a segment that isn't yet fully replicated is unmapped? If it's a large message I think the unmap may occur in the middle of the lz4 compression.
> 
> Will these unmaps be necessary if we handle unmapping properly, which this PR hopefully solves?

True, there's a chance for that, if the unmap is triggered just when we're rolling over to a new segment. 

And yes, shouldn't be needed anymore.

carlhoerberg:
also started to tinker with a ref counter as you did, but yes, it fast gets messy, as we return a `BytesMessage` from `MessageStore`, but it points into the mmaping.. so we don't know when the client has sent them.. 

```crystal
class MFile
  @counter = 0
  @counter_lock = Mutex.new(:unchecked)

  def borrow : self
    @counter_lock.synchronize do
      counter = @counter
      if counter.zero?
        mmap
      end
      @counter = counter + 1
    end
    self
  end

  def unborrow : Nil
    @counter_lock.synchronize do
      counter = @counter -= 1
      if counter.zero?
        unmap
      end
    end
  end
end
```

carlhoerberg:
This is still a problem: https://github.com/cloudamqp/lavinmq/blob/e13e7af6a46ef75646184e4f3128ad562d562412/src/lavinmq/amqp/queue/message_store.cr/#L261

spuun:
> This is still a problem: https://github.com/cloudamqp/lavinmq/blob/e13e7af6a46ef75646184e4f3128ad562d562412/src/lavinmq/amqp/queue/message_store.cr/#L261
Do we need this unmap? Why not let a file be mapped as long as it's opened? I.e. only unmap in `#close`?

carlhoerberg:
> Do we need this unmap? Why not let a file be mapped as long as it's opened? I.e. only unmap in `#close`?

You can't mmap unlimited number of files, eg. in linux you're limited by `vm.max_map_count`. If you have a queue that you only publish to no segment would never be unmapped without this

carlhoerberg:
> This is still a problem: https://github.com/cloudamqp/lavinmq/blob/e13e7af6a46ef75646184e4f3128ad562d562412/src/lavinmq/amqp/queue/message_store.cr/#L261

This is not a problem anymore, now that `MFile#unmap` is safe by default, `MFile#unsafe_unmap` is not, it doesn't check the waitgroup

PR #880, opened 2024-12-13
Exchange abstraction

spuun:
Move existing exchanges to AMQP namespace and introduce a clean base class for all exchanges, similar to what's been done with queues. This enables better separation of amqp and mqtt code.

Run specs to see that nothing breaks.

NOTE:
rebase+merge, not squash, to not screw up git history


viktorerlingsson (approved):
Haven't looked through every line of code in the moved files, but I'm assuming there's no big changes to logic, just adding the AMQP module? 

I think this makes sense 👍

PR #886, opened 2024-12-16
exit 5 instead of raising error if no etcd endpoint responds

### LINKED ISSUE #872: Crashes on startup if etcd isn't available
If etcd isn't running lavinmq crashes with `Unhandled exception: No endpoint responded (LavinMQ::Etcd::Error)` on startup.

This need to be handled better, but exit with non-zero is fine.

viktorerlingsson:
Exit 5 (5th letter, E, for etcd) instead of raising error when no etcd endpoint responds. 

fixes #872 

Try starting LavinMQ with clustering enabled and no etcd running.

### DIFF
--- a/spec/etcd_spec.cr
+++ b/spec/etcd_spec.cr
@@ -27,7 +27,7 @@ describe LavinMQ::Etcd do
         etcd.watch("foo") do |val|
           w.send val
         end
-      rescue LavinMQ::Etcd::Error
+      rescue SpecExit
         # expect this when etcd nodes are terminated
       end
       w.receive # sync
@@ -51,7 +51,7 @@ describe LavinMQ::Etcd do
         etcd.elect_listen(key) do |value|
           leader.send value
         end
-      rescue LavinMQ::Etcd::Error
+      rescue SpecExit
         # expect this when etcd nodes are terminated
       end
       lease = etcd.elect(key, "bar", 1)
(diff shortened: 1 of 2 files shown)


carlhoerberg:
I did this for exit 3: https://github.com/cloudamqp/lavinmq/pull/887/commits/5fd0ea00678ace02539faeb4c2109293496e4c3c So it won't exit specs.. it's not pretty though :/

viktorerlingsson:
> I did this for exit 3: [5fd0ea0](https://github.com/cloudamqp/lavinmq/commit/5fd0ea00678ace02539faeb4c2109293496e4c3c) So it won't exit specs.. it's not pretty though :/

Did the same here for now 👍 

IMO let's go with that solution until we figure out something better!

carlhoerberg:
I reverted that, exiting in code, if the code is running in a fiber nothing would rescue and exit it.

viktorerlingsson:
> Check if we can override `abort` in specs

Yes we can, just something like the following in `spec_helper` works. But we should probably raise an exception that we can catch in the specs where we want LavinMQ to exit. 

```
def abort(reason = "")
end
```

spuun:
> Yes we can, just something like the following in `spec_helper` works. But we should probably raise an exception that we can catch in the specs where we want LavinMQ to exit.
> 

So we can just do:
```
def abort(reason = "")
raise reason
end
```

But maybe it's worth defining a `SpecExitError` or something.

carlhoerberg (approved):

PR #893, opened 2024-12-18
Streams filters

viktorerlingsson:
Adds support for filtering in streams. 

Filters can be set on messages by providing `x-stream-filter-value` when publishing messages to a stream. 
When a consumer tries to consume with `x-stream-filter` set, only messages matching the filter value will be returned. `x-stream-filter` on consumers can be a single value as a string, or a set of values separated by commas. Any message matching any of the filters on the consumer will be returned. 

You can also choose to return all messages without a filter by providing `x-stream-match-unfiltered` as an argument on the consumer. 
If you do not provide `x-stream-filter` when consuming, all messages in the stream will be returned as if no filters existed. 

Run specs

### DIFF
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -136,8 +136,8 @@ module LavinMQ::AMQP
       end
 
       private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
-        if msg_filter = filter_value_from_headers(msg_headers)
-          return consumer_filter.includes? msg_filter
+        if filter_value = filter_value_from_headers(msg_headers)
+          return consumer_filter.bsearch { |f| f >= filter_value } == filter_value
         else
           return match_unfiltered
         end
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -137,9 +137,9 @@ module LavinMQ::AMQP
 
       private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
         if filter_value = filter_value_from_headers(msg_headers)
-          return consumer_filter.bsearch { |f| f >= filter_value } == filter_value
+          consumer_filter.bsearch { |f| f >= filter_value } == filter_value
         else
-          return match_unfiltered
+          match_unfiltered
         end
       end
 
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -126,6 +126,18 @@ module LavinMQ::AMQP
           sp = SegmentPosition.new(consumer.segment, consumer.pos, msg.bytesize.to_u32)
           consumer.pos += sp.bytesize
           consumer.offset += 1
+          if consumer_filter = consumer.filter # can be array of strings
+            matched = false
+            if (msg_filters = filter_values_from_headers(msg.properties.headers))
+              # can msg_filters be array of strings?
+              consumer_filter.split(",").each do |filter|
+                matched = true if msg_filters == filter
+              end
+            #else
+            # matched = true if x-stream-match-unfiltered
+            end
+            return unless matched
+          end
           Envelope.new(sp, msg, redelivered: false)
         rescue ex
           raise Error.new(rfile, cause: ex)
@@ -234,6 +246,14 @@ module LavinMQ::AMQP
 
       private def offset_from_headers(headers) : Int64
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
+      end
+
+      private def filter_values_from_headers(headers) : String?
+        if filters = headers.not_nil!("Message lacks headers")["x-stream-filter-value"]?
+          filters.to_s
+        else
+          nil
+        end
       end
 
       private def build_segment_indexes
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -108,7 +108,7 @@ module LavinMQ::AMQP
         seg
       end
 
-      def shift?(consumer : AMQP::StreamConsumer) : Envelope? # ameba:disable Metrics/CyclomaticComplexity
+      def shift?(consumer : AMQP::StreamConsumer) : Envelope?
         raise ClosedError.new if @closed
 
         if env = shift_requeued(consumer.requeued)
@@ -126,20 +126,10 @@ module LavinMQ::AMQP
           sp = SegmentPosition.new(consumer.segment, consumer.pos, msg.bytesize.to_u32)
           consumer.pos += sp.bytesize
           consumer.offset += 1
-          if consumer_filter = consumer.filter
-            return unless matching?(msg.properties.headers, consumer_filter, consumer.match_unfiltered?)
-          end
+          return unless consumer.filter_match?(msg.properties.headers)
           Envelope.new(sp, msg, redelivered: false)
         rescue ex
           raise Error.new(rfile, cause: ex)
-        end
-      end
-
-      private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
-        if filter_value = filter_value_from_headers(msg_headers)
-          consumer_filter.bsearch { |f| f >= filter_value } == filter_value
-        else
-          match_unfiltered
         end
       end
 
@@ -245,14 +235,6 @@ module LavinMQ::AMQP
 
       private def offset_from_headers(headers) : Int64
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
-      end
-
-      private def filter_value_from_headers(headers) : String?
-        if filter = headers.try &.["x-stream-filter-value"]?
-          filter.to_s
-        else
-          nil
-        end
       end
 
       private def build_segment_indexes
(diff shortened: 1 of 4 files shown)


src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -37,6 +39,18 @@ module LavinMQ
         when Nil, Int, Time, "first", "next", "last"
         else raise LavinMQ::Error::PreconditionFailed.new("x-stream-offset must be an integer, a timestamp, 'first', 'next' or 'last'")
         end
+        case frame.arguments["x-stream-filter-value"]?
+        when String
+          @filter = frame.arguments["x-stream-filter-value"].to_s
```
spuun, line 44: What about
```suggestion
          @filter = frame.arguments["x-stream-filter-value"].to_s.split(",")
```
to get rid of the `split` in `StreamQueueMessageStore#matching?`? That would get rid of some allocations for every call to `#matching?`.

spuun (commented):
See inline comments.

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -236,6 +250,14 @@ module LavinMQ::AMQP
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
       end
 
+      private def filter_value_from_headers(headers) : String?
+        if filter = headers.not_nil!("Message lacks headers")["x-stream-filter-value"]?
```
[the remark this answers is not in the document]
viktorerlingsson: Yeah, looks better, thanks! 👍

src/lavinmq/amqp/stream_consumer.cr
viktorerlingsson: Yes, that's better 👍

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -126,12 +126,26 @@ module LavinMQ::AMQP
...
         rescue ex
           raise Error.new(rfile, cause: ex)
         end
       end
 
+      private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
+        if msg_filter = filter_value_from_headers(msg_headers)
+          consumer_filter.each do |filter|
+            return true if msg_filter == filter
+          end
```
spuun, line 142:
```suggestion
            return consumer_filter.includes? msg_filter
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -126,12 +126,26 @@ module LavinMQ::AMQP
           sp = SegmentPosition.new(consumer.segment, consumer.pos, msg.bytesize.to_u32)
           consumer.pos += sp.bytesize
           consumer.offset += 1
+          if consumer_filter = consumer.filter # can be a string or an array of comma-separated strings
```
spuun, line 129: Always an array now. Probably no need to explain it anymore?
```suggestion
          if consumer_filter = consumer.filter
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -126,12 +126,26 @@ module LavinMQ::AMQP
...
 
+      private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
+        if msg_filter = filter_value_from_headers(msg_headers)
+          consumer_filter.each do |filter|
+            return true if msg_filter == filter
+          end
+        else
+          return match_unfiltered
+        end
+        false
```
spuun, line 146: With the above change this is never reached
```suggestion
```

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -37,6 +39,18 @@ module LavinMQ
         when Nil, Int, Time, "first", "next", "last"
         else raise LavinMQ::Error::PreconditionFailed.new("x-stream-offset must be an integer, a timestamp, 'first', 'next' or 'last'")
         end
+        case frame.arguments["x-stream-filter-value"]?
+        when String
+          @filter = frame.arguments["x-stream-filter-value"].to_s.split(",")
+        when Nil
```
spuun, line 45: At first I managed to interpret the code to raise on nil.
```suggestion
        when Nil
          # nop
```

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -37,6 +39,18 @@ module LavinMQ
...
+        else raise LavinMQ::Error::PreconditionFailed.new("x-stream-filter-value must be a string")
+        end
+        case frame.arguments["x-stream-match-unfiltered"]?
+        when Bool
+          @match_unfiltered = frame.arguments["x-stream-match-unfiltered"].as(Bool)
+        when Nil
```
spuun, line 51:
```suggestion
        when Nil
          # nop
```

spuun (approved):

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -37,6 +39,20 @@ module LavinMQ
         when Nil, Int, Time, "first", "next", "last"
         else raise LavinMQ::Error::PreconditionFailed.new("x-stream-offset must be an integer, a timestamp, 'first', 'next' or 'last'")
         end
+        case frame.arguments["x-stream-filter-value"]?
+        when String
+          @filter = frame.arguments["x-stream-filter-value"].to_s.split(",")
```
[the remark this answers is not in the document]
carlhoerberg: by sorting the filter bsearch could be used to find matches faster. 
```suggestion
        case filter = frame.arguments["x-stream-filter-value"]?
        when String
          @filter = filter.split(',').sort!
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -236,6 +247,14 @@ module LavinMQ::AMQP
         headers.not_nil!("Message lacks headers")["x-stream-offset"].as(Int64)
       end
 
+      private def filter_value_from_headers(headers) : String?
+        if filter = headers.try &.["x-stream-filter-value"]?
+          filter.to_s
+        else
+          nil
+        end
+      end
```
carlhoerberg, line 256:
```suggestion
      private def filter_value_from_headers(headers) : String?
        headers.try &.fetch("x-stream-filter-value", nil).try &.to_s
      end
```

carlhoerberg: Or maybe this is more readable? 
```suggestion
      private def filter_value_from_headers(headers : Hash) : String?
        if value = headers["x-stream-filter-value"]?
          value.to_s
        end
      end
      
      private def filter_value_from_headers(headers : Nil) : Nil
      end
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -126,12 +126,23 @@ module LavinMQ::AMQP
...
+          end
           Envelope.new(sp, msg, redelivered: false)
         rescue ex
           raise Error.new(rfile, cause: ex)
         end
       end
 
+      private def matching?(msg_headers, consumer_filter, match_unfiltered) : Bool
+        if msg_filter = filter_value_from_headers(msg_headers)
+          return consumer_filter.includes? msg_filter
```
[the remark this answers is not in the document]
carlhoerberg: Could be speed up with bsearch (just have to make sure filters array is sorted). 
```suggestion
        if filter_value = filter_value_from_headers(msg_headers)
          return consumer_filter.bsearch { |f| f >= filter_value } == filter_value
```

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -126,12 +126,23 @@ module LavinMQ::AMQP
           sp = SegmentPosition.new(consumer.segment, consumer.pos, msg.bytesize.to_u32)
           consumer.pos += sp.bytesize
           consumer.offset += 1
+          if consumer_filter = consumer.filter
+            return unless matching?(msg.properties.headers, consumer_filter, consumer.match_unfiltered?)
+          end
```
carlhoerberg, line 131: I think it would be nicer to move this to the consumer. Then the consumer doesn't need to make `filter` and `match_unfiltered` public. 
```suggestion
          return unless consumer.filter_match?(msg.properties.headers)
```

carlhoerberg (changes requested):

src/lavinmq/amqp/queue/stream_queue_message_store.cr
viktorerlingsson: I think the first suggestion is readable enough!

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -9,8 +9,8 @@ module LavinMQ
       property segment : UInt32
       property pos : UInt32
       getter requeued = Deque(SegmentPosition).new
-      getter filter : Array(String)? = nil
-      getter? match_unfiltered : Bool = false
+      @filter : Array(String)? = nil
```
carlhoerberg, line 12: My last question is, should the filter be an empty array by default? Then  `filter_match?` could return early with `return true if @filter.empty?

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -9,6 +9,8 @@ module LavinMQ
       property segment : UInt32
       property pos : UInt32
       getter requeued = Deque(SegmentPosition).new
+      @filter : Array(String) = Array(String).new
+      @match_unfiltered : Bool = false
```
carlhoerberg, line 13:
```suggestion
      @filter = Array(String).new
      @match_unfiltered = false
```

carlhoerberg (approved):

PR #896, opened 2024-12-19
Log: strip logged user input for cleaner logs

kickster97:

It strip the protocol string from newlines, so we get cleaner logs if users send in bad input. 

EX: 

> Dec 19 06:07:33Z lmq.amqp.connection_factory: "Unexpected protocol '# 003
Dec 19 06:07:33Z  *%�
Dec 19 06:07:33Z  ', closing socket

send in a header with a bad protocol, with newlines, and see that the log is not devided.

### DIFF
--- a/src/lavinmq/amqp/connection_factory.cr
+++ b/src/lavinmq/amqp/connection_factory.cr
@@ -47,7 +47,10 @@ module LavinMQ
         elsif proto != AMQP::PROTOCOL_START_0_9_1 && proto != AMQP::PROTOCOL_START_0_9
           socket.write AMQP::PROTOCOL_START_0_9_1.to_slice
           socket.flush
-          log.warn { "Unexpected protocol '#{String.new(proto.to_slice)}', closing socket" }
+          log.warn {
+            bad_protocol = String.new(proto.to_slice).gsub(/\n/, "\\n")
+            "Unexpected protocol '#{bad_protocol}', closing socket"
+            }
           false
         else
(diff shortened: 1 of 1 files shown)


kickster97:
This should probably be done in all places where we log user input, but i'll put it up like this so we can discuss.

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -47,7 +47,10 @@ module LavinMQ
         elsif proto != AMQP::PROTOCOL_START_0_9_1 && proto != AMQP::PROTOCOL_START_0_9
           socket.write AMQP::PROTOCOL_START_0_9_1.to_slice
           socket.flush
-          log.warn { "Unexpected protocol '#{String.new(proto.to_slice)}', closing socket" }
+          log.warn {
+            bad_protocol = String.new(proto.to_slice).gsub(/\n/, "\\n")
+            "Unexpected protocol '#{bad_protocol}', closing socket"
+            }
```
carlhoerberg, line 53: only stringify the number of bytes we read, the rest will always be garbage. Then use `String#inspect` to render `\n` as `"\n"`, but also all other non non rendable characters, like `\t`, `\0` etc
```suggestion
          log.warn { "Unexpected protocol '#{String.new(proto.to_unsafe, count).inspect}', closing socket" }
```

carlhoerberg (changes requested):

PR #897, opened 2024-12-24
Use CRC32 for the replication protocol

carlhoerberg:

Uses CRC32 instead of SHA1 for calculating file hashes, for a follower to compare if they got all files or not. 

Specs


carlhoerberg:
Should maintain backwards compatibility with older followers that only support sha1.

spuun (approved):

johanrhodin:
Motivation is only speed (reduced complexity), or something else too?

carlhoerberg:
Yes, only motivation is that it's faster, but it's also more collision prone, should at least send the file size too then, which changes the protocol further.

The hash performance is one thing, but any algorithm still needs to read the whole file, which causes significant IO usage when hundreds of GB has to be read, so caching the hash results will always be faster than any algorithm, hence: #1088

PR #900, opened 2025-01-08
Store non-durable queues in separate transient dir

### LINKED ISSUE #796: Non-durable queues' datadir not removed on non-graceful shutdown
1. Create a non-durable queue
1. kill -9
1. Start lavinmq again
1. Queue data dir still exists

This is because frames for non-durable queues aren't written to definitions so there is no reference to the queues and no cleanup is being done on start.

viktorerlingsson:
Stores data directory for non-durable queues in a separate directory: `transient`. The transient dir is removed on both `VHost#init` and `VHost#close` to make sure we don't store unnecessary data. 

Also deletes files for non-durable queues on creation, meaning they will be automatically deleted once they are unmapped. This makes sure no segments are kept, even if lavinmq is shut down non-gracefully. 

Fixes #796 

Run specs

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -154,9 +154,10 @@ module LavinMQ::AMQP
     end
 
     private def make_data_dir : String
-      data_dir = File.join(@vhost.data_dir, Digest::SHA1.hexdigest @name)
-      unless durable?
-        data_dir = File.join(@vhost.data_dir, "transient", Digest::SHA1.hexdigest @name)
+      data_dir = if durable?
+        File.join(@vhost.data_dir, Digest::SHA1.hexdigest @name)
+      else
+        File.join(@vhost.data_dir, "transient", Digest::SHA1.hexdigest @name)
       end
       if Dir.exists? data_dir
         # delete left over files from transient queues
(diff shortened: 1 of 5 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -155,6 +155,9 @@ module LavinMQ::AMQP
 
     private def make_data_dir : String
       data_dir = File.join(@vhost.data_dir, Digest::SHA1.hexdigest @name)
+      unless durable?
+        data_dir = File.join(@vhost.data_dir, "transient", Digest::SHA1.hexdigest @name)
+      end
```
kickster97, line 160: minor: this way we don't do unnecessary assigns.. but since this is done once at init its really no big deal :)
 
```suggestion
      if durable?
        data_dir = File.join(@vhost.data_dir, Digest::SHA1.hexdigest @name)
      else
        data_dir = File.join(@vhost.data_dir, "transient", Digest::SHA1.hexdigest @name)
      end
```

kickster97 (approved):

src/lavinmq/amqp/queue/queue.cr
carlhoerberg: or:
```suggestion
      data_dir = if durable?
        File.join(@vhost.data_dir, Digest::SHA1.hexdigest @name)
      else
        File.join(@vhost.data_dir, "transient", Digest::SHA1.hexdigest @name)
      end
```

PR #901, opened 2025-01-08
Adds cleanup_datadir to lavinmqctl

viktorerlingsson:
Adds a `cleanup_datadir` cmd that removes any queue dirs without a corresponding queue in LavinMQ. Mainly useful for removing any "orphaned" non-durable queues that can have been left because of an earlier non-graceful shutdown of LavinMQ. 

Somewhat related to #900 

Add a directory to a vhost data directory, run `lavinmqctl cleanup_datadir <datadir>`, see that your new directory is removed.

### DIFF
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -4,6 +4,8 @@ require "./lavinmq/definitions_generator"
 require "http/client"
 require "json"
 require "option_parser"
+require "digest/sha1"
+require "file_utils"
 
 class LavinMQCtl
   @options = {} of String => String
@@ -41,6 +43,7 @@ class LavinMQCtl
     {"delete_exchange", "Delete exchange", "<name>"},
     {"set_vhost_limits", "Set VHost limits (max-connections, max-queues)", "<json>"},
     {"set_permissions", "Set permissions for a user", "<username> <configure> <write> <read>"},
+    {"cleanup_datadir", "Cleans up the datadir, removing orphaned files", "<datadir> <vhost>"},
   }
 
   def initialize
@@ -178,6 +181,7 @@ class LavinMQCtl
     when "set_vhost_limits"      then set_vhost_limits
     when "set_permissions"       then set_permissions
     when "definitions"           then definitions
+    when "cleanup_datadir"       then cleanup_datadir
     when "stop_app"
     when "start_app"
     else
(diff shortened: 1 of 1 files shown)


src/lavinmqctl.cr
```diff
@@ -700,6 +704,56 @@ class LavinMQCtl
...
+    vhosts.compact.each do |current_vhost|
+      resp = http.get "/api/queues/#{URI.encode_www_form(current_vhost)}", @headers
+      handle_response(resp, 200)
+      queue_list = [] of String
+      if queues = JSON.parse(resp.body).as_a?
+        queue_list = queues.map do |q|
+          next unless v = q.as_h?
+          Digest::SHA1.hexdigest(v["name"].to_s)
+        end
+      end
```
spuun, line 733: You'll need to make loop here to make sure you get all queues since the result is paginated. It's probably good to use `LavinMQ::HTTP::Controller::MAX_PAGE_SIZE` as page size to reduce the number of calls.

spuun (commented):
I think this is very risky operation. What if a client adds a queue while this is running?

Maybe a set of all queue dirs should be compiled first, then fetch all queues and remove existing queues from the set and remove the dirs that's left? Probably still a risk that a queue is recreated while this is happening.

viktorerlingsson:
> I think this is very risky operation. What if a client adds a queue while this is running?

Yes, it could be risky. I guess this isn't something that is intended to be run often, but rather something you can do once before migrating to a new VM or copying data to a new disk. 

> Maybe a set of all queue dirs should be compiled first, then fetch all queues and remove existing queues from the set and remove the dirs that's left? Probably still a risk that a queue is recreated while this is happening.

Yes, that should be more safe!

src/lavinmqctl.cr
viktorerlingsson: Ah. That seems to be true for pretty much all GETs in lavinmqctl 😬 
I'll take a look at that!

viktorerlingsson: This should fix that for all GETs https://github.com/cloudamqp/lavinmq/pull/904

carlhoerberg:
another offline approach is to parse the `definitions.amqp` file. could then use the DataDirLock so make sure no active server instance is running.

viktorerlingsson:
> another offline approach is to parse the `definitions.amqp` file. could then use the DataDirLock so make sure no active server instance is running.

Hmm, yeah maybe doing it offline is the only real safe option 🤔

viktorerlingsson:
This should not be needed once #900 is merged. Closing.

PR #904, opened 2025-01-09
Use pagination to get all results in lavinqmctl

viktorerlingsson:
No GETs in lavinmqctl supports pagination, so only the first page of data (10000 items) was ever received. This adds pagination support to all GETs. 

Create 10k+ queues, run `lavinmqctl list_queues -p <vhost>` (or any other list_x cmd), look at results

### DIFF
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -306,6 +306,23 @@ class LavinMQCtl
     URI.encode_www_form(@options["vhost"])
   end
 
+  MAX_PAGE_SIZE = 10_000
+
+  private def get(url, page = 1, items = [] of JSON::Any)
+    resp = http.get("#{url}?page=#{page}&page_size=#{MAX_PAGE_SIZE}", @headers)
+    handle_response(resp, 200)
+    if data = JSON.parse(resp.body).as_h?
+      items += data["items"].as_a
+      page = data["page"].as_i
+      if page < data["page_count"].as_i
+        return get(url, page + 1, items)
+      end
+    else
+      abort "invalid data"
+    end
+    items
+  end
+
   private def import_definitions
     file = ARGV.shift? || ""
     resp = if file == "-"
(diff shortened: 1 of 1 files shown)


spuun (approved):

kickster97 (approved):

src/lavinmqctl.cr
```diff
@@ -306,6 +306,23 @@ class LavinMQCtl
     URI.encode_www_form(@options["vhost"])
   end
 
+  MAX_PAGE_SIZE = 10_000
+
+  private def get(url, page = 1, items = [] of JSON::Any)
```
carlhoerberg, line 311: prefer `Array(...)` over `[] of ...`

src/lavinmqctl.cr
```diff
@@ -306,6 +306,23 @@ class LavinMQCtl
...
+    resp = http.get("#{url}?page=#{page}&page_size=#{MAX_PAGE_SIZE}", @headers)
+    handle_response(resp, 200)
+    if data = JSON.parse(resp.body).as_h?
+      items += data["items"].as_a
+      page = data["page"].as_i
+      if page < data["page_count"].as_i
+        return get(url, page + 1, items)
+      end
+    else
+      abort "invalid data"
```
carlhoerberg, line 321: maybe a more detailed error message could be helpful here

`abort "Unexpected response from #{url}\n#{resp.body}"`

src/lavinmqctl.cr
carlhoerberg: I would probably have used `Iterator.chain` instead of concating arrays.

src/lavinmqctl.cr
```diff
@@ -306,6 +306,23 @@ class LavinMQCtl
     URI.encode_www_form(@options["vhost"])
   end
 
+  MAX_PAGE_SIZE = 10_000
```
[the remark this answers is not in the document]
viktorerlingsson: yes!

PR #909, opened 2025-01-14
Make Launcher the starting point in all cases

spuun:
After other refactoring lavinmq no longer released its leader lease on graceful shutdown. The problem was that `Launcher` handled the signal, but didn't have a reference to `Clustering::Controller` or the `Leadership`, and therefore couldn't release the lease.

When looking into different solutions I realized that we had two ways of starting lavinmq: `Launcher` and `Clustering::Controller`.

This PR will refactor the upstart to make `Launcher` responsible for the upstart. This also makes launcher the signal handler instead of having this in both `Launcher` and `Clustering::Client`. `Launcher` will now have a reference to `Clustering::Controller` instead of the other way around.

Specs still passes, but some things needs to be tested manually.

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -19,6 +19,10 @@ class LavinMQ::Clustering::Controller
                       "tcp://#{System.hostname}:#{@config.clustering_port}"
   end
 
+  # This method is called by the Launcher#run.
+  # The block will be yielded when the controller's prerequisites for a leader
+  # to start are met, i.e when the current node has been elected leader.
+  # The method is blocking.
   def run(&)
     spawn(follow_leader, name: "Follower monitor")
     wait_to_be_insync
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -1,25 +1,31 @@
-require "../launcher"
 require "../etcd"
 require "./client"
 
 class LavinMQ::Clustering::Controller
   Log = LavinMQ::Log.for "clustering.controller"
 
-  @id : Int32
+  getter id : Int32
 
-  def initialize(@config = Config.instance)
-    @etcd = Etcd.new(@config.clustering_etcd_endpoints)
+  @repli_client : Client? = nil
+
+  def self.new(config : Config)
+    etcd = Etcd.new(config.clustering_etcd_endpoints)
+    new(config, etcd)
+  end
+
+  def initialize(@config : Config, @etcd : Etcd)
     @id = clustering_id
     @advertised_uri = @config.clustering_advertised_uri ||
                       "tcp://#{System.hostname}:#{@config.clustering_port}"
   end
 
-  def run
+  def run(&)
     spawn(follow_leader, name: "Follower monitor")
     wait_to_be_insync
     @lease = lease = @etcd.elect("#{@config.clustering_etcd_prefix}/leader", @advertised_uri) # blocks until becoming leader
+    @repli_client.try &.close
     # TODO: make sure we still are in the ISR set
-    @launcher = Launcher.new(@config, @etcd).start
+    yield
     loop do
       if lease.wait(30.seconds)
         break if @stopped
@@ -35,7 +41,7 @@ class LavinMQ::Clustering::Controller
 
   def stop
     @stopped = true
-    @launcher.try &.stop
+    @repli_client.try &.close
     @lease.try &.release
   end
 
@@ -74,7 +80,7 @@ class LavinMQ::Clustering::Controller
           break
         end
       end
-      repli_client = r = Clustering::Client.new(@config, @id, secret)
+      @repli_client = repli_client = r = Clustering::Client.new(@config, @id, secret)
       spawn r.follow(uri), name: "Clustering client #{uri}"
       SystemD.notify_ready
     end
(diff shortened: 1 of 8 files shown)


src/lavinmq/launcher.cr
```diff
@@ -27,28 +41,36 @@ module LavinMQ
...
+        @data_dir_lock = DataDirLock.new(@config.data_dir)
       end
-      @replicator = replicator = etcd ? Clustering::Server.new(@config, etcd) : Clustering::NoopServer.new
-      @amqp_server = LavinMQ::Server.new(@config, replicator)
-      @http_server = LavinMQ::HTTP::Server.new(@amqp_server)
+
+      if @config.clustering?
+        etcd = Etcd.new(@config.clustering_etcd_endpoints)
+        @runner = controller = Clustering::Controller.new(@config, etcd)
+        @replicator = Clustering::Server.new(@config, etcd, controller.id)
```
spuun, line 50: I think we should make `Clustering::Server` accept a `Clustering::Controller` instead of etcd and id, and then add methods to `Clustering::Controller` to handle cluster secret and isr. That would isolate all etcd code to `Clustering::Controller`.

src/lavinmq/clustering/controller.cr
```diff
@@ -74,7 +80,7 @@ class LavinMQ::Clustering::Controller
           break
         end
       end
-      repli_client = r = Clustering::Client.new(@config, @id, secret)
+      @repli_client = repli_client = r = Clustering::Client.new(@config, @id, secret)
       spawn r.follow(uri), name: "Clustering client #{uri}"
```
carlhoerberg, line 84:
```suggestion
      @repli_client = repli_client = Clustering::Client.new(@config, @id, secret)
      spawn repli_client.follow(uri), name: "Clustering client #{uri}"
```

src/lavinmq/launcher.cr
```diff
@@ -27,28 +41,36 @@ module LavinMQ
...
       @tls_context = create_tls_context if @config.tls_configured?
       reload_tls_context
       setup_signal_traps
-      setup_log_exchange
     end
 
-    def start : self
-      listen
+    private def start : self
+      @data_dir_lock.try &.acquire
```
carlhoerberg, line 62: Does the replication client acquire the datadirlock independently?

src/lavinmq/launcher.cr
```diff
@@ -7,16 +7,30 @@ require "./http/http_server"
 require "./in_memory_backend"
 require "./data_dir_lock"
 require "./etcd"
+require "./clustering/controller"
 
 module LavinMQ
+  struct DefaultRunner
```
carlhoerberg, line 13:
```suggestion
  struct StandaloneRunner
```
maybe?

src/lavinmq/clustering/controller.cr
```diff
@@ -1,25 +1,31 @@
...
   end
 
-  def run
+  def run(&)
     spawn(follow_leader, name: "Follower monitor")
     wait_to_be_insync
     @lease = lease = @etcd.elect("#{@config.clustering_etcd_prefix}/leader", @advertised_uri) # blocks until becoming leader
+    @repli_client.try &.close
     # TODO: make sure we still are in the ISR set
-    @launcher = Launcher.new(@config, @etcd).start
```
carlhoerberg, line 22: maybe comment what the yield is for (for the next one reading the code)

src/lavinmq/clustering/controller.cr
```diff
@@ -1,25 +1,31 @@
...
+  end
+
+  def initialize(@config : Config, @etcd : Etcd)
     @id = clustering_id
     @advertised_uri = @config.clustering_advertised_uri ||
                       "tcp://#{System.hostname}:#{@config.clustering_port}"
   end
 
-  def run
+  def run(&)
```
carlhoerberg, line 22: Maybe document what the yield is for, and that the method will block after yielding

carlhoerberg (approved):

src/lavinmq/launcher.cr
spuun: Yes, and I think that's good. I actually had a "bug" where the client didn't exit which I discovered when the server tried to startup after it was elected leader but couldn't aquire the lock.

src/lavinmq/clustering/controller.cr
spuun: This won't work because `repli_client` is nilable, so had to keep the `r`.

src/lavinmq/launcher.cr
carlhoerberg: So the clustering client will modify the data dir without a lock? What about acquiring the lock in the launcher in the constructor?

spuun: Both `Launcher` and `Clustering::Client` will lock, and it was thanks to this I noticed the bug that the `Client` didn't exit properly when the follower was elected leader, because `Launcher` couldn't aquire the lock.

It has always been likes this. (The bug was just during my refactoring, nothing that has been commited.)

carlhoerberg: oh, i misread! yes, good that replication client also acquire :)

src/lavinmq/launcher.cr
carlhoerberg: ok, but you would like to do that in another PR or?

carlhoerberg (approved):

src/lavinmq/launcher.cr
spuun: Yes

PR #910, opened 2025-01-14
Catch exits in spec_helper and raise an exception that can be handled in other specs

viktorerlingsson:
Defines exit in spec_helper so we can catch any program exits and instead raise an exception that we can handle in other specs. 

Run specs.

### DIFF
--- a/spec/etcd_spec.cr
+++ b/spec/etcd_spec.cr
@@ -2,6 +2,7 @@ require "spec"
 require "../src/lavinmq/etcd"
 require "file_utils"
 require "http/client"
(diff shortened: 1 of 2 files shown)


spuun:
`exit` should be `exit(status = 0)` and the exception should accept the status and expose it in a property. Then one can rescue it and check the status.

viktorerlingsson:
> `exit` should be `exit(status = 0)` and the exception should accept the status and expose it in a property. Then one can rescue it and check the status.

Yes, agreed 👍

carlhoerberg (approved):

PR #913, opened 2025-01-17
Keep Etcd dependencies to `Clustering::Controller`

spuun:
Tiny refactor to move all `Etcd` actions to `Clustering::Controller`.

Specs

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -17,6 +17,22 @@ class LavinMQ::Clustering::Controller
     @id = clustering_id
     @advertised_uri = @config.clustering_advertised_uri ||
                       "tcp://#{System.hostname}:#{@config.clustering_port}"
+    @isr_key = "#{@config.clustering_etcd_prefix}/isr"
+  end
+
+  def password : String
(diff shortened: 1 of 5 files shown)


src/lavinmq/clustering/controller.cr
```diff
@@ -17,6 +17,22 @@ class LavinMQ::Clustering::Controller
     @id = clustering_id
     @advertised_uri = @config.clustering_advertised_uri ||
                       "tcp://#{System.hostname}:#{@config.clustering_port}"
+    @isr_key = "#{@config.clustering_etcd_prefix}/isr"
+  end
+
+  def password : String
```
carlhoerberg, line 23: this can/will only be called by the leader? otherwise we have a racecondition in this method.

spuun: Yes. The `Clustering::Client` is created by the `Clustering::Controller` and the password is passed as constructor argument. `Clustering::Client` doesn't know about neither `Etcd` nor `Clustering::Controller`.

Maybe `Clustering::Server` should be created by `Clustering::Controller` too. Then `#password` wont have to be public and `Clustering::Controller` handles it all.

spuun:
I don't think this is necessary at the moment.

PR #914, opened 2025-01-17
UI: Load svgs from files instead of from data tags

viktorerlingsson:
Load svg's from files instead of from data-tag to abide by Content Security Policy.

Fixes console error: 
```
Refused to load the image 'data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='none' viewBox='0 0 20 20'%3e%3cpath stroke='%23f7f5f2' stroke-linecap='round' stroke-linejoin='round' stroke-width='1.5' d='M6 8l4 4 4-4'/%3e%3c/svg%3e' because it violates the following Content Security Policy directive: "img-src 'self'".
``` 
Start LavinMQ, make sure the arrow in the vhost dropdown is shown. And usermenu button in mobile. 

image

### DIFF
--- a/static/img/arrow-down.svg
+++ b/static/img/arrow-down.svg
@@ -1,0 +1,1 @@
+<svg xmlns='http://www.w3.org/2000/svg' fill='none' viewBox='0 0 20 20'><path stroke='#f7f5f2' stroke-linecap='round' stroke-linejoin='round' stroke-width='1.5' d='M6 8l4 4 4-4'/></svg>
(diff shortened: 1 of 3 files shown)


viktorerlingsson:
> Shouldn't the same error appear due to this line? :)
> 
> https://github.com/cloudamqp/lavinmq/blob/fb038060c259cbcfb46e35b2df349d296c9c555b/static/main.css#L1919

Hmm, yes, sort of. But that is never loaded because `#usermenu-button` is `display:none`, so there's no error. Maybe that can just be removed then @bengtmagnus ?

carlhoerberg:
An alternative could be to modify the CSP: `Content-Security-Policy: img-src data:` avoids a http request for a tiny img, but might have other downsides..

viktorerlingsson:
> An alternative could be to modify the CSP: `Content-Security-Policy: img-src data:` avoids a http request for a tiny svg, but might have other downsides..

Yeah, I think allowing `data:` _might_ increase the risks for some XSS, so this felt like the safest option.

carlhoerberg:
A third alternative could be to replace it with a utf8 character, if there's one similar enough

carlhoerberg:
Eg. ˅ (`"\u{2c5}"`)

carlhoerberg (approved):
Good! But let's fix all `content: url()` in the css as @ThomasSarlin mentioned while we're at it.

bengtmagnus:
Yes! The #usermenu-button is because it is only visible on mobile. I can check and fix them all 👍

PR #921, opened 2025-01-22
Abstract authenticator

kickster97:
This is a preparatory step towards introducing multiple authentication backends, which can be tried and run on a chain of command. (It breaks up [#861 ](https://github.com/cloudamqp/lavinmq/pull/861) so it dosn't become to large) 
This PR introduces the basic authenticator, which entails the basic authentication method we already support. 

These modules will be extended with oauth and http support in the future. 

includes some very basic specs, will be able to test more when we add multiple authentication backends.

### DIFF
--- a/spec/auth_spec.cr
+++ b/spec/auth_spec.cr
@@ -1,14 +1,12 @@
 require "./spec_helper"
 require "../src/lavinmq/auth/chain"
 require "../src/lavinmq/auth/authenticators/basic"
-require "../src/lavinmq/user_store"
 
 describe LavinMQ::Auth::Chain do
 
   it "Creates a default authentication chain if not configured" do
      with_amqp_server do |s|
-      chain = LavinMQ::Auth::Chain.create(s.@users)
-
+      chain = LavinMQ::Auth::Chain.create(s.@config, s.@users)
       chain.@backends.should be_a Array(LavinMQ::Auth::Authenticator)
       chain.@backends.size.should eq 1
      end
@@ -16,7 +14,7 @@ describe LavinMQ::Auth::Chain do
 
   it "Successfully authenticates and returns a basic user" do
     with_amqp_server do |s|
-      chain = LavinMQ::Auth::Chain.create(s.@users)
+      chain = LavinMQ::Auth::Chain.create(s.@config, s.@users)
       user = chain.authenticate("guest", "guest")
       user.should_not be_nil
     end
@@ -24,7 +22,7 @@ describe LavinMQ::Auth::Chain do
 
   it "Does not authenticate when given invalid credentials" do
     with_amqp_server do |s|
-      chain = LavinMQ::Auth::Chain.create(s.@users)
+      chain = LavinMQ::Auth::Chain.create(s.@config, s.@users)
       user = chain.authenticate("guest", "invalid")
       user.should be_nil
     end
--- a/spec/auth_spec.cr
+++ b/spec/auth_spec.cr
@@ -1,0 +1,32 @@
+require "./spec_helper"
+require "../src/lavinmq/auth/chain"
+require "../src/lavinmq/auth/authenticators/basic"
+require "../src/lavinmq/user_store"
+
+describe LavinMQ::Auth::Chain do
+
+  it "Creates a default authentication chain if not configured" do
+     with_amqp_server do |s|
+      chain = LavinMQ::Auth::Chain.create(s.@users)
+
+      chain.@backends.is_a?(Array(LavinMQ::Auth::Authenticator)).should be_true
+      chain.@backends.size.should eq 1
+     end
+  end
+
+  it "Successfully authenticates and returns a basic user" do
+    with_amqp_server do |s|
+      chain = LavinMQ::Auth::Chain.create(s.@users)
+      user = chain.authenticate("guest", "guest")
+      user.should_not be_nil
+    end
+  end
+
+  it "Does not authenticate when given invalid credentials" do
+    with_amqp_server do |s|
+      chain = LavinMQ::Auth::Chain.create(s.@users)
+      user = chain.authenticate("guest", "invalid")
+      user.should be_nil
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


spec/auth_spec.cr
```diff
@@ -0,0 +1,32 @@
...
+require "../src/lavinmq/auth/authenticators/basic"
+require "../src/lavinmq/user_store"
+
+describe LavinMQ::Auth::Chain do
+
+  it "Creates a default authentication chain if not configured" do
+     with_amqp_server do |s|
+      chain = LavinMQ::Auth::Chain.create(s.@users)
+
+      chain.@backends.is_a?(Array(LavinMQ::Auth::Authenticator)).should be_true
```
spuun, line 12:
```suggestion
      chain.@backends.should be_a Array(LavinMQ::Auth::Authenticator)
```

src/lavinmq/config.cr
```diff
@@ -62,6 +62,7 @@ module LavinMQ
     property default_consumer_prefetch = UInt16::MAX
     property yield_each_received_bytes = 131_072    # max number of bytes to read from a client connection without letting other tasks in the server do any work
     property yield_each_delivered_bytes = 1_048_576 # max number of bytes sent to a client without tending to other tasks in the server
+    property auth_backends : Array(String)? = ["basic"]
```
spuun, line 65:
```suggestion
    property auth_backends : Array(String) = ["basic"]
```
No need to be nilable when it has a default value

src/lavinmq/auth/chain.cr
```diff
@@ -0,0 +1,41 @@
...
+  module Auth
+    class Chain < Authenticator
+      @backends : Array(Authenticator)
+
+      def initialize(backends : Array(Authenticator))
+        @backends = backends
+      end
+
+      def self.create(users : UserStore) : Chain
+        backends = Config.instance.auth_backends
```
spuun, line 14:
```suggestion
      def self.create(config : Config, users : UserStore) : Chain
        backends = config.auth_backends
```
and inject the config

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -2,13 +2,15 @@ require "../version"
...
 require "../client/connection_factory"
+require "../auth/chain"
 
 module LavinMQ
   module AMQP
     class ConnectionFactory < LavinMQ::ConnectionFactory
       Log = LavinMQ::Log.for "amqp.connection_factory"
 
       def initialize(@users : UserStore, @vhosts : VHostStore)
+        @auth_chain = LavinMQ::Auth::Chain.create(@users)
```
spuun, line 13:
```suggestion
        @auth_chain = LavinMQ::Auth::Chain.create(Config.instance, @users)
```

src/lavinmq/auth/authenticators/basic.cr
```diff
@@ -0,0 +1,21 @@
...
+  module Auth
+    class BasicAuthenticator < Authenticator
+      def initialize(@users : UserStore)
+      end
+
+      def authenticate(username : String, password : String)
+        begin
+          user = @users[username]
+          return user if user && user.password && user.password.not_nil!.verify(password)
+          Log.info { "Basic authentication failed" }
```
spuun, line 14: I think you can remove this. The `ConnectionFactory` will log failed authentications with more details.

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -18,7 +20,7 @@ module LavinMQ
         logger = Logger.new(Log, metadata)
         if confirm_header(socket, logger)
           if start_ok = start(socket, logger)
-            if user = authenticate(socket, remote_address, start_ok, logger)
+            if user = authenticate(socket, remote_address, start_ok, logger, @auth_chain)
```
spuun, line 23:
```suggestion
            if user = authenticate(socket, remote_address, start_ok, logger)
```
No need to pass it here, you can use it directly in `authenticate`.

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -103,10 +105,10 @@ module LavinMQ
         end
       end
 
-      def authenticate(socket, remote_address, start_ok, log)
+      def authenticate(socket, remote_address, start_ok, log, auth_chain)
```
spuun, line 108:
```suggestion
      def authenticate(socket, remote_address, start_ok, log)
```

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -103,10 +105,10 @@ module LavinMQ
...
         username, password = credentials(start_ok)
         user = @users[username]?
-        return user if user && user.password && user.password.not_nil!.verify(password) &&
+        return user if user && auth_chain.authenticate(username, password) &&
```
spuun, line 111:
```suggestion
        return user if user && @auth_chain.authenticate(username, password) &&
```

spuun (commented):
Just minor stuff.

src/lavinmq/config.cr
kickster97: That default was mostly put there for manual testing hehe, idk if we should have a default value like this? if we do, we don't need to have the safety clause in the chain that checks if its empty etc

src/lavinmq/auth/chain.cr
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def authenticate(username : String, password : String)
+        @backends.each do |backend|
+          if user = backend.authenticate(username, password)
+            return user
+          end
+        end
+        nil
+      end
```
spuun, line 38: I think this could be changed to

```suggestion
      def authenticate(username : String, password : String)
        @backends.find &.authenticate(username, password)
      end
```

src/lavinmq/auth/authenticator.cr
```diff
@@ -0,0 +1,9 @@
+module LavinMQ
+  module Auth
+    abstract class Authenticator
+      Log = LavinMQ::Log.for "auth.handler"
+
+      abstract def authenticate(username : String, password : String)
```
spuun, line 6: Maybe it would be nice to be explicit about the return type here
```suggestion
      abstract def authenticate(username : String, password : String) : User?
```

spuun (approved):
:+1:
The comments are optional.

src/lavinmq/auth/chain.cr
kickster97: we should use `find_value`, but yes!

PR #925, opened 2025-01-27
Tooltips added for details table

bengtmagnus:
Adds tooltips with html title for better describing queue features.

Check if it looks ok and works.


kickster97 (approved):
looks good

kickster97:
b ut with further consideration, we spoke about that it would be nice if they were consistent with the rest of the tooltips, and that we could put the tooltip on the features box instead

but also in general the tooltips is something we can improve

PR #929, opened 2025-02-04
add message rates to /exchanges

kickster97:
Adds message rates to the exchange view

publish some messages, maybe try some with immediate deliver or try publishing to a queue that doesn't exist to get different values.


johanrhodin (commented):
reates->rates

spuun (approved):

PR #933, opened 2025-02-05
log and re-raise error if load unexpectedly fails

kickster97:
Add clearer logs if lavinmq unexpectedly fails to load a user, parameter or vhost file during launch

corrupt users.json and read logs

### DIFF
--- a/src/lavinmq/parameter_store.cr
+++ b/src/lavinmq/parameter_store.cr
@@ -76,6 +76,9 @@ module LavinMQ
         end
       end
       @log.debug { "#{size} items loaded from #{@file_name}" }
+    rescue ex
+      @log.error(exception: ex) { "Failed to load #{@file_name}" }
+      raise ex
     end
   end
 end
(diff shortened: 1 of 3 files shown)


src/lavinmq/parameter_store.cr
```diff
@@ -76,6 +76,9 @@ module LavinMQ
         end
       end
       @log.debug { "#{size} items loaded from #{@file_name}" }
+    rescue ex
+      @log.error { "Failed to load #{@file_name}: #{ex}" }
```
spuun, line 80: Maybe
```suggestion
      @log.error(exception: ex) { "Failed to load #{@file_name}" }
```
Or dunno how much we want to log, but I guess the more the merrier in a "rescue all" clause?

kickster97: yes! nice

baelter: And include the exception in the other rescues as well I suppose? :)

kickster97: yes, got stuck in other stuff but will fix the rest too!

carlhoerberg (approved):

PR #934, opened 2025-02-07
Use MFiles in followers

### LINKED ISSUE #931: Follower slowing down cluster - switch to MFiles
Adding a follower affects the throughput heavily. After digging around I discovered that the usage of `File` instead of `MFile` in the followers penalties the performance a lot.

A switch to `MFile` also requires handling of

- increase capacity if needed
- unmap (and close) files at some point
  - unmap oldest/first file when a threshold of mapped files is reached?
  - unmap based on some "last access" timestamp?

spuun:

This fixes #931

Followers will be using MFiles. It will keep track on last access time for a file and close it when it's been unused for at least 10 seconds (this can be discussed). Files are only closed when a new one is opened, meaning that a lot of files may stay open for a good period of time.

Run specs (still passes). Should probably add specs to verify that files are closed and unmapped.

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -32,7 +32,7 @@ module LavinMQ
       end
 
       def size
-        @file.try &.size || File.size(@path)
+        @file.size
       end
 
       def rename(new_filename)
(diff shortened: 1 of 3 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -19,14 +71,28 @@ module LavinMQ
       def initialize(@config : Config, @id : Int32, @password : String, proxy = true)
         System.maximize_fd_limit
         @data_dir = config.data_dir
-        @files = Hash(String, File).new do |h, k|
-          path = File.join(@data_dir, k)
-          Dir.mkdir_p File.dirname(path)
-          h[k] = File.open(path, "a").tap &.sync = true
```
carlhoerberg, line 25: If you just changed this line instead, to sync false. What kind of performance do you then get?

spuun: Seems to be on par with m files.

spuun:
Probably not needed.

PR #940, opened 2025-02-10
Improve connection negotiation and channel_max enforcement

### LINKED ISSUE #939: Server channel_max is not respected
**Describe the bug**
The server sends `channel_max` in the `Tune` frame, which is defined as:
> Specifies highest channel number that the server permits. Usable channel numbers are in the range 1..channel-max. Zero indicates no specified limit. 

The client responds with `channel_max` in the `TuneOk` frame, which is defined as:
> The maximum total number of channels that the client will use per connection.
> If the client specifies a channel max that is higher than the value provided by the server, the server MUST close the connection without attempting a negotiated close. The server may report the error in some fashion to assist implementors.

The spec also includes assertions that the client can't specify a value larger that the server:

```
<assert check="notnull"/>
<assert check="le" method="tune" field="channel-max"/>
``` 

This is also mentioned in RabbitMQs documentation:
> The lower value of the two is used: the client cannot be configured to allow for more channels than the server configured maximum

https://www.rabbitmq.com/docs/channels#channel-max

LavinMQ currently just accepts the value the client sends in `TuneOK`

https://github.com/cloudamqp/lavinmq/blob/20ebdb0517208a6ca6102367cb1015914a44f2f6/src/lavinmq/amqp/client.cr#L45

**Describe your setup**
Please describe the environment you are using, e.g. LavinMQ version, operating system, client software. 

**How to reproduce**
Only verified by reading the code so far, we should create a spec

**Expected behavior**
The client should not be able to set a value higher that what the server suggests

baelter:

- Fixing `channel_max` and `frame_max` negotiation. 
- Enforces `channel_max`. Closes #939 
- Improves readability by using enum for Channel and Connection Close reply-code
- Adds support for specs that need custom config values

Specs

### DIFF
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -308,14 +308,14 @@ module LavinMQ
             frame.body.skip(frame.body_size)
           else
             @log.error { "Channel #{frame.channel} not open while processing #{frame.class.name}" }
-            close_connection(frame, 504_u16, "CHANNEL_ERROR - Channel #{frame.channel} not open")
+            close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "Channel #{frame.channel} not open")
           end
         end
       end
 
       private def open_channel(frame)
         if @channels.has_key? frame.channel
-          close_connection(frame, 504_u16, "CHANNEL_ERROR - second 'channel.open' seen")
+          close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "second 'channel.open' seen")
         else
           @channels[frame.channel] = AMQP::Channel.new(self, frame.channel)
           @vhost.event_tick(EventType::ChannelCreated)
@@ -399,7 +399,7 @@ module LavinMQ
         end
       rescue ex : LavinMQ::Error::UnexpectedFrame
         @log.error { ex.inspect }
-        close_channel(ex.frame, 505_u16, "UNEXPECTED_FRAME - #{ex.frame.class.name}")
+        close_channel(ex.frame, ChannelReplyCode::UNEXPECTED_FRAME, ex.frame.class.name)
       end
 
       private def cleanup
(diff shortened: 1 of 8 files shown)


baelter:
This is not actually  a failing spec since `amqp-client.cr` guards against this: https://github.com/cloudamqp/amqp-client.cr/blob/main/src/amqp-client/connection.cr#L296

src/lavinmq/amqp/client.cr
```diff
@@ -440,24 +445,60 @@ module LavinMQ
         !@running
       end
 
-      def close_channel(frame : AMQ::Protocol::Frame, code, text)
-        return close_connection(frame, code, text) if frame.channel.zero?
+      enum ChannelReplyCode : UInt16
```
spuun, line 448: I wouldn't mind to put this in its own file. It could be put in AMQP namspace, no need for it to exist under `Client` (so no need for a client directory).

src/lavinmq/amqp/client.cr
```diff
@@ -440,24 +445,60 @@ module LavinMQ
...
+          send AMQP::Frame::Channel::Close.new(frame.channel, code.value, text, frame.class_id, frame.method_id)
         else
-          send AMQP::Frame::Channel::Close.new(frame.channel, code, text, 0, 0)
+          send AMQP::Frame::Channel::Close.new(frame.channel, code.value, text, 0, 0)
         end
         @channels.delete(frame.channel).try &.close
       end
 
-      def close_connection(frame : AMQ::Protocol::Frame?, code, text)
+      enum ConnectionReplyCode : UInt16
```
spuun, line 477: ditto

spuun (approved):
👍

src/lavinmq/amqp/client.cr
viktorerlingsson: 👍 
`ChannelReplyCode` and `ConnectionReplyCode` could also maybe be merged to a single `ReplyCode`?

baelter: Did the first thing, but want to keep them separate to make it clear which replycodes are meant for connection close and channel close.

PR #941, opened 2025-02-11
Separate process to keepalive etcd lease

### LINKED ISSUE #935: Losing leadship when disconnecting a client with tons of channels/consumers
Create 20k consumers on one connection. Disconnect it and observe how lavinmq is busy handling it and then losing leadership.

carlhoerberg:
During high load lavinmq can struggle to get renew the etcd lease because the fiber is not run often enough. By putting it in a subprocess we can ensure that it gets more resources. When proper MT is added to LavinMQ we can revert this patch.

It introduces a dependency on `etcdctl`, thus for example in Ubuntu/Debian the etcd-client package has to be installed.

Fixes #935

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY ./src ./src
 
 # Run specs on build platform
 FROM base AS spec
-RUN apt-get install -y etcd-server
+RUN apt-get install -y etcd-server etcd-client
 COPY ./spec ./spec
 ARG spec_args="--order random"
 RUN crystal spec ${spec_args}
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Maintainer: CloudAMQP <contact@cloudamqp.com>
 Package: lavinmq
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, adduser
-Suggests: etcd-server (>= 3.4.0)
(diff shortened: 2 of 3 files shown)


viktorerlingsson:
Should we also remove `keepalive_loop` and `lease_keepalive`? they aren't used anymore with this change, so might be confusing to leave them in

spuun:
I end up with dangling `etcdctl` processes if I kill lavinmq. Maybe it's acceptable?

spuun:
I end up with dangling  `etcdctl` processes, but maybe that's acceptable?

carlhoerberg:
Ok, need to call `prctl(PR_SET_PDEATHSIG, SIGTERM)` between `fork` and `exec` to prevent that...

spuun:
A dangling process prevents the node from starting again since the old lease is never released. In a multi-node cluster no failover will happen.

carlhoerberg:
Good find! But as mentioned, prctl inbetween fork and exec should solve it.

carlhoerberg:
We don't have a good solution to prevent dangling `etcdctl` processes on other platforms than linux. We could hold off until Crystal 1.16 with thread support.

carlhoerberg:
Let's use a separate execution context for the keepalive instead.

PR #942, opened 2025-02-12
Retain store bugfixes

spuun:
Fixed the flaky retain store clustering spec. It wasn't really flaky in itself, but worked if you had a running etcd.

When etcd connection string was fixed another bug was exposed: messages files got truncated. It turned out files was opened with wrong mode.

Also, `#read` only worked once since IO wasn't rewinded.

Run specs


kickster97 (approved):
nice catch!

Issue #943, opened 2025-02-12
MQTT connections are missing `client_properties` in api response for `/api/connections`


snichme:
**Describe the bug**

MQTT connections are missing `client_properties` in api response for `/api/connections`  which causes the javascript that renders the connection to fail resulting in no connections are listed. 

![Image](https://github.com/user-attachments/assets/3ff15ed4-34a9-4da7-91c4-1f4f7df75989)

**How to reproduce**

Run LavinMQ from main (currently commit f01449d9) and connect some MQTT clients, then list the connections in the mgmt interface and see if fail 

**Expected behavior**

It should either handle the missing property in the javascript that renders the view or add the property from in the API response. 
The OpenAPI spec says the property is there so I think it should be added even it it's empty for mqtt connections.

PR #944, opened 2025-02-12
MQTT client handles client_properties

kickster97:
It adds handling of client_properties and adds some more details to the details tuple for a smother UI experience with MQTT connections

image

addresses #943 

connect an mqtt client and look at the `/connections` view

### DIFF
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -145,7 +145,7 @@ module LavinMQ
           ssl:               @connection_info.ssl?,
           tls_version:       @connection_info.ssl_version,
           cipher:            @connection_info.ssl_cipher,
-          client_properties: AMQP::Table.new,
+          client_properties: NamedTuple.new,
         }.merge(stats_details)
       end
 
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -135,16 +135,16 @@ module LavinMQ
 
       def details_tuple
         {
-          vhost:        @broker.vhost.name,
-          user:         @user.name,
-          protocol:     "MQTT 3.1.1",
-          client_id:    @client_id,
-          name:         @name,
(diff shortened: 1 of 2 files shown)


static/js/connections.js
```diff
@@ -37,11 +33,14 @@ Table.renderTable('table', tableOptions, function (tr, item, all) {
     Table.renderCell(tr, 9, item.channel_max, 'right')
     Table.renderCell(tr, 10, item.timeout, 'right')
     const clientDiv = document.createElement('span')
-    if (item?.client_properties) {
-      clientDiv.textContent = `${item.client_properties.product} / ${item.client_properties.platform || ''}`
-      clientDiv.appendChild(document.createElement('br'))
-      clientDiv.appendChild(document.createElement('small')).textContent = item.client_properties.version
+    console.log(item.client_properties)
```
spuun, line 36:
```suggestion
```

kickster97: haha how does this always slip in

src/lavinmq/mqtt/client.cr
```diff
@@ -135,11 +135,17 @@ module LavinMQ
...
+          user:              @user.name,
+          protocol:          "MQTT 3.1.1",
+          client_id:         @client_id,
+          name:              @name,
+          connected_at:      @connected_at,
+          state:             state,
+          ssl:               @connection_info.ssl?,
+          tls_version:       @connection_info.ssl_version,
+          cipher:            @connection_info.ssl_cipher,
+          client_properties: AMQP::Table.new,
```
spuun, line 148: Does this really need to be a Table? It's not possible to just put an empty named tuple here?
If it must be a Table it's probably better to have one Table instance as a class variable and use that as an "empty" Table.

kickster97: no, you're right! lets use empty NamedTuple

spuun (approved):

PR #945, opened 2025-02-12
Channel max part2

baelter:

Was i little trigger happy with the prio PR....

- Improving channel_max specs
- Actually implementing channel_max 0 = unlimited

Question: How do we handle the "hard" limit of `UInt16::MAX`?
Should we communicate that? Or have it as a "hidden" limit?

Specs


spuun:
> Question: How do we handle the "hard" limit of UInt16::MAX?
Should we communicate that? Or have it as a "hidden" limit?

It's probably insane to open more than (or even close to) `UInt16::MAX`, but the spec allows it. I guess it would be good to support real unlimited.

baelter:
> real unlimited

Well, that is not possible. There is a limit to how many electrons there are in the universe. Or maybe not. But a limit to how many we have access to :)

spuun (commented):
With `channel_max = 0` in config:
```
# lavinmqperf connection-count -x 1 -c 100000 --uri 'amqp://guest:guest@localhost/%2F?channel_max=0' 
Unhandled exception in spawn: channel_max reached (Exception)
```

baelter:
> channel_max reached

`UInt16::MAX = 65535` so it's expects. And that error is coming from amqp-client, but if it would not lavinmq would also stop at 65535 and close the connection. 

Do we want channel ids to be `UInt32` instead?

baelter:
Actually 0 is not "unlimited" it's

> Zero indicates no specified limit.

And channel ids are `short` i.e uint16 per definition. So theoretical max is 65535. In RabbitMQ as well.

spuun:
> > channel_max reached
> 
> `UInt16::MAX = 65535` so it's expected. And that error is coming from amqp-client, but if it would not lavinmq would also stop at 65535 and close the connection.
> 
> Do we want channel ids to be `UInt32` instead?

I was confused by the commit message `Implement channel_max 0 = unlimited`. But "unlimited" can't exist since channel id is a "short" (uint16), so unlimited means `UInt16::MAX`.

spuun:
> Actually 0 is not "unlimited" it's
> 
> > Zero indicates no specified limit.
> 
> And channel ids are `short` i.e uint16 per definition. So theoretical max is 65535. In RabbitMQ as well.

Oh, didn't update before i wrote my last comment! 👍

spuun (approved):

PR #947, opened 2025-02-13
Prevent lavinmq from freezing when closing consumers and channels

spuun:
When closing a connection with tons of consumers lavinmq will freeze because the fiber closing the connection will be CPU bound for a very long time.

This PR adds yields to the loops closing consumers and channels.

Start lavinmq then run `lavinmqperf connection-count -x 1 -c 10000 -C 1`. Close perf and observe that e.g. mgmt ui is responsive while lavinmq is closing the connection.

### DIFF
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -636,7 +636,10 @@ module LavinMQ
 
       def close
         @running = false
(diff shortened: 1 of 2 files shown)


viktorerlingsson (approved):
LGTM

PR #954, opened 2025-02-14
Add GC metrics to prometheus endpoints

### LINKED ISSUE #936: Add GC stats to prometheus metrics
It would be nice with `GC.stats` in prometheus metrics.

spuun:
This will make stats loop save GC.prof_stats which is then added to prometheus `metrics`

Metrics exported is what's in https://github.com/ivmai/bdwgc/blob/47e0ef445d1aa48b8c48ef166b117386e4e0100c/include/gc/gc.h#L795

Fixes #936

Spin up lavinmq, then `curl -u guest:guest http://localhost:15672/metrics | grep gc_`

### DIFF
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -334,7 +334,7 @@ module LavinMQ
         writer.write({name: "gc_since_last_collection_explicitly_freed_bytes", value: gc_stats.expl_freed_bytes_since_gc,
                       type: "counter",
                       help: "Number of bytes freed explicitly since the recent GC"})
-        k
+
         writer.write({name: "gc_from_os_obtained_bytes_total", value: gc_stats.obtained_from_os_bytes,
                       type: "counter",
                       help: "Total amount of memory obtained from OS, in bytes"})
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -39,7 +39,8 @@ module LavinMQ
       def write(m : Metric)
         return if m[:value].nil?
         io = @io
-        name = "#{@prefix}_#{m[:name]}"
+
+        name = name_writer(m)
         if t = m[:type]?
           io << "# TYPE " << name << " " << t << "\n"
         end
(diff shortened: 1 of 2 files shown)


spuun:
TODO: fix naming according to https://prometheus.io/docs/practices/naming/

jage:
Took it for a spin in Prometheus + Grafana via Snowstorm, a few `0` values but maybe expected.

Screenshot 2025-02-17 at 11 46 33

Screenshot 2025-02-17 at 11 49 58

I guess there's no timings for how long it spends in GC, e.g. mark, sweep?

spuun:
> Took it for a spin in Prometheus + Grafana via Snowstorm, a few `0` values but maybe expected.

Yeah. Probably. It may depend on the gc lib version.

> I guess there's no timings for how long it spends in GC, e.g. mark, sweep?

Probably not. This is what I've found.

src/lavinmq/http/controller/prometheus.cr
```diff
@@ -291,40 +291,51 @@ module LavinMQ
...
                       type: "gauge",
                       help: "Number of bytes not considered candidates for GC"})
-        writer.write({name: "gc_no", value: gc_stats.gc_no,
+
+        writer.write({name: "gc_cycles_total", value: gc_stats.gc_no,
                       type: "counter",
                       help: "Garbage collection cycle number (value may wrap)"})
-        writer.write({name: "gc_markers_m1", value: gc_stats.markers_m1,
+
+        writer.write({name: "gc_marker_threads_total", value: gc_stats.markers_m1,
```
jage, line 322:
```suggestion
        writer.write({name: "gc_marker_threads", value: gc_stats.markers_m1,
```

I think the unit is "threads" and it's not a counter so no need for total, or am I overthinking this? :)

spuun: IMO "threads" is enough. And yeah, not counter. It's probably a static number in our case 🤷

spuun:
> @jage Oups, missed that you had commented already.

Well... Dunno about the names. Tried to shuffle around the words and make labels that make sense ending with unit.

jage (approved):
👍

snichme:
Would it make sense to run `gc_stats = GC.prof_stats` inside the `gc_metrics` methods instead?  So we get fresh data everytime we query the endpoint? Does that come with any downsides?

spuun:
> Would it make sense to run `gc_stats = GC.prof_stats` inside the `gc_metrics` methods instead? So we get fresh data everytime we query the endpoint? Does that come with any downsides?

Yeah, it was something I was thinking about, but dunno. If it's collected at the same time as other data maybe GC metrics will align with stuff in the other metrics?

snichme:
> Yeah, it was something I was thinking about, but dunno. If it's collected at the same time as other data maybe GC metrics will align with stuff in the other metrics?

aha, makes sense

snichme (approved):

PR #955, opened 2025-02-16
lavinmqperf queue-count test scenario

annieblomgren:
Adding a queue-count test scenario to lavinmqperf (similar to the connection-count) that will create x number of queues, using 1 connection and 1 channel. How many queues it will create can be set with -q flag. 

It's really simple now, but we can extend with more arguments if we find it useful. 

run `lavinmqperf queue-count`

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -19,6 +19,7 @@ module LavinMQPerf
   when "channel-churn"    then ChannelChurn.new.run
   when "consumer-churn"   then ConsumerChurn.new.run
   when "connection-count" then ConnectionCount.new.run
+  when "queue-count"      then QueueCount.new.run
   when /^.+$/             then Perf.new.run([mode.not_nil!])
   else                         abort Perf.new.banner
   end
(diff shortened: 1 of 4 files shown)


annieblomgren:

annieblomgren:
ah TIL thanks! what a handy tool! have fixed format now but left `require "amqp-client"` in there as all the other test cases use have it. Is that ok?

Yeah moving rss method to helper sounds good, I can do that in another PR?

viktorerlingsson:
> ah TIL thanks! what a handy tool! have fixed format now but left `require "amqp-client"` in there as all the other test cases use have it. Is that ok?

That's fine! 

> 
> Yeah moving rss method to helper sounds good, I can do that in another PR?

Sure. But you could do it in this PR as well IMO. 
Just move the `rss` method to `perf.cr` and remove it from `connection_count` and `queue_count`

snichme (approved):
Tested locally and it looks good

Nice!

kickster97:

PR #956, opened 2025-02-18
Add http authenticator

JadeKharats:
This PR adds the ability to configure a "rabbit auth backend" in addition to the basic one.

To work, the user must exist in the AMQP server UserStore.

This PR also adds a client for the "Rabbit Auth Backend" because it provides other services than authentication. It also manages authorization on elements such as vhosts.

This PR does not fully cover issue 246 but meets the need of feature 952

For testing, `crystal spec` is enough.
The PR adds, in the `spec_helper`, a stubbed `Rabbit Auth Backend`

### DIFF
--- a/src/lavinmq/auth/authenticators/rabbit_http_auth.cr
+++ b/src/lavinmq/auth/authenticators/rabbit_http_auth.cr
@@ -1,0 +1,29 @@
+require "../authenticator"
+require "../../server"
+
+module LavinMQ
+  module Auth
+    class RabbitHttpAuthenticator < Authenticator
+      def initialize(@users : UserStore)
+      end
+
+      def authenticate(username : String, password : Bytes) : User?
+        user = @users[username]
+        return user if user && rabbit_http_authenticate?(username, password)
+      rescue ex : Exception
+        Log.error { "Rabbit Http authentication failed: #{ex.message}" }
+      end
+
+      private def rabbit_http_authenticate?(username : String, password : Bytes) : Bool
+        payload = {
+          "username" => username,
+          "password" => password.to_s,
+        }.to_json
+
+        ::HTTP::Client.post(Config.instance.auth_http_user_path,
+          headers: ::HTTP::Headers{"Content-Type" => "application/json"},
+          body: payload).success?
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


src/lavinmq/auth/authenticators/rabbit_http_auth.cr
```diff
@@ -0,0 +1,41 @@
...
+require "../../server"
+
+module LavinMQ
+  module Auth
+    class RabbitBackendAuthenticator < Authenticator
+      def initialize(@users : UserStore)
+      end
+
+      def authenticate(username : String, password : Bytes) : User?
+        user = @users[username]
```
kickster97, line 11: I don’t think we should require a user to be pre-registered in LavinMQ just to authorize. Instead, if they can authenticate via the HTTP backend, we can the go on to create a "temporary/remote" user with some permissions that are sent in with the connect packet.
This would be a new user class, and needs to be added before we can move forward with implementing "remote" authenticators. (OAuth2 will also need it)

src/lavinmq/auth/authenticators/rabbit_http_auth.cr
```diff
@@ -0,0 +1,41 @@
+require "../authenticator"
+require "../../server"
+
+module LavinMQ
+  module Auth
+    class RabbitBackendAuthenticator < Authenticator
```
kickster97, line 6: Also, the naming should be HTTPBackendAuthenticator? avoid using rabbit in naming :)

src/lavinmq/auth/authenticators/rabbit_http_auth.cr
JadeKharats: The authenticate method is defined to return a `User`. I think this is a good thing. To add the notion of `RemoteUser` and `LocalUser`, I think I'll make an abstract `User` class, transform the current `User` into `LocalUser`.
Is this what you had in mind?

src/lavinmq/auth/authenticators/rabbit_http_auth.cr
JadeKharats: Ok, I'm removing all mentions of Rabbit.

kickster97:
closing due to inactivity

PR #958, opened 2025-02-20
automatically rerun failed jobs in debian workflow

kickster97:
We've been dealing with random workflow failures on ubuntu-24.04 for linux/arm64, which usually get fixed when we manually rerun them. To cut down on these hiccups and the need for manual restarts, this pull request adds an automatic retry feature for any jobs that fail.

observe deb workflow

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -108,3 +108,13 @@ jobs:
               https://packagecloud.io/api/v1/repos/${{ github.repository }}-head/packages.json
           done
         if: github.event_name != 'pull_request'
+  rerun_on_failure:
+    needs: build_deb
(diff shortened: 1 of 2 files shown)


jage:
Heisenbug, now it worked on first try so we can't test retry. :)

jage (approved):
I think it looks good, low risk to merge and see if it helps?

viktorerlingsson:
I think it's OK to merge as well. 

Guess there's a case where we actually break something where re-running won't help, but since max retries is low it shouldn't be that much of an issue?

kickster97:
reverted this merge, had something wrong in the references.. making changes trying again after lunch.. might try to run stuff locally so its not bothering anyone, but have to do some research about how to run gha locally

PR #959, opened 2025-02-20
Auto rerun failed deb workflows

kickster97:
We've been dealing with random workflow failures on ubuntu-24.04 for linux/arm64, which usually get fixed when we manually rerun them. To cut down on these hiccups and the need for manual restarts, this pull request adds an automatic retry feature for any jobs that fail.

It only retries 2 times before it gives up.


viktorerlingsson (approved):
LGTM!

PR #968, opened 2025-02-28
Handle missing files in follower sync

spuun:
Haven't been able to reproduce it, but we've seen errors where the leader gets a `FileNotFound` when it hashes files (an ack file was missing) causing the sync to fail and the follower to disconnect.

This PR will add check existence of files before trying to read them.

Make sure an ack file exists when the leader is started, then remove it. Start follower and observe no hash being sent for the removed file.

### DIFF
--- a/src/lavinmq/clustering/follower.cr
+++ b/src/lavinmq/clustering/follower.cr
@@ -117,7 +117,6 @@ module LavinMQ
         loop do
           filename_len = socket.read_bytes Int32, IO::ByteFormat::LittleEndian
           break if filename_len.zero?
-
           filename = socket.read_string(filename_len)
           send_requested_file(filename)
           @lz4.flush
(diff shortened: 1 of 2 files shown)


snichme (approved):

carlhoerberg:
When it happens so in frequently it's cheaper to catch the exception instead of checking the for the existence of each. 

But it also sounds more like a race condition and that we need more locks somewhere. So now it was a file that did exist but doesnt any more, what about new files that has been added? Those will not be synced at all then!

spuun:
> When it happens so in frequently it's cheaper to catch the exception instead of checking the for the existence of each.

True. I'll change that.
 
> But it also sounds more like a race condition and that we need more locks somewhere. So now it was a file that did exist but doesnt any more, what about new files that has been added? Those will not be synced at all then!

I guess the file may be deleted after the hash has been sent, but before the follower has requested it. But that should be ok, and just an empty file may be sent (as this PR does it)?

PR #969, opened 2025-03-03
Validate expiration when publishing via http api

### LINKED ISSUE #962: Force expiration property to be a number
(This is not a feature request but also not a bug since it does follow the AMQP 0.9.1 spec)

According to the AMQP 9.0.1 protocol, the expiration property on a message is a `String` property.
Today If LavinMQ gets a message with the expiration property "hello", it will try to convert it to an integer but if it fails it will let it be nil. 
It could be argued that its expected to raise something telling the user that "hello" is not a valid expiration value. 

For inspiration, RabbitMQ enforces expiration to be a number with: `Precondition failed reason="Expiration not a number"`

Raise precondition failed  and inform the user if the expiration string is not a value we can handle.

spuun:
Fixes #962

Manually for now.

### DIFF
--- a/src/lavinmq/http/controller/exchanges.cr
+++ b/src/lavinmq/http/controller/exchanges.cr
@@ -138,6 +138,15 @@ module LavinMQ
             unless properties && routing_key && payload && payload_encoding
               bad_request(context, "Fields 'properties', 'routing_key', 'payload' and 'payload_encoding' are required")
             end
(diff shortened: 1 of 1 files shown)


viktorerlingsson:
Changes looks good to me! But I think it would be nice to add a spec for this to avoid regression in the future

carlhoerberg (approved):

snichme (approved):

PR #980, opened 2025-03-10
Add Deleted vhost Message Counters and Prometheus Metrics

### LINKED ISSUE #916: Add redelivery rate to metrics
During operations of LavinMQ, the re-delivery rate is an important indicator, if sth is wrong with the overall client application.

Currently the metrics do not show the re-delivery rate (or count?).

An alternative could be to check if the minimum `lavinmq_queue_messages` never reaches 0 within 1 hour, but this could easily be true in larger applications, anyway.

Please add it to the `/metrics` endpoint, which provides the Prometheus export.

For reference, see the RabbitMQ metric: `rabbitmq_global_messages_redelivered_total`

Best Regards, Felix

kickster97:
Adds the module GlobalCounters, keeping track of global stats that should persist when deleting a vhost. 

Fixes #916 
 

deliver some messages to a client and observe the metrics are still present when deleting the vhost.

### DIFF
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -88,6 +88,7 @@ module LavinMQ
             writer = PrometheusWriter.new(context.response, prefix)
             overview_broker_metrics(vhosts, writer)
             overview_queue_metrics(vhosts, writer)
+            global_metrics(writer)
             custom_metrics(writer)
             gc_metrics(writer)
           end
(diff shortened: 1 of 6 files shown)


src/lavinmq/http/controller/prometheus.cr
```diff
@@ -152,8 +153,19 @@ module LavinMQ
                       value:  1,
                       help:   "System information",
                       labels: {
+                        "#{writer.prefix}_node"                 => System.hostname,
+                        "#{writer.prefix}_cluster"              => System.hostname,
```
viktorerlingsson, line 157: Oh, this should probably be fixed 🤔

viktorerlingsson: or removed

viktorerlingsson:
I think it would be good to add a spec for at least one global counter that deletes a vhost and makes sure the counter is correct. 

Other than that I think it looks good!

snichme (changes requested):
I get the idea of the PR but I think we should do this in another way. Now, for example, `global_messages_delivered_total` will only hold the number of delivered messages for deleted vhosts which is rarely a metric we are interested in. So this leads to the need to sum up the global metric with all the active vhosts. 

Either rename `GlobalCounters` to `DeletedVhostCounters` or have the global metrics count both active vhosts and deleted ones. Either update the metric on each event or reuse some rate fiber to update these. 
The current solution will probably work for now but will be difficult to understand and make use of later on.

kickster97:
Converting this to a draft, might be good to do more extensive refactor instead of adding this for a small part of metrics. 

A more centralized metrics collection might be good if we want global metrics

Issue #971, opened 2025-03-03
MQTT scenarios for lavinmqperf


kickster97:
**Is your feature request related to a problem? Please describe.**
To be able to performance test MQTT connections and throughput 

**Describe the solution you'd like**
You should be able to specify what protocall you want to test, and lavinmqperf should create a client and run tests accordingly

**Describe alternatives you've considered**
you could configure a port, or just specify protocol as a string "amqp" or "mqtt"

carlhoerberg:
Good input @annieblomgren, so yes makes sense to keep it one binary @kickster97. should then break it up to several files, in the `lavinmqperf` sub directory

kickster97:
added with #983

PR #983, opened 2025-03-12
Add mqtt to lavinmqperf

kickster97:
This PR abstracts LavinMQPerf into AMQP and MQTT modules. 
lavinmqperf can now be run for both AMQP and MQTT by specifying the protocol in the cli like this: 

```
lavinmqperf [protocol] [throughput | bind-churn | queue-churn | connection-churn | connection-count | queue-count] [arguments]
```
e.g. 

`lavinmqperf mqtt throughput`

It adds the throughput test for the MQTT module, largely similar to the AMQP throughput test. 

MQTT-specific features:
  - QoS levels
  - Retain flag for messages
  - Clean session option

addresses #971 

```bash
lavinmqperf mqtt throughput 
```


snichme:
Is there a way to run the throughput test using non-durable queues?

carlhoerberg:
:scream: 
```
✗ bin/lavinmqperf mqtt throughput -y 300 -x 100
Publish rate: 0 msgs/s Consume rate: 25063 msgs/s
Publish rate: 0 msgs/s Consume rate: 25941 msgs/s
Publish rate: 0 msgs/s Consume rate: 27455 msgs/s
Publish rate: 0 msgs/s Consume rate: 31092 msgs/s
Publish rate: 0 msgs/s Consume rate: 2382 msgs/s
Publish rate: 0 msgs/s Consume rate: 397 msgs/s
Publish rate: 0 msgs/s Consume rate: 3165 msgs/s
Publish rate: 0 msgs/s Consume rate: 1033 msgs/s
Publish rate: 0 msgs/s Consume rate: 3452 msgs/s
Publish rate: 0 msgs/s Consume rate: 10074 msgs/s
Publish rate: 0 msgs/s Consume rate: 12846 msgs/s
Publish rate: 0 msgs/s Consume rate: 15797 msgs/s
Publish rate: 0 msgs/s Consume rate: 16273 msgs/s
Invalid memory access (signal 11) at address 0x1060
[0x55e33f940eb9] ?? +94434512604857 in bin/lavinmqperf
[0x55e33f94058b] ?? +94434512602507 in bin/lavinmqperf
[0x7f1790141050] ?? +139739178209360 in /lib64/libc.so.6
[0x55e33f919810] ?? +94434512443408 in bin/lavinmqperf
[0x55e33f941eee] ?? +94434512609006 in bin/lavinmqperf
[0x55e33f943e6c] ?? +94434512617068 in bin/lavinmqperf
[0x55e33f943f6e] ?? +94434512617326 in bin/lavinmqperf
[0x55e33fa6785e] GC_inner_start_routine +94 in bin/lavinmqperf
[0x55e33fa58b98] GC_call_with_stack_base +24 in bin/lavinmqperf
[0x7f1790197ba8] ?? +139739178564520 in /lib64/libc.so.6
[0x7f179021bb8c] ?? +139739179105164 in /lib64/libc.so.6
[0x0] ???
```

didn't get a coredump unfortunately, will try to reproduce again

carlhoerberg:
Maybe this is a crystal 1.16.0 issue, this is the coredump backtrace: 

```
#0  0x0000556fffdc834d in wake_at () at /usr/share/crystal/src/crystal/event_loop/polling/event.cr:31
#1  heap_compare () at /usr/share/crystal/src/crystal/event_loop/polling/event.cr:64
#2  meld () at /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:92
#3  merge_pairs () at /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:130
#4  shift? () at /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:38
#5  run () at /usr/share/crystal/src/crystal/event_loop/timers.cr:35
#6  reschedule () at /usr/share/crystal/src/crystal/scheduler.cr:144
#7  0x0000556fffdf09ee in reschedule () at /usr/share/crystal/src/crystal/scheduler.cr:62
#8  suspend () at /usr/share/crystal/src/fiber.cr:351
#9  run_loop () at /usr/share/crystal/src/crystal/scheduler.cr:150
#10 0x0000556fffdf296c in -> () at /usr/share/crystal/src/crystal/scheduler.cr:150
#11 0x0000556fffdf2a6e in start () at /usr/share/crystal/src/crystal/system/thread.cr:231
#12 thread_proc () at /usr/share/crystal/src/crystal/system/unix/pthread.cr:47
#13 -> () at /usr/share/crystal/src/crystal/system/unix/pthread.cr:22
#14 0x0000556ffff1689e in GC_inner_start_routine ()
#15 0x0000556ffff07bd8 in GC_call_with_stack_base ()
#16 0x00007f8e864f7ba8 in start_thread () from /lib64/libc.so.6
#17 0x00007f8e8657bb8c in __clone3 () from /lib64/libc.so.6
```

carlhoerberg:
With debug symbols compiled in: 
```
#0  0x00005645d3f6ed2f in merge_pairs (self=0x7fe033d0fce0, node=...) at /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:128
        tail = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fdf607ff7b8}}
        a = {type_id = 869402120, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0xf700000001}}
        b = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fde927ff5d8}}
        root = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fdf607ff7b8}}
#1  0x00005645d3f6ebfd in shift? (self=0x7fe033d0fce0) at /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:38
        node = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fe02b1ffdd0}}
#2  0x00005645d3f7fe53 in run (self=0x7fe033d0de70, blocking=true) at /usr/share/crystal/src/crystal/event_loop/timers.cr:35
        fiber = 0x7fe02a190a80
        blocking = true
        buffer = {{events = 1, data = {ptr = 0xb, fd = 11, u32 = 11, u64 = 11}}, {events = 5, data = {ptr = 0xfc00000001, fd = 1, u32 = 1,
              u64 = 1082331758593}}, {events = 1, data = {ptr = 0xa, fd = 10, u32 = 10, u64 = 10}}, {events = 4, data = {ptr = 0x18c00000001, fd = 1,
              u32 = 1, u64 = 1700807049217}}, {events = 4, data = {ptr = 0x17400000001, fd = 1, u32 = 1, u64 = 1597727834113}}, {events = 1, data = {
              ptr = 0xa, fd = 10, u32 = 10, u64 = 10}}, {events = 4, data = {ptr = 0x14400000001, fd = 1, u32 = 1, u64 = 1391569403905}}, {events = 5,
            data = {ptr = 0x18500000001, fd = 1, u32 = 1, u64 = 1670742278145}}, {events = 5, data = {ptr = 0x18c00000001, fd = 1, u32 = 1,
              u64 = 1700807049217}}, {events = 5, data = {ptr = 0x19200000001, fd = 1, u32 = 1, u64 = 1726576852993}}, {events = 5, data = {
              ptr = 0x19500000001, fd = 1, u32 = 1, u64 = 1739461754881}}, {events = 5, data = {ptr = 0x16000000002, fd = 2, u32 = 2, u64 = 1511828488194}}, {
            events = 5, data = {ptr = 0x16700000002, fd = 2, u32 = 2, u64 = 1541893259266}}, {events = 5, data = {ptr = 0x17b00000002, fd = 2, u32 = 2,
              u64 = 1627792605186}}, {events = 5, data = {ptr = 0x18200000002, fd = 2, u32 = 2, u64 = 1657857376258}}, {events = 5, data = {
              ptr = 0x14e00000003, fd = 3, u32 = 3, u64 = 1434519076867}}, {events = 5, data = {ptr = 0x19b00000001, fd = 1, u32 = 1, u64 = 1765231558657}}, {
            events = 5, data = {ptr = 0x19c00000001, fd = 1, u32 = 1, u64 = 1769526525953}}, {events = 5, data = {ptr = 0x19d00000001, fd = 1, u32 = 1,
              u64 = 1773821493249}}, {events = 5, data = {ptr = 0x19e00000001, fd = 1, u32 = 1, u64 = 1778116460545}}, {events = 5, data = {
              ptr = 0x19f00000001, fd = 1, u32 = 1, u64 = 1782411427841}}, {events = 5, data = {ptr = 0x1a000000001, fd = 1, u32 = 1, u64 = 1786706395137}}, {
            events = 5, data = {ptr = 0x1a100000001, fd = 1, u32 = 1, u64 = 1791001362433}}, {events = 5, data = {ptr = 0x1a200000001, fd = 1, u32 = 1,
              u64 = 1795296329729}}, {events = 5, data = {ptr = 0x1a300000001, fd = 1, u32 = 1, u64 = 1799591297025}}, {events = 8221, data = {
              ptr = 0x5a00000000, fd = 0, u32 = 0, u64 = 386547056640}}, {events = 8221, data = {ptr = 0x5d00000000, fd = 0, u32 = 0, u64 = 399431958528}}, {
            events = 8221, data = {ptr = 0x6100000000, fd = 0, u32 = 0, u64 = 416611827712}}, {events = 8221, data = {ptr = 0x6200000000, fd = 0, u32 = 0,
              u64 = 420906795008}}, {events = 8221, data = {ptr = 0x6900000000, fd = 0, u32 = 0, u64 = 450971566080}}, {events = 8221, data = {
              ptr = 0x6d00000000, fd = 0, u32 = 0, u64 = 468151435264}}, {events = 8221, data = {ptr = 0x7100000000, fd = 0, u32 = 0, u64 = 485331304448}}, {
            events = 8221, data = {ptr = 0x7400000000, fd = 0, u32 = 0, u64 = 498216206336}}, {events = 8221, data = {ptr = 0x7900000000, fd = 0, u32 = 0,
              u64 = 519691042816}}, {events = 8221, data = {ptr = 0x7c00000000, fd = 0, u32 = 0, u64 = 532575944704}}, {events = 8221, data = {
              ptr = 0x8000000000, fd = 0, u32 = 0, u64 = 549755813888}}, {events = 8221, data = {ptr = 0x8400000000, fd = 0, u32 = 0, u64 = 566935683072}}, {
            events = 8221, data = {ptr = 0x8a00000000, fd = 0, u32 = 0, u64 = 592705486848}}, {events = 8221, data = {ptr = 0x8d00000000, fd = 0, u32 = 0,
              u64 = 605590388736}}, {events = 8221, data = {ptr = 0x9200000000, fd = 0, u32 = 0, u64 = 627065225216}}, {events = 8221, data = {
              ptr = 0x9400000000, fd = 0, u32 = 0, u64 = 635655159808}}, {events = 8221, data = {ptr = 0x9800000000, fd = 0, u32 = 0, u64 = 652835028992}}, {
            events = 8221, data = {ptr = 0x9e00000000, fd = 0, u32 = 0, u64 = 678604832768}}, {events = 8221, data = {ptr = 0x9a00000000, fd = 0, u32 = 0,
              u64 = 661424963584}}, {events = 8221, data = {ptr = 0xa300000000, fd = 0, u32 = 0, u64 = 700079669248}}, {events = 8221, data = {
              ptr = 0xa900000000, fd = 0, u32 = 0, u64 = 725849473024}}, {events = 8221, data = {ptr = 0xad00000000, fd = 0, u32 = 0, u64 = 743029342208}}, {
            events = 8221, data = {ptr = 0xaa00000000, fd = 0, u32 = 0, u64 = 730144440320}}, {events = 8221, data = {ptr = 0xb400000000, fd = 0, u32 = 0,
              u64 = 773094113280}}, {events = 8221, data = {ptr = 0xb900000000, fd = 0, u32 = 0, u64 = 794568949760}}, {events = 8221, data = {
              ptr = 0xbd00000000, fd = 0, u32 = 0, u64 = 811748818944}}, {events = 8221, data = {ptr = 0xc000000000, fd = 0, u32 = 0, u64 = 824633720832}}, {
            events = 8221, data = {ptr = 0xc600000000, fd = 0, u32 = 0, u64 = 850403524608}}, {events = 8221, data = {ptr = 0xc500000000, fd = 0, u32 = 0,
              u64 = 846108557312}}, {events = 8221, data = {ptr = 0xcc00000000, fd = 0, u32 = 0, u64 = 876173328384}}, {events = 8221, data = {
              ptr = 0xd000000000, fd = 0, u32 = 0, u64 = 893353197568}}, {events = 8221, data = {ptr = 0xd400000000, fd = 0, u32 = 0, u64 = 910533066752}}, {
            events = 8221, data = {ptr = 0xd800000000, fd = 0, u32 = 0, u64 = 927712935936}}, {events = 8221, data = {ptr = 0xdc00000000, fd = 0, u32 = 0,
              u64 = 944892805120}}, {events = 8221, data = {ptr = 0xde00000000, fd = 0, u32 = 0, u64 = 953482739712}}, {events = 8221, data = {
              ptr = 0xe100000000, fd = 0, u32 = 0, u64 = 966367641600}}, {events = 8221, data = {ptr = 0xe500000000, fd = 0, u32 = 0, u64 = 983547510784}}, {
            events = 8221, data = {ptr = 0xeb00000000, fd = 0, u32 = 0, u64 = 1009317314560}}, {events = 8221, data = {ptr = 0xef00000000, fd = 0, u32 = 0,
              u64 = 1026497183744}}, {events = 8221, data = {ptr = 0xf300000000, fd = 0, u32 = 0, u64 = 1043677052928}}, {events = 8221, data = {
              ptr = 0xf500000000, fd = 0, u32 = 0, u64 = 1052266987520}}, {events = 8221, data = {ptr = 0xf900000000, fd = 0, u32 = 0, u64 = 1069446856704}},
          {events = 8221, data = {ptr = 0xfc00000000, fd = 0, u32 = 0, u64 = 1082331758592}}, {events = 8221, data = {ptr = 0xff00000000, fd = 0, u32 = 0,
              u64 = 1095216660480}}, {events = 8221, data = {ptr = 0x10300000000, fd = 0, u32 = 0, u64 = 1112396529664}}, {events = 8221, data = {
              ptr = 0x10500000000, fd = 0, u32 = 0, u64 = 1120986464256}}, {events = 8221, data = {ptr = 0x10b00000000, fd = 0, u32 = 0,
              u64 = 1146756268032}}, {events = 8221, data = {ptr = 0x11000000000, fd = 0, u32 = 0, u64 = 1168231104512}}, {events = 8221, data = {
              ptr = 0x11300000000, fd = 0, u32 = 0, u64 = 1181116006400}}, {events = 8221, data = {ptr = 0x11100000000, fd = 0, u32 = 0,
              u64 = 1172526071808}}, {events = 22085, data = {ptr = 0x5645d42634e8 <__pthread_register_cancel@got.plt>, fd = -735693592, u32 = 3559273704,
              u64 = 94857912005864}}, {events = 1705638384, data = {ptr = 0x2bbef9f000007ffd, fd = 32765, u32 = 32765, u64 = 3152231598393425917}}, {
            events = 32736, data = {ptr = 0x7fe0347b101d <_dl_fixup+269>, fd = 880480285, u32 = 880480285, u64 = 140600929882141}}, {events = 5, data = {
              ptr = 0x0, fd = 0, u32 = 0, u64 = 0}}, {events = 0, data = {ptr = 0x7fe033f941a8, fd = 871973288, u32 = 871973288, u64 = 140600921375144}}, {
            events = 870818224, data = {ptr = 0x7fe0, fd = 32736, u32 = 32736, u64 = 32736}}, {events = 0, data = {ptr = 0x7fe02bbefdc0, fd = 733937088,
              u32 = 733937088, u64 = 140600783338944}}, {events = 733939392, data = {ptr = 0xffffff7000007fe0, fd = 32736, u32 = 32736,
              u64 = 18446743455234293728}}, {events = 4294967295, data = {ptr = 0x0, fd = 0, u32 = 0, u64 = 0}}, {events = 1705638384, data = {
              ptr = 0x2bbefe1000007ffd, fd = 32765, u32 = 32765, u64 = 3152236133878890493}}, {events = 32736, data = {
              ptr = 0x7fe0347b365e <_dl_runtime_resolve_xsavec+126>, fd = 880490078, u32 = 880490078, u64 = 140600929891934}}, {events = 0, data = {
              ptr = 0x33e856d500000000, fd = 0, u32 = 0, u64 = 3740334963359219712}}, {events = 32736, data = {ptr = 0x4cc3b26e1a6517ee, fd = 442832878,
              u32 = 442832878, u64 = 5531460953300080622}}, {events = 0, data = {ptr = 0x2bbefe1000000000, fd = 0, u32 = 0, u64 = 3152236133878857728}}, {
            events = 32736, data = {ptr = 0x7fe033d41000, fd = 869535744, u32 = 869535744, u64 = 140600918937600}}, {events = 0, data = {ptr = 0x0, fd = 0,
              u32 = 0, u64 = 0}}, {events = 0, data = {ptr = 0x0, fd = 0, u32 = 0, u64 = 0}}, {events = 0, data = {ptr = 0x2bbefa5800000000, fd = 0, u32 = 0,
              u64 = 3152232045069991936}}, {events = 32736, data = {ptr = 0x5645d3f89d50 <->>, fd = -738681520, u32 = 3556285776, u64 = 94857909017936}}, {
--Type <RET> for more, q to quit, c to continue without paging--
            events = 0, data = {ptr = 0x2000000000, fd = 0, u32 = 0, u64 = 137438953472}}, {events = 0, data = {ptr = 0x7fe02bbf06c0, fd = 733939392,
              u32 = 733939392, u64 = 140600783341248}}, {events = 3560143616, data = {ptr = 0x33d13a8000005645, fd = 22085, u32 = 22085,
              u64 = 3733829887497098821}}, {events = 32736, data = {ptr = 0x7fe033d13a80, fd = 869350016, u32 = 869350016, u64 = 140600918751872}}, {
            events = 1, data = {ptr = 0xd41ae0a300000000, fd = 0, u32 = 0, u64 = 15283775276119490560}}, {events = 22085, data = {ptr = 0x7fe033d13a80,
              fd = 869350016, u32 = 869350016, u64 = 140600918751872}}, {events = 32, data = {ptr = 0x2bbefe1000000000, fd = 0, u32 = 0,
              u64 = 3152236133878857728}}, {events = 32736, data = {ptr = 0x7fe02bbf06c0, fd = 733939392, u32 = 733939392, u64 = 140600783341248}}, {
            events = 4294967152, data = {ptr = 0xd3eefe0effffffff, fd = -1, u32 = 4294967295, u64 = 15271422726837895167}}, {events = 22085, data = {
              ptr = 0x20, fd = 32, u32 = 32, u64 = 32}}, {events = 32, data = {ptr = 0x2000000000, fd = 0, u32 = 0, u64 = 137438953472}}, {events = 0,
            data = {ptr = 0x133d13a98, fd = 869350040, u32 = 869350040, u64 = 5164317336}}, {events = 32, data = {ptr = 0xd41d061000000000, fd = 0, u32 = 0,
              u64 = 15284379376154574848}}, {events = 22085, data = {ptr = 0x55555554, fd = 1431655764, u32 = 1431655764, u64 = 1431655764}}, {
            events = 1179670597, data = {ptr = 0x2000005645, fd = 22085, u32 = 22085, u64 = 137438975557}}, {events = 0, data = {
              ptr = 0x5645d3f6faed <malloc+93>, fd = -738788627, u32 = 3556178669, u64 = 94857908910829}}, {events = 4, data = {ptr = 0x33d0fc800000082b,
              fd = 2091, u32 = 2091, u64 = 3733761717776156715}}, {events = 32736, data = {ptr = 0x7fe033d13a80, fd = 869350016, u32 = 869350016,
              u64 = 140600918751872}}, {events = 3556178162, data = {ptr = 0x2b1ffff000005645, fd = 22085, u32 = 22085, u64 = 3107483674166187589}}, {
            events = 32736, data = {ptr = 0x564500000000, fd = 0, u32 = 0, u64 = 94854352732160}}, {events = 0, data = {ptr = 0x400000004, fd = 4, u32 = 4,
              u64 = 17179869188}}, {events = 0, data = {ptr = 0x7fe000000000, fd = 0, u32 = 0, u64 = 140600049401856}}, {events = 0, data = {
              ptr = 0x33d13a8000000000, fd = 0, u32 = 0, u64 = 3733829887497076736}}, {events = 32736, data = {ptr = 0x5645d41d4730, fd = -736278736,
              u32 = 3558688560, u64 = 94857911420720}}, {events = 5, data = {ptr = 0xd3f2370f00000005, fd = 5, u32 = 5, u64 = 15272329823930810373}}, {
            events = 22085, data = {ptr = 0x500000000, fd = 0, u32 = 0, u64 = 21474836480}}, {events = 3556241734, data = {ptr = 0x33d0fc8000005645,
              fd = 22085, u32 = 22085, u64 = 3733761717776176709}}, {events = 32736, data = {ptr = 0x100000005, fd = 5, u32 = 5, u64 = 4294967301}}, {
            events = 3559276380, data = {ptr = 0x2a10df0800005645, fd = 22085, u32 = 22085, u64 = 3031167774673098309}}, {events = 32736, data = {
              ptr = 0x500000005, fd = 5, u32 = 5, u64 = 21474836485}}, {events = 3555866383, data = {ptr = 0x33d3ef2800000000, fd = 0, u32 = 0,
              u64 = 3734591471098003456}}, {events = 32736, data = {ptr = 0x5645d3f26110 <set+128>, fd = -739090160, u32 = 3555877136,
              u64 = 94857908609296}}, {events = 4, data = {ptr = 0x500000001, fd = 1, u32 = 1, u64 = 21474836481}}, {events = 0, data = {
              ptr = 0x7fe033d3ef28, fd = 869527336, u32 = 869527336, u64 = 140600918929192}}}
        epoll_events = {size = 1, read_only = false, pointer = 0x7fe02bbef5f8}
        timer_triggered = true
        i = 0
        epoll_event = 0x7fe02bbef5f8
        __temp_1037 = 11
        i = 1
        fiber = 0x7fe02a190a80
        epoll_event = 0x7fe02bbef5f8
        index = {data = 1082331758593}
        events = 5
        pd = 0x7fe033d20720
        index = {data = 1082331758593}
        entry = 0x7fe033d20720
        index = {data = 1082331758593}
        entry = {type_id = 503, union = {Pointer(Crystal::EventLoop::Polling::Arena::Entry(Crystal::EventLoop::Polling::PollDescriptor)) = 0x7fe033d20720}}
        event = 0x3feccccccccccccd
        event = 0x0
        node = 0x404b000000000000
        node = 0x7fe02bbefe10
        _next = 0x1
        fiber = 0x0
        event = 0x7fe033cc6600
        __temp_1042 = false
        event = 0x0
        fiber = 0x0
        __temp_1044 = false
        event = 0x0
        fiber = 0x0
        __temp_1045 = false
        event = 0x7fdf6a7ff5d8
        event = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fdf6a7ff5d8}}
        fiber = 0x7fe02a190a80
        __temp_1046 = false
        event = 0x0
        fiber = 0x0
        __temp_1047 = true
        fiber = 0x0
        timer_triggered = true
        buffer = {0x0 <repeats 11 times>, 0xa8c46505853, 0x207, 0xa88, 0x0, 0x0, 0x0, 0x203, 0x0 <repeats 110 times>}
        size = 0
        __temp_1049 = false
        event = 0x0
        __temp_1048 = {[0] = 227483, [1] = 739632449}
        seconds = 227483
--Type <RET> for more, q to quit, c to continue without paging--
        nanoseconds = 739632449
        now = {seconds = 227483, nanoseconds = 739632449}
        event = {type_id = 424, union = {Pointer(Crystal::EventLoop::Polling::Event) = 0x7fe02b1ffdd0}}
        event = 0x0
        i = -1
        i = 0
        fiber = 0x0
        event = 0x7fe033e2731d <sigsuspend+93>
        fiber = 0x7fe033d13a80
        __temp_1055 = 22085
        select_action = 0x0
        pd = 0x0
        pd = 0x0
#3  0x00005645d3f2623c in reschedule (self=0x7fe033d3ef00) at /usr/share/crystal/src/crystal/scheduler.cr:144
        runnable = 0x0
        section = Sched
        operation = 0x5645d41d3f30
        time = 869347040
        metadata = {<No data fields>}
#4  0x00005645d3f261c7 in reschedule () at /usr/share/crystal/src/crystal/scheduler.cr:62
No locals.
#5  0x00005645d3f26196 in suspend () at /usr/share/crystal/src/fiber.cr:351
No locals.
#6  0x00005645d3f890fa in run_loop (self=0x7fe033d3ef00) at /usr/share/crystal/src/crystal/scheduler.cr:150
        runnable = 0x0
        section = Sched
        operation = 0x5645d41d48c0
        time = 869388256
        metadata = {<No data fields>}
#7  0x00005645d3f8dcef in -> (__temp_1318=0x7fe033d12ee0) at /usr/share/crystal/src/crystal/scheduler.cr:150
        scheduler = 0x7fe033d3ef00
        pending = {value = 0}
#8  0x00005645d3f8e09e in start (self=0x7fe033d12ee0) at /usr/share/crystal/src/crystal/system/thread.cr:231
        fiber = 0x7fe033d13b40
        name = 0x7fe033d14f60
        ex = 0x388
#9  0x00005645d3f8dfe0 in thread_proc (data=0x7fe033d12ee0) at /usr/share/crystal/src/crystal/system/unix/pthread.cr:47
        th = 0x7fe033d12ee0
#10 0x00005645d3f8df9e in -> (__temp_1320=0x7fe033d12ee0) at /usr/share/crystal/src/crystal/system/unix/pthread.cr:22
No locals.
#11 0x00005645d41c35de in GC_inner_start_routine ()
No symbol table info available.
#12 0x00005645d41b4918 in GC_call_with_stack_base ()
No symbol table info available.
#13 0x00007fe033e7dba8 in start_thread () from /lib64/libc.so.6
No symbol table info available.
#14 0x00007fe033f01b8c in __clone3 () from /lib64/libc.so.6
No symbol table info available.
```

kickster97:
Hmm not good.. 

forgot to take notes here, so doing it now: last week me & @spuun saw multi-thread issues the throughput tool, we think we have it down to mqtt-client.  
The consuming client disconnects after a while, we could not nail down where the issue was, but it can't be replicated with mosquitto. If we run `read_loop` and `message_loop` in `mqtt_client` with `same_thread: true` we see no issues.

spuun:
> Hmm not good..
> 
> forgot to take notes here, so doing it now: last week me & @spuun saw multi-thread issues the throughput tool, we think we have it down to mqtt-client. The consuming client disconnects after a while, we could not nail down where the issue was, but it can't be replicated with mosquitto. If we run `read_loop` and `message_loop` in `mqtt_client` with `same_thread: true` we see no issues.

I narrowed that down to https://github.com/crystal-lang/crystal/issues/15647

ysbaddaden:
This might be caused by https://github.com/crystal-lang/crystal/issues/15647. With the fix from https://github.com/crystal-lang/crystal/issues/15650 `lavinmqperf` now prints lots of:

```
Unhandled exception in spawn(name: mqtt-client read_loop): BUG: transfering fd=340 to another evloop with pending reader/writer fibers (RuntimeError)
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling/poll_descriptor.cr:30:5 in 'take_ownership'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:413:31 in 'wait_readable'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:318:9 in 'evented_read'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:199:5 in 'read'
  from /home/julien/work/crystal-lang/crystal/src/crystal/system/socket.cr:85:5 in 'system_read'
  from /home/julien/work/crystal-lang/crystal/src/socket.cr:446:5 in 'unbuffered_read'
  from /home/julien/work/crystal-lang/crystal/src/io/buffered.cr:272:5 in 'fill_buffer'
  from /home/julien/work/crystal-lang/crystal/src/io/buffered.cr:62:5 in 'read_byte'
  from lib/mqtt-client/src/mqtt-client/connection.cr:133:15 in 'read_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:131:15 in 'read_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:43:9 in '->'
```

As explained in https://github.com/crystal-lang/crystal/issues/15658#issuecomment-2801312520 this means that a thread's evloop is trying to read or write, while another thread's evloop is already having a pending read or write.

ysbaddaden:
This might be caused by https://github.com/crystal-lang/crystal/issues/15647. With the fix from https://github.com/crystal-lang/crystal/issues/15650 `lavinmqperf` now prints lots of:

```
Unhandled exception in spawn(name: mqtt-client read_loop): BUG: transfering fd=340 to another evloop with pending reader/writer fibers (RuntimeError)
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling/poll_descriptor.cr:30:5 in 'take_ownership'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:413:31 in 'wait_readable'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:318:9 in 'evented_read'
  from /home/julien/work/crystal-lang/crystal/src/crystal/event_loop/polling.cr:199:5 in 'read'
  from /home/julien/work/crystal-lang/crystal/src/crystal/system/socket.cr:85:5 in 'system_read'
  from /home/julien/work/crystal-lang/crystal/src/socket.cr:446:5 in 'unbuffered_read'
  from /home/julien/work/crystal-lang/crystal/src/io/buffered.cr:272:5 in 'fill_buffer'
  from /home/julien/work/crystal-lang/crystal/src/io/buffered.cr:62:5 in 'read_byte'
  from lib/mqtt-client/src/mqtt-client/connection.cr:133:15 in 'read_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:131:15 in 'read_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:43:9 in '->'
```

As explained in https://github.com/crystal-lang/crystal/issues/15658#issuecomment-2801312520 this means that a thread's evloop is trying to read or write, while another thread's evloop is already having a pending read or write.

Or maybe there's an issue with the fix. I'm investigating to understand. :thinking:

ysbaddaden:
The fix exposes a limitation from the "lifetime event loop": we can't share a `fd` across event loops, and I understand from tracing that there are two fibers created: one is dedicated to read from the the `fd` while the other is dedicated to write to the `fd`.

Problem is, those two fibers can't end up in two different event loops at the same time. This is a known limitation of the "lifetime" design.

Due to `preview_mt` spawning fibers into whatever thread and each thread having their own event loop, the fibers end up in different threads, and thus different event loops, and both try to wait on read and write at the same time, so one evloop takes ownership of the `fd` while another evloop crashes because it also tries to take ownership while there's a pending reader.

The situation can be avoided for `preview_mt` with `same_thread: true` to make sure both fibers are on the same thread (and thus the same event loop).

It can also be avoided with a `Fiber::ExecutionContext::MultiThreaded` context that have one event loop per context (not per thread), as long as both fibers are running in the same context.

ysbaddaden:
I started a MT execution context + the patch from https://github.com/crystal-lang/crystal/issues/15647 and I can't reproduce the "can't transfer fd" bug nor the SEGFAULT.

```patch
diff --git a/src/lavinmqperf/mqtt/throughput.cr b/src/lavinmqperf/mqtt/throughput.cr
index 5bff1522..e5363a5f 100644
--- a/src/lavinmqperf/mqtt/throughput.cr
+++ b/src/lavinmqperf/mqtt/throughput.cr
@@ -100,16 +100,18 @@ def initialize
       def run
         super

+        mt = Fiber::ExecutionContext::MultiThreaded.new("MT", maximum: 4)
+
         done = WaitGroup.new(@consumers + @publishers)

         @consumers.times do |i|
-          spawn { reconnect_on_disconnect(done) { consume(i) } }
+          mt.spawn { reconnect_on_disconnect(done) { consume(i) } }
         end

         sleep 1.seconds # Give consumers time to connect

         @publishers.times do |i|
-          spawn { reconnect_on_disconnect(done) { pub(i) } }
+          mt.spawn { reconnect_on_disconnect(done) { pub(i) } }
         end

         if @timeout != Time::Span.zero
```

carlhoerberg:
Ok, we "fixed" mqtt-client.cr and spawned the `message_loop` with `same_thread: true` in the `read_loop` fiber but now I get:

```
Unhandled exception in spawn(name: mqtt-client message_loop): Crystal::EventLoop::Polling::Event#wake_at cannot be nil (NilAssertionError)
  from /usr/share/crystal/src/crystal/event_loop/polling/event.cr:31:3 in 'wake_at'
  from /usr/share/crystal/src/crystal/event_loop/polling/event.cr:64:15 in 'heap_compare'
  from /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:92:8 in 'meld'
  from /usr/share/crystal/src/crystal/pointer_pairing_heap.cr:67:19 in 'delete'
  from /usr/share/crystal/src/crystal/event_loop/timers.cr:54:7 in 'delete'
  from /usr/share/crystal/src/crystal/event_loop/polling.cr:460:34 in 'delete_timer'
  from /usr/share/crystal/src/crystal/event_loop/polling.cr:479:33 in 'run'
  from /usr/share/crystal/src/crystal/scheduler.cr:144:11 in 'reschedule'
  from /usr/share/crystal/src/crystal/scheduler.cr:62:5 in 'reschedule'
  from /usr/share/crystal/src/fiber.cr:351:5 in 'suspend'
  from /usr/share/crystal/src/channel.cr:221:7 in 'receive?'
  from lib/mqtt-client/src/mqtt-client/connection.cr:144:21 in 'message_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:142:7 in 'message_loop'
  from lib/mqtt-client/src/mqtt-client/connection.cr:154:9 in '->'
  from /usr/share/crystal/src/fiber.cr:170:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:105:3 in '->'
  from ???
```

ysbaddaden:
Allright. I'll try to reproduce. It sounds like a `Event#clear` set `@wake_at` to `nil` before removing the event from the timers' structure... which should have been fixed by https://github.com/crystal-lang/crystal/issues/15647

ysbaddaden:
But then you're using a single thread for everything, so you're back to ST.

You should be able to open in a thread/context, do some things (like authentication) then pass the `fd` to another thread/context.

Having the `message_loop` and `read_loop` fibers in the same thread should be enough.

ysbaddaden:
I patched mqtt-client/connection, so both fibers would be on the same thread:

```crystal
spawn same_thread: false do
  spawn read_loop, name: "mqtt-client:read_loop", same_thread: true
  spawn message_loop, name: "mqtt-client:message_loop", same_thread: true
end
```

But I still get `BUG: transfering fd=302 to another evloop with pending reader/writer fibers` :disappointed:

carlhoerberg:
Maybe should have a `write_loop` and a `Channel` with all write requests, instead of a Mutex around the socket write operations

ysbaddaden:
No, it was me not thinking correctly. I took a break for dinner and realized we need the fix (to be released in 1.16.1) for the evloop timers to work correctly with MT, otherwise we get canceled timeouts on the wrong event loop and that leads to the above segfaults and nil wake_at.

kickster97:
For this PR's sake, and because its good to be close to the packets, i'll go with using mqtt-protocol instead of mqtt-clent. 

this change is interesting though, running `lavinmqperf mqtt throughput` (default values) crashes lavinmq with 
`Process terminated because of an invalid memory access`

src/lavinmqperf.cr
```diff
@@ -11,17 +12,28 @@ Signal::SEGV.reset # Let the OS generate a coredump
...
+    when "connection-count" then AMQP::ConnectionCount.new.run
+    when "queue-count"      then AMQP::QueueCount.new.run
+    when /^.+$/             then Perf.new.run([mode.not_nil!])
+    else                         abort Perf.new.banner
+    end
+  when "mqtt"
+    mode = ARGV.shift?
+    case mode
+    when "throughput" then MQTT::Throughput.new.run
+    else                   abort Perf.new.banner
```
snichme, line 35: Different banner needed for mqtt since it doesn't support the same modes

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+          @random_bodies = true
+        end
+        @parser.on("--retain", "Set retain flag on published messages") do
+          @retain = true
+        end
+        @parser.on("--clean-session", "Use clean session") do
+          @clean_session = true
+        end
+        @parser.on("-u uri", "--uri=uri", "MQTT broker URI (default mqtt://localhost:1883)") do |v|
+          @uri = URI.parse(v)
```
snichme, line 78: Not sure `URI.parse` is the best here, it fails quietly when given somethign it doesn't understand. 

For example: 
`-u mqtt://hostname:8822` works but since the protocol isn't used is tempting to write just `-u hostname:8822` but that doesn't work and it defaults to localhost:1883. 

If we want the scheme to matter, one option is to remove the protocol check in LavinMQPerf and only look at what schema is used for the URI. defaulting to `amqp://localhhost` as we did before. This also gives us backwards compatibility.

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+        socket.keepalive = true
+        socket.tcp_nodelay = false
+        socket.sync = true
+        io = LavinMQ::MQTT::IO.new(socket)
+
+        client_id = "#{role}-#{id}"
+        connect_packet = LavinMQ::MQTT::Connect.new(
+          client_id: client_id,
+          clean_session: @clean_session,
+          keepalive: 30_u16,
```
snichme, line 103: Maybe an improvement for later but configure keepalive using query params like `-u mqtt://hostname?keepalive=10`

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+
+      def run
+        super
+
+        done = WaitGroup.new(@consumers + @publishers)
+        @consumers.times do |i|
+          spawn { reconnect_on_disconnect(done) { consume(i) } }
+        end
+
+        sleep 1.seconds # Give consumers time to connect
```
snichme, line 133:
```suggestion
        sleep 1.second # Give consumers time to connect
```

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+               response.as(LavinMQ::MQTT::Connack).return_code == LavinMQ::MQTT::Connack::ReturnCode::Accepted
+          socket.close rescue nil
+          raise "Failed to connect: #{response.inspect}"
+        end
+
+        {socket, io}
+      rescue ex : Exception
+        socket.try &.close rescue nil
+        puts "Error creating connection: #{ex.message}"
+        raise ex
```
snichme, line 122: We log and raise here, catch it in `consume` or `pub`  which logs and raises which in the end will crash the fiber with a stacktrace. 
Should be maybe just not rescue it and in the spawn block rescue any error and print the `e.message` 

What I'm thinking is, is there an idea to how to handle exceptions or have we just added rescue when needed? :)

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+            end
+          end
+        end
+
+        LavinMQ::MQTT::Disconnect.new.to_io(io) if socket && !socket.closed?
+      rescue ex : Exception
+        puts "Publisher error: #{ex.message}"
+        raise ex
+      ensure
+        socket.try &.close rescue nil
```
snichme, line 254: this is also done in `create_client` do we need it in both places?

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+        topic_filter = LavinMQ::MQTT::Subscribe::TopicFilter.new(@topic, @qos.to_u8)
+        LavinMQ::MQTT::Subscribe.new([topic_filter], packet_id: 1_u16).to_io(io)
+
+        suback = LavinMQ::MQTT::Packet.from_io(io)
+        unless suback.is_a?(LavinMQ::MQTT::SubAck)
+          raise "Expected SUBACK but got #{suback.inspect}"
+        end
+
+        done_channel = Channel(Nil).new(1)
+        spawn do
```
snichme, line 274: Why `spawn` here when we do it in `run` for the whole method?

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -0,0 +1,345 @@
...
+
+        LavinMQ::MQTT::Disconnect.new.to_io(io) if socket && !socket.closed?
+      rescue ex : Exception
+        puts "Consumer error: #{ex.message}"
+        raise ex
+      ensure
+        socket.try &.close rescue nil
+      end
+
+      private def reconnect_on_disconnect(done, &)
```
snichme, line 333:
```suggestion
      private def rerun_on_exception(done, &)
```

maybe more accurate name

snichme (commented):
Has anything changed in the amqp perf tools or are they just moved?

src/lavinmqperf/mqtt/throughput.cr
kickster97: Huh, TIL! didn't know there was a `.second`, wonder what the difference from `.seconds` is 😮

kickster97:
> Has anything changed in the amqp perf tools or are they just moved?

they are just moved 👍

src/lavinmqperf/mqtt/throughput.cr
kickster97: If i understand correctly, then you would always need to specify a uri if you want to run lavinmqperf with mqtt? 
so just running locally that would be
for amqp: lavinmqperf throughput
but for mqtt: lavinmqperf throguhtput --uri=mqtt://localhost:1883

kickster97: I took inspo from the omq rabbitmq client 
https://github.com/cloudamqp/lavinmq/issues/971#issuecomment-2695720805

src/lavinmqperf/mqtt/throughput.cr
kickster97: Haha yes, no plan just vibecoding ^^ you're right! i'll clean this up

kickster97: check out commit: f0cd62819a1523953f2c6fe29e2926294499dcf2
is this what you mean? :)

kickster97: Or rather we need to do it like in amqp, where reconnect_on_disconnect handles everything: 68a56c0f17d38cab51582dec8b3c680798d65775

src/lavinmqperf/mqtt/throughput.cr
kickster97: agree.. copied over from amqp throughput. I'll change both!

src/lavinmqperf/mqtt/throughput.cr
snichme: Yes, Both the commits are great improvments and having it in `reconnect_on_disconnect` makes sense. You just have to remember to set the message on the exception but that is no problem.

src/lavinmqperf/mqtt/throughput.cr
snichme: Yes, that would be the result. 
Benefits are that this change will be backward compatible but since you already discussed this I'm fine with the current change. Makes it nice to run it locally.

snichme: But back to the issue, URI.parse since it fails silently. 
I like that when I enter a badly formatted argument, the program tells me whats wrong instead of using defaults and just continue. Is there a `URI.validate` or something we can use? 
Or just raise on line 88 instead of `|| "localhost"`  to tell that something is wrong.

kickster97: yeah I get what you mean.. I'm not sure either what is best. First i thought its not really something we integrate anywhere, more of a local tool, but now that we use it with the benchmarking terraform tool its a different story

carlhoerberg:
```
 u = URI.parse "localhost:1883"
 => #<URI:0x7f799d52f9c0 @scheme="localhost", @host=nil, @port=nil, @path="1883", @query=nil, @user=nil, @password=nil, @fragment=nil>
```

you can have some logic when `u.host == u.port == nil` maybe?

carlhoerberg:
```crystal 
if u.host == u.port == nil
  u.host = u.scheme
  u.scheme = "mqtt"
  u.port = u.path.to_i
  u.path = nil
end

kickster97: I added a conditional reformatting if there is no uri host/port, so now it should help a user when they enter `host:port` 
I alos added a log line printing out what uri the connection has used so its clear, idk @mange if you think thats enough or do you still think it should raise instead of defaulting to localhost?

kickster97: I added a conditional reformatting if there is no uri host/port, so now it should help a user when they enter `host:port` 
I alos added a log line printing out what uri the connection has used so its clear, idk @snichme  if you think thats enough or do you still think it should raise instead of defaulting to localhost?

kickster97:
I got some strange errors after rebasing and adopting execution context, the consuming client would randomly disconnect itself without catching any errors, and LavinMQ would either log "connection reset by peer.." or nothing at all. 

Me & @spuun looked at this today, and this seems to not happen when I use two separate execution contexts for the consumer and publisher. One execution context should be ok here (?), so we are a bit confused because using one causes the errors (at least to happen more frequently, with 2 ECs we don't see errors even if we keep it going for minutes)

kickster97:
> Ok, yes, that's why i did `wg.done rescue nil` in the amqp throughput. Needed when the connection is reconnected.

I get this for the amqp throughput thats merged to main as well. If I start lavinmq and lavinmqperf and then stop lavinmq and start it again, lavinmqperf crashes

PR #990, opened 2025-03-14
Always read cluster secret from etcd

### LINKED ISSUE #981: Followers failed to authenticate
We've seen followers failing to authenticate when connecting to the leader. For some reason the secret the leader is using is different from the secret in etcd.

In `Clustering::Server` we store the secret in an instance variable `@password`. For some reason this seems to be out of sync with the value stored in etcd. There's probably no reason to have this instance variable, instead we should always read the value from etcd.

spuun:
Fixes #981

Run specs

### DIFF
--- a/spec/clustering/server_spec.cr
+++ b/spec/clustering/server_spec.cr
@@ -5,6 +5,18 @@ require "../../src/lavinmq/clustering/server"
 describe LavinMQ::Clustering::Server do
   add_etcd_around_each
 
+  describe "#password" do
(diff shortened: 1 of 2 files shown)


spuun:
I should add that I'm sure that it fixes #981, but most likely.

carlhoerberg:
That's very odd, who/what changes the password? Is there some kind of race condition here? Should we use a lease when we set the password instead so only the leader can update it?

Issue #991, opened 2025-03-16
MQTT binds to 127.0.0.1 despite the -b 0.0.0.0


tomekit:
**Is your feature request related to a problem? Please describe.**
When LavinMQ is started: `lavinmq -b 0.0.0.0`, both AMQP and HTTP use `0.0.0.0` bind as expected, however MQTT seem to be sticking to `127.0.0.1`: 
```
2025-03-16T09:00:33.872213Z  INFO lmq.server Listening for MQTT on 127.0.0.1:1883
```

It seems that this is inline with the `-h`
```
-b BIND, --bind=BIND             IP address that both the AMQP and HTTP servers will listen on (default: 127.0.0.1)
```

however it doesn't seem to be inline with what website describes: https://lavinmq.com/documentation/mqtt-in-lavinmq

```
New Configuration Variables

bind 	IP address that AMQP, MQTT, and HTTP servers will listen on.
```

As you can see `bind` also applies to MQTT as well.
Please note that information on website is under the `Clustering with MQTT`, whereas I am running standalone.

**Describe the solution you'd like**
Provide a way to bind MQTT to different interface than `127.0.0.1`

**Potential options**
a) Modify `-b` behavior, so it also applies to MQTT,
b) Introduce: `--mqtt-bind=BIND` in a similar way to existing: `--amqp-bind=BIND` and `--http-bind=BIND`

tomekit:
The `mqtt` bind to `127.0.0.1` is hardcoded here:
```
https://github.com/cloudamqp/lavinmq/blob/2ff1f8f9bce3b8d940f3f18015d32c32c240bf4b/src/lavinmq/config.cr#L21
```

and not included in `-b` behavior
https://github.com/cloudamqp/lavinmq/blob/2ff1f8f9bce3b8d940f3f18015d32c32c240bf4b/src/lavinmq/config.cr#L89

I am wondering if there was any reason for that.

In our case, `127.0.0.1` would be normally fine if it was running on the host OS, but since we're using containerized environment we can't "expose" this to the host OS, as it's listening locally inside the container.

kickster97:
Hi Tom, thanks for reporting! Your right, the `-b` flag does not set the `mqtt_bind` config, it probably should. We'll patch that! 

The `mqtt_bind` is set to `127.0.0.1` by default (what you reference as hardcoded), you can configure it by changing the config value in `lavinmq.ini`. 
Read more about the config file here: https://lavinmq.com/documentation/configuration-files

PR #992, opened 2025-03-16
let -b also set the mqtt_port

kickster97:
let @mqtt_port be set by the -b port flag. Also introduces the --mqtt_bind flag.

Adresses #991 

Start lavinmq while setting the -b flag, or --mqtt_bind


src/lavinmq/config.cr
```diff
@@ -89,6 +89,7 @@ module LavinMQ
         p.on("-b BIND", "--bind=BIND", "IP address that both the AMQP and HTTP servers will listen on (default: 127.0.0.1)") do |v|
```
carlhoerberg, line 89: Edit this text too

PR #996, opened 2025-03-17
Drop msg on requeue if above delivery_limit

### LINKED ISSUE #976: Message dropped on delivery not requeue when using `x-delivery-limit`
When a message has been re(delivered) for more than `x-delivery-limit` times it should be dropped. RabbitMQ seems to do this on requeue while LavinMQ does it on the next delivery attempt.

It doesn't make sense to requeue the message if it won't be delivered, instead it should be dropped on requeue.

One implication of this is that the message may stay in the queue instead of being dead-lettered at once.

viktorerlingsson:
Drops messages directly on requeue if they have been redelivered more than `x-delivery-limit` times, instead of waiting for the next time we try to deliver them. Also drops messages when a new policy is applied and removes the check on delivery-count on delivery. 

Fixes #976 

Run specs

### DIFF
--- a/spec/policies_spec.cr
+++ b/spec/policies_spec.cr
@@ -204,6 +204,25 @@ describe LavinMQ::VHost do
     end
   end
 
+  it "should drop messages if above delivery-limit" do
+    with_amqp_server do |s|
+      defs = {"delivery-limit" => JSON::Any.new(0_i64)} of String => JSON::Any
+      with_channel(s) do |ch|
+        args = AMQP::Client::Arguments.new
+        args["x-delivery-limit"] = 5
+        q = ch.queue("delivery-limit", exclusive: true, args: args)
+        q.publish_confirm "m1"
+        q.publish_confirm "m2"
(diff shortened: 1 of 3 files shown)


carlhoerberg:
Should we drop the check on delivery then?

viktorerlingsson:
> Should we drop the check on delivery then?

Yeah I guess we can. I was thinking we needed to keep it to handle if the delivery_limit was changed by a policy, but that's not allowed, so I think it should be safe to remove.

viktorerlingsson:
> > Should we drop the check on delivery then?
> 
> Yeah I guess we can. I was thinking we needed to keep it to handle if the delivery_limit was changed by a policy, but that's not allowed, so I think it should be safe to remove.

With the changes in #1000 we need to either keep the check on delivery or drop on apply policy. Maybe we keep it like it is for now and handle dropping on apply policy in another PR?

spuun (approved):

PR #997, opened 2025-03-17
dont publish to dockerhub for PR's or for release candidate versions

### LINKED ISSUE #985: Don't push rc-versions to 'latest' tag on dockerhub
The `latest` tag is the default tag that is pulled when you do docker `docker pull cloudamqp/lavinmq`, we should only have stable versions on latest. Instead we should publish rc-versions to a specific rc tag

kickster97:
Removes pushing to the main branch as a trigger for this workflow, now it will only be for tags 
Also specifies that only versions that follow "v1.2.3", and not e.g. "v.1.2.3-rc.1" will be tagged as latest

Fixes #985

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,8 +12,6 @@ on:
       - views/**
       - openapi/**
   push:
-    branches:
-      - main
     tags:
       - 'v*'
     paths:
@@ -30,6 +28,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   main:
+    if: "!contains(github.ref, '-rc')" # Skip the job if the tag includes '-rc'
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -44,8 +43,7 @@ jobs:
         with:
           images: cloudamqp/lavinmq
           tags: |
-            type=raw,value=latest,enable={{is_default_branch}}
-            type=semver,pattern={{version}}
+            type=semver,pattern={{version}},value=latest
```
carlhoerberg, line 46: Actually, maybe we want to keep the `main` tag still:

```suggestion
            type=raw,enable={{is_default_branch}}
            type=semver,pattern={{version}},value=latest
```

.github/workflows/docker.yml
```diff
@@ -12,8 +12,6 @@ on:
       - views/**
       - openapi/**
   push:
-    branches:
-      - main
     tags:
       - 'v*'
```
carlhoerberg, line 16: with this you don't need the `if contains -rc` 

```suggestion
      - 'v[0-9]+\.[0-9]+\.[0-9]+$'
```

carlhoerberg (approved):

PR #1010, opened 2025-03-21
Add Prometheus metrics for `global_message_*` counters 

kickster97:
Introduce reset_stats into the rate_stats macro, so we can reset all counters when a vhost gets deleted. 
This ensures consistency for our variables and allows us to add global variables. 

adresses #916 

create multiple vhosts, and publish so stats go up, then delete one of the vhosts and see that all stats are zeroed

### DIFF
--- a/src/lavinmq/http/controller/vhosts.cr
+++ b/src/lavinmq/http/controller/vhosts.cr
@@ -53,7 +53,8 @@ module LavinMQ
         delete "/api/vhosts/:name" do |context, params|
           refuse_unless_administrator(context, user(context))
           with_vhost(context, params, "name") do |vhost|
-            @amqp_server.vhosts.each_value(&.reset_stats) # Reset all vhosts to maintain accurate metrics after deletion
+            # Reset all vhosts to maintain accurate Prometheus counters
+            @amqp_server.vhosts.each_value(&.reset_stats)
             @amqp_server.vhosts.delete(vhost)
             context.response.status_code = 204
           end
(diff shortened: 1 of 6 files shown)


src/lavinmq/vhost_store.cr
```diff
@@ -42,8 +42,9 @@ module LavinMQ
       vhost
     end
 
-    def delete(name) : Nil
+    def delete(name, &) : Nil
       if vhost = @vhosts.delete name
+        yield vhost
```
carlhoerberg, line 47:
```suggestion
    def delete(name) : Nil
      if vhost = @vhosts.delete name
```

src/lavinmq/http/controller/vhosts.cr
```diff
@@ -53,6 +53,7 @@ module LavinMQ
         delete "/api/vhosts/:name" do |context, params|
           refuse_unless_administrator(context, user(context))
           with_vhost(context, params, "name") do |vhost|
+            @amqp_server.vhosts.each_value(&.reset_stats) # Reset all vhosts to maintain accurate metrics after deletion
```
carlhoerberg:
```suggestion
            # Reset all vhosts to maintain accurate Prometheus counters
            @amqp_server.vhosts.each_value(&.reset_stats)
```

carlhoerberg:
I like the minimal change, but maybe we should go back to either having global counters or we have global counters only for deleted vhosts. That's a bit more involved and adding to another counter should be minimal overhead even though it's in the hot path.

carlhoerberg:
Added counters on the server level for deleted vhosts, I'm afraid that if we made global counters which included all message rates we will have problems with thread synchronization and it would be a very contentious lock.

kickster97:
> Added counters on the server level for deleted vhosts, I'm afraid that if we made global counters which included all message rates we will have problems with thread synchronization and it would be a very contentious lock.

sounds good, similar to the original idéa @viktorerlingsson had 👍

PR #1016, opened 2025-03-25
Dont run specs that LavinMQ doesn't support

snichme:

There are specs in Bunny that LavinMQ doesn't support, this change will stop them from being run in CI

Specs removed: 
* Dont raise PreconditionFailed when changing `x-consumer-timeout`: LavinMQ will raise PreconditionFailed as stated by the AMQP protocol
* Dont raise PreconditionFailed when changing `x-alternate-exchange`: LavinMQ will raise PreconditionFailed as stated by the AMQP protocol
* Specs related to defaulting to classic queue type: LavinMQ doesn't have classic and quorum queues so no need for this

Check CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,8 @@ jobs:
           sed -i /rabbitmq_http_api_client/d Gemfile
           bundle add rabbitmq_http_api_client --version=">= 3" --group=test --require=rabbitmq/http/client
           bundle install
+          sed -i '211,212d' spec/lower_level_api/integration/queue_declare_spec.rb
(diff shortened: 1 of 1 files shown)


viktorerlingsson (approved):
Nice with green checkmarks! ✅ 

Guess this will/might need to be updated if bunny updates their specs? But it looks like we use a specific version of Bunny so things shouldn't break unless we update?

walro:
> Guess this will/might need to be updated if bunny updates their specs? But it looks like we use a specific version of Bunny so things shouldn't break unless we update?

Had the same thought, but locking should make it stable at least. I wonder if an alternative could be to have a patch applied that removes these specs intead, thinking it *may* handle changes to the file a tad bit better but haven't tested its "resilience".

Issue #974, opened 2025-03-06
Allow operator to change prefetch


baelter:
As an operator it would be nice to be able to speed up a slow consumer on the fly by increasing it's prefetch.

**Describe the solution you'd like**
This can be an opt-in feature for clients that send a specific capability during connection negotiation, or a feature that is always available.
A operator can change prefetch through the UI, API or CLI.

May require slight tweak on the client to allow more concurrency and the initial prefetch.

PR #1020, opened 2025-03-27
API endpoint to adjust the consumer prefetch

snichme:

Adds an API endpoint that allows the user to adjust the prefetch of a consumer 

Partial fix for #974 

specs added for the new endpoint


src/lavinmq/http/controller/consumers.cr
```diff
@@ -32,31 +32,48 @@ module LavinMQ
...
+            refuse_unless_management(context, user, vhost)
+            consumer = find_consumer_from_params(user, vhost, params)
             unless consumer
               context.response.status_code = 404
               break
             end
-            consumer.cancel
+            body = parse_body(context)
+            if prefetch = body["prefetch"]?.try(&.as_i?)
+              prefetch = 1 if prefetch < 1
```
carlhoerberg, line 57: I think we should allow 0 too

src/lavinmq/http/controller/consumers.cr
```diff
@@ -32,31 +32,49 @@ module LavinMQ
...
+            consumer = find_consumer_from_params(user, vhost, params)
             unless consumer
               context.response.status_code = 404
               break
             end
-            consumer.cancel
+            body = parse_body(context)
+            if prefetch = body["prefetch"]?.try(&.as_i?)
+              prefetch = 0 if prefetch < 0
+              prefetch = UInt16::MAX if prefetch > UInt16::MAX
```
viktorerlingsson, line 58: Should we abort with an error message instead if prefetch is `< 0` or `> UInt16::MAX` ? I think that's easier to understand for a user

PR #1021, opened 2025-03-27
Adjust consumer prefetch UI

snichme:

Example on how you could edit prefetch 

image

+/- adjust prefetch by 10 for each click. 

In this example the callback has a debounce function if the user clicks a lot of times, only one request is sent in the end. 

Related:  #974 

Manual 

Run this branch, connect a consumer and try it out.

### DIFF
--- a/spec/api/consumers_spec.cr
+++ b/spec/api/consumers_spec.cr
@@ -107,4 +107,35 @@ describe LavinMQ::HTTP::ConsumersController do
       end
     end
   end
+
+  describe "PUT /api/consumers/vhost/connection/channel/consumer" do
+    it "should allow to update the prefetch" do
+      with_http_server do |http, s|
+        with_channel(s) do |ch|
+          ch.prefetch(5)
+          q = ch.queue("")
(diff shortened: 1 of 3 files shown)


carlhoerberg:
Why not an input field? The +/- seems complicated and what if you want to go from 1000 to 2000?

carlhoerberg:
What if when you click the number it changes into an input field?

snichme:
> What if when you click the number it changes into an input field?

Yes, I like that idea. 

This PR was mostly to spark a discussion on how to do it. Didn't want an input field rendered on each row, but toggle the number to an input field is a nice solution.

carlhoerberg:
From offline discussion: Make it possible to edit the prefetch level on the Channel page, replicate the AMQP protocol level: https://github.com/cloudamqp/lavinmq/blob/3622d1a4681603084eadcc180244a033cab1b3ce/src/lavinmq/amqp/channel.cr#L580-L596

snichme:
https://github.com/cloudamqp/lavinmq/pull/1033

PR #1022, opened 2025-04-01
Multiple filter values on messages in stream queues. 

viktorerlingsson:
Adds support for multiple filter values on messages in streams. 

This also changes the behavior when supplying multiple filter values on a consumer to matching all provided filter values with AND. 

Run specs.

### DIFF
--- a/src/lavinmq/amqp/stream_consumer.cr
+++ b/src/lavinmq/amqp/stream_consumer.cr
@@ -133,7 +133,7 @@ module LavinMQ
         return true if @filter.empty?
         if msg_filters = filter_value_from_msg_headers(msg_headers)
           @filter.each do |consumer_filter|
-            return false unless msg_filters.bsearch { |f| f >= consumer_filter } == consumer_filter
+            return false unless msg_filters.find { |f| f == consumer_filter }
           end
           true
         else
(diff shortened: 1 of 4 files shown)


src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -131,15 +131,18 @@ module LavinMQ
...
+          true
         else
           @match_unfiltered
         end
       end
 
-      private def filter_value_from_msg_headers(msg_headers) : String?
-        msg_headers.try &.fetch("x-stream-filter-value", nil).try &.to_s
+      private def filter_value_from_msg_headers(msg_headers) : Array(String)?
+        msg_headers.try &.fetch("x-stream-filter-value", nil).try &.to_s.split(',')
```
carlhoerberg, line 145: Can we do some kind of optimization here? If the consumer only filters on a single value for example? So that we don't have to do a split on each message? Or use the `split` with a block so that no `Array` allocation is done.

carlhoerberg:
Should the readme be updated with this feature?

src/lavinmq/amqp/stream_consumer.cr
viktorerlingsson: Updated with a more efficient `filter_match` method with less allocations

viktorerlingsson:
> Should the readme be updated with this feature?

Yes!

carlhoerberg (approved):

carlhoerberg:
i guess we could still do bsearch, but i doubt it's any faster for the most common case where there are very few elements in `@consumer_filter`.

carlhoerberg:
Yes, it's about 2.25 times slower with bsearch where there are two values and two filters.

PR #1031, opened 2025-04-03
make it possible to filter on connection_name and user from /connections

### LINKED ISSUE #1025: Filter on more fields on /connections in GUI
It's only possible to filter on `name` on connections in the GUI. It would be nice to be able to filter on `user` and `connection-name` as well.

viktorerlingsson:
Make it possible to filter on `connection_name` and `user` from /connections in the GUI.

Fixes #1025 

Manual

### DIFF
--- a/src/lavinmq/http/controller/connections.cr
+++ b/src/lavinmq/http/controller/connections.cr
@@ -17,6 +17,12 @@ module LavinMQ
     class ConnectionsController < Controller
       include ConnectionsHelper
 
+      protected def match_value(value)
+        if client_properties = (value[:client_properties]? || value["client_properties"]?)
+          "#{value[:name]? || value["name"]?} #{client_properties["connection_name"]?} #{value[:user]? || value["user"]?}"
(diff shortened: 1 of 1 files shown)


src/lavinmq/http/controller/connections.cr
```diff
@@ -17,6 +17,12 @@ module LavinMQ
     class ConnectionsController < Controller
       include ConnectionsHelper
 
+      protected def match_value(value)
+        if client_properties = (value[:client_properties]? || value["client_properties"]?)
```
carlhoerberg, line 21: so if there are not client_properties it's not searchable at all?

viktorerlingsson: Yeah, but client_properties should always exist as far as I can tell.

kickster97: maybe not for mqtt connections..

carlhoerberg (approved):

PR #1033, opened 2025-04-04
Update prefetch on channel

snichme:

Introduce a new API endpoint that can update the prefetch for a channel
Adds UI to update the prefetch

image

Clicking on the prefetch count will replace that with a form so you can update the value. 

image

Issue #974

specs and checking the UI (channel page)

### DIFF
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -595,8 +595,8 @@ module LavinMQ
       end
 
       def prefetch_count=(value)
-        @prefetch_count = value
         @consumers.each(&.prefetch_count = value)
+        @prefetch_count = value
       end
 
       def basic_recover(frame) : Nil
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -589,10 +589,14 @@ module LavinMQ
             notify_has_capacity(frame.prefetch_count.to_i - unacked_by_consumers)
           end
         else
-          @prefetch_count = frame.prefetch_count
-          @consumers.each(&.prefetch_count = frame.prefetch_count)
+          self.prefetch_count = frame.prefetch_count
         end
         send AMQP::Frame::Basic::QosOk.new(frame.channel)
+      end
+
+      def prefetch_count=(value)
+        @prefetch_count = value
+        @consumers.each(&.prefetch_count = value)
       end
 
       def basic_recover(frame) : Nil
--- a/src/lavinmq/http/controller/channels.cr
+++ b/src/lavinmq/http/controller/channels.cr
@@ -28,6 +28,19 @@ module LavinMQ
             }).to_json(context.response)
           end
         end
+
+        put "/api/channels/:name" do |context, params|
+          with_channel(context, params) do |channel|
+            body = parse_body(context)
+            if prefetch = body["prefetch"]?.try(&.as_i?)
+              unless 0 <= prefetch <= UInt16::MAX
+                bad_request(context, "prefetch must be between 0 and #{UInt16::MAX}")
+              end
+              channel.prefetch_count = prefetch.to_u16
+            end
+            context.response.status_code = 204
+          end
+        end
       end
 
       private def all_channels(user)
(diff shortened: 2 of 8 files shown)


src/lavinmq/http/controller/channels.cr
```diff
@@ -28,6 +28,19 @@ module LavinMQ
             }).to_json(context.response)
           end
         end
+
+        put "/api/channels/:name" do |context, params|
+          with_channel(context, params) do |channel|
+            body = parse_body(context)
+            if prefetch = body["prefetch"]?.try(&.as_i?)
```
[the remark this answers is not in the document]
carlhoerberg: Ah, yeah, we show it in the UI. But maybe it can come in another PR, it's not used that often.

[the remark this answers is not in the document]
snichme: Agree can be another PR

src/lavinmq/amqp/channel.cr
```diff
@@ -589,12 +589,16 @@ module LavinMQ
...
-          @prefetch_count = frame.prefetch_count
-          @consumers.each(&.prefetch_count = frame.prefetch_count)
+          prefetch_count(frame.prefetch_count)
         end
         send AMQP::Frame::Basic::QosOk.new(frame.channel)
       end
 
+      def prefetch_count(value)
+        @prefetch_count = value
+        @consumers.each(&.prefetch_count = value)
```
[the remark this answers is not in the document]
snichme: Doing this is nicer, but cannot call `prefetch_count = frame.prefetch_count` from the `basic_qos` method above. 
It doesn't say anything just seeing the that consumers doesn't get updated with new prefetch.

snichme:
@bengtmagnus need some help with designing the small form to update prefetch. 
Instead of the Save button, we want two small buttons, one green checkmark to save the change and a red cross to abort the change.

bengtmagnus:
@snichme on it!

bengtmagnus:
@snichme Made some alternatives. Anything along these lines you had in mind?
Skärmavbild 2025-04-08 kl  10 10 59
Skärmavbild 2025-04-08 kl  10 11 05
Skärmavbild 2025-04-08 kl  10 11 11

snichme:
Yes, the last one with all of them on the same row.

static/js/channel.js
```diff
@@ -31,11 +31,76 @@ Table.renderTable('table', consumerTableOpts, function (tr, item, all) {
...
+  const form = document.createElement('form')
+  form.classList.add('prefetch-form')
+  form.addEventListener('submit', (event) => {
+    event.preventDefault()
+    const prefetch = parseInt(input.value)
+    HTTP.request('PUT', channelUrl, { body: { prefetch } })
+      .then((r) => {
+        if (!(r && r.is_error)) {
+          cb(prefetch)
+        }
```
carlhoerberg, line 60: Show any alert if it fails? 

Can `r` be undefined?

static/js/channel.js
```diff
@@ -31,11 +31,76 @@ Table.renderTable('table', consumerTableOpts, function (tr, item, all) {
...
+const prefetchView = (cb) => {
+  const text = document.createElement('div')
+  text.title = 'Click to edit'
+  text.style.cursor = 'pointer'
+  text.addEventListener('click', cb)
+  const updateView = (value) => { text.textContent = Helpers.formatNumber(value) }
+  return { text, updateView }
+}
+
+const prefetchHandler = () => {
```
carlhoerberg, line 76: does this have to be a method?

static/js/channel.js
snichme: There is an automatic alert, triggering an popup, if we get back an error. Handled in http.js 

yes, `r` is undefined when no body is returned.

static/js/channel.js
snichme: no not really, just leftovers from earlier implementations :)

snichme: But on the other hand, having it in a function means that it doesn't add a bunch of variables on the global scope. 

Alternative would be 


```js
let prefetchHandler
{ // create new scope
  const el = ...
  const update = () => {}
  .... 
  prefetchHandler = { el, update }
}
```

But I find that harder to read

PR #1038, opened 2025-04-08
Replace chart.js with uPlot

carlhoerberg:

https://github.com/leeoniya/uPlot


carlhoerberg:
Will vibe code a d3.js implementation instead.

PR #1039, opened 2025-04-09
handle keepalive set to 0 in `MQTT::client`

kickster97:
Disclaimer, i had a hard time finding documentation for this, but I think setting read_timeout with a TimeSpan of 0 does not set an unlimited timer? rather it sets it to fire instantly? Since its a nilable value, we can set it to nil for an unlimited value.

### DIFF
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -52,7 +52,7 @@ module LavinMQ
       private def read_loop
         socket = @socket
         if socket.responds_to?(:"read_timeout=")
-          socket.read_timeout = (@keepalive * 1.5).seconds # 1.5 according to [MQTT-3.1.2-24].
+          socket.read_timeout = @keepalive.zero? ? nil : (@keepalive * 1.5).seconds
         end
         loop do
           @log.trace { "waiting for packet" }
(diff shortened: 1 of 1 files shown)


carlhoerberg:
`read_timeout = 0` is basically like saying `sleep 0`: https://github.com/crystal-lang/crystal/blob/27cefcec77d9bd0e9cbe37c2b9fb414f6733d2cd/src/io/evented.cr#L61-L64

src/lavinmq/mqtt/client.cr
```diff
@@ -52,7 +52,7 @@ module LavinMQ
       private def read_loop
         socket = @socket
         if socket.responds_to?(:"read_timeout=")
-          socket.read_timeout = (@keepalive * 1.5).seconds # 1.5 according to [MQTT-3.1.2-24].
+          socket.read_timeout = @keepalive.zero? ? nil : (@keepalive * 1.5).seconds
```
carlhoerberg, line 55: I think it would be good to keep the comment :)
```suggestion
          # 1.5 times grace period according to [MQTT-3.1.2-24].
          socket.read_timeout = @keepalive.zero? ? nil : (@keepalive * 1.5).seconds
```

Issue #1035, opened 2025-04-08
MQTT: Session tries to write to closed socket


snichme:
**Describe the bug**

In the deliver loop in the mqtt session, when closing the client connection we might end up trying to deliver a packet to that socket anyway causing the send method in `MQTT::Client` to raise a IO::Error. 

We catch that in the [Session](https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/mqtt/session.cr#L44) but only log that and continue try to deliver the packets which sometimes ends up in a forever loop.

Second time the loop runs the packet that LavinMQ tries to deliver is marked as redeliver and then we fail with an ArgumentError here https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/mqtt/session.cr#L47 because if the original message had QoS=0 `dup` is not allowed to be not zero. [source](https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/mqtt/session.cr#L152)
This ArgumentError is caught but only logs and the loop continues

**Describe your setup**

Linux, latest version of LavinMQ 

**How to reproduce**

Found the issue by running `crystal run src/lavinmqperf.cr -- mqtt throughput -y 300 -x 100` from the mqtt-perf branch #983 a couple of times

**Expected behavior**

Not sure, but stop the deliver_loop on IO::Error, maybe not all errors but Broken pipe at least I think should do it.

snichme:
It happened during throughput testing with both 1 consumer and one producer and mulitple.

kickster97:
We requeue QoS 0 messages if the message cannot be delivered..(https://github.com/cloudamqp/lavinmq/blob/0d8055467818055abd6fee761a483aadb57ea827/src/lavinmq/mqtt/session.cr#L122) 

we should be able to just raise and throw the message if i'm not mistaken? 
If we want to requeue the message, we need to override queue methods to not requeue messages for QoS 0 etc..

snichme:
From the specs: 

> The DUP flag MUST be set to 1 by the Client or Server when it attempts to re-deliver a PUBLISH Packet [MQTT-3.3.1.-1]. The DUP flag MUST be set to 0 for all QoS 0 messages [MQTT-3.3.1-2]. 

And QoS=0 means `At most once delivery` so we can either just throw away the message or try again but set DUP=0. 

Setting DUP=0 on redelivery feels like it's wrong, WDYT?

kickster97:
[...]

PR #1040, opened 2025-04-09
Exit `MQTT::Session` deliver loop on `IO::Error` and `Argument::Error`

kickster97:
we log that the deliver loop exits, but then we dont break out of the loop so it will continue to try and deliver the loop. 
Adresses #1035 

Checkout the mqtt-perf branch, and then run lavinmqperf mqtt throughput and kill the program.

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -36,28 +36,26 @@ module LavinMQ
             end
             next
           end
-          consumer = consumers.first.as(MQTT::Consumer)
+          consumer = @consumers.first.as(MQTT::Consumer)
           get_packet do |pub_packet|
             consumer.deliver(pub_packet)
           end
           Fiber.yield if (i &+= 1) % 32768 == 0
-        rescue ::IO::Error
-          @log.error { "deliver loop exited due to IO error" }
-        rescue ArgumentError
-          @log.error { "deliver loop exited due to argument error" }
-        rescue ex
-          @log.error(exception: ex) { "Unexpected error in deliver loop" }
         end
-      rescue ::Channel::ClosedError
-        @log.debug { "deliver loop exited due to channel closed" }
+      rescue ex : ::IO::Error
+        @log.debug(exception: ex) { "deliver loop exited due to IO error" }
+      rescue ex : ::Channel::ClosedError
+        @log.debug(exception: ex) { "deliver loop exited due to channel closed" }
       rescue ex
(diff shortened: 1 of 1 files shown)


snichme:
Should we also address the issue of DUP=1 with QoS=0 since that is what raises the ArgumentError? Or maybe in another PR

kickster97:
> Should we also address the issue of DUP=1 with QoS=0 since that is what raises the ArgumentError? Or maybe in another PR

That can be in another pr :)

src/lavinmq/mqtt/session.cr
```diff
@@ -43,8 +43,10 @@ module LavinMQ
           Fiber.yield if (i &+= 1) % 32768 == 0
         rescue ::IO::Error
           @log.error { "deliver loop exited due to IO error" }
+          break
         rescue ArgumentError
           @log.error { "deliver loop exited due to argument error" }
+          break
         rescue ex
           @log.error(exception: ex) { "Unexpected error in deliver loop" }
```
spuun, line 51: We should probably break here too? Dunno how to recover from something unexpected.

src/lavinmq/mqtt/session.cr
```diff
@@ -43,8 +43,10 @@ module LavinMQ
           Fiber.yield if (i &+= 1) % 32768 == 0
         rescue ::IO::Error
           @log.error { "deliver loop exited due to IO error" }
+          break
         rescue ArgumentError
           @log.error { "deliver loop exited due to argument error" }
```
spuun, line 48: I think the exception message should be included in the error. Maybe even log the entire error?

src/lavinmq/mqtt/session.cr
```diff
@@ -43,8 +43,10 @@ module LavinMQ
           Fiber.yield if (i &+= 1) % 32768 == 0
         rescue ::IO::Error
           @log.error { "deliver loop exited due to IO error" }
```
spuun, line 45: This may be expected, so dunno if we should log this as error? Or use different levels depending on `@closed`?

src/lavinmq/mqtt/session.cr
kickster97: yeah true

src/lavinmq/mqtt/session.cr
kickster97: feel like it should at least be warn, bc it also might be unexpected 
or do you think it can go down to an info?

carlhoerberg:
Should the socket not be closed when the deliver_loop exits?

PR #1042, opened 2025-04-09
Refactor paging and details_tuple to avoid SEGV and improve performance

carlhoerberg:
# Big refactor Controller#page

Improved performance for listing many entities, searching for one queue
among 25000 now uses 70KiB instead of 1.52MiB.

Filter out items that doesn't match before calling the more expensive
`details_tuple`.

Each class that implemented SortableJSON now have to have a
`search_match?`, finding if it maches either a string or regex.

Calculate correct total and filter count (wasn't before).

Sorting is a bit easier to read now and avoid macros (which have the
added benefit that it will be easier to debug if this part is involved
in the segfaults).

# Align MQTT Consumers details_tuple with the AMQP one

A segfault stacktrace indicated that this discrepancy
might been the issue, could not replicate but let's
fix it anyway.

# ConnectionInfo changes

ConnectionInfo is a class (so it can be referenced and don't need to be
copied).

ConnectionInfo#remote_/local_address instead of src/dst, clearer that it
is from the perspective of the server.

At one point we suspected a memory problem with Socket::IPAddress so
created a local replacement.

# Log memory usage and time for each http request

Used when optimizing the paging method

### DIFF
--- a/spec/proxy_protocol_spec.cr
+++ b/spec/proxy_protocol_spec.cr
@@ -8,8 +8,8 @@ describe "ProxyProtocol" do
       w.write "PROXY TCP 1.2.3.4 127.0.0.2 34567 1234\r\n".to_slice
 
       conn_info = LavinMQ::ProxyProtocol::V1.parse(r)
-      conn_info.src.to_s.should eq "1.2.3.4:34567"
-      conn_info.dst.to_s.should eq "127.0.0.2:1234"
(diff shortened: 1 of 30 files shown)


viktorerlingsson (approved):
Changes looks good to me! 
I guess the title is pretty self-explanatory, but always nice with some info in description/comment as well 🙂

carlhoerberg:
This PR morphed into something else, but now it's cleaned up and each commit is independently important.

PR #1043, opened 2025-04-09
let restore_index logging have trace level, instead of info

kickster97:
Some log lines in `restore_index`in `RetainStore` have an unnecessarily high log level, this PR sets it down to `trace`

Also open to removing the lines, wdyt?

Inspect the logs.


snichme (approved):

spuun:
> What's the policy for `trace` vs `debug`?

It's very undefined.

I'd go with debug here. Also include `@dir` in the log messages.

PR #1047, opened 2025-04-14
Don't let `Session` redeliver messages with `QoS=0`

kickster97:
Before, the `get_packet`loop would requeue a message that fails to deliver (setting the DUP-argument to 1), even if it has QoS set to 0. According to the MQTT 3.1.1 spec, this should not be allowed. 

> The DUP flag MUST be set to 1 by the Client or Server when it attempts to re-deliver a PUBLISH Packet [MQTT-3.3.1.-1]. The DUP flag MUST be set to 0 for all QoS 0 messages [MQTT-3.3.1-2].

Adresses #1035


spuun (approved):
I wonder if it's possible to do a spec for this?

PR #1050, opened 2025-04-15
Bugfix: federation upstream exchange doesn't route messages

### LINKED ISSUE #190: Exchange federation configured on rabbitmq with lavinmq as upstream does not create bindings on the temp queue created on the upstream

spuun:
#190 states that bindings are not added to the queue, but the problem was that the federation upstream exchange was a topic exchange and not a fanout exchange.

Bindings are added between the federated exchange and the upstream exchange, and the latter should just forward everything to the queue.
z
Fixes #190


snichme (approved):
Haven't tested, but the change make sense

carlhoerberg (approved):

Issue #1009, opened 2025-03-21
Filter expressions / filtering broker side


filipproch:
With upcoming RabbitMQ support for AMQP Filter Expressions
https://www.rabbitmq.com/blog/2024/12/13/amqp-filter-expressions
I am curious what is LavinMQ stance on this / any plans for support.

**Is your feature request related to a problem? Please describe.**
The problem is we use MQ for change notification distribution. 

We have one queue(stream) per entity type and utilize the bloom filter/single value filter so that we are able to filter them (at least somewhat) based on single value (data for specific customer).

However in several cases (entities that change frequently), further filtering on the broker side would be required to avoid either excessive traffic to backend or the need for secondary non-LavinMQ broker, handling the multicast distribution of filtered messages to backend.

```
FRONTEND <-- ( websocket ) -- BACKEND <-- ( consumer connected to MQ ) -- LavinMQ
```
Where we start a new consumer for each websocket connection.

With current options, we can apply narrow filtering only on BACKEND, still resulting in excessive amount of messages going from LavinMQ -> BACKEND, especially when hundreds of FRONTENDs are connected.

**Describe the solution you'd like**
Have ability to apply more filters on a stream. Either RabbitMQ filter expressions or an alternative solution.

**Describe alternatives you've considered**
- Switching to RabbitMQ (waiting for stable release of filter expressions), plus we like LavinMQ for its potential over RabbitMQ, so would like to stay with LavinMQ if possible
- Single consumer per backend + backend filtering - however this is infeasible due to data volume ( all change data for all customers would unnecessarily reach our backend ), we need to apply at least the currently available bloom filter (allows us to filter data for specific customer, eliminating most of the traffic).
- Custom/own broker service that would consume data with bloom filter, do the narrow filtering and multicast to backend over websocket / however since RabbitMQ will support the filter expressions, this feels like unnecessary complexity we can avoid

viktorerlingsson:
Hi @filipproch 
Not sure if you know yet, but we just added our first version of filtering for streams in [v2.2.0](https://github.com/cloudamqp/lavinmq/releases/tag/v2.2.0). This implementation is pretty similar to how the bloom filters work in RMQ, with the exception that filtering in LavinMQ does not return any false positives. There's some information about [filtering in the docs here](https://lavinmq.com/documentation/server-side-offset-tracking-and-stream-filtering#stream-filtering).

[...]

PR #1053, opened 2025-04-15
Streams - Filtering on headers

viktorerlingsson:
Enhances the stream consumer filtering capabilities in LavinMQ by making it possible to filter messages based on any message header value.

Added `x-filter-match-type` to make filtering more flexible: 
- `all` (default): All consumer filters must match for a message to be delivered
- `any`: At least one consumer filter must match

Consumer filters can be any of the following: 
- `String` (comma-separated), backwards-compatible
- `AMQ::Protocol::Table`
- `Array` of `String` & `AMQ::Protocol::Table`

Addresses #1009 

**Todo:** 
- Update documentation on lavinmq.com

Run specs.


viktorerlingsson:
Just noting: If we merge the Readme-overhaul first we can move the description from readme to contributing.md

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -9,7 +9,8 @@ module LavinMQ
       property segment : UInt32
       property pos : UInt32
       getter requeued = Deque(SegmentPosition).new
-      @consumer_filters = Array(String).new
+      @consumer_filters = Array(Hash(String, String)).new
+      @filter_match_type = "all" # all/any
```
carlhoerberg, line 13: use symbols or enum if there are only two values, or even a bool.

src/lavinmq/amqp/stream_consumer.cr
```diff
@@ -66,10 +68,41 @@ module LavinMQ
...
         when String
-          @consumer_filters = arg.split(',')
+          arg.split(',').each do |f|
+            @consumer_filters << {"x-stream-filter" => f.strip}
+          end
+        when AMQ::Protocol::Table
+          arg.each do |k, v|
+            if k.to_s == "x-stream-filter"
+              v.to_s.split(',').each do |f|
+                @consumer_filters << {k.to_s => f.strip}
```
carlhoerberg, line 78: the hash has ever only one key/value pair? then use a tuple instead

carlhoerberg (changes requested):

carlhoerberg (approved):

PR #1061, opened 2025-04-16
move user and password files to auth

kickster97:
A suggestion to move user & password files into the auth folder, so we can get some structure when we create more user types.

When this is merged, i will create an abstract user class and have a basic_user and a temp_user (this could be broken up into oauth_user, http_user etc. if we need that)

this does not affect functionality, So just compiling should be good.


kickster97:
Yeah, sounds good! The user folder maybe is jumping the gun.. I'm going to want a user folder for when we introduce multiple types if users, like we do with `authenticator` and `authenticators`.

viktorerlingsson:
Looks like there's a reference in lavinmqctl that needs updating
```
In src/lavinmqctl.cr:7:1

 7 | require "./lavinmq/user"
     ^
Error: can't find file './lavinmq/user' relative to '/__w/lavinmq/lavinmq/src/lavinmqctl.cr'
```

viktorerlingsson (approved):

PR #1065, opened 2025-04-17
Run etcd lease keepalive loop in a separate thread

carlhoerberg:
Enable execution context, but only use one Isolated 

`bin/lavinmqperf throughput -y 300 -e amq.fanout -u ""` and monitor that the server isn't loosing the lease.


spuun:
We must make sure we're using Crystal 1.16.1 everywhere for this.

viktorerlingsson:
Tried running this with `bin/lavinmqperf throughput -y 300 -e amq.fanout -u ""` for a while and it seems to work great! 👍 

We should be be running specs with `-Dpreview_mt -Dexecution_context` as well since we're building with those flags, right? 

I tried running specs with `-Dpreview_mt -Dexecution_context`, and [this spec](https://github.com/cloudamqp/lavinmq/blob/etcd-lease-thread/spec/queue_spec.cr#L5-L20) sometimes fails. Changing [this sleep](https://github.com/cloudamqp/lavinmq/blob/etcd-lease-thread/spec/queue_spec.cr#L14) to 50ms seems to fix it. Some timing that changes with new execution context? 
[This spec](https://github.com/cloudamqp/lavinmq/blob/etcd-lease-thread/spec/etcd_spec.cr#L97-L109) also fails sometimes. (And `etcd_spec.cr` won't run at all without `-Dpreview_mt -Dexecution_context`)

Issue #957, opened 2025-02-19
Separate MQTT::Session from AMQP::Queue


kickster97:
**Is your feature request related to a problem? Please describe.**
Currently the MQTT::Session is inheriting AMQP::Queue, forcing some AMQP behaviours and characteristics on the MQTT::Session. This was done as a POC and works fine, but it puts constraints on the session. 
 
**Describe the solution you'd like**
The MQTT::Session should be a stand alone class with its own publish method, that does not need to be tied to a consumer class (Having the Client class should be sufficient for an MQTT connection since you can only have one consumer). 
Overall, breaking out the MQTT::Session would make the MQTT solution more efficient and easier to work with. 

This issue is a WIP

kickster97:
fixed by #1920

PR #1069, opened 2025-04-24
Don't reset the sessions client on Subscribe on existing sessions

kickster97:
We have a bug where when a subscribe is done, the sessions client is re-set, meaning the old consumer(client) is removed, and the new client is added as a consumer on the session. For a clean session this causes issues, because when the consumer pool is empty it will auto delete the session. This means we end up with a variety of bad cases when subscribing to multiple topics when `clean_session = true` for the client. 

We fix this by only re-setting the client on a subscribe if needed, that is: if the session is new and needs to be declared. 

Fixes: #1074

Run MQTT specs
Run with and without changes, and try to subscribe to multiple topics with a client that has `clean_session = true` and see that without the changes you end up with only one topic subscription, and with the changes you have multiple.

### DIFF
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -75,8 +75,7 @@ module LavinMQ
       end
 
       def subscribe(client, topics)
-        session = sessions[client.client_id]? || sessions.declare(client)
-        session.client = client
+        session = sessions[client.client_id]? || sessions.declare(client).tap { |s| s.client = client }
         headers = AMQP::Table.new({RETAIN_HEADER => true})
         topics.map do |tf|
           session.subscribe(tf.topic, tf.qos)
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -75,7 +75,7 @@ module LavinMQ
       end
 
       def subscribe(client, topics)
-        session = sessions[client.client_id]? || sessions.declare(client).tap { |s| s.client = client }
+        session = sessions.declare(client)
         headers = AMQP::Table.new({RETAIN_HEADER => true})
         topics.map do |tf|
           session.subscribe(tf.topic, tf.qos)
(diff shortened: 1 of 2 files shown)


src/lavinmq/mqtt/broker.cr
```diff
@@ -75,8 +75,7 @@ module LavinMQ
       end
 
       def subscribe(client, topics)
-        session = sessions[client.client_id]? || sessions.declare(client)
-        session.client = client
+        session = sessions[client.client_id]? || sessions.declare(client).tap { |s| s.client = client }
```
carlhoerberg, line 78: Why doesn't the `declare` method do the assignment? it has the `client` available to it.

src/lavinmq/mqtt/broker.cr
```diff
@@ -75,8 +75,7 @@ module LavinMQ
       end
 
       def subscribe(client, topics)
-        session = sessions[client.client_id]? || sessions.declare(client)
-        session.client = client
```
carlhoerberg, line 79: Why does this reset the session and remove the subscriptions?

src/lavinmq/mqtt/broker.cr
kickster97: oh yeah, thats true. lets do that

src/lavinmq/mqtt/broker.cr
kickster97: because when we trigger the client= setter, it removes the consumer on the session (we only have one consumer in mqtt, the old client) and replace it with the new one. But if you have a client with clean_session set to true, it will delete the session when the consumer is removed (it inherits the auto_delete functionality from Queue)

carlhoerberg: oh i see

spuun: That should probably be refactored too, probably as a part of #957

kickster97: Yes definitely, probably time for that soon.. this solves the bug in the mean time though

PR #1072, opened 2025-04-25
Show active arguments on queues & exchanges

### LINKED ISSUE #1024: Show which arguments are "active" on a queue/exchange
A queue or exchange in LavinMQ can have any arguments but only some of them are used by LavinMQ. 

Having the arguments that impacts how LavinMQ works highlighted in some way to separate them from other arguments would make it easier to identify them and if they have any typos.

viktorerlingsson:
Displays arguments as active or inactive on both queues and exchanges. An active argument is an argument that enables certain functionality in LavinMQ, whereas an inactive argument is any other argument. 
image

Fixes #1024 

Check in GUI.

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -29,6 +29,7 @@ module LavinMQ
       @delayed_queue : Queue?
       @deleted = false
       @deduper : Deduplication::Deduper?
+      @effective_args = Array(String).new
 
       rate_stats({"publish_in", "publish_out", "unroutable", "dedup"})
       property publish_in_count, publish_out_count, unroutable_count, dedup_count
@@ -67,15 +68,22 @@ module LavinMQ
       end
 
       def handle_arguments
+        @effective_args = Array(String).new
         @alternate_exchange = (@arguments["x-alternate-exchange"]? || @arguments["alternate-exchange"]?).try &.to_s
+        @effective_args << "x-alternate-exchange" if @alternate_exchange
         if @arguments["x-delayed-exchange"]?.try &.as?(Bool)
           @delayed = true
           init_delayed_queue
+          @effective_args << "x-delayed-exchange"
         end
         if @arguments["x-message-deduplication"]?.try(&.as?(Bool))
+          @effective_args << "x-message-deduplication"
           ttl = parse_header("x-cache-ttl", Int).try(&.to_u32)
+          @effective_args << "x-cache-ttl" if ttl
           size = parse_header("x-cache-size", Int).try(&.to_u32)
+          @effective_args << "x-cache-size" if size
           header_key = parse_header("x-deduplication-header", String)
+          @effective_args << "x-deduplication-header" if header_key
           cache = Deduplication::MemoryCache(AMQ::Protocol::Field).new(size)
           @deduper = Deduplication::Deduper.new(cache, ttl, header_key)
         end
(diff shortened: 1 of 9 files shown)


kickster97:
Hmm, how do we handle arguments like: 
image
image

viktorerlingsson:
> looks very good! Hmm, how do we handle arguments like: 

Ah, they need some special handling since they aren't handed like other arguments on queues. I'll fix!

kickster97 (approved):

carlhoerberg:
Looks really good!

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -95,6 +103,7 @@ module LavinMQ
           operator_policy: @operator_policy.try &.name,
           effective_policy_definition: Policy.merge_definitions(@policy, @operator_policy),
           message_stats: current_stats_details,
+          effective_arguments: @effective_args,
```
carlhoerberg, line 106: is that property present in rabbit too?

carlhoerberg, line 106: is that property present in rabbit too?

carlhoerberg, line 106: is that property present in rabbit too? in queue this is a comma separated string, here an array, which one do we want to use?

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -252,34 +253,56 @@ module LavinMQ::AMQP
     end
 
     private def handle_arguments
+      @effective_args = Array(String).new
```
carlhoerberg, line 256: can `handle_arguments` be called multiple times throughout the lifetime of the Queue or why is the array reassigned?

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -404,6 +427,7 @@ module LavinMQ::AMQP
         state:                       @state,
         effective_policy_definition: Policy.merge_definitions(@policy, @operator_policy),
         message_stats:               current_stats_details,
+        effective_arguments:         @effective_args.join(','),
```
carlhoerberg, line 430: or keep it as an array?

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: Yes, `handle_arguments` is called every time `apply_policy` runs.

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: yes!

src/lavinmq/amqp/exchange/exchange.cr
viktorerlingsson: > is that property present in rabbit too? 

No, this functionality does not exist in rabbit. 

>  in queue this is a comma separated string, here an array, which one do we want to use?

I started with a string and then moved on to array, but l guess I forgot to remove that `join(',')` in queue. I'll do that!

static/main.css
```diff
@@ -1769,11 +1769,22 @@ fieldset.inline {
...
+pre.arguments .active-argument {
   color: var(--font-color-light);
 }
+pre.arguments .active-argument:before,
+pre.arguments .inactive-argument:before {
+  padding-right: 5px;
+}
+pre.arguments .active-argument:before {
+  content: "●";
+  color: #44FC70;
```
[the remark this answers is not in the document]
viktorerlingsson: No, guess we can add that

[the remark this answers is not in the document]
carlhoerberg: It just looked some other color we already use, and i was thinking that it makes it easier when we add light mode, if the color palette are in variables.

Issue #1073, opened 2025-04-26
Problem on setting default username and password from environment


abdullahenesoncu-s4e:
As https://github.com/cloudamqp/lavinmq/pull/919, I expected to be able to set default username and password from environment.

I am getting below errors when trying to set default user and password from environment:

`docker run -e LAVINMQ_DEFAULT_USER=guest -e LAVINMQ_DEFAULT_PASSWORD=guest -p 5672:5672 -p 15672:15672 cloudamqp/lavinmq`

```
WARNING: 'guest_only_loopback' is deprecated, use '--default-user-only-loopback' instead
WARNING: 'guest_only_loopback' is deprecated, use '--default-user-only-loopback' instead
2025-04-26T21:25:34.047313Z  INFO lmq[level: "Info", target: "stdout"] Logger settings
2025-04-26T21:25:34.047783Z  INFO lmq.launcher LavinMQ 2.3.0
2025-04-26T21:25:34.047792Z  INFO lmq.launcher Crystal 1.16.0 [53cead6] (2025-04-09)
2025-04-26T21:25:34.047792Z  INFO lmq.launcher LLVM: 18.1.3
2025-04-26T21:25:34.047793Z  INFO lmq.launcher Default target: aarch64-unknown-linux-gnu
2025-04-26T21:25:34.047794Z  INFO lmq.launcher Build flags: --release --debug
2025-04-26T21:25:34.047796Z  INFO lmq.launcher Pid: 1
2025-04-26T21:25:34.047796Z  INFO lmq.launcher Data directory: /var/lib/lavinmq
2025-04-26T21:25:34.047891Z  INFO lmq.launcher Max mmap count: 262144
2025-04-26T21:25:34.047902Z  INFO lmq.launcher FD limit: 1048576
2025-04-26T21:25:34.048556Z ERROR lmq.user_store Failed to load users
Wrong size (Base64::Error)
  from /usr/share/crystal/src/base64.cr:283:7 in 'decode'
  from /tmp/src/lavinmq/password.cr:63:17 in 'new'
  from /tmp/src/lavinmq/user.cr:61:30 in 'update_password_hash'
  from /tmp/src/lavinmq/user_store.cr:40:7 in 'add'
  from /tmp/src/lavinmq/config.cr:78:7 in 'initialize'
  from /usr/share/crystal/src/gc/boehm.cr:192:7 in 'start'
[...]

PR #1078, opened 2025-04-28
Verify default_password hash on launch

viktorerlingsson:
Verifies default_password hash on launch and provides a more useful error message if not valid. 

Related to #1073 

Run LavinMQ with a bad password hash `lavinmq --default_password=not-a-hash` should exit early with an error message.


carlhoerberg:
Should the argument be `--default-password-hash=` to not confuse ppl?

viktorerlingsson:
> Should the argument be `--default-password-hash=` to not confuse ppl?

Yeah, that's probably better! 
Added `default-password-hash` and marked `default-password` as deprecated.

kickster97 (approved):
lgtm :)

PR #1081, opened 2025-04-29
Do not delete message before delivering packet

kickster97:
The build_packet method constructs an `MQTT::Publish` packet using the message data stored in `env.message`.
If the message has already been deleted (via `delete_message(sp)`), the `env.message` reference becomes invalid, leading to a segmentation fault when accessed.

This PR also fixes a bug where the `@consumers.empty?` never got triggered because the `deliver_loop` didn't yield in time for the `read_loop` to take over and set the `client=nil`, so the `deliver_loop` would continue to loop errors. 
Now we set `client=nil `when the `deliver_loop` experiences an error, so the `consumers.empty?` channel triggers and the loop waits for a new client connection. 

run `lavinmqperf mqtt throughput` via the `mqtt-perf` branch and make sure there are no segment faults.

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -36,15 +36,11 @@ module LavinMQ
             consumer.deliver(pub_packet)
           end
           Fiber.yield if (i &+= 1) % 32768 == 0
+        rescue ex
+          @log.error(exception: ex) { "Failed to deliver message in deliver_loop" }
+          @consumers.each &.close
+          client = nil
         end
-      rescue ex : ::IO::Error
-        @log.debug(exception: ex) { "deliver loop exited due to IO error" }
-      rescue ex : ::Channel::ClosedError
-        @log.debug(exception: ex) { "deliver loop exited due to channel closed" }
-      rescue ex
-        @log.error(exception: ex) { "deliver loop exited unexpectedly" }
-      ensure
-        @consumers.each &.close
       end
 
       def client=(client : MQTT::Client?)
@@ -108,9 +104,12 @@ module LavinMQ
           sp = env.segment_position
           no_ack = env.message.properties.delivery_mode == 0
           if no_ack
-            delete_message(sp)
-            packet = build_packet(env, nil)
-            yield packet
+            begin
+              packet = build_packet(env, nil)
+              yield packet
+            ensure
+              delete_message(sp)
+            end
           else
             id = next_id
             return false unless id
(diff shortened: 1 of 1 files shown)


src/lavinmq/mqtt/session.cr
```diff
@@ -113,9 +109,12 @@ module LavinMQ
...
           no_ack = env.message.properties.delivery_mode == 0
           if no_ack
-            delete_message(sp)
-            packet = build_packet(env, nil)
-            yield packet
+            begin
+              packet = build_packet(env, nil)
+              yield packet
+            ensure
+              delete_message(sp)
```
carlhoerberg, line 116: you want to delete it even if we know that the delivery failed? Or do we want to requeue it then? That's what we do in AMQP

kickster97: We don't want to requeue qos=0 messages, they are fire & forget 

> 4.3.1 QoS 0: At most once delivery
The message is delivered according to the capabilities of the underlying network. No response is sent by the receiver and no retry is performed by the sender. The message arrives at the receiver either once or not at all.

kickster97: if `!no_ack`, then we will `mark_unacked(sp)` which will requeue the message.

carlhoerberg: https://github.com/cloudamqp/lavinmq/blob/d34655fd2c3ea78ae48ef484dc36e67b9914f318/src/lavinmq/amqp/queue/queue.cr#L714-L726

carlhoerberg: i think requeuing if we know the delivery failed (exception raised) makes more sense, at least for non clean sessions. clean sessions will delete all messages if the connection is lost, right?

kickster97: ok!

src/lavinmq/mqtt/session.cr
```diff
@@ -36,15 +36,11 @@ module LavinMQ
             consumer.deliver(pub_packet)
           end
           Fiber.yield if (i &+= 1) % 32768 == 0
+        rescue ex
+          @log.error(exception: ex) { "Failed to deliver message in deliver_loop" }
+          @consumers.each &.close
+          client = nil
```
carlhoerberg, line 42: I think `self` is required here, otherwise it's a local variable assignment. 
```suggestion
          self.client = nil
```

carlhoerberg (changes requested):

src/lavinmq/mqtt/session.cr
kickster97: ah good catch!

PR #1083, opened 2025-04-29
Faster queue purge if no unacked messages

### LINKED ISSUE #1079: Optimize purging queues
Purging a queue is slower than deleting it and re-creating it, because purging deletes one message at a time, whereas deleting just deletes the files. Could purging be optimized? At least in some cases? 

Delete entire message store if possible when purging all messages. 

There might be a reason we went with the current solution though. For example we need to keep track of unacked messages. Is there anything else?

viktorerlingsson:
When all messages in a queue is to be purged, and there's no unacked messages, we can just delete the files instead of deleting each message one by one. This makes purging large queues **much** faster. 

Before:
`Purged 50000000 messages in 00:00:04.349915974`
After: 
`Purged 50000000 messages in 00:00:00.002632290`

This is sort of going back to previous functionality that was changed [here](https://github.com/cloudamqp/lavinmq/commit/465e1f768af0b04b78fd378b051e14ec3663f4cc), but now we make sure there's no unacked messages. 

Also makes sure that stream_queues are not affected, since they can not be completely purged. 

Fixes #1079 

Specs should cover most cases.

### DIFF
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -200,6 +200,25 @@ module LavinMQ
         count
       end
 
+      def purge_all
+        while f = @segments.shift?
+          delete_file(f[1])
+        end
+        while f = @acks.shift?
+          delete_file(f[1])
+        end
+        @deleted.clear
+        @segment_msg_count.clear
+        @requeued.clear
+        @bytesize = 0_u64
+        @size = 0_u32
+
+        open_new_segment
+        @wfile_id = @rfile_id = @segments.first_key
+        @wfile = @rfile = @segments.first_value
+        notify_empty(true)
+      end
+
       def delete
         close
         @segments.each_value { |f| delete_file(f) }
@@ -409,6 +428,7 @@ module LavinMQ
       end
 
       private def delete_unused_segments : Nil
+        return if @segments.empty?
         current_seg = @segments.last_key
         @segments.reject! do |seg, mfile|
           next if seg == current_seg # don't the delete the segment still being written to
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -201,12 +201,10 @@ module LavinMQ
       end
 
       def purge_all
-        while f = @segments.shift?
-          delete_file(f[1])
-        end
-        while f = @acks.shift?
-          delete_file(f[1])
-        end
+        @segments.each_value { |f| delete_file(f) }
+        @segments.clear
+        @acks.each_value { |f| delete_file(f) }
+        @acks.clear
         @deleted.clear
         @segment_msg_count.clear
         @requeued.clear
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -203,20 +203,21 @@ module LavinMQ
       end
 
       def purge_all
-        @segments.each_value { |f| delete_file(f) }
+        @segments.each_value { |f| delete_file(f); f.close }
         @segments = Hash(UInt32, MFile).new
-        @acks.each_value { |f| delete_file(f) }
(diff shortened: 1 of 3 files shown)


viktorerlingsson:
Is there any other case besides unacked messages where we don't want to purge all by deleting like this? 🤔

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -200,6 +200,23 @@ module LavinMQ
         count
       end
 
+      def purge_all
+        @segments.each_value { |f| delete_file(f) }
+        @segments.clear
```
carlhoerberg, line 205: Probably a good idea to assign a new hash instead. `Hash#clear` doesn't shrink if it was large. Same for all other `clear` in this method

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -409,6 +426,7 @@ module LavinMQ
       end
 
       private def delete_unused_segments : Nil
+        return if @segments.empty?
```
carlhoerberg, line 429: how can it be empty? isn't there always a write segment?

carlhoerberg (approved):

carlhoerberg:
Not sure why this spec started to fail:
```
crystal spec spec/server_spec.cr:522 # LavinMQ::Server splits frames into max frame sizes

**Comment by carlhoerberg** [comment]
Not sure why this spec started to fail:
```
crystal spec spec/server_spec.cr:522 # LavinMQ::Server splits frames into max frame sizes
```

**Comment by carlhoerberg** [comment]
On the server side the error is: 
```
Unexpected error, while reading: Missing hash key: 2 (KeyError)
  from /usr/share/crystal/src/hash.cr:1195:11 in '[]'
  from src/lavinmq/amqp/queue/message_store.cr:164:9 in 'delete'
  from src/lavinmq/amqp/queue/queue.cr:799:9 in 'delete_message'
  from src/lavinmq/amqp/queue/queue.cr:723:11 in 'basic_get'
  from src/lavinmq/amqp/client.cr:882:31 in 'basic_get'
  from src/lavinmq/amqp/client.cr:393:11 in 'process_frame'
  from src/lavinmq/amqp/client.cr:143:15 in 'read_loop'
  from src/lavinmq/amqp/client.cr:68:9 in '->'
  from /usr/share/crystal/src/fiber.cr:170:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:105:3 in '->'
  from ???
```

viktorerlingsson:
> On the server side the error is:
> 
> ```
> Unexpected error, while reading: Missing hash key: 2 (KeyError)
>   from /usr/share/crystal/src/hash.cr:1195:11 in '[]'
>   from src/lavinmq/amqp/queue/message_store.cr:164:9 in 'delete'
>   from src/lavinmq/amqp/queue/queue.cr:799:9 in 'delete_message'
>   from src/lavinmq/amqp/queue/queue.cr:723:11 in 'basic_get'
>   from src/lavinmq/amqp/client.cr:882:31 in 'basic_get'
>   from src/lavinmq/amqp/client.cr:393:11 in 'process_frame'
>   from src/lavinmq/amqp/client.cr:143:15 in 'read_loop'
>   from src/lavinmq/amqp/client.cr:68:9 in '->'
>   from /usr/share/crystal/src/fiber.cr:170:11 in 'run'
>   from /usr/share/crystal/src/fiber.cr:105:3 in '->'
>   from ???
> ```

Good catch!

viktorerlingsson:
> Not sure why this spec started to fail:
> 
> ```
> crystal spec spec/server_spec.cr:522 # LavinMQ::Server splits frames into max frame sizes
> ```

That spec is kinda weird? Why would it need to purge directly after creating a queue? 🤔

carlhoerberg:
> That spec is kinda weird? Why would it need to purge directly after creating a queue? 🤔 

Yes, dont know but it caught a bug! :)

PR #1084, opened 2025-04-30
Bugfix: append to x-received-from instead of replacing

### LINKED ISSUE #1075: Federation: x-received-from is replaced instead of appended
When a message is federated the `x-received-from` header is updated. The header should be a list contaning all jumps a message has done, i.e. a new values is appended to the header for each jump.

What happens today is that the old value is replaced with a new value each jump, meaning that only the last jump is in the header.

spuun:
This will make sure that values are appended to the `x-received-from` header instead of replacing the header each hop.

It will also fix bindings on the upstream for exchange federation. Bindings was added to the queue but should be added to the ´x-upstream-federation` exchange. Dunno how, but it was missed in #1050...

Fixes #1075


carlhoerberg (approved):

carlhoerberg:
Fixed the merge conflict, i think

carlhoerberg:
Oh, probably not: 

```
1) LavinMQ::Federation::Upstream Exchange federation should reflect all bindings to upstream q
Error running at_exit handler: Exiting with code 1
     Failure/Error: upstream_q.bindings.size.should eq queues.size + 1 # +1 for the default exchange

       Expected: 11
            got: 2

     # spec/upstream_spec.cr:484

PR #1089, opened 2025-05-02
Bugfix: bindings added despite invalid weight in ConsitentHashExchange

spuun:
When adding bindings to a consistent hash exchange routing key must be a number. This is validated in both `#bind` and `#unbind`. However, it was done too late in `#bind` so binding information was added to `@bindings` and thus visible in the binding list in the management UI.

The validation was however done before the destination was added to the `@hasher` meaning that even though a binding was visible in the UI, no message is routed.

This PR will make sure that weight is calculated and validated first in `#bind` and `#unbind`.

It also reorganizes the specs for consistent hash exchange.

Run specs

### DIFF
--- a/spec/consistent_hash_exchange_spec.cr
+++ b/spec/consistent_hash_exchange_spec.cr
@@ -1,78 +1,35 @@
 require "./spec_helper"
 require "../src/lavinmq/consistent_hasher.cr"
 
-describe "Consistent Hash Exchange" do
-  describe "Hasher" do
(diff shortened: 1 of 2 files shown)


viktorerlingsson (approved):
LGTM

snichme:
Just looking at the diff of this pull request it was difficult to understand why it mattered to move the call to wait higher up, but when looking at the whole file it made sense.  
So `weight` both parses and validate the input and but the name indicates that it's only calculate the weight.

PR #1090, opened 2025-05-03
Faster boot by storing number of messages in a segment at the end of each segment

carlhoerberg:
By appending the number of messages a segment contains at the end of the segment file we can speed up the boot process significantly. Then when restoring the queue only that end of the segment has to be read, not the whole file as before.

### DIFF
--- a/src/lavinmq/amqp/queue/message_store.cr
+++ b/src/lavinmq/amqp/queue/message_store.cr
@@ -267,7 +267,7 @@ module LavinMQ
 
       private def write_to_disk(msg) : SegmentPosition
         wfile = @wfile
-        if wfile.capacity < wfile.size + msg.bytesize
+        if wfile.capacity < wfile.size + msg.bytesize + 20
           wfile = open_new_segment(msg.bytesize)
         end
         wfile_id = @wfile_id
@@ -279,6 +279,10 @@ module LavinMQ
       end
 
       private def open_new_segment(next_msg_size = 0) : MFile
+        append_msg_count(@wfile, @segment_msg_count[@wfile_id])
+        @wfile.read_only = true
+        @wfile.truncate(@wfile.size)
+
         @wfile.unmap unless @wfile == @rfile
         next_id = @wfile_id + 1
         path = File.join(@queue_data_dir, "msgs.#{next_id.to_s.rjust(10, '0')}")
@@ -298,7 +302,7 @@ module LavinMQ
       private def open_ack_file(id) : MFile
         path = File.join(@queue_data_dir, "acks.#{id.to_s.rjust(10, '0')}")
         capacity = Config.instance.segment_size // BytesMessage::MIN_BYTESIZE * 4 + 4
-        mfile = MFile.new(path, capacity, writeonly: true)
+        mfile = MFile.new(path, capacity, write_only: true)
         mfile.delete unless @durable # mark as deleted if non-durable
         @replicator.try &.register_file mfile
         mfile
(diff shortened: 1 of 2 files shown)


src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -399,33 +413,85 @@ module LavinMQ
...
-            break mfile.resize(pos) if ts.zero? # This means that the rest of the file is zero, so resize it
-            bytesize = BytesMessage.skip(mfile)
-            count += 1
-            next if deleted?(seg, pos)
+          next if mfile.size <= 4 # only got the schema version
+          # check if message count in segment is appended to the end
+          begin
+            count = read_msg_count(mfile)
+            @segment_msg_count[seg] = count
+            bytesize = mfile.size - 4 # minus schema version
```
spuun, line 421: Subtract the 20 bytes for count+magic+zeros?

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -399,33 +413,85 @@ module LavinMQ
...
+            append_msg_count(f, count)
+          end
+        end
+      end
+
+      private def append_msg_count(file, count)
+        file.write_bytes 0i64 # ts zero
+        file.write_bytes MAGIC
+        file.write_bytes count.to_u32
+        file.write_bytes 0u32 # reserved
```
spuun, line 481: Instead of a reserved integer write a CRC32 of ts, MAGIC and count? Or keep the reserved integer and add a CRC32 field?

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -399,33 +413,85 @@ module LavinMQ
...
           mfile.pos = 4
           mfile.unmap # will be mmap on demand
           if is_long_queue
             @log.info { "Loaded #{counter}/#{@segments.size} segments, #{@size} messages" } if (counter &+= 1) % 128 == 0
-          else
-            @log.debug { "Loaded #{counter}/#{@segments.size} segments, #{@size} messages" } if (counter &+= 1) % 128 == 0
           end
-          Fiber.yield
-          @segment_msg_count[seg] = count
+          @log.info { "Loaded #{counter} segments, #{@size} messages" }
```
spuun, line 441: `counter` isn't incremented except for long queues.

src/lavinmq/amqp/queue/message_store.cr
carlhoerberg: yes!

src/lavinmq/amqp/queue/message_store.cr
carlhoerberg: i think so, should have a spec that catches that too

src/lavinmq/amqp/queue/message_store.cr
```diff
@@ -279,10 +285,16 @@ module LavinMQ
       end
 
       private def open_new_segment(next_msg_size = 0) : MFile
-        @wfile.unmap unless @wfile == @rfile
+        unless @wfile_id.zero?
+          append_msg_count(@wfile, @segment_msg_count[@wfile_id])
+          @wfile.read_only = true
```
carlhoerberg, line 290: This is only so that we don't by mistake edit any data in these files, but it adds a lot of code `MFile`, so that could be optional if we still want that in this PR.

spuun:
Now now segfault i linux (CI) too? Hm...

carlhoerberg:
Ok, probably because we're missing a `reserve`/`unreserve` somewhere

viktorerlingsson:
Thoughts from our meeting: Write this to a separate meta file for each msg file instead. Should be safer and easier to implement, and also super useful for s3 streams.

carlhoerberg:
Superseeded by #1163

PR #1091, opened 2025-05-05
Update docs for running & building LavinMQ with -Dpreview_mt -Dexecution_context

viktorerlingsson:
Updates CONTRIBUTING.md to reflect how to build and test LavinMQ with `-Dpreview_mt -Dexecution_context.` 

Run cmds specified in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,10 @@ The first step to making a code contribution, is starting a conversation around 
 
 ### Develop
 
-1. Run specs with `crystal spec`
-1. Compile and run locally with `crystal run src/lavinmq.cr -- -D /tmp/amqp`
+1. Run specs with `make test`
+1. Run a specific spec file with `make test SPEC=spec/storage_spec.cr`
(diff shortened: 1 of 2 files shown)


viktorerlingsson:
Just adding for clarity: We've had some discussions regarding if this is something that should be in the Makefile or a separate script. Or just handled by each dev locally however they see fit - for example by always building before running, with a script or an alias. 
(I'm trying out using an alias now `alias lavin-local='crystal run src/lavinmq.cr -Dpreview_mt -Dexecution_context --stats --'` )

carlhoerberg:
Agree. My workflow is `make bin/lavinmq CRYSTAL_FLAGS=` and then `bin/lavinmq --config config.ini`, beneficial to have a stored binary when when testing restarts, clustering etc.

viktorerlingsson:
Reverted the changes to the Makefile, just specifying how to build & run in CONTRIBUTING.md instead

kickster97 (approved):

PR #1092, opened 2025-05-05
Exchanges must support binding arguments

spuun:
All exchanges must store binding arguments.

Specs should cover it


carlhoerberg (approved):
Maybe some of the `Set`s can be `Array`s instead? Should be cheaper, but not dealbreaker for this PR.

PR #1094, opened 2025-05-05
fixes a bug where using routing_key or exchange with length 255 causes an ArithmeticOverflow

### LINKED ISSUE #1093: ArithmeticOverflow when starting LavinMQ with routing keys that are 255 chars long
**Describe the bug**
LavinMQ crashes on startup if a message has a routing key that is 255 chars long

**Describe your setup**
LavinMQ v2.3.0. 

**How to reproduce**
Publish a message with a routing key that is 255 chars long. Then restart LavinMQ. 
LavinMQ then crashes with an ArithmeticOverflow error. 

This probably also applies if using an exchange that is 255 chars long.

The problem is probably this row https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/message.cr#L42
`skipped += io.skip(io.read_byte || raise IO::EOFError.new) + 1 # rk`

Because the return type from `io.read_byte` is `UInt8`, and if the value is 255, it will raise if we try to add 1 to it. 

**Expected behavior**
LavinMQ should be able to read routing keys that are 255 chars long.

viktorerlingsson:
fixes a bug where using routing_key or exchange with length 255 causes an ArithmeticOverflow

Fixes #1093 

### DIFF
--- a/src/lavinmq/message.cr
+++ b/src/lavinmq/message.cr
@@ -37,9 +37,11 @@ module LavinMQ
 
     def self.skip(io, format = IO::ByteFormat::SystemEndian) : UInt64
       skipped = 0_u64
-      skipped += io.skip(sizeof(UInt64))                             # ts
-      skipped += io.skip(io.read_byte || raise IO::EOFError.new) + 1 # ex
-      skipped += io.skip(io.read_byte || raise IO::EOFError.new) + 1 # rk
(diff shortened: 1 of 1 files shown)


carlhoerberg (approved):

carlhoerberg:
Or just put the 1 before io.skip instead of after

PR #1095, opened 2025-05-06
Abstract MessageStore out of Queue

kickster97:
I want to move the message store out of `Queue` and put it in a place so it makes more sense when we break `MQTT::Session` free from `Queue`. 

### How to test this 
no functionality should change with this refactor, so LavinMQ needs to compile and the specs should pass


carlhoerberg:
I have nothing against this change, it makes sense! But as always, put the file move `git mv` in one commit and any changes in another, and keep the two commits when merging the PR.

kickster97:
> I have nothing against this change, it makes sense! But as always, put the file move `git mv` in one commit and any changes in another, and keep the two commits when merging the PR.

Yes roger that!

> The reason I called it `queue_data_dir` was to not confuse it with the global `data_dir`, how valid that still is I don't know.

Ok, I see that it could be confusing to have it called `data_dir`, instead it can be called `msg_data_dir`, makes it a bit more general but still separates it from `data_dir`.

baelter:
> instead it can be called `msg_data_dir`

or just `msg_dir`?

kickster97:
> > instead it can be called `msg_data_dir`
> 
> or just `msg_dir`?

yes!!

viktorerlingsson (approved):
LGTM 👍 

I think we could move more files around to make our general file structure more intuitive, but that can be another PR.

PR #1099, opened 2025-05-09
[wip] started adding get_no_ack_count when basic_get with no_ack

lukas8219:
Please fill me in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,3 +163,6 @@ clean:
 .PHONY: watch
 watch:
 	while true; do inotifywait -qqr -e modify -e create -e delete src/ && $(MAKE); done
+
+run-dev:
+	crystal run -Dpreview_mt -Dexecution_context src/lavinmq.cr -- -D /tmp/amqp
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -726,6 +726,7 @@ module LavinMQ::AMQP
       @queue_expiration_ttl_change.try_send? nil
       @get_count.add(1)
       @deliver_get_count.add(1)
+      @get_no_ack_count.add(1) if no_ack
       get(no_ack) do |env|
         yield env
       end


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -726,6 +726,7 @@ module LavinMQ::AMQP
       @queue_expiration_ttl_change.try_send? nil
       @get_count.add(1)
       @deliver_get_count.add(1)
+      @get_no_ack_count.add(1) if no_ack
```
lukas8219, line 729: missing tests

lukas8219:

Makefile
```diff
@@ -163,3 +163,6 @@ clean:
 .PHONY: watch
 watch:
 	while true; do inotifywait -qqr -e modify -e create -e delete src/ && $(MAKE); done
+
+run-dev:
+	crystal run -Dpreview_mt -Dexecution_context src/lavinmq.cr -- -D /tmp/amqp
```
kickster97, line 168: This is good, we have discussed adding something like this internally, but it should be in a separate PR with some explanation and possibly docs aswell :)

kickster97 (changes requested):
Looks good 👍

Makefile
viktorerlingsson: I think we basically decided against having a run cmd in the Makefile and instead just updating the docs https://github.com/cloudamqp/lavinmq/pull/1091

kickster97: ok cool!

viktorerlingsson:
Looking at where we increase `get_count`, does this also need to be handled for the [channel](https://github.com/lukas8219/lavinmq/blob/824/src/lavinmq/amqp/channel.cr#L390-L404), [vhost](https://github.com/lukas8219/lavinmq/blob/824/src/lavinmq/vhost.cr#L675-L678) and added to [message_details](https://github.com/lukas8219/lavinmq/blob/824/src/lavinmq/vhost.cr#L145-L177) for a `vhost`?

lukas8219:
@viktorerlingsson 

I think it makes a lot of sense. I'll add this logic to both channel and vhost in the next commit.

Regarding `message_details` - just so we are on the same page. [It seems like vhost already has this setup](https://github.com/lukas8219/lavinmq/blob/824/src/lavinmq/vhost.cr#L156)?

Makefile
lukas8219: Thank you guys. I'll remove from this PR

src/lavinmq/amqp/channel.cr
```diff
@@ -388,9 +388,16 @@ module LavinMQ
           elsif q.is_a? StreamQueue
             @client.send_not_implemented(frame, "Stream queues does not support basic_get")
           else
-            @get_count.add(1)
-            @deliver_get_count.add(1)
-            @client.vhost.event_tick(EventType::ClientGet)
+            #TODO 824 check frame for `frame.no_ack` to emit proper event
```
lukas8219, line 391:
```suggestion
```

lukas8219:
I also hate force-pushes @kickster97 @viktorerlingsson but I had to do because of rebases. The CLA Assistant was requiring more people sign-off when pulling their changes. (Any suggestions in the future to rebase branches?)

I've addressed the points of queue, channel and vhost `no_ack` metrics.

kickster97:
> The CLA Assistant was requiring more people sign-off when pulling their changes



We usually work with force pushes to rebase, or to clean up a commit log that should not be squashed. I don't really have another suggestion :) Where is the CLA assistant requiring that? I can't see it in this PR.

viktorerlingsson:
I enabled the GitHub workflows to have specs and stuff run. Seems like not all if them runs properly on external branches.

@lukas8219 you can ignore most of the failing CI since that's because of authorization (and OS X specs are flaky), but please have a look at the Lint check.

lukas8219:
@kickster97 Since I rebased I think it got lost within the old commits.

@viktorerlingsson About the cyclomatic complexity - may I split the `Lavin::AMQP:Queue::get` function?

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -748,6 +747,12 @@ module LavinMQ::AMQP
     # returns true if a message was deliviered, false otherwise
     # if we encouncer an unrecoverable ReadError, close queue
     private def get(no_ack : Bool, & : Envelope -> Nil) : Bool
+      case no_ack
+      when true
+        @get_no_ack_count.add(1)
+      when false
+        @get_count.add(1)
+      end
```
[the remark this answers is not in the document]
viktorerlingsson: @lukas8219 it was easier to respond here than directly to your comment 🙂

spec/api/channels_spec.cr
```diff
@@ -60,6 +60,27 @@ describe LavinMQ::HTTP::ChannelsController do
...
+          channel = JSON.parse(response.body)[0]
+          name = URI.encode_www_form(channel["name"].as_s)
+          vhost = channel["vhost"].as_s
+          s.vhosts[vhost].declare_queue("q1", false, false)
+          s.vhosts[vhost].queues["q1"].basic_get(false) do |_|
+            response = http.get("/api/channels/#{name}")
+            response.status_code.should eq 200
+            body = JSON.parse(response.body)
+            message_stats = body["message_stats"].not_nil!
+            message_stats["get_no_ack"].should eq(1)
```
viktorerlingsson, line 79: You can avoid the lint error on this by doing something like
```suggestion
            if message_stats = body["message_stats"]
              message_stats["get_no_ack"].should eq(1)
            else
              fail "No message stats"
            end
```

src/lavinmq/amqp/queue/queue.cr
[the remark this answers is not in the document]
lukas8219: It was on `basic_get` before but isn't this metric also handled for `consume_get`? Should this be specifically for basic.get frames?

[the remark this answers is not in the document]
viktorerlingsson: No, consume_get is added to deliver (and should be added to deliver_no_ack, but we haven't implemented that for some reason). It's designed to work the same way as it does in rabbitmq, see Message_stats objects here for reference: https://www.rabbitmq.com/docs/http-api-reference#metrics-returned-by-the-http-api#message_stats-objects

So get and get_no_ack should be just for basic get.

lukas8219:
@viktorerlingsson Per your recommendation I tried running the `ameba` for linting but all it outputted was some weird stuff regarding `require "../helpers"` . Not sure I am missing something on this.

spec/api/channels_spec.cr
```diff
@@ -60,6 +60,30 @@ describe LavinMQ::HTTP::ChannelsController do
...
+          s.vhosts[vhost].queues["q1"].basic_get(false) do |_|
+            response = http.get("/api/channels/#{name}")
+            response.status_code.should eq 200
+            body = JSON.parse(response.body)
+            if message_stats = body["message_stats"]?
+              message_stats["get_no_ack"].should eq(1)
+            else
+              fail "message_stats is nil"
+            end
+          end
```
viktorerlingsson, line 83: The block after `s.vhosts[vhost].queues["q1"].basic_get(false) do |_|` doesn't run here because there is no messages in the queue, so the assertion is never done. 

Also I think it would be a good thing assert that `get == 0 `and `deliver_get == 1`. 

So I would suggest doing something like this instead: 

```suggestion
        with_channel(s) do |ch|
          response = http.get("/api/channels")
          response.status_code.should eq 200
          channel = JSON.parse(response.body)[0]
          name = URI.encode_www_form(channel["name"].as_s)
          
          q1 = ch.queue("q1")
          q1.publish_confirm "msg"
          q1.get(no_ack: true)

          response = http.get("/api/channels/#{name}")
          response.status_code.should eq 200
          body = JSON.parse(response.body)
          if message_stats = body["message_stats"]?
            message_stats["get_no_ack"].should eq(1)
            message_stats["get"].should eq(0)
            message_stats["deliver_get"].should eq(1)
          else
            fail "message_stats is nil"
          end
        end
```

spec/api/channels_spec.cr
```diff
@@ -60,6 +60,30 @@ describe LavinMQ::HTTP::ChannelsController do
         end
       end
     end
+
+    it "should return get_no_ack count 1" do
+      with_http_server do |http, s|
+        with_channel(s) do |c|
```
viktorerlingsson, line 66: Very minor, but we pretty much always use `ch` in all other specs, so I think it's easier if we keep to that tradition 🙂

spec/api/channels_spec.cr
```diff
@@ -60,6 +60,30 @@ describe LavinMQ::HTTP::ChannelsController do
...
+          vh_name = "/"
+          q = c.queue
+          q.publish "get_no_ack_message"
+          message = q.get(true)
+          if message = nil
+            fail "No message returned"
+          end
```
viktorerlingsson, line 73: This isn't needed - if no message is returned the assertions below would fail because `get_no_ack` would be 0

viktorerlingsson: so you can also skip assigning message, just do `q.get(true)`

spec/vhost_stats_spec.cr
```diff
@@ -65,5 +65,23 @@ describe LavinMQ::VHost do
...
+    it "should count get_no_ack" do
+      with_amqp_server do |s|
+        with_channel(s) do |c|
+          vhost = s.vhosts["/"]
+          s.update_stats_rates
+          vhost.get_no_ack_count.should eq(0)
+          q = c.queue("q1")
+          q.publish "not acked message"
+
+          q.get(true).not_nil!
```
viktorerlingsson, line 78: `not_nil!` shouldn't be needed here?

lukas8219:
@viktorerlingsson About those linting issues you mentioned - I noticed the `Makefile` didn't have them. May I add in this PR?

spec/api/channels_spec.cr
lukas8219: Hmm yeah maybe that was redundant. It was mainly to deal with `nil` scenarios - same [as this](https://github.com/cloudamqp/lavinmq/pull/1099#discussion_r2097265991)

viktorerlingsson:
> @viktorerlingsson About those linting issues you mentioned - I noticed the `Makefile` [didn't have /spec](https://github.com/cloudamqp/lavinmq/blob/main/Makefile#L107). May I add in this PR?

Hmm 🤔 I think that should be a separate PR if we do that, we should aim for only doing one thing per PR. 

I'm not sure if I think we need to lint specs in the Makefile though, since specs aren't really part of the actual build. Maybe it's best to create an issue/discussion for that and have the rest of the team weigh in!

lukas8219:
@viktorerlingsson  Yup. Totally agree with separate PR - my comment was confusing indeed. Regarding CI - it does run `ameba spec` in [the CI](https://github.com/cloudamqp/lavinmq/actions/runs/15123477681/workflow?pr=1099#L48) - at least for PR's

viktorerlingsson:
Sorry for the delay, I found an existing bug in `channel.cr` while looking through the code here. Problem was that we where incrementing `deliver_count` in the `deliver()` method, which was used by both `basic_get()` and `consume()`, meaning `deliver_count` and `deliver_get_count` would be increased when it shouldn't be. Fix for it here: #1119

carlhoerberg (approved):

viktorerlingsson:
Sorry for the delay @lukas8219 , but now that #1119 is merged, please rebase the changes from main. And then we can hopefully get this merged pretty soon! 👍

lukas8219:
Getting some weird output from `http_api_spec.cr` after rebasing. I'll try looking into it tomorrow

viktorerlingsson:
> Getting some weird output from `http_api_spec.cr` after rebasing. I'll try looking into it tomorrow

Hmm, yeah you probably need to increment deliver_get_count on the get_no_ack event in vhost.cr 🤔 

(I'll be AFK for a couple of days, so I might not check in here until Monday 🙂)

spec/api/queues_spec.cr
```diff
@@ -71,9 +71,10 @@ describe LavinMQ::HTTP::QueuesController do
         response.status_code.should eq 200
         body = JSON.parse(response.body)
         body["message_stats"]["publish_details"]["rate"].nil?.should be_false
-        body["message_stats"]["get"].should eq 2
+        body["message_stats"]["get"].should eq 1
         body["message_stats"]["deliver"].should eq 1
         body["message_stats"]["deliver_get"].should eq 3
+        body["message_stats"]["get_no_ack"].should eq 1
```
viktorerlingsson: Very minor, but In think it looks better and is more clear to check all "single" stats first, and then `deliver_get`
```suggestion
        body["message_stats"]["get"].should eq 1
        body["message_stats"]["deliver"].should eq 1
        body["message_stats"]["get_no_ack"].should eq 1
        body["message_stats"]["deliver_get"].should eq 3
```

spec/vhost_stats_spec.cr
```diff
@@ -65,5 +65,23 @@ describe LavinMQ::VHost do
         vhost.connection_closed_count.should eq(initial_connections_closed + 1)
       end
     end
+
+    it "should count get_no_ack" do
+      with_amqp_server do |s|
+        with_channel(s) do |c|
+          vhost = s.vhosts["/"]
+          s.update_stats_rates
+          vhost.get_no_ack_count.should eq(0)
```
viktorerlingsson: No need for `()`
```suggestion
          vhost.get_no_ack_count.should eq 0
```

spec/vhost_stats_spec.cr
```diff
@@ -65,5 +65,23 @@ describe LavinMQ::VHost do
...
+          q = c.queue("q1")
+          q.publish "not acked message"
+
+          q.get(true)
+          s.update_stats_rates
+          vhost.get_no_ack_count.should eq(1)
+          vhost.deliver_get_count.should eq(1)
+          vhost.get_count.should eq(0)
```
viktorerlingsson: No need for `()`
```suggestion
          vhost.get_no_ack_count.should eq 1
          vhost.deliver_get_count.should eq 1
          vhost.get_count.should eq 0
```

viktorerlingsson:
Looks good to me! Nice work👍 
Just a couple very minor style things, then I think we're ready to merge this 🙂

PR #1102, opened 2025-05-12
Fix(mqtt/session): Remove consumer after requeueing unacked messages when setting client

kickster97:

When a new MQTT client tries to connect with an ID already in use by a MQTT client with unacknowledged messages, the old connection isn't properly removed and gets stuck. This happens because the Session tries to requeue the unacked messages after it has removed the consumer.  

This PR makes sure we try to requeue messages frist, and remove the consumer after that is finished. 
It also adds so we don't requeue messages if the session is an auto_delete/unclean session

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -47,9 +47,11 @@ module LavinMQ
         return if @closed
         @last_get_time = RoughTime.monotonic
 
-        @msg_store_lock.synchronize do
-          @unacked.values.each do |sp|
-            @msg_store.requeue(sp)
+        if !clean_session?
+          @msg_store_lock.synchronize do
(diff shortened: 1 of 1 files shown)


src/lavinmq/mqtt/session.cr
```diff
@@ -46,17 +46,20 @@ module LavinMQ
...
         return if @closed
         @last_get_time = RoughTime.monotonic
-        @consumers.each do |c|
-          rm_consumer c
-        end
 
-        @msg_store_lock.synchronize do
-          @unacked.values.each do |sp|
-            @msg_store.requeue(sp)
+        if !clean_session?
```
spuun, line 50:
```suggestion
        unless clean_session?
```

spuun (approved):
LGTM

Issue #911, opened 2025-01-14
Pause/Edit/Resume Shovel


kickster97:
**Is your feature request related to a problem? Please describe.**
It would be nice to have more control over a shovel, being able to pause/resume the shovel and also being able to edit the shovel. 

**Describe the solution you'd like**
an endpoint that pauses/resumes a shovel from shovelling messages. 
And an interface that allows you to update and edit the information on a shovel.

lukas8219:
@kickster97 I am also working on this one - Halfway through

kickster97:
Awesome 🔥

PR #1103, opened 2025-05-12
feat(shovel): add ability to pause and resume a shovel #911

lukas8219:
Adds ability to pause and resume a running shovel
Fixes https://github.com/cloudamqp/lavinmq/issues/911

- Specs:
 - spec/shovel_spec.cr

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -16,6 +16,7 @@ module LavinMQ
       Starting
       Running
       Stopped
+      Paused
       Terminated
       Error
     end
@@ -383,7 +384,7 @@ module LavinMQ
       def run
         Log.context.set(name: @name, vhost: @vhost.name)
         loop do
-          break if terminated?
+          break if shouldStopLoop?
           @state = State::Starting
           unless @source.started?
             if @source.last_unacked
@@ -393,7 +394,7 @@ module LavinMQ
           end
           @destination.start unless @destination.started?
 
-          break if terminated?
+          break if shouldStopLoop?
           Log.info { "started" }
           @state = State::Running
           @retries = 0
@@ -404,7 +405,7 @@ module LavinMQ
           @vhost.delete_parameter("shovel", @name) if @source.delete_after.queue_length?
           break
         rescue ex : ::AMQP::Client::Connection::ClosedException | ::AMQP::Client::Channel::ClosedException | Socket::ConnectError
-          break if terminated?
+          break if shouldStopLoop?
           @state = State::Error
           # Shoveled queue was deleted
           if ex.message.to_s.starts_with?("404")
@@ -414,14 +415,14 @@ module LavinMQ
           @error = ex.message
           exponential_reconnect_delay
         rescue ex
-          break if terminated?
+          break if shouldStopLoop?
           @state = State::Error
           Log.warn { ex.message }
           @error = ex.message
           exponential_reconnect_delay
         end
       ensure
-        terminate
+        terminate if !paused?
       end
 
       def exponential_reconnect_delay
@@ -443,6 +444,19 @@ module LavinMQ
         }
       end
 
+      def resume
+        @state = State::Running
+        run
+        Log.info &.emit("Resumed", name: @name, vhost: @vhost.name)
+      end
+
+      def pause
+        @state = State::Paused
+        @source.stop
+        @destination.stop
+        Log.info &.emit("Paused", name: @name, vhost: @vhost.name)
+      end
+
       # Does not trigger reconnect, but a graceful close
       def terminate
         @state = State::Terminated
@@ -450,6 +464,14 @@ module LavinMQ
         @destination.stop
         return if terminated?
         Log.info &.emit("Terminated", name: @name, vhost: @vhost.name)
+      end
+
+      def shouldStopLoop?
+        terminated? || paused?
+      end
+
+      def paused?
+        @state.paused?
       end
 
       def terminated?
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -444,8 +444,8 @@ module LavinMQ
         }
       end
 
+      #TODO 911 review this - I don't think we need this - we can just call `run`
       def resume
-        @state = State::Running
         run
         Log.info &.emit("Resumed", name: @name, vhost: @vhost.name)
       end
--- a/src/lavinmq/shovel/shovel_store.cr
+++ b/src/lavinmq/shovel/shovel_store.cr
@@ -16,6 +16,7 @@ module LavinMQ
     end
 
     def create(name, config)
+      #TODO 911 review this - feels like we could have an update method here
       @shovels[name]?.try &.terminate
       delete_after_str = config["src-delete-after"]?.try(&.as_s.delete("-")).to_s
       delete_after = Shovel::DeleteAfter.parse?(delete_after_str) || Shovel::DEFAULT_DELETE_AFTER
--- a/src/lavinmq/shovel/shovel_store.cr
+++ b/src/lavinmq/shovel/shovel_store.cr
@@ -15,9 +15,27 @@ module LavinMQ
       end
     end
 
-    def create(name, config)
-      #TODO 911 review this - feels like we could have an update method here
-      @shovels[name]?.try &.terminate
+    def upsert(name, config)
+      shovel = @shovels[name]?
+      return upsert0(name, config) unless shovel
+      case {shovel.state, config["state"]}
+      when {Shovel::State::Running, "paused"}
+        shovel.pause
+      when {Shovel::State::Paused, "running"}
+        spawn(shovel.resume, name: "Shovel name=#{name} vhost=#{@vhost.name}")
+      end
+      shovel.terminate
+      return upsert0(name, config)
+    end
+
+    def delete(name)
+      if shovel = @shovels.delete name
+        shovel.terminate
+        shovel
+      end
+    end
+
+    private def upsert0(name, config)
       delete_after_str = config["src-delete-after"]?.try(&.as_s.delete("-")).to_s
       delete_after = Shovel::DeleteAfter.parse?(delete_after_str) || Shovel::DEFAULT_DELETE_AFTER
       ack_mode_str = config["ack-mode"]?.try(&.as_s.delete("-")).to_s
@@ -40,13 +58,6 @@ module LavinMQ
       shovel
     rescue KeyError
       raise JSON::Error.new("Fields 'src-uri' and 'dest-uri' are required")
-    end
-
-    def delete(name)
-      if shovel = @shovels.delete name
-        shovel.terminate
-        shovel
-      end
     end
 
     private def destination(name, config, ack_mode)
--- a/spec/shovel_spec.cr
+++ b/spec/shovel_spec.cr
@@ -600,6 +600,48 @@ describe LavinMQ::Shovel do
         end
       end
     end
+
+    it "should pause and resume shovel" do
+      with_amqp_server do |s|
+        vhost = s.vhosts.create("pause:resume:vhost")
+        queue_name = "shovel:pause:resume"
+        source = LavinMQ::Shovel::AMQPSource.new(
+          "#{queue_name}q1",
+          [URI.parse(s.amqp_url)],
+          "#{queue_name}q1",
+          delete_after: LavinMQ::Shovel::DeleteAfter::QueueLength,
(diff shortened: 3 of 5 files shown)


src/lavinmq/shovel/shovel.cr
```diff
@@ -16,6 +16,7 @@ module LavinMQ
       Starting
       Running
       Stopped
+      Paused
```
lukas8219, line 19: TODO: Re-order

src/lavinmq/shovel/shovel.cr
```diff
@@ -414,14 +415,14 @@ module LavinMQ
...
-          break if terminated?
+          break if should_stop_loop?
           @state = State::Error
           Log.warn { ex.message }
           @error = ex.message
           exponential_reconnect_delay
         end
       ensure
-        terminate
+        terminate if !paused?
```
lukas8219, line 425: Lint: cyclomatic complex

src/lavinmq/shovel/shovel_store.cr
```diff
@@ -15,8 +15,33 @@ module LavinMQ
...
+      shovel
+    end
+
+    private def update_status_or_terminate(shovel, config)
+      case {shovel.state, config["state"]?.try &.as_s}
+      when {"Running", "Paused"}
+        shovel.pause
+        return shovel
+      when {"Paused", "Running"}
+        spawn(shovel.resume, name: "Shovel name=#{shovel.name} vhost=#{@vhost.name}")
```
lukas8219, line 30: 1. Check wether this strategy is OK
2. If so, check wether we need this `as_s` on [shovel.cr](https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/shovel/shovel.cr#L379)

lukas8219: Honestly feels like Enum lose it's sense if they are treated as simple strings?

lukas8219:
@kickster97 I am almost done with this one - all I need now is write a proper HTTP Layer Integration Testing as I found issues when doing E2E testing manually.

If you guys wanna start reviewing it - I did some small structural changes in the `shovel.cr`

src/lavinmq/shovel/shovel.cr
lukas8219: Hmmm I'll have give it more though to solve this one to be fair

lukas8219:
@viktorerlingsson @kickster97 moving back to draft as I do have some leftover questions about the structure.

inb4: This is pretty related to how I designed the solution. Re-thinking it should be enough so this doesn't become an issue. (**Or maybe I should've asked from the beginning**)

While writing the tests I noticed two things:
1. When I first developed the solution I tested mainly with `shovels_spec.cr` - which gave me the expected outcome.
2. When testing on UI - [it reveled a bug](https://github.com/cloudamqp/lavinmq/pull/1103/commits/fc7b6538b237a55e70b655166a0698db23f1c15d) - given the string -> enum + my pattern match logic.

It is clear that the `shovels_spec.cr` is not enough to test it - but creating HTTP layer tests seems a bit weird wth current API design (or at least breaks **file/module->test** convention).

It really feels like
```crystal
# file `definitions_spec.cr`
it "should be able to pause shovel via API" do
#logic
end
```

Is just wrong? As `definitions` API should be agnostic about Shovels behavior. Yet - we don't have `api/shovels` separated into it's own file/API.



Question:

Should we add this API - e.g make it backwards compatible with RabbitMQ while enhancing with a new API? Feels easier to test and even use.

src/lavinmq/shovel/shovel_store.cr
spuun: Yeah, agree. I guess that state getter initially was only used as a display value in the mgmt ui? It should probably be changed.

src/lavinmq/shovel/shovel.cr
lukas8219: kept it simple so i don't touch too much stuff

spuun:
> It is clear that the shovels_spec.cr is not enough to test it - but creating HTTP layer tests seems a bit weird wth current API design (or at least breaks file/module->test convention).

As you maybe have noticed we haven't really followed any convention; initially we just went for "move fast and break things" :)

> Is just wrong? As `definitions` API should be agnostic about Shovels behavior. Yet - we don't have `api/shovels` separated into it's own file/API.
> 
> Question:
> 
> Should we add this API - e.g make it backwards compatible with RabbitMQ while enhancing with a new API? Feels easier to test and even use.

Since pausing at starting a shovel will be a lavinmq feature it makes sense to make a new API for this. Let's do that!

Queue has `PUT /api/queues/{vhost}/{name}/resume` and `PUT /api/queues/{vhost}/{name}/pause` for the same functionality, so I guess it makes sense to mimic this.

src/lavinmq/shovel/shovel_store.cr
```diff
@@ -15,8 +15,33 @@ module LavinMQ
...
+      shovel ||= create(name, config)
+      shovel
+    end
+
+    private def update_status_or_terminate(shovel, config)
+      case {shovel.state, config["state"]?.try &.as_s}
+      when {"Running", "Paused"}
+        shovel.pause
+        return shovel
+      when {"Paused", "Running"}
```
lukas8219, line 29: Moving this to Enum didn't break the test. So I'll revisit it

src/lavinmq/http/controller/shovels.cr
```diff
@@ -0,0 +1,66 @@
...
+              end
+              current_shovel.pause
+              context.response.status_code = 204
+            else
+              context.response.status_code = 404
+            end
+          end
+        end
+
+        #TODO 911: found a bug when we restart the LMQ instance - shovel gets back to running
```
lukas8219, line 48: This is an interesting topic. It seems like the `upsert` behavior is causing this - aka the restart of instance is `recreating` shovels - causing paused to actually be on running state.

lukas8219:
FYI: @spuun There's one thing I think I need more opinion - it is really obvious but y'all are the one's to decide. Currently, with the current `upsert|create` behavior from `shovel_store.cr` - whenever we restart LMQ, the shovel returns back to `Running`.

Gut feeling is -> Persist 'Paused' state across restarts. How do you guys feel?

(Also if there's any other `@` I can tag so I don't overload a single person with Q's and reviews - lmk!)

spec/api/shovels.cr
```diff
@@ -0,0 +1,80 @@
...
+          "reconnect-delay":5,
+          "ack-mode":"on-confirm"
+        }
+      })
+      response = http.put("/api/parameters/shovel/#{vhost_url_encoded}/name", body: body)
+      response.status_code.should eq 201
+      wait_for {
+        current_state = JSON.parse(http.get("/api/shovels/#{vhost_url_encoded}/name").body)["state"]
+        current_state === "Running"
+      }
```
lukas8219, line 26: Hmm I think I should migrate to the `should_eventually`

spuun:
> FYI: @spuun There's one thing I think I need more opinion - it is really obvious but y'all are the one's to decide. Currently, with the current `upsert|create` behavior from `shovel_store.cr` - whenever we restart LMQ, the shovel returns back to `Running`.
> 
> Gut feeling is -> Persist 'Paused' state across restarts. How do you guys feel?
We should persist the state, yes. I think one can 


> (Also if there's any other `@` I can tag so I don't overload a single person with Q's and reviews - lmk!)

spuun:
> @spuun Without diving more into import/export code - it sounds feasible - on current impl - if we allow "State" property to be added in the parameters.value - and also update it to reflect the shovel state.
> 
> But that sounds odd, atleast for me

Yes, that's what I think too.

> By "new config" you mean adding a new '.json' file with the shovels state? Much cleaner

At first I was thinking about a property in the parameter, but as we've agreed on that's odd. (Overall, that shovels and federation are generic entities like "parameters" is odd enough too me!)

I think the only thing we need to care about is if a shovel is paused, so maybe we can just introduce a `.paused` file in the vhost data dir for this? `shovel.<name>.paused`? If the file exists the shovel is paused, else not. It's also very easy to unpause a shovel "offline" then by just removing the file.

> If so, how do you feel about feature flagging this current PR and implement the config+persistence in a next PR?

Yes, I think it makes sense to make the persistence a new PR!

kickster97:
> Ok. is there any examples of feature flags in the code which I can replicate in this PR?

No, we have not worked a lot with feature flags like this before but maybe it would be time to start. 
something like: unless compiled with the -pause_shovel flag, pause/resume methods/endpoints do nothing, and the buttons are not rendered. 

another angle is that adding the .paused file is a straight forward implementation and could be merged very quickly after, so might not be worth the hassle to add all the feature flag logic.

lukas8219:
@kickster97  It's mostly up to you guys - this week I should have plenty of time to work on this feature and the following PR's while I stay in Stockholm.

I am not sure I am believer of `straightforward` as I often fail to predict complexity in projects 🤣

kickster97:
> @kickster97 It's mostly up to you guys - this week I should have plenty of time to work on this feature and the following PR's while I stay in Stockholm.
> 
> I am not sure I am believer of `straightforward` as I often fail to predict complexity in projects 🤣

I don't think a feature flag is necessary this time, I would say merge when ready and then let's make sure to document it carefully in the changelog and docs(we can do the website docs because they are maintained internally by 84codes) so it's clear that the state is not persisted in this commit 😊

src/lavinmq/http/controller/shovels.cr
```diff
@@ -0,0 +1,67 @@
...
+        get "/api/shovels" do |context, _params|
+          itrs = vhosts(user(context)).flat_map do |v|
+            v.shovels.each_value
+          end
+          page(context, itrs)
+        end
+
+        get "/api/shovels/:vhost" do |context, params|
+          with_vhost(context, params) do |vhost|
+            page(context, @amqp_server.vhosts[vhost].shovels.each_value)
```
spuun, line 18: We can and should use the `vhost` instance from `with_vhost` (so wrong already in main controller)
```suggestion
            page(context, vhost.shovels.each_value)
```

src/lavinmq/http/controller/shovels.cr
```diff
@@ -0,0 +1,67 @@
...
+          end
+        end
+
+        get "/api/shovels/:vhost/:name" do |context, params|
+          with_vhost(context, params) do |vhost|
+            shovel_name = params["name"]
+            if shovel = @amqp_server.vhosts[vhost].shovels[shovel_name]?
```
spuun, line 25:
```suggestion
            if shovel = vhost.shovels[shovel_name]?
```

src/lavinmq/http/controller/shovels.cr
```diff
@@ -0,0 +1,67 @@
...
+            else
+              context.response.status_code = 404
+            end
+          end
+        end
+
+        put "/api/shovels/:vhost/:name/pause" do |context, params|
+          with_vhost(context, params) do |vhost|
+            shovel_name = params["name"]
+            if current_shovel = @amqp_server.vhosts[vhost].shovels[shovel_name]?
```
spuun, line 36:
```suggestion
            if current_shovel = vhost.shovels[shovel_name]?
```

src/lavinmq/http/controller/shovels.cr
```diff
@@ -0,0 +1,67 @@
...
+              context.response.status_code = 404
+            end
+          end
+        end
+
+        # TODO 911: found a bug when we restart the LMQ instance - shovel gets back to running
+        put "/api/shovels/:vhost/:name/resume" do |context, params|
+          with_vhost(context, params) do |vhost|
+            shovel_name = params["name"]
+            if current_shovel = @amqp_server.vhosts[vhost].shovels[shovel_name]?
```
spuun, line 52:
```suggestion
            if current_shovel = vhost.shovels[shovel_name]?
```

spuun (commented):
I noticed an old mistake you inherited.

lukas8219:
@spuun Seems like `with_vhost` is actually just a parsing of the incoming request Path so not sure we can access any resources out of it

spuun:
> @spuun Seems like `with_vhost` is actually just a parsing of the incoming request Path so not sure we can access any resources out of it

Lol, i see... My mistake 🤦 

Something to refactor in the future.

lukas8219:
@spuun Created follow-up issue for Paused behavior + removed all TODO's and leftovers of doing shovel_store + parameter behavior

spuun (changes requested):
LGTM!

If you could fix what the JS linter is complaing on before merge, that would be great! I think `npx standard --fix static/js` is enough :)

PR #1106, opened 2025-05-18
feat(amqp/channels): support consumer_max_per_channel configuration to limit amount of consumers #938

### LINKED ISSUE #938: max_consumers configuration for a channel
Its good to be able to limit the amount of consumers you can create on a channel. 

a config value max_consumers that  will set the max amount of consumers you can create on a single channel.

lukas8219:
Adding a new configuration - at broker-level that ensures a limit for consumers per channel
Closes #938

Mainly new spec on the PR that ensures the Client throws a Channel exception when going over the threshold

### DIFF
--- a/spec/channel_spec.cr
+++ b/spec/channel_spec.cr
@@ -2,25 +2,36 @@ require "./spec_helper"
 require "amqp-client"
 
 describe LavinMQ::AMQP::Channel do
-  describe "Configurations" do
-    it "should respect consumer_max_per_channel" do
-      config = LavinMQ::Config.new
-      config.consumer_max_per_channel = 1
-      with_amqp_server(config: config) do |s|
-        connection = AMQP::Client.new(port: amqp_port(s)).connect
-        channel = connection.channel
-        channel.queue("test:queue:1")
-        channel.queue("test:queue:2")
-        channel.basic_consume("test:queue:1") { }
-        begin
-          channel.basic_consume("test:queue:2") { }
-          fail "Expected error when trying to surpass consumer_max_per_channel"
-        rescue ex : Exception
-          ex.should be_a(AMQP::Client::Channel::ClosedException)
-          error_message = ex.message || ""
-          error_message.should contain "RESOURCE_ERROR"
-        end
+  it "should respect consumer_max_per_channel config" do
+    config = LavinMQ::Config.new
+    config.max_consumers_per_channel = 1
+    with_amqp_server(config: config) do |s|
+      connection = AMQP::Client.new(port: amqp_port(s)).connect
+      channel = connection.channel
+      channel.queue("test:queue:1")
+      channel.queue("test:queue:2")
+      channel.basic_consume("test:queue:1") { }
+      begin
+        channel.basic_consume("test:queue:2") { }
+        fail "Expected error when trying to surpass consumer_max_per_channel"
+      rescue ex : Exception
+        ex.should be_a(AMQP::Client::Channel::ClosedException)
+        error_message = ex.message || ""
+        error_message.should contain "RESOURCE_ERROR"
       end
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


spec/channel_spec.cr
```diff
@@ -0,0 +1,26 @@
+require "./spec_helper"
+require "amqp-client"
+
+describe LavinMQ::AMQP::Channel do
+  describe "Configurations" do
+    it "should respect consumer_max_per_channel" do
+      config = LavinMQ::Config.new
+      config.consumer_max_per_channel = 1
+      with_amqp_server(config: config) do |s|
+        puts "Creating resources"
```
lukas8219, line 10: TODO i'll remove this in next commit

src/lavinmq/config.cr
```diff
@@ -66,6 +66,7 @@ module LavinMQ
     property max_deleted_definitions = 8192         # number of deleted queues, unbinds etc that compacts the definitions file
     property consumer_timeout : UInt64? = nil
     property consumer_timeout_loop_interval = 60 # seconds
+    property consumer_max_per_channel = UInt64::MAX
```
lukas8219, line 69: Not sure if we should do `0` as some parallel to `infinity` or just use this MAX. I mean, hard to believe someone will ever reach such number of consumers??

src/lavinmq/amqp/channel.cr
```diff
@@ -340,6 +340,10 @@ module LavinMQ
       end
 
       def consume(frame)
+        if (@consumers.size + 1) > Config.instance.consumer_max_per_channel
```
carlhoerberg, line 343:
```suggestion
        if @consumers.size >= Config.instance.consumer_max_per_channel
```

src/lavinmq/amqp/channel.cr
```diff
@@ -340,6 +340,10 @@ module LavinMQ
...
+          @client.send_not_allowed(frame, "Channel consumers #{@consumers.size} reached max consumers #{Config.instance.consumer_max_per_channel}")
```
carlhoerberg, line 344:
```suggestion
          @client.send_not_allowed(frame, "Max #{Config.instance.consumer_max_per_channel} consumers per channel reached")
```

src/lavinmq/amqp/channel_reply_code.cr
```diff
@@ -7,6 +7,7 @@ module LavinMQ
       NOT_FOUND           = 404
       RESOURCE_LOCKED     = 405
       PRECONDITION_FAILED = 406
+      NOT_ALLOWED         = 530
```
carlhoerberg, line 10: I think "506 - resource-error" might be a better fit:
> The server could not complete the method because it lacked sufficient
> resources. This may be due to the client creating too many of some type
> of entity.

src/lavinmq/config.cr
carlhoerberg: Or have a reasonable default limit? I don't anyone will/should reach even UInt16::MAX, 65K, but even that might be crazy high. Not sure how much memory one consumer uses?

src/lavinmq/amqp/channel_reply_code.cr
lukas8219: Used 530 to follow Rabbit implementation. But I'll do 506, makes sense

src/lavinmq/config.cr
lukas8219: Agreed, considering this limit is by channel it makes sense to limit by a lesser value. Altough the price for such decision (2 vs 8 bytes) weighted against limiting (as weird as it might be) clients implementation - seems to payoff?

src/lavinmq/amqp/channel.cr
lukas8219: oof. great catch, leftover from brainrot.

was tired from brazil -> italy flight

PR #1107, opened 2025-05-19
Use `Hash` instead of `AMQP::Table` in headers exchange

spuun:
The header exchange stores stores bindings in a `Hash` with the binding arguments as key. This works as long as the hash contains eight or less elements, because then the hash will only use the `==` operator when comparing keys. When the hash has more than eight elements it will rely on `#hash(hasher)` instead.

Binding arguments is a `AMQP::Table` which doesn't implement `#hash(hasher)` which means that the hash won't treat two equal (==) tables as the same key meaning, that the hash may container many equal keys (the "same" key multiple times). This also makes it impossible to delete entries etc.

This PR will change header exchange to store its bindings as `Hash` which can be used as a key in a `Hash`. Since bind and unbind isn't in the hot path this is fine.

This PR will also remove the merging of exchange arguments with bind arguments, which must be seen as a bug (or an odd and undocmented feature).

Specs

### DIFF
--- a/spec/message_routing_spec.cr
+++ b/spec/message_routing_spec.cr
@@ -197,16 +197,56 @@ module MessageRoutingSpec
         "user"    => "test",
       })
 
+      describe "bind" do
+        it "shouldn't care about argument order" do
+          q = LavinMQ::AMQP::Queue.new(vhost, "q")
+          args1 = {"foo": "bar", "f00": "baz"}
+          args2 = {"f00": "baz", "foo": "bar"}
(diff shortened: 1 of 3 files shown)


spuun:
> But `AMQP::Table` does implement `#hash(hasher)`: https://github.com/cloudamqp/amq-protocol.cr/blob/93b79e89057be48302ba983d28b4b65dc741ded5/src/amq/protocol/table.cr#L171-L185

Oh, I wonder how I missed that 🤦
So, we just need to make a new release of amq-protocol then!

PR #1109, opened 2025-05-20
Dont log twice when user not found 

viktorerlingsson:
Rescues `KeyError` in `BasicAuthenticator` so we dont log "Missing hash key" since we already log "User abc not found"

```
2025-05-20 13:46:42.090 lmq.amqp.connection_factory[address: "1.2.3.4:12345"] User "abc" not found 
2025-05-20 13:46:41.654 lmq.auth.handler Basic authentication failed: Missing hash key: "abc"
```

Manual

### DIFF
--- a/src/lavinmq/auth/authenticators/basic.cr
+++ b/src/lavinmq/auth/authenticators/basic.cr
@@ -10,6 +10,7 @@ module LavinMQ
       def authenticate(username : String, password : Bytes) : User?
         user = @users[username]
         return user if user && user.password && user.password.not_nil!.verify(password)
+      rescue KeyError # User not found
       rescue ex : Exception
         Log.error { "Basic authentication failed: #{ex.message}" }
       end


src/lavinmq/auth/authenticators/basic.cr
```diff
@@ -10,6 +10,7 @@ module LavinMQ
       def authenticate(username : String, password : Bytes) : User?
         user = @users[username]
         return user if user && user.password && user.password.not_nil!.verify(password)
+      rescue KeyError # User not found
```
spuun, line 13: What about 
```suggestion
        user = @users[username]?
        return user if user && user.password && user.password.not_nil!.verify(password)
```

spuun (commented):
See inline comment.

src/lavinmq/auth/authenticators/basic.cr
viktorerlingsson: Yeah, maybe even

```
if user = @users[username]?
  return user if user.password && user.password.not_nil!.verify(password)
end
```

spuun (approved):

src/lavinmq/auth/authenticators/basic.cr
```diff
@@ -8,8 +8,9 @@ module LavinMQ
       end
 
       def authenticate(username : String, password : Bytes) : User?
-        user = @users[username]
-        return user if user && user.password && user.password.not_nil!.verify(password)
+        if user = @users[username]?
+          return user if user.password && user.password.not_nil!.verify(password)
+        end
```
carlhoerberg, line 13:
```suggestion
        if user = @users[username]?
          if p = user.password
            if p.verify(password)
              return user
            end
          end
        end
```

PR #1119, opened 2025-05-22
bugfix: channel.basic_get no longer increments deliver_count

viktorerlingsson:
Fixes a bug where `basic_get` (and probably `basic_return` as well) would increment `deliver_count`, because `deliver_count` was incremented in `deliver()`, which is used by both `basic_get` and `consume`. 
This fixes it by increasing `deliver_count` in the `deliver_loop` on the consumer instead. 

Run included spec

### DIFF
--- a/spec/api/channels_spec.cr
+++ b/spec/api/channels_spec.cr
@@ -69,7 +69,7 @@ describe LavinMQ::HTTP::ChannelsController do
           ch.prefetch(1)
 
           ch.basic_get(q.name, no_ack: false)
-          q.subscribe {}
+          q.subscribe { }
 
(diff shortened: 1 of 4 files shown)


viktorerlingsson:
> I see now that `basic_return` also increments `deliver_count`, is that right?

Yeah, I don't think it should. `deliver_count` should be "Count of messages delivered in acknowledgement mode to consumers.", so it should only be incremented in deliver_loop as far as I can tell? (And we should have a `deliver_no_ack_count` for messages delivered in no-ack mode)

kickster97 (approved):
looks good to me!

PR #1124, opened 2025-05-23
Add `deliver_no_ack` metric

### LINKED ISSUE #1123: Add and track deliver_no_ack
We should track `deliver_no_ack` for queues, channels and vhosts and provide it in `message_stats`.

> `deliver_no_ack` Count of messages delivered in no-acknowledgement mode to consumers.

kickster97:
Adds and tracks deliver_no_ack in queues, vhosts and channels

Fixed #1123 

Publish and consume unacked messages and check graphs and api endpoints to see that the metric increases.

### DIFF
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -40,7 +40,7 @@ module LavinMQ
       @tx = false
       @next_msg_body_tmp = IO::Memory.new
 
-      rate_stats({"ack", "get", "publish", "deliver", "deliver_get", "redeliver", "reject", "confirm", "return_unroutable"})
+      rate_stats({"ack", "get", "publish", "deliver", "deliver_get", "deliver_no_ack", "redeliver", "reject", "confirm", "return_unroutable"})
 
       Log = LavinMQ::Log.for "amqp.channel"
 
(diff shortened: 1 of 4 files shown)


viktorerlingsson:
First, there are a couple other PR's up that makes changes to related metrics and how/where they are added. I think maybe it's best to get them merged first ( #1099 and #1119 ) so we get those bugfixes in. 

Secondly, `deliver_count` and `deliver_no_ack_count` should not both be updated at the same time, so where we increment, it should probably be something like
```
if no_ack
  @deliver_no_ack_count.add(1)
else
  @deliver_count.add(1)
end
```

And thirdly, `deliver_no_ack_count` should only be updated when consuming (so basically where we update `deliver_count`), and not when doing basic.get (so not where we update `get_count`). 

(`deliver_get_count` is a summary of `get_count + get_no_ack_count + deliver_count + deliver_no_ack_count`)

kickster97:
> First, there are a couple other PR's up that makes changes to related metrics and how/where they are added. I think maybe it's best to get them merged first ( #1099 and #1119 ) so we get those bugfixes in.
> 
> Secondly, `deliver_count` and `deliver_no_ack_count` should not both be updated at the same time, so where we increment, it should probably be something like
> 
> ```
> if no_ack
>   @deliver_no_ack_count.add(1)
> else
>   @deliver_count.add(1)
> end
> ```
> 
> And thirdly, `deliver_no_ack_count` should only be updated when consuming (so basically where we update `deliver_count`), and not when doing basic.get (so not where we update `get_count`).
> 
> (`deliver_get_count` is a summary of `get_count + get_no_ack_count + deliver_count + deliver_no_ack_count`)
> 
> Edit: oh, and we should add this to a spec as well 🙃

Thanks for the feedback :) converting this to draft until the other PRs are merged!

PR #1130, opened 2025-05-28
Respect max-hops in federation

### LINKED ISSUE #206: Support max-hops in federation
Federation does not respect max-hops, it should.

spuun:

- Bindings are not propagated beyond max-hops
- Messages are not traveling more than max-hops (only exchange federation)

Fixes #206 

Run specs


spec/upstream_spec.cr
```diff
@@ -683,6 +737,208 @@ describe LavinMQ::Federation::Upstream do
...
+            # vhost v9.
+            4.to(9) do |i|
+              fe = s.vhosts["v#{i}"].exchanges["fe"]
+              fe.bindings_details.size.should eq 1
+              fe_bk = fe.bindings_details.first.binding_key
+              args = fe_bk.arguments.should_not be_nil
+              x_bound_from = args["x-bound-from"].should be_a(Array(AMQ::Protocol::Field))
+              # in v4 there should be 6 entries
+              # in v5 there should be 5 entries etc
+              x_bound_from.size.should eq(9 - i + 1), "Expected #{6 - i + 1} entries in v#{i} but got #{x_bound_from.size}"
```
viktorerlingsson: Should this be `"Expected #{9 - i + 1}..."` (instead of `"Expected #{6 - i + 1}..."`) to match the comparison?

src/lavinmq/federation/link.cr
```diff
@@ -285,23 +285,38 @@ module LavinMQ
...
             @upstream.stop_link(@federated_ex)
           in .bind?
-            with_consumer_ex do |ex|
-              b = data_as_binding_details(data)
-              args = b.arguments || ::AMQP::Client::Arguments.new
-              ex.bind(@upstream_exchange, b.routing_key, args: args)
+            b = data_as_binding_details(data)
+            # Must clone args here, else we'll add stuff to
+            # the arguments to the source exchange's binding
+            args = b.arguments.clone || ::AMQP::Client::Arguments.new
```
viktorerlingsson: maybe move this to `update_bound_from?` so we don't have to do it (and comment on why it's being done) in multiple places?

spuun: Yeah,  I don't remember why i decided not to do that. But since `update_bound_from?` returns updated arguments I guess it's ok that will clone the given arguments. 👍

PR #1133, opened 2025-06-01
Use more refined synchronization primitives

carlhoerberg:
Use the sync library by @ysbaddaden with nsync inspired synchronization primitives.

First @followers was protected using a RWLock, yielding a big performance improvment, around 30% higher throughput, when clustering is enabled.

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -9,6 +9,7 @@ require "../mfile"
 require "crypto/subtle"
 require "lz4"
 require "../etcd"
+require "sync/shared"
 
 module LavinMQ
   module Clustering
@@ -28,10 +29,10 @@ module LavinMQ
       Log = LavinMQ::Log.for "clustering.server"
 
       @lock = Mutex.new(:unchecked)
-      @followers = Array(Follower).new(4)
+      @followers : Sync::Shared(Array(Follower)) = Sync::Shared.new(Array(Follower).new(4))
       @password : String
       @files = Hash(String, MFile?).new
-      @dirty_isr = true
+      @dirty_isr = Atomic(Bool).new true
       @id : Int32
       @config : Config
 
@@ -122,9 +123,7 @@ module LavinMQ
       end
 
       def followers : Array(Follower)
-        @lock.synchronize do
-          @followers.dup # for thread safety
-        end
+        @followers.read &.dup
       end
 
       def password : String
@@ -157,22 +156,22 @@ module LavinMQ
           Log.error { "Disconnecting follower with the clustering id of the leader" }
           return
         end
-        if stale_follower = @followers.find { |f| f.id == follower.id }
+        if stale_follower = @followers.read &.find { |f| f.id == follower.id }
           Log.error { "Disconnecting stale follower with id #{follower.id.to_s(36)}" }
           stale_follower.close
         end
         follower.full_sync # sync the bulk
-        @lock.synchronize do
+        @followers.write do |followers|
           follower.full_sync # sync the last
-          @followers << follower
-          update_isr
+          followers << follower
         end
+        update_isr
         begin
           follower.action_loop
         ensure
-          @lock.synchronize do
-            @followers.delete(follower)
-            @dirty_isr = true
+          @followers.write do |followers|
+            followers.delete(follower)
+            @dirty_isr.set true
           end
         end
       rescue ex : AuthenticationError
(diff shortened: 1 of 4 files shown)


src/lavinmq/clustering/server.cr
```diff
@@ -187,36 +183,36 @@ module LavinMQ
         follower.try &.close
       end
 
-      private def update_isr
+      private def update_isr(followers)
         isr_key = "#{@config.clustering_etcd_prefix}/isr"
         ids = String.build do |str|
```
ysbaddaden: Doing less within a lock usually helps improve performance:

- `isr_key` looks like a constant?
- `String.build` could pre-allocate a string with enough capacity: `(followers.size + 1) * 37`.

src/lavinmq/clustering/server.cr
```diff
@@ -27,11 +28,10 @@ module LavinMQ
       include Replicator
       Log = LavinMQ::Log.for "clustering.server"
 
-      @lock = Mutex.new(:unchecked)
-      @followers = Array(Follower).new(4)
+      @followers : Sync::Shared(Array(Follower)) = Sync::Shared.new(Array(Follower).new(4))
       @password : String
       @files = Hash(String, MFile?).new
-      @dirty_isr = true
+      @dirty_isr = Atomic(Bool).new true
```
ysbaddaden, line 34: I don't think you need `:acquire` / `:release` semantics for it: you only need `#get` and `#set` to be atomic by themselves, other memory accesses don't need to be synchronized along with it (that's the lock's job). The `:relaxed` memory order should be enough for all cases

src/lavinmq/clustering/server.cr
```diff
@@ -187,36 +183,36 @@ module LavinMQ
...
       end
 
       private def each_follower(& : Follower -> Nil) : Nil
-        @lock.synchronize do
-          update_isr if @dirty_isr
-          @followers.each do |f|
+        # remove stale follower from ISR set as new changes are coming
+        any_closed = false
+        @followers.read do |followers|
+          update_isr(followers) if @dirty_isr.get(:acquire)
```
ysbaddaden: Multiple readers can acquire, so multiple readers can see `dirty_isr` as true, and update the isr in parallel, until one eventually sets `dirty_isr` to false. A `swap` could make sure that only one will update at a time:

```suggestion
          update_isr(followers) if @dirty_isr.swap(false, :relaxed)
```

Then move ` @dirty_isr.set(false)` out of `#update_isr` and into `#handle_socket`?

src/lavinmq/clustering/server.cr
carlhoerberg: Thanks!

src/lavinmq/clustering/server.cr
carlhoerberg: Ahh! Of course

src/lavinmq/clustering/server.cr
carlhoerberg: Smart!

carlhoerberg:
@kickster97 observe the changes to the MQTT RetainStore: https://github.com/cloudamqp/lavinmq/pull/1133/commits/02a6755eab23b1e8f9bfd751f4837235c3010479

carlhoerberg:
This is the only commit I'm unsure about, by applying polices without a fiber/spawn, those operations might take a long time, eg. max-length limiting a very long queue. 
https://github.com/cloudamqp/lavinmq/pull/1133/commits/eba559f76a9bb40be346c71af506e992590bc344

viktorerlingsson:
> This is the only commit I'm unsure about, by applying polices without a fiber/spawn, those operations might take a long time, eg. max-length limiting a very long queue. [eba559f](https://github.com/cloudamqp/lavinmq/commit/eba559f76a9bb40be346c71af506e992590bc344)

Yeah, this is probably pretty risky. Policies can take a pretty long time to apply if the queue is long. Could this be handled with a lock or something on a queue while applying policies? I feel like you shouldn't be able to do anything with a queue while a policy is being applied.

carlhoerberg:
Policies are only applied via HTTP API, not the AMQP protocol. But yes, client's might timeout..

viktorerlingsson (approved):
Code looks good to me 👍 Should be pretty safe to implement with the "feature flag" IMO. But maybe @spuun wants to check this as well? I think he knows MT and locks better than me 🙂

carlhoerberg:
Main
```sh
✗ bin/lavinmqperf amqp queue-churn
create/delete transient queue   1.24k (805.19µs) (± 7.44%)  208B/op        fastest
  create/delete durable queue 427.07  (  2.34ms) (± 4.94%)  323B/op   2.91× slower
```
CoW:
```sh
✗ bin/lavinmqperf amqp queue-churn
create/delete transient queue 785.88  (  1.27ms) (± 7.54%)  207B/op        fastest
  create/delete durable queue 326.81  (  3.06ms) (± 6.46%)  318B/op   2.40× slower
```

RWLock:
```sh
✗ bin/lavinmqperf amqp queue-churn
create/delete transient queue   1.53k (651.68µs) (± 5.47%)  208B/op        fastest
  create/delete durable queue 483.00  (  2.07ms) (± 5.71%)  322B/op   3.18× slower
```

PR #1138, opened 2025-06-04
use a BoolChannel for @is_leader with a timeout to make sure it is set

viktorerlingsson:
Use a `BoolChannel` for @`is_leader` with a timeout to make sure its set before checking its value when a node is trying to become leader.

Fixes a bug where sometimes a node would raise an error `FATAL lmq.clustering.controller Another node in the cluster is advertising the same URI` when trying to become leader, because of a timing issue where @is_leader had not been set before it needed to be used. 

Manual

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -23,12 +23,13 @@ class LavinMQ::Clustering::Controller
   # The block will be yielded when the controller's prerequisites for a leader
   # to start are met, i.e when the current node has been elected leader.
   # The method is blocking.
+  @is_leader = BoolChannel.new(true)
   def run(&)
     lease = @lease = @etcd.lease_grant(id: @id)
     spawn(follow_leader, name: "Follower monitor")
     wait_to_be_insync
     @etcd.election_campaign("#{@config.clustering_etcd_prefix}/leader", @advertised_uri, lease: @id) # blocks until becoming leader
-    @is_leader = true
+    @is_leader.set(true)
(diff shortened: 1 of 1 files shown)


src/lavinmq/clustering/controller.cr
```diff
@@ -23,12 +23,14 @@ class LavinMQ::Clustering::Controller
   # The block will be yielded when the controller's prerequisites for a leader
   # to start are met, i.e when the current node has been elected leader.
   # The method is blocking.
+  @is_leader = BoolChannel.new(true)
```
carlhoerberg: shouldn't it be false by default?

carlhoerberg:
```suggestion
  @is_leader = BoolChannel.new(false)
```

src/lavinmq/clustering/controller.cr
```diff
@@ -84,10 +86,11 @@ class LavinMQ::Clustering::Controller
         next
       end
       if uri == @advertised_uri # if this instance has become leader
-        if @is_leader
+        select
+        when @is_leader.when_true.receive
           Log.debug { "Is leader, don't replicate from self" }
           return
```
carlhoerberg, line 93: We won't need it any more, saves an extra fiber. 
```suggestion
          @is_leader.close
          return
```

src/lavinmq/clustering/controller.cr
viktorerlingsson: Yes, my bad 👍

PR #1155, opened 2025-06-17
frontend: use tagged template literal instead of encodeURIComponent

spuun:
The frontend code is sprinkled with `encodeURIComponent` to construct URLs. It makes the code cluttered and it's easy to forget to encode what's necessary.

This PR will replace all `encodeURIComponent` with [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) and a custom `HTTP.url` tag that will automatically encode all expressions in the template literal.

```javascript
const url = HTTP.url`/path/${myVar}/slug/${otherVar}`
// is equivalent to
const url = `/path/${encodeURIComponent(myVar)}/slug/${encodeURIComponent(otherVar)}`
```

I'm not sure about the name, maybe it should say something about encode?

It's possible to make an expression not being encoded by using the `HTTP.noencode` (rename to `raw`?) method like 

```javascript
const url = HTTP.url`/path/${HTTP.noencode(myVar)}/slug/${otherVar}`
```

Specs and manually

### DIFF
--- a/static/js/channel.js
+++ b/static/js/channel.js
@@ -6,8 +6,7 @@ import * as Chart from './chart.js'
 import { DataSource } from './datasource.js'
 
 const channel = new URLSearchParams(window.location.hash.substring(1)).get('name')
-const urlEncodedChannel = encodeURIComponent(channel)
-const channelUrl = 'api/channels/' + urlEncodedChannel
+const channelUrl = HTTP.url`api/channels/${channel}`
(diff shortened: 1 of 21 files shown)


baelter (approved):

viktorerlingsson (approved):
LGTM! 👍 

Not really loving having to use `noencode()` in some places, but IMO it's much better than all the usage of `encodeURIComponent` before!

PR #1161, opened 2025-06-19
Don't reset x-cache-ttl when expiring msg with delayed exchange

### LINKED ISSUE #1153: Deduplication with delayed messages
**Describe the bug**

When using the [delayed exchange](https://lavinmq.com/documentation/delayed-message) in combination with [exchange-level message deduplication](https://lavinmq.com/documentation/message-deduplication), the cache TTL for deduplication gets doubled. This might be due to the round trip of delayed messages through the exchange (exchange --> dead letter queue --> exchange).

**Describe your setup**

I am using the following setup:
* LavinMQ 2.4.0 (released 3 days ago) as published on [docker hub](https://hub.docker.com/r/cloudamqp/lavinmq)
* amqplib 0.10.8 (currently latest) from [NPM registry](https://www.npmjs.com/package/amqplib)
* NodeJS v22.15.0 

**How to reproduce**

Please find [attached](https://github.com/user-attachments/files/20741229/dedupe.tar.gz) a minimal reproducer written in NodeJS.
1. untar to a directory of your liking
2. change to that directory
3. `npm install`
4. `npm run lavinmq`
5. `npm run test`
6. observe test logs

**Expected behavior**

Using message deduplication alone will process the first received message, and then drop the future messages in the deduplication window (cache TTL). Imagine however a CI/CD system triggered by code changes: we do not want to trigger CI/CD too often, but we also want to deploy the latest version of the code, i.e. the last message in the deduplication window.

To handle this scenario, I tried to combine delayed messages with deduplication, so that 
1. the first message arrives, and is delayed by, say, 5 minutes
2. deduplication is set to the same 5 minutes, so if other messages arrive during this window, they are ignored
3. when the delay expires, the first message is processed (at the leading edge of the delay window) thus deploying the latest code changes

What I am observing is instead that
1. the first message arrives, and is correctly delayed
2. if more messages arrive within 5 minutes they are correctly dropped
3. when the delay expires, the first message is processed
4. if more message arrive in the following 5 minutes, they are still dropped
5. deduplication cache expires after 10 minutes, and new messages are processed as above

viktorerlingsson:
Previously, when using deduplication and delayed exchange together, the `x-cache-ttl` would be reset when a message was expired and re-published (if cache-ttl and delay was the same), meaning the `cache_ttl` was effectively more like `cache_ttl + msg_delay`. This changes the behavior so that a delayed message will not trigger deduplication by calling `exchange.route_msg` instead of `vhost.publish`. 

Also refactors `exchange.publish` to separate the msg routing to a separate `route_msg` function (was `do_publish`) that just does the msg routing. 

Fixes #1153 

Run spec

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -196,43 +196,50 @@ module LavinMQ
 
       def publish(msg : Message, immediate : Bool,
                   queues : Set(LavinMQ::Queue) = Set(LavinMQ::Queue).new,
-                  exchanges : Set(LavinMQ::Exchange) = Set(LavinMQ::Exchange).new) : UInt32
+                  exchanges : Set(LavinMQ::Exchange) = Set(LavinMQ::Exchange).new) : Bool
         @publish_in_count.add(1, :relaxed)
         if d = @deduper
           if d.duplicate?(msg)
             @dedup_count.add(1, :relaxed)
-            return 0u32
+            return false
           end
           d.add(msg)
         end
-        count = do_publish(msg, immediate, queues, exchanges)
-        @unroutable_count.add(1, :relaxed) if count.zero?
-        @publish_out_count.add(count, :relaxed)
-        count
-      end
-
-      private def do_publish(msg : Message, immediate : Bool, queues : Set(LavinMQ::Queue), exchanges : Set(LavinMQ::Exchange)) : UInt32
-        headers = msg.properties.headers
-        if should_delay_message?(headers)
+        if should_delay_message?(msg.properties.headers)
           if q = @delayed_queue
             q.publish(msg)
-            return 1u32
+            @publish_out_count.add(1, :relaxed)
+            return true
           else
-            return 0u32
+            @unroutable_count.add(1, :relaxed)
+            return false
           end
         end
+        route_msg(msg, immediate, queues, exchanges)
+      end
+
+      def route_msg(msg : Message) : Bool
+        route_msg(msg, false, Set(LavinMQ::Queue).new, Set(LavinMQ::Exchange).new)
+      end
+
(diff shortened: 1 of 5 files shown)


spuun:
I wonder if this is the best solution or if some changes to `Exchange#publish` is the way to go. I'm thinking that this change add an unexpected behavior to `Deduper`? E.g. it will shift on entry from the cache, even though the message shouldn't be added because ttl = 0.

Maybe change
https://github.com/cloudamqp/lavinmq/blob/fa5fb2665f2b76702a03a81c9e6e1108d3c82e20/src/lavinmq/amqp/exchange/exchange.cr#L201

to 
```crystal
if !has_been_delayed?(msg.properties.headers) && (d = @deduper)
```
and add something like
```crystal
private def has_been_delayed?(msg)
    # probably more nil checks needed...
    headers["x-death"]?.try &.first["queue"]? == "amq.delayed.#{@name}"
end
```

viktorerlingsson:
> I wonder if this is the best solution or if some changes to `Exchange#publish` is the way to go. I'm thinking that this change add an unexpected behavior to `Deduper`? E.g. it will shift on entry from the cache, even though the message shouldn't be added because ttl = 0.
> 

Okay, did some refactoring after our chat. Refactored `Exchange#publish` and `Exchange#do_publish` to `publish`/`route_msg` instead, where `publish` handles delay and deduplication, and `route_msg` just routes the msg to any matching queues. 

Reverted the earlier changes in `Deduplication`, instead calling `exchange#route_msg` directly when expiring a msg from a `delayed_exchange_queue`. 

WDYT?

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -196,43 +196,49 @@ module LavinMQ
...
+          return false
+        end
 
-        count = 0u32
+        routed = false
         queues.each do |queue|
           if queue.publish(msg)
-            count += 1u32
+            routed = true
+            @publish_out_count.add(1)
```
spuun: I think it's worth having the local count variable and then add it to `@publish_out_count` after the loop.

PR #1163, opened 2025-06-23
Speed up boot by storing message count in a file next to the message segment 

carlhoerberg:
## Summary
Improves LavinMQ boot performance by storing message metadata alongside message segments:

- Creates `.meta` files for each message segment containing message count, avoiding the need to read entire segments during boot
- For stream queues, stores the offset and timestamp of the first message in each segment  
- Automatically deletes `.meta` files when corresponding message segments are removed
- Uses metadata to build segment indexes efficiently during startup

## Changes
- **Message segments**: Write `.meta` files with message count for faster boot times
- **Stream queues**: Store first message offset and timestamp in metadata
- **Cleanup**: Remove `.meta` files when segments are deleted
- **Index building**: Use metadata instead of reading full segments during startup

## Performance Impact
Significantly reduces boot time by eliminating the need to read and parse message segments just to count messages during initialization.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/spec/msg_segment_meta_files_spec.cr
+++ b/spec/msg_segment_meta_files_spec.cr
@@ -1,0 +1,316 @@
+require "./spec_helper"
+
+describe "Message segment metadata files" do
+  describe "Metadata file creation" do
+    it "creates .meta file when segment becomes full" do
+      with_amqp_server do |s|
+        vhost = s.vhosts.create("test_vhost")
+        with_channel(s, vhost: vhost.name) do |ch|
+          q = ch.queue("meta_test")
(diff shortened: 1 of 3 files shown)


viktorerlingsson:
Nice! 👍 

For streams it would also be very useful to write the offset (and timestamp) of the first message in the segment to this file, so we can build our indexes without having to open the segments. Maybe the file could be called .meta or something to account for that?

carlhoerberg:
@viktorerlingsson Good idea, implemented

carlhoerberg:
Todo: delete the .meta files when segment files are

PR #1168, opened 2025-06-25
CI: allow java client tests to fail again

dentarg:
We allowed them to fail until e35d582d9daa6862a8e01ce84ccf78f7cbbf1b33

There was some effort making them pass, but I don't think they have not been passing for quite some time now.

I don't think they will start to pass without a directed effort

> Tests run: 611, Failures: 3, Errors: 9, Skipped: 50

so until that happens, I think it is a better look for the LavinMQ project with green CI.


viktorerlingsson:
Agreed - I think we should either make sure they pass or allow them to fail. Having them just fail and leaving it like that doesn't provide any extra value IMO.

PR #1169, opened 2025-06-25
Fix: default lavinmqperf protocol to amqp if not specified

kickster97:

If you do not specify a protocol when starting lavinmqperf, it will default to amqp.  

run lavinmqperf throughput

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -33,9 +33,9 @@ module LavinMQPerf
   when "mqtt"
     mode = ARGV.shift?
     case mode
-    when "throughput"       then MQTT::Throughput.new.run
-    when /^.+$/             then Perf.new.run([mode.not_nil!])
-    else                         abort Perf.new.mqtt_banner
+    when "throughput" then MQTT::Throughput.new.run
+    when /^.+$/       then Perf.new.run([mode.not_nil!])
+    else                   abort Perf.new.mqtt_banner
     end
   when /^.+$/
     Perf.new.run([protocol.not_nil!])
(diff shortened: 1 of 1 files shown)


kickster97:
also added the `/^.+$/` case for mqtt

src/lavinmqperf.cr
```diff
@@ -12,6 +12,9 @@ Signal::SEGV.reset # Let the OS generate a coredump
 Log.setup_from_env
 
 module LavinMQPerf
+  protocol = ARGV[0]?
```
viktorerlingsson: This isn't used, right?

viktorerlingsson (approved):

src/lavinmqperf.cr
```diff
@@ -12,6 +12,9 @@ Signal::SEGV.reset # Let the OS generate a coredump
...
+  # Default to 'amqp' if the first argument is not a protocol or an option
+  ARGV.unshift("amqp") if ARGV[0]?.try { |a| a != "amqp" && a != "mqtt" && !a.starts_with?("-") } || ARGV.empty?
   protocol = ARGV.shift? || "amqp"
```
spuun: This feels overcomplicated. What about
```suggestion
protocols = {"amqp", "mqtt"}
protocol = protocols.includes?(ARGV[0]?) ? ARGV.shift : "amqp"
```

kickster97: nice! :))

src/lavinmqperf.cr
kickster97: correct hehe i'll fix

PR #1170, opened 2025-06-25
bugfix: Streams - read msg files if there is no meta files, also read index data from meta files if they exist

viktorerlingsson:
Fixes two bugs in #1163 : 
- Stream queues without meta files did not read segments, added `super` to `produce_metadata` to make sure that happens properly.
- Streams with metadata did not read `offset_index` and `timestamp_index` from meta files. 

-

### DIFF
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -342,12 +342,35 @@ module LavinMQ::AMQP
       end
 
       private def produce_metadata(seg, mfile)
+        super
         msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
         @offset_index[seg] = offset_from_headers(msg.properties.headers)
         @timestamp_index[seg] = msg.timestamp
       rescue IndexError
         @offset_index[seg] = @last_offset
         @timestamp_index[seg] = RoughTime.unix_ms
+      end
+
+      private def read_metadata_file(seg, mfile)
+        File.open("#{mfile.path}.meta") do |file|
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -342,13 +342,36 @@ module LavinMQ::AMQP
       end
 
       private def produce_metadata(seg, mfile)
+        super
```
carlhoerberg, line 345: Good catch!

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -342,13 +342,36 @@ module LavinMQ::AMQP
...
         msg = BytesMessage.from_bytes(mfile.to_slice + 4u32)
         @offset_index[seg] = offset_from_headers(msg.properties.headers)
         @timestamp_index[seg] = msg.timestamp
       rescue IndexError
         @offset_index[seg] = @last_offset
         @timestamp_index[seg] = RoughTime.unix_ms
       end
+
+      private def read_metadata_file(seg, mfile)
```
carlhoerberg, line 354: wow, was convinced i did that but no :shrug: haha. Thanks!

carlhoerberg (approved):

src/lavinmq/amqp/queue/stream_queue_message_store.cr
viktorerlingsson: Yeah, honestly I was pretty sure you did as well - I feel like I thought about that when reviewing, but maybe not! 🤷

PR #1173, opened 2025-06-25
Consumer fairness

carlhoerberg:


carlhoerberg:
With these changes all java client test passes on my local machine, CI still fails for some reason..

src/lavinmq/amqp/channel.cr
```diff
@@ -674,11 +669,16 @@ module LavinMQ
       end
 
       protected def next_delivery_tag(queue : Queue, sp, no_ack, consumer) : UInt64
-        @unack_lock.synchronize do
-          tag = @delivery_tag &+= 1
-          @unacked.push Unack.new(tag, queue, sp, consumer, RoughTime.monotonic) unless no_ack
-          tag
+        tag = @delivery_tag.add(1, :relaxed)
```
lukas8219, line 672: just couple q's while glaring at this. i love the idea of lock-free stuff and it coincidentally matched the topic which i was studying last Friday (mmr ordering while cloning the `pthread_once` syscall - [for ref of my notes](https://github.com/lukas8219/linux-api-exercises-notes/commit/75a98e68668cc0cbe4623e5aeb385dca4e66c7f1#diff-e402ca3f76c5834f1b8fb1bdc4dff59ab890ae08b61cb2625b849ee4230f6080R5-R31))

First thing that pops into my mind is either 1) i didnt understand mmr ordering properly or 2) is the global QoS or even, in the above case `delivery_tag` something that CAN be inconsistent?

Hypothetically 2 running threads (or Fibers in Crystal?) may return the same value (e.g via Atomic.add) if under high CPU pressure. Am I completely nuts here?

carlhoerberg: Yes, when MT is enabled the `Client#read_loop` (for `basic_get` requests) and multiple `Consumer#deliver_loop` fibers can potentially run in different threads, and hence the `@delivery_tag` has to be MT safely incremented. `Atomic#add(1)` should not be able to return the same value for multiple threads, that's the guarantee that we rely on here.

lukas8219: So that’s what’s still a bit unclear for me. Does add(1, :relaxed) truly guarantee a unique and visible increment across threads?

From my understanding, using :relaxed means that the atomic increment happens, but the visibility of the result across Fibers isn't guaranteed immediately — so two threads could both do add(1, :relaxed) and each get the same previous value.

With `:relaxed` this scenario is actually possible:

```crystal
#MainThread
counter = Atomic.new(Uint32);
#ThreadA
counter.Atomic.add(1, :relaxed) # returns 0, but eventually increments +1
#ThreadB
counter.add(1, :relaxed) # returns 0, but eventually increments +1
#ThreaC
counter.add(1, :acquire) #returns 2, since has highe guarantee, and eventually increments +1
```

This could cause repeated delivery tags if I understand correctly.

Or I am completely nuts.

lukas8219: Ok. A bit more of research while ensuring ^ was right gave me the correct answer lol. `fetch-and-add` should be good (eg Add) but the `Get(:relaxed)` might have weird numbers then?

PR #1174, opened 2025-06-25
Relaxed ordering for atomic counters

carlhoerberg:

Relaxed ordering is ok for counters.


carlhoerberg:
~~Or should we add a `RelaxedAtomic` instead which defaults to `:relaxed` ordering?~~

No, way to much code has to be copied..

PR #1179, opened 2025-06-26
ReplaceAction will handle underlaying file being replaced again

carlhoerberg:
When a file replacement is replicated using ReplaceAction, there was a bug that if the file was replaced again by the time the ReplaceAction was sent to a slow follower, the filesize could mismatch. This fixes it by opening the File when the ReplaceAction is constructed.

In LavinMQ replaced files are always written to a tempfile file and then renamed, the content of them are not modified.

This also drops support for `MFile` in `ReplaceAction`, it was already dropped from `Replicator` and `Follower`.

### DIFF
--- a/spec/clustering/actions_spec.cr
+++ b/spec/clustering/actions_spec.cr
@@ -64,6 +64,24 @@ describe LavinMQ::Clustering::Action do
           read_and_verify_data(io, "foo")
         end
       end
+
+      it "sends the original file even though it has been appended to" do
+        with_datadir do |data_dir|
+          filename = "file1"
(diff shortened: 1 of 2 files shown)


kickster97 (approved):

carlhoerberg:
Great! But are we appending to files that we then replace? I think we should avoid that

spuun:
> Great! But are we appending to files that we then replace? I think we should avoid that

It happens in `RetainStore`. When it's restored it will replicate the new index file by doing a replace. It may start to append to the new index file before the replace has been sent.

PR #1183, opened 2025-06-27
Refactor: abstract password into Auth and Password namespace

kickster97:
I want to abstract password out of the User class, into its own classes, because I see no reason for it to be a part of the User class. 
This is something i stumbled into when working with breaking the user out into an abstract class so we can have multiple user types. 

This should not change any functionality, run specs


viktorerlingsson (approved):
LGTM 👍

PR #1198, opened 2025-07-25
Add unauthenticated /metrics endpoint on port 15692

### LINKED ISSUE #1162: Ability to serve /metrics without auth

carlhoerberg:
This implements the ability to serve Prometheus metrics without authentication on localhost:15692 (configurable), addressing the need for simplified metrics scraping as described in issue #1162.

## Key Changes
- Add metrics_port configuration option (default: 15692)
- Create MetricsServer class for unauthenticated metrics serving
- Expose necessary methods in PrometheusController
- Integrate metrics server into launcher
- Add comprehensive tests for new functionality

The metrics server only binds to localhost for security and serves both /metrics and /metrics/detailed endpoints. Followers expose the same metrics they can provide (CPU/RAM/GC metrics and replication-related data).

Fixes #1162

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/spec/metrics_server_spec.cr
+++ b/spec/metrics_server_spec.cr
@@ -1,0 +1,86 @@
+require "./spec_helper"
+
+describe LavinMQ::HTTP::MetricsServer do
+  it "should handle /metrics without authentication" do
+    with_amqp_server do |s|
+      metrics_server = LavinMQ::HTTP::MetricsServer.new(s)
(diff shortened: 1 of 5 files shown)


carlhoerberg:
@claude modify this pr, when MetricServer is run in a replication follower an amqp_server can't be injected, so make it optional.

PR #1199, opened 2025-07-28
validate input for operator policies

markusgust:
Currently only policies validate input for values for max-lenght, max-length-bytes, expires,  message-ttl and delivery-limit, making it possible to create an operator policy with a string value and cause all policies to break.

I have compiled locally and tested manually


markusgust:

PR #1206, opened 2025-08-07
fine-tune dead letter loop detection

### LINKED ISSUE #1134: Second dispatch fails on TTL queue
Hello and thank you for lavinmq!

**Describe the bug**
I'll try to describe as well as I can but let me know if something is unclear or details are missing.

I'm using symfony/amqp-messenger and testing retries/delays. It uses the established pattern of a direct exchange and a queue that has x-message-ttl set to the intended delay, x-expires set to delay + 10000, and x-dead-letter-exchange to send the message back to the original exchange after the ttl.

Given a fixed delay of X millisecs, the first message gets through with the delay, but the second message with the same delay disappears without a trace.
When the delay varies, for example I use an exponential 1,2,4 seconds, it works without a problem.
As far as I can tell, the 10s expiry is not reached before the second dispatch, and I do see the queue during that expiry, but it never gets the second message.
The same setup/code works without a problem with rabbitmq

**Describe your setup**
Reproduces with both LavinMQ 2.3 and 2.4.0-rc.1
I'm running it in docker on MacOS.

**How to reproduce**
See above.

**Expected behavior**
Each delayed message should get back onto the main exchange and then the bound queue.

kickster97:

This PR fixes the dead letter loop detection algorithm. The previous implementation was overly aggressive - it would block dead lettering on the first occurrence of a message that had already died in the same queue for the same reason, even when no actual infinite loop existed. Messages with single x-death entries (e.g., count: 1, reason: "expired") were incorrectly blocked from dead lettering, preventing legitimate message routing scenarios.

This PR Updates the algorithm to only prevent dead lettering when there's a genuine cycle (message has been through the same queue multiple times), while maintaining proper handling of rejected messages that break automatic cycles.
  
  
 Fixes #1134

@antondalgren and i did this together but for some reason git is not letting me put him as co-author 🙃
Run `make test SPEC=spec/dlx_spec.cr`

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -641,6 +641,8 @@ module LavinMQ::AMQP
         queue_matches += 1 if table["queue"]? == @name
       end
 
+      # For maxlen/maxlenbytes, prevent loop on first occurrence (threshold=0) since they trigger immediately
+      # For other reasons like TTL, allow one occurrence before blocking (threshold=1)
       threshold = reason.in?(:maxlen, :maxlenbytes) ? 0 : 1
       if queue_matches > threshold && !has_rejected
         @log.debug { "preventing dead letter loop" }


spec/dlx_spec.cr
```diff
@@ -107,4 +107,176 @@ describe "Dead lettering" do
...
+            {"x-message-ttl" => 1, "x-dead-letter-exchange" => "", "x-dead-letter-routing-key" => "q2"}
+          ))
+          q2 = ch.queue("q2", args: AMQP::Client::Arguments.new(
+            {"x-message-ttl" => 1, "x-dead-letter-exchange" => "", "x-dead-letter-routing-key" => "q1"}
+          ))
+
+          ch.default_exchange.publish_confirm("msg", q1.name)
+
+          # Message should cycle once but then get dropped on second cycle
+          sleep 0.1 # Allow messages to cycle
```
viktorerlingsson: should be `sleep 0.1.seconds`

spec/dlx_spec.cr
```diff
@@ -107,4 +107,176 @@ describe "Dead lettering" do
...
+          ] of AMQ::Protocol::Field,
+        })
+        props = AMQ::Protocol::Properties.new(headers: headers)
+        msg = LavinMQ::Message.new(RoughTime.unix_ms, "", "q1", props, 4, IO::Memory.new("msg2"))
+
+        initial_count = v.queues["q2"].message_count
+        v.publish msg
+
+        # Should be blocked (genuine cycle)
+        sleep 0.1
```
viktorerlingsson: here as well `sleep 0.1.seconds`

spec/dlx_spec.cr
kickster97: thanks! 
Maybe we could tweak our linter to catch these?

kickster97:
> I think `threshold` isn't super clear - maybe add a comment on why that's different depending on if reason is `maxlen`/`maxlenbytes` or not?
> 
> Besides that, looks good IMO! 👍

hehe thought the same so already updated that :)

viktorerlingsson:
> hehe thought the same so already updated that :)

Haha, nicely done! 🚀

TamasSzigeti:
Sorry to bother, not sure where else to ask, whenabouts may I expect this fix to get into a release and be available on CloudAMQP? I assume it would go into a 2.4.2 patch release as a bugfix?
Thank you!

kickster97:
Hi @TamasSzigeti! CloudAMQP will get it as part of the 2.5.0 release, as we currently have it as a part of version 2.5.0-rc.1. We have not finished the testing yet, but will try to get something out on CloudAMQP next week so you don't have to wait too long! :)

PR #1207, opened 2025-08-07
Fix pagination bounds check to prevent out-of-bounds pages

### LINKED ISSUE #1201: When going to a page that it is out of bound, redirect to the last valid page
**Summary**  
Create a bunch of queues, go to the last page. Add a policy that deletes all but a few queues. Go back to the Queues page and you will get an empty table without the paginator at the bottom.

viktorerlingsson:
## Summary
- Fixes pagination issue where users could end up on pages that no longer exist
- Adds automatic redirect to the appropriate page when current page exceeds total pages
- Prevents showing empty result sets when page is out of bounds

## Test plan
- Navigate to a high page number with many queues
- Delete queues to reduce the total page count below current page
- Verify that the page automatically redirects to the last valid page

Fixes #1201

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/static/js/datasource.js
+++ b/static/js/datasource.js
@@ -106,6 +106,7 @@ class DataSource {
   get reverseOrder () { return this._queryState.sort_reverse }
   set searchTerm (value) { this._queryState.name = value }
   get searchTerm () { return this._queryState.name }
+
   get filteredCount () { return this._filteredCount }
   get itemCount () { return this._itemCount }
   get pageCount () { return this._pageCount }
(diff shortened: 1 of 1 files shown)


kickster97 (approved):
It works well, code looks good! but idk if the comments are super nessecary :)

viktorerlingsson:
> It works well, code looks good! but idk if the comments are super nessecary :)

Yeah, but maybe one does not really expect a redirect to happen in that method, so I think it might be nice to highlight that a little 🤷

PR #1209, opened 2025-08-08
use paused instead of .paused as filename for paused queues

### LINKED ISSUE #1208: Don't start filenames with `.` for paused queues (.paused)
Files starting with a `.` are hidden by default in linux.

viktorerlingsson:
Use `paused` instead of `.paused` as filename for paused queues. Files starting with `.` are hidden by default in linux, meaning they can be missed. 

This will also rename any existing `.paused` file in a queue dir to `paused`. 

Fixes #1208

Run specs.

### DIFF
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -91,11 +91,24 @@ describe LavinMQ::AMQP::Queue do
         v.declare_queue("q", true, false)
         data_dir = s.vhosts["/"].queues["q"].as(LavinMQ::AMQP::Queue).@msg_store.@msg_dir
         s.vhosts["/"].queues["q"].pause!
-        File.exists?(File.join(data_dir, ".paused")).should be_true
+        File.exists?(File.join(data_dir, "paused")).should be_true
         s.restart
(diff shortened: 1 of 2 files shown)


snichme:
I understand what the PR is doing and why, but I don't understand why it's an issue with the file being hidden. 
There are other files that are hidden as well and I think that it's nice to have these files hidden, similar as you have config files in your home directory in linux hidden.  You can see them when needed otherwise they dont take your attention.

viktorerlingsson:
> I understand what the PR is doing and why, but I don't understand why it's an issue with the file being hidden. There are other files that are hidden as well and I think that it's nice to have these files hidden, similar as you have config files in your home directory in linux hidden. You can see them when needed otherwise they dont take your attention.

In my opinion the issue is that it's a hidden file but it has a function. So you could miss that there's a .paused file and not realize a queue will be paused. Most other hidden files are just information (at least in LavinMQ).

Some forms of copying files will also ignore hidden files IIRC, which could cause unexpected behavior when copying data dirs.

Issue #1082, opened 2025-04-29
Cluster: Don't forward `/metrics` to leader node


jage:
Current behaviour in clustered LavinMQ is for followers to forward HTTP calls to the leader node, this issue only focuses on issues with `/metrics` (no comment on forwarding other API calls)

In Prometheus I would argue it's a common assumption that one target endpoint only returns metrics for itself. Each node in a cluster should be scraped and and if we want a cluster wide stats it should be summed up in the time series database.

When scraping multiple nodes in a LavinMQ cluster we get duplicated data, e.g. number of messages are reported once per node. This is not only creates unnecessary metric series but is also problematic when analysing the metrics, is there `sum(lavinmq_queue_messages)` or `max(lavinmq_queue_messages)` total in a cluster? (I know it's `max()` but that can change in the future).

I'd like to see leader nodes report metrics as is, but followers should not forward any metrics from the leader. Followers should only report relevant metrics like CPU, memory, GC from itself, not any metrics for queues/messages that they are not in control of.

Side-node: There are multi-target use cases within Prometheus, but they are explicit like `/metrics?target=leader`, for example prometheus-blackbox.

carlhoerberg:
Ok, together with #1162 this should be straight forward

PR #1213, opened 2025-08-18
allow /metrics to be open from config

### LINKED ISSUE #1162: Ability to serve /metrics without auth

snichme:

Adds a config option `allow_open_metrics` that allows `/metrics/*` endpoints to be accessible without authentication

Fixes #1162 

Manually

### DIFF
--- a/src/lavinmq/config.cr
+++ b/src/lavinmq/config.cr
@@ -118,7 +118,7 @@ module LavinMQ
         end
         p.on("--no-data-dir-lock", "Don't put a file lock in the data directory (default: true)") { @data_dir_lock = false }
         p.on("--raise-gc-warn", "Raise on GC warnings (default: false)") { @raise_gc_warn = true }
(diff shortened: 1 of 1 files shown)


snichme:
Is the name of the config `allow_open_metrics` good?  Should we change how it has been working and have the endpoints open by default (which is usually how prometheus endpoints are)?

carlhoerberg:
Shouldn't we provide prometheus metrics on another port instead, that only listens on localhost? Followers could also then have that port open, with the only metrics a follower can provide.

carlhoerberg:
See https://github.com/cloudamqp/lavinmq/issues/1082

snichme:
> Shouldn't we provide prometheus metrics on another port instead, that only listens on localhost? Followers could also then have that port open, with the only metrics a follower can provide.

Makes sense, looking at a solution for that as well

snichme:
Here is another attempt: https://github.com/cloudamqp/lavinmq/pull/1217

Issue #1203, opened 2025-08-04
Use actual unix socket address instead of faking local connection info


kickster97:
Currently, when handling UNIX socket connections without proxy protocol, LavinMQ uses `ConnectionInfo.local` which creates fake local connection information instead of using the actual UNIX socket address. 

We should extract the actual UNIX socket path information and create proper ConnectionInfo

lukas8219:
@kickster97 I'll start a poc to understand LOE on this later this week.

lukas8219:
[Ref for Connection.local usage](https://github.com/cloudamqp/lavinmq/blob/64d091d73171dd135f6dfee6d16fbaead22a7872/src/lavinmq/connection_info.cr#L21) + [Server.cr path](https://github.com/cloudamqp/lavinmq/blob/64d091d73171dd135f6dfee6d16fbaead22a7872/src/lavinmq/server.cr#L170)

lukas8219:
@kickster97 There's a couple places I didn't touch that uses `port` from `IP` protocol such as the `client.cr` files from both [AMQP](https://github.com/cloudamqp/lavinmq/blob/fa0cbdc8322a63536776719a2807beca33781d1b/src/lavinmq/amqp/client.cr#L95)/[MQTT](https://github.com/cloudamqp/lavinmq/blob/fa0cbdc8322a63536776719a2807beca33781d1b/src/lavinmq/mqtt/client.cr#L244) protocol

wondering if it makes sense to introduce a breaking change to change to `peer` instead of `peer_host|port` as the later is protocol-specific.

Not sure this is worth only to support UNIX Protocol - since it has way less use-cases

PR #1214, opened 2025-08-18
feature(Unix): enable AMQP via Unix Socket by replacing mock `ConnectionInfo.local`

lukas8219:
Replacing the `ConnectionInfo.local` that mocked a SRC/DST via localhost for a proper UNIX Socket.

Specs:
- Already has a proper accepts UNIX Sockets

Closes https://github.com/cloudamqp/lavinmq/issues/1203

### DIFF
--- a/src/lavinmq/server.cr
+++ b/src/lavinmq/server.cr
@@ -170,7 +170,7 @@ module LavinMQ
           case @config.unix_proxy_protocol
           when 1 then ProxyProtocol::V1.parse(client)
           when 2 then ProxyProtocol::V2.parse(client)
-          else ConnectionInfo.new(client.remote_address.path, client.local_address.path)
+          else        ConnectionInfo.new(client.remote_address.path, client.local_address.path)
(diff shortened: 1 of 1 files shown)


src/lavinmq/connection_info.cr
```diff
@@ -34,12 +28,26 @@ module LavinMQ
...
+        @port = parts[1].to_u16
+      end
+
       def to_s(io)
         io << @address << ':' << @port
       end
 
       def loopback?
-        @address == "::1" || @address.starts_with? "127."
+        @address == "::1" || @address.starts_with?("127.") || @address.starts_with?("/")
```
lukas8219, line 50: Mostly because of localhost/guest login. I prefer moving into `def is_unix` ?

lukas8219:
This is mostly for UI purposes. The logs are confusing as they indicate the `remote_address` is 127.0.0.1 but it's not going via TPC Stack

Issue #1137, opened 2025-06-04
Improvments to HTTP shovel with exponential backoff and retry limits


snichme:
**Describe the solution you'd like**

When I add a HTTP shovel that sends message to a http server, if for some reason that server doesn't respond or respond with a non-200 status I want to limit the number of retries, right now it will just keep trying until the endpoint is up again. 

Some ideas are: 

* exponential backoff: So it retries less often 
* retry limits, drop the message if it has tried sending it for X times
* connect timeout: if the http server if down, limit the amount of time lavinmq should wait before rejecting the message

I guess this things should go as settings on the shovel.

lukas8219:
@snichme Do you think I can own this one? Or is it already planned in the upcoming weeks?

kickster97:
@lukas8219 It has not been planned yet. We have a few weeks where most of the team is on vacation, so you can go ahead if you have a solution in mind :)

lukas8219:
@kickster97 What I envision is simple as accepting new parameters config's:
- `http-backoff-time`
- `http-timeout`
- `http-max-retries`
- `http-jitter` (should we also add jitter?)

It sounds simple if it's acceptable that, whenever the target server is down, the shovel consumption rate will be slowed.

carlhoerberg:
Should we make it protocol agnostic and use the same logic for AMQP shovel targets?

lukas8219:
Personally I like that. But how would you envision each parameter for `AMQPDestination` as you exemplified?

lukas8219:
@carlhoerberg @kickster97 What about prefixing those with `dest-` e.g `dest-backoff-time`?

lukas8219:
I made a very naive implementation on this draft and I've been thinking how AMQP Destination would behave - or either if AMQPDestination is on this scope.

Would we only apply the retries depending on wether the publish mode is `OnConfirm`? The retry being `recreate channel  -> publish message`

carlhoerberg:
What should happen at max retries? Should it reject the message and try the
next one? Unlikely that it will be sent successfully either if the host is
down. Or should it stop the whole shovel and it have to be manually
restarted later?

lukas8219:
@carlhoerberg I would say to scope the failover scenario to a future issue - along the lines of adding a prop like `dest-fail-strategy` where it either keeps the current behavior (moving to `Error`) or just keep going.

Just add the retry + backoff and scope this to another issue

PR #1215, opened 2025-08-18
feat(Shovels): implement retry + backoff in Shovel destinations

lukas8219:
1. Implements Retrier in HTTPDestination Shovel - configured by parameters
  - `dest-backoff`
  - `dest-jitter`
  - `dest-timeout`
  - `dest-max-retries`
2. Fixes HTTPDestination not taking `no_ack` mode into consideration - as all HTTPDestinations were defaulted as OnConfirm/OnPublish.

Specs:
- [x] Expose mock server to receive HTTP requests. Fail at least 2 times and accept third time.
- [x] Tests focused on the Retrier logic 

### RFC
- Should we actually implement a proper Circuit Breaker instead of just Retrier?
- Currently using `?.try` is [misleading](https://github.com/cloudamqp/lavinmq/pull/1215#discussion_r2595426705). Is there a better way to handle it? 

Closes https://github.com/cloudamqp/lavinmq/issues/1137

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -330,8 +330,8 @@ module LavinMQ
         new(
           jitter: parameters["dest-jitter"]?.try &.as_f? || 1.0,
           backoff: parameters["dest-backoff"]?.try &.as_f? || 2.0,
-          timeout: parameters["dest-timeout"]?.try &.as_f? || 10.0,
-          max_retries: parameters["dest-max-retries"]?.try &.as_i? || 1
+          timeout: parameters["dest-timeout"]?.try &.as_f? || 30.0,
+          max_retries: parameters["dest-max-retries"]?.try &.as_i? || 0
         )
       end
     end
--- a/src/lavinmq/shovel/shovel_retrier.cr
+++ b/src/lavinmq/shovel/shovel_retrier.cr
@@ -1,11 +1,6 @@
 module LavinMQ
   module Shovel
     module Retrier
-      struct Either
-        getter error
-        getter data
-      end
-
       def self.push_with_retry(max_retries : Int32, jitter : Float64, backoff : Float64, & : -> Bool) : Bool
         retries = 0
         while retries <= max_retries
--- a/spec/shovel_spec.cr
+++ b/spec/shovel_spec.cr
@@ -864,7 +864,7 @@ describe LavinMQ::Shovel do
           LavinMQ::Shovel::HTTPDestinationParameters.from_parameters(JSON.parse(%({
             "dest-timeout": 0.1,
             "dest-backoff": 0.1,
-            "dest-max-retries": 3.0,
+            "dest-max-retries": 3,
             "dest-jitter": 0.1
           }))),
           LavinMQ::Shovel::AckMode::OnConfirm
@@ -903,7 +903,7 @@ describe LavinMQ::Shovel do
           LavinMQ::Shovel::HTTPDestinationParameters.from_parameters(JSON.parse(%({
             "dest-timeout": 0.1,
             "dest-backoff": 0.1,
-            "dest-max-retries": 3.0,
+            "dest-max-retries": 3,
             "dest-jitter": 0.1
           }))),
           LavinMQ::Shovel::AckMode::OnPublish


src/lavinmq/shovel/shovel.cr
```diff
@@ -352,14 +351,28 @@ module LavinMQ
...
+          response = @client.not_nil!.post(path, headers: headers, body: body)
+          return true if response.success?
+          retries += 1
+          # Exponential backoff: 2^retries seconds with jitter
+          base_delay = (2.0 ** retries).seconds
+          jitter_delay = jitter.seconds * Random.rand(0.0..1.0)
+          sleep base_delay + jitter_delay
+        end
+        false
+      end
```
lukas8219: Ideally I think this could be a macro? trait? that receives the `max_retries` and `jitter` and executes a `&block`

kickster97:
Have not looked closely at the code, but does this adress issue [#1357](https://github.com/cloudamqp/lavinmq/issues/1357) aswell?

lukas8219:
@kickster97 Not related. It was meant for only the HTTP retries

src/lavinmq/shovel/shovel.cr
carlhoerberg: yes, like this: 
```suggestion
      private def push_with_retry(path, headers, body, max_retries = 3, jitter = 0.5.seconds, &) : Bool
        retries = 0
        while retries < max_retries
          return true if yield
          retries += 1
          # Exponential backoff: 2^retries seconds with jitter
          base_delay = (2.0 ** retries).seconds
          jitter_delay = jitter * rand(1.0)
          sleep base_delay + jitter_delay
        end
        false
      end
```

src/lavinmq/shovel/shovel.cr
```diff
@@ -359,14 +358,28 @@ module LavinMQ
                else
                  "/"
                end
-        response = c.post(path, headers: headers, body: msg.body_io)
+        success = push_with_retry(path, headers, msg.body_io, max_retries: 3, jitter: 0.5)
```
snichme: `push_with_retry`  makes sense for `AckMode::OnConfirm` and `AckMode::OnPublish` but `NoAck` is more like "fire-and-forget" so there we don't care about the results. 

What I mean is that we should only retry on OnConfirm and OnPublish

spec/shovel_retrier_spec.cr
```diff
@@ -0,0 +1,139 @@
+require "./spec_helper"
+require "../src/lavinmq/shovel/shovel_retrier"
+
+describe LavinMQ::Shovel::Retrier do
```
lukas8219, line 4: It got me wondering wether `Retrier` shouldn't just be some sort of Circuit Breaker...

src/lavinmq/shovel/shovel_retrier.cr
```diff
@@ -0,0 +1,22 @@
+module LavinMQ
+  module Shovel
+    module Retrier
+      struct Either
+        getter error
+        getter data
+      end
```
lukas8219: Remove

src/lavinmq/shovel/shovel_store.cr
```diff
@@ -53,7 +53,12 @@ module LavinMQ
       destinations = uris.map do |uri|
         case uri.scheme
         when "http", "https"
-          Shovel::HTTPDestination.new(name, uri)
+          Shovel::HTTPDestination.new(
+            name,
+            uri,
+            Shovel::HTTPDestinationParameters.from_parameters(config),
+            ack_mode
```
lukas8219, line 60: ⚠️ 🔥  This might be a breaking change for some customers. I need to check deployed LMQ definitions and adding in release notes a comment about this. ACK mode was not being considered in HTTPDest

src/lavinmq/shovel/shovel.cr
```diff
@@ -316,17 +317,36 @@ module LavinMQ
...
+
+      def initialize(@jitter, @backoff, @timeout, @max_retries)
+      end
+
+      def self.from_parameters(parameters : JSON::Any)
+        new(
+          jitter: parameters["dest-jitter"]?.try &.as_f? || 1.0,
+          backoff: parameters["dest-backoff"]?.try &.as_f? || 2.0,
+          timeout: parameters["dest-timeout"]?.try &.as_f? || 10.0,
+          max_retries: parameters["dest-max-retries"]?.try &.as_i? || 1
```
lukas8219: 💡 ⚠️  `?.try` may fail to transform user input like `1.0` or `1` between Float and Integer. Not sure if `?.try` is the best idea here.

lukas8219: Meaning, it could be misleading for the user if the Shovel Parameter accepts their input but behind the scenes, it uses the default values. Notice it via Spec

snichme: This is a reoccurring thing, how and when to validate input data and also handle it when it's invalid. 
What you have here is common pattern that we try to parse the input and if it fails we use a default value, not saying it's the best since the shovel might end up being configured in a wrong way. But trying to find a common solution to all input is also difficult.  @spuun did some refactoring on validating input for a queue here: https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/amqp/queue/queue.cr#L27 might give some inspiration.

spec/shovel_retrier_spec.cr
snichme: Could you explain some more what you are thinking here? :)

lukas8219: @snichme If we are doing HTTP Retries, there's a high possibility that blindly retrying is not the best approach. What if the user has N HTTP shovels, and their downstreams start failing? 

This behavior of blindly retrying could just make the situation worse...And it's not a good UX to simply have the user deleting/pausing shovels one by one? (on the other hand for that specific scenario, we could make it simple as having a `Pause All` sorta button)

snichme: Aha agree. 
I like it. I would prefer to have a checkbox next to each shovel, so you can select some of them and have batch operations on the selected ones, like Pause, Delete etc. 
Like we have with Queues and Streams. 

But this should be another PR IMO. so this wont run away in size or time

snichme:
Draft PR but still something that will impact this PR if merged before: https://github.com/cloudamqp/lavinmq/pull/1591 I can help rebase if needed.

lukas8219:
@snichme Ok. I am still waiting on that PR to be merged so I can push this further.

lukas8219:
Closed in favor of https://github.com/cloudamqp/lavinmq/pull/1667

PR #1216, opened 2025-08-19
Allow MQTT/AMQP cross protocol binding

### LINKED ISSUE #1136: Support cross protocol exchange to exchange binding
You can currently not bind the MQTT exchange or session to an AMQP exchange or queue. 

I would like to be able to bind a mqtt exchange/session to an AMQP exchange or queue. 

Supporting this probably means we need to adapt the way the exchanges routes messages, have some sort of AMQP/MQTT conversion.

kickster97:
Allow AMQP queues and exchanges bind to MQTT exchanges and session and vice versa. 

Today the mqtt exchange is an internal exchange, we can consider to not have it this way 

Note that AMQP queues will not receive retained messages on bind("subscribe"). But it will receive messages as usual when published. 

Fixes #1136 

Run specs

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -197,7 +197,6 @@ module LavinMQ
       private def route_msg(msg : Message, immediate : Bool, queues : Set(LavinMQ::Queue), exchanges : Set(LavinMQ::Exchange)) : Bool
         headers = msg.properties.headers
         find_queues(msg.routing_key, headers, queues, exchanges)
-        pp queues
         if queues.empty? || (immediate && !queues.any? &.immediate_delivery?)
           @unroutable_count.add(1, :relaxed)
           return false
--- a/src/lavinmq/mqtt/exchange.cr
+++ b/src/lavinmq/mqtt/exchange.cr
@@ -8,12 +8,12 @@ require "./retain_store"
 module LavinMQ
   module MQTT
     class Exchange < AMQP::Exchange
-      # Transform MQTT topic to AMQP routing key
-      # MQTT uses '/' as separator, AMQP uses '.'
-      # MQTT '+' wildcard becomes AMQP '*' wildcard
-      # MQTT '#' wildcard stays '#' in AMQP
       private def mqtt_topic_to_amqp_routing_key(mqtt_topic : String) : String
         mqtt_topic.gsub('/', '.').gsub('+', '*')
+      end
+
+      private def amqp_routing_key_to_mqtt_topic(amqp_routing_key : String) : String
+        amqp_routing_key.gsub('.', '/').gsub('*', '+')
       end
 
       # In MQTT only topic/routing key is used in routing, but arguments is used to
@@ -90,11 +90,14 @@ module LavinMQ
               queues.add(destination)
             end
           in LavinMQ::AMQP::Exchange
-            amqp_routing_key = mqtt_topic_to_amqp_routing_key(packet.topic)
-            amqp_msg = Message.new(timestamp, EXCHANGE, amqp_routing_key, properties, bodysize, body)
-            amqp_msg.properties.delivery_mode = qos
             if exchanges.add?(destination)
-              destination.publish(amqp_msg, false, queues, exchanges)
+              amqp_routing_key = mqtt_topic_to_amqp_routing_key(packet.topic)
+              amqp_msg = Message.new(timestamp, EXCHANGE, amqp_routing_key, properties, bodysize, body)
+              amqp_msg.properties.delivery_mode = qos
+              # Use route_msg directly to avoid creating new sets
+              if destination.route_msg(amqp_msg)
+                count += 1
+              end
               amqp_msg.body_io.rewind
             end
           in LavinMQ::MQTT::Exchange
@@ -155,7 +158,10 @@ module LavinMQ
         qos = arguments.try { |h| h[QOS_HEADER]?.try(&.as(UInt8)) } || 0u8
         binding_key = BindingKey.new(routing_key, arguments)
         @bindings[binding_key].add destination
-        @tree.subscribe(routing_key, destination, qos)
+
+        # Convert AMQP routing key to MQTT topic pattern for subscription tree
+        mqtt_topic_pattern = amqp_routing_key_to_mqtt_topic(routing_key)
+        @tree.subscribe(mqtt_topic_pattern, destination, qos)
 
         data = BindingDetails.new(name, vhost.name, binding_key.inner, destination)
         notify_observers(ExchangeEvent::Bind, data)
@@ -168,7 +174,9 @@ module LavinMQ
         rk_bindings.delete destination
         @bindings.delete binding_key if rk_bindings.empty?
 
-        @tree.unsubscribe(routing_key, destination)
+        # Convert AMQP routing key to MQTT topic pattern for subscription tree
+        mqtt_topic_pattern = amqp_routing_key_to_mqtt_topic(routing_key)
+        @tree.unsubscribe(mqtt_topic_pattern, destination)
 
         data = BindingDetails.new(name, vhost.name, binding_key.inner, destination)
         notify_observers(ExchangeEvent::Unbind, data)


src/lavinmq/mqtt/exchange.cr
```diff
@@ -55,8 +55,36 @@ module LavinMQ
 
         msg = Message.new(timestamp, EXCHANGE, packet.topic, properties, bodysize, body)
         count = 0u32
-        @tree.each_entry(packet.topic) do |queue, qos|
+        delivered_to = Set(LavinMQ::Destination).new
```
carlhoerberg: Can it be reused? Allocating a new Set for each publish will not perform well

src/lavinmq/mqtt/exchange.cr
```diff
@@ -55,8 +55,36 @@ module LavinMQ
...
+
+        # First, handle MQTT sessions using topic pattern matching
+        @tree.each_entry(packet.topic) do |destination, qos|
+          # Use the subscription QoS for MQTT sessions (as per MQTT spec)
           msg.properties.delivery_mode = qos
+          case destination
+          when MQTT::Session
```
carlhoerberg: Use `in` instead of  `when` if possible

carlhoerberg (commented):
Cool feature! But allocations in the hot path is no good

snichme:
I had some ideas that I wrote down in another PR: https://github.com/cloudamqp/lavinmq/pull/1224 

Mainly it was setting up `@queues` and `@exchanges` in Exchange class instead of sending them from the client all the way down. But this might break some cases, not sure. 

Also one more suggestion, not use the `find_queues_internal` as it's only used one time and it will make the diff for the PR larger. I the method is used more than one time we can extract it to a method. 

WDYT?

kickster97:
> I had some ideas that I wrote down in another PR: #1224
> 
> Mainly it was setting up `@queues` and `@exchanges` in Exchange class instead of sending them from the client all the way down. But this might break some cases, not sure.
> 
> Also one more suggestion, not use the `find_queues_internal` as it's only used one time and it will make the diff for the PR larger. I the method is used more than one time we can extract it to a method.
> 
> WDYT?

As we discussed, the `find_queues_internal` method is needed for the MQTT exchange to work together with the `SubscriptionTree`. 

`@queues` and `@exchanges` need to be created at the client level, so they are isolated to that client, otherwise multiple clients publishing simultaneously would share the same sets and there would be modification of same sets from different threads.

src/lavinmq/mqtt/exchange.cr
```diff
@@ -55,13 +63,55 @@ module LavinMQ
...
+              count += 1
+              msg.body_io.rewind
+              queues.add(destination)
+            end
+          in LavinMQ::AMQP::Exchange
+            amqp_routing_key = mqtt_topic_to_amqp_routing_key(packet.topic)
+            amqp_msg = Message.new(timestamp, EXCHANGE, amqp_routing_key, properties, bodysize, body)
+            amqp_msg.properties.delivery_mode = qos
+            if exchanges.add?(destination)
+              destination.publish(amqp_msg, false, queues, exchanges)
```
snichme: Shouldn't we increment the count here as well? Doesn't `publish` here return the count for how many queues it managed to publish to? So we can use that

src/lavinmq/mqtt/exchange.cr
```diff
@@ -55,13 +63,55 @@ module LavinMQ
...
+              amqp_msg.body_io.rewind
+            end
+          in LavinMQ::MQTT::Exchange
+            if exchanges.add?(destination)
+              destination.publish(msg, false, queues, exchanges)
+              msg.body_io.rewind
+            end
+          in LavinMQ::Exchange
```
snichme, line 108: What is the difference between `LavinMQ::Exchange`  and `LavinMQ::AMQP::Exchange`?  

also `count` here as well.

src/lavinmq/mqtt/exchange.cr
```diff
@@ -55,13 +63,55 @@ module LavinMQ
...
+          case destination
+          in LavinMQ::AMQP::Queue
+            amqp_routing_key = mqtt_topic_to_amqp_routing_key(packet.topic)
+            amqp_msg = Message.new(timestamp, EXCHANGE, amqp_routing_key, properties, bodysize, body)
+            amqp_msg.properties.delivery_mode = qos
+            if destination.publish(amqp_msg)
+              count += 1
+              amqp_msg.body_io.rewind
+              queues.add(destination)
+            end
```
snichme, line 79: These four lines are repeated three times, can we restructure to reduce repeating?

src/lavinmq/mqtt/exchange.cr
```diff
@@ -8,6 +8,14 @@ require "./retain_store"
 module LavinMQ
   module MQTT
     class Exchange < AMQP::Exchange
+      # Transform MQTT topic to AMQP routing key
+      # MQTT uses '/' as separator, AMQP uses '.'
+      # MQTT '+' wildcard becomes AMQP '*' wildcard
+      # MQTT '#' wildcard stays '#' in AMQP
+      private def mqtt_topic_to_amqp_routing_key(mqtt_topic : String) : String
+        mqtt_topic.gsub('/', '.').gsub('+', '*')
```
snichme, line 12: @spuun IIRC there are more things  to consider for this transformation?

src/lavinmq/http/controller/bindings.cr
```diff
@@ -137,7 +137,7 @@ module LavinMQ
               access_refused(context, "User doesn't have write permissions to exchange '#{destination.name}'")
             elsif source.name.empty? || destination.name.empty?
               access_refused(context, "Not allowed to bind to the default exchange")
-            elsif destination.internal?
+            elsif destination.internal? && destination.name != "mqtt.default"
```
snichme, line 140: is `mqtt.default` an internal or default exchange? the `if` above checks for default exchange and based on the name this is a default exchange. 

Is there a way to not have the string "mqtt.default" here but rather some method on `destination` to check wether we can bind/unbind to the change? Less hard coded strings is my goal here.

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -225,6 +197,7 @@ module LavinMQ
       private def route_msg(msg : Message, immediate : Bool, queues : Set(LavinMQ::Queue), exchanges : Set(LavinMQ::Exchange)) : Bool
         headers = msg.properties.headers
         find_queues(msg.routing_key, headers, queues, exchanges)
+        pp queues
```
snichme: debug statment 
```suggestion

```

src/lavinmq/mqtt/exchange.cr
kickster97: this part is a wip now, found some bugs yesterday so needed to rewrite this. the difference is that LavinMQ::Exchange counts for both amqp and mqtt exchanges. but i need better code overall in this case section

src/lavinmq/mqtt/exchange.cr
kickster97: Same as above, need to refactor this, just needed the case to work as expected for a while so we had a build that worked.

src/lavinmq/mqtt/exchange.cr
kickster97: maybe, from what i remember it was pretty simple :o

src/lavinmq/http/controller/bindings.cr
kickster97: Yes, i agree this is unclear. Its been a "good enough" solution up untill now. I think we should have it as a default exchange and let it behave more like an amq.topic exchange. @antondalgren lifts a good issue about bindings not persiting for mqtt exchange bound to amqp exchange/queues. this is because we onyl built it to support the mqtt use case. 

tldr; we need to refactor the exchange's place in lavinmq :)

kickster97: but new features like this might be less prioritized next week/2 weeks to prioritize squashing bugs we found in 2.4.0

src/lavinmq/mqtt/exchange.cr
snichme: Ok, then this is all good :)

kickster97:
closing this in favor of doing a refactor first and getting a better starting point for this feature

PR #1217, opened 2025-08-19
/metrics on a separate port

### LINKED ISSUE #1162: Ability to serve /metrics without auth

snichme:

Fixes #1162  and #1082 

Adds a MetricsServer that serves /metrics endpoints on a separate port. 
This also resolves the issue where /metrics requests was forwarded to the leader, now we can collect metrics from a follower as well. Current it only reports GC metrics on the follow. 

Specs and manual

### DIFF
--- a/src/lavinmq/config.cr
+++ b/src/lavinmq/config.cr
@@ -36,6 +36,8 @@ module LavinMQ
     property http_unix_path = ""
     property http_systemd_socket_name = "lavinmq-http.socket"
     property amqp_systemd_socket_name = "lavinmq-amqp.socket"
+    property metrics_http_bind = "127.0.0.1"
+    property metrics_http_port = 15693
     property heartbeat = 300_u16                     # second
     property frame_max = 131_072_u32                 # bytes
     property channel_max = 2048_u16                  # number
@@ -159,6 +161,12 @@ module LavinMQ
         end
         p.on("--mqtt-unix-path=PATH", "MQTT UNIX path to listen to") do |v|
           @mqtt_unix_path = v
+        end
+        p.on("--metrics-http-bind=BIND", "IP address that the Prometheus server will bind to (default: 127.0.0.1)") do |v|
+          @metrics_http_bind = v
+        end
+        p.on("--metrics-http-port=PORT", "HTTP port that prometheus will listen to (default: 15692)") do |v|
+          @metrics_http_port = v.to_i
         end
         p.on("--cert FILE", "TLS certificate (including chain)") { |v| @tls_cert_path = v }
         p.on("--key FILE", "Private key for the TLS certificate") { |v| @tls_key_path = v }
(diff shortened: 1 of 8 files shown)


src/lavinmq/http/metrics_server.cr
```diff
@@ -0,0 +1,44 @@
...
+        addr
+      end
+
+      def listen
+        @http.listen
+      end
+
+      def close
+        @http.try &.close
+        File.delete?(INTERNAL_UNIX_SOCKET)
```
carlhoerberg: Copy paste error?

carlhoerberg (approved):

src/lavinmq/config.cr
```diff
@@ -36,6 +36,8 @@ module LavinMQ
     property http_unix_path = ""
     property http_systemd_socket_name = "lavinmq-http.socket"
     property amqp_systemd_socket_name = "lavinmq-amqp.socket"
+    property metrics_http_bind = "127.0.0.1"
```
carlhoerberg, line 39: Could be localhost? So that it also listens on IPv6 ::1

snichme:
This is a breaking change as we remove the PrometheusController from the default http server https://github.com/cloudamqp/lavinmq/pull/1217/files#diff-b82ac321d4e4fce66f0c1c37f6be02a6a8e672679f934b8aa9ce394df48328dbL27 Should we keep that so we can release this now?

viktorerlingsson:
> This is a breaking change as we remove the PrometheusController from the default http server https://github.com/cloudamqp/lavinmq/pull/1217/files#diff-b82ac321d4e4fce66f0c1c37f6be02a6a8e672679f934b8aa9ce394df48328dbL27 Should we keep that so we can release this now?

Maybe keep it in the default http server unless you have configured a separate port/bind for it?

snichme:
After thinking about this, I think it's easier to keep the PrometheusController on the default http server and also have it on the metrics server.  
But be clear in the changelog that it will be removed, lets say in v3.0.0. Then everyone have time to migrate without issues

src/lavinmq/config.cr
```diff
@@ -160,6 +162,12 @@ module LavinMQ
         p.on("--mqtt-unix-path=PATH", "MQTT UNIX path to listen to") do |v|
           @mqtt_unix_path = v
         end
+        p.on("--metrics-http-bind=BIND", "IP address that the Prometheus server will bind to (default: 127.0.0.1)") do |v|
```
carlhoerberg:
```suggestion
        p.on("--metrics-http-bind=BIND", "IP address that the Prometheus server will bind to (default: #{@metrics_http_bind})") do |v|
```

src/lavinmq/config.cr
```diff
@@ -160,6 +162,12 @@ module LavinMQ
...
+          @metrics_http_bind = v
+        end
+        p.on("--metrics-http-port=PORT", "HTTP port that prometheus will listen to (default: 15692)") do |v|
```
carlhoerberg: So that the docs don't get out of sync if the default were to be changed 

```suggestion
        p.on("--metrics-http-port=PORT", "HTTP port that prometheus will listen to (default: #{@metrics_http_port})") do |v|
```

carlhoerberg (approved):

src/lavinmq/config.cr
snichme: I did 127.0.0.1 since we have that for `amqp_bind` and `http_bind` for defaults.

PR #1224, opened 2025-08-28
suggestions

snichme:
Suggestions on some code changes


kickster97 (approved):

kickster97:
we need to use `@tree.each_entry` instead of `each_destination`

Issue #1223, opened 2025-08-27
UI/API: Get message does not work for Stream queues


jage:
**Summary**  

When clicking "Get message" it just responds "No messages in queue", even though there are messages there.

**Current Behavior**  

Pretty sure this is due to us using `basic_get` which isn't a thing for Stream Queues: 

https://github.com/cloudamqp/lavinmq/blob/fa0cbdc8322a63536776719a2807beca33781d1b/src/lavinmq/amqp/queue/stream_queue.cr#L65-L68

**Desired Behavior**  

I'd like to be able to "look at" any message in the stream, e.g. first, last or position N.

**Screenshots/Mockups**  

<img width="1761" height="1372" alt="Image" src="https://github.com/user-attachments/assets/a50a910f-46a1-4043-8e0f-18a326dec159" />

viktorerlingsson:
Yes, it's done because streams requires a consumer with some argument (x-stream-offset). But that can probably be worked around. I think it would be nice to be able to do this 👍

kickster97:
We should build a nice way to have different queue views for different queue types like Streams or MQTT sessions, so we don't have to make lots of conditionals in the JS. Something generic so we can add more queue types in the future.

snichme:
The form for get messages for a queue has Mode, Encoding and Messages (how many messages to get). 
But in order to read messages from a stream there should also Offset for it to make sense. 

It's even so that when you Get messages from a Queue you read from the end of the queue and backwards, since you pop the messages of the queue when getting them but for a Stream you set a starting point and you say how many messages you should get from the starting point and forwards. 

But besides the semantic difference, we need a way to have different forms, and also some different handling in the `queue.js` which posts the request to backend. 

Looking at https://github.com/cloudamqp/lavinmq/pull/1284 it checks if the `x-queue-type` exists and if it's `stream`, we can do the same here and hide some DOM elements unless it's a stream but I see this as getting messy for future cases, like "Move messages" should probably have a starting offset and should not be "Move" it should be "Copy", "Purge Queue" should either not exists since a stream is persistent but if we want the possibilty (it can be nice to have) it should be called "Purge Stream" and in this IMO we should rename it from Queue to Stream. 

Anyway, I'm landing in that maybe it's a good idea to have a separate view for a stream, but if so, how should we handle routing to `/queue` or `/stream` and how can we share html and js between these two?

viktorerlingsson:
[...]

PR #1236, opened 2025-09-05
read messages from stream

snichme:

Introduce a StreamReader where you give it an offset and then it will yield each message until end-of-stream. 
You get a reader by `q.stream(offset)` 

This is intended to be used from the Queues controller to read X messages of a stream #1223

Specs

### DIFF
--- a/src/lavinmq/amqp/stream/stream_reader.cr
+++ b/src/lavinmq/amqp/stream/stream_reader.cr
@@ -1,0 +1,34 @@
+require "../stream/stream_message_store"
+require "../stream/stream"
+
+module LavinMQ::AMQP
+  class StreamReader
+    Log = LavinMQ::Log.for "stream_reader"
+
+    def initialize(@store : StreamStore, @start_offset : String | Int64 | Time)
+    end
+
+    def each(&)
+      store = @store
+      msg_offset, segment, position = store.find_offset(@start_offset)
+      loop do
+        env = store.read(segment, position)
+        if env
+          position += env.segment_position.bytesize
+        else
+          # try read from new segment
+          _, s, p = store.find_offset(msg_offset)
+          env = store.read(s, p)
+          break unless env # No more messages to read, i.e we reached last segment
+          segment = s
+          position = p
+        end
+        msg_offset += 1
+        yield env
+      rescue e : StreamMessageStore::Error
+        Log.error { "store error: #{e}" }
+        return
+      end
+    end
+  end
+end
--- a/src/lavinmq/amqp/stream/stream_reader.cr
+++ b/src/lavinmq/amqp/stream/stream_reader.cr
@@ -5,29 +5,25 @@ module LavinMQ::AMQP
   class StreamReader
     Log = LavinMQ::Log.for "stream_reader"
 
-    def initialize(@store : StreamStore, @start_offset : String | Int64 | Time)
+    def initialize(@store : StreamMessageStore, @start_offset : String | Int64 | Time)
     end
 
     def each(&)
       store = @store
-      msg_offset, segment, position = store.find_offset(@start_offset)
+      _, segment, position = store.find_offset(@start_offset)
       loop do
+        break if store.closed
         env = store.read(segment, position)
         if env
           position += env.segment_position.bytesize
         else
           # try read from new segment
-          _, s, p = store.find_offset(msg_offset)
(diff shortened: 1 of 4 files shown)


src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -3,7 +3,11 @@ require "../stream_consumer"
 
 module LavinMQ::AMQP
   class StreamQueue < DurableQueue
+    module MsgStore
```
snichme: This is to allow a MockStore from the specs.

src/lavinmq/http/controller/queues.cr
```diff
@@ -13,6 +13,34 @@ module LavinMQ
         not_found(context, "Not Found") unless q
         q
       end
+
+      private def env_as_json(j, q, env, truncate, encoding)
```
snichme: Not sure about this. 
Different ways of getting the messages from a Queue and a Stream but rendering is the same..

src/lavinmq/http/controller/queues.cr
```diff
@@ -215,6 +219,40 @@ module LavinMQ
...
+            # if q.state != QueueState::Running && q.state != QueueState::Paused
+            #   forbidden(context, "Can't get from queue that is not in running state")
+            # end
+            unless q.is_a?(AMQP::StreamQueue)
+              access_refused(context, "Can only read from a stream queue")
+            end
+            truncate = nil
+            body = parse_body(context)
+            encoding = body["encoding"]?.try(&.as_s) || "auto"
+            count = body["count"]?.try(&.as_i) || 1
```
snichme: also maybe a max count here

src/lavinmq/amqp/stream/stream_reader.cr
```diff
@@ -0,0 +1,34 @@
...
+
+module LavinMQ::AMQP
+  class StreamReader
+    Log = LavinMQ::Log.for "stream_reader"
+
+    def initialize(@store : StreamStore, @start_offset : String | Int64 | Time)
+    end
+
+    def each(&)
+      store = @store
```
kickster97, line 12: Why do we do this assignment?

src/lavinmq/amqp/stream/stream_reader.cr
```diff
@@ -0,0 +1,34 @@
...
+        if env
+          position += env.segment_position.bytesize
+        else
+          # try read from new segment
+          _, s, p = store.find_offset(msg_offset)
+          env = store.read(s, p)
+          break unless env # No more messages to read, i.e we reached last segment
+          segment = s
+          position = p
+        end
```
viktorerlingsson, line 25: This is probably works just fine, but I think it should pretty much always be enough to just set `segment+=1` and `position=4` when switching to the next segment for streams (that's basically what we do in `next_segment` in `stream_message_store`) 🤔 Maybe refactor next_segment so we can re-use the logic instead? 

Maybe something like:
```
        else
          # try read from new segment
          segment = store.next_segment_id(segment) || break
          position = 4_u32
          env = store.read(segment, position)
          break unless env # No more messages to read, i.e we reached last segment
        end
```

```
    def next_segment_id(segment)
      @segments.each_key.find { |sid| sid > segment }
    end

    private def next_segment(consumer) : MFile?
      consumer.segment = next_segment_id(consumer.segment) || return nil
      consumer.pos = 4_u32
      @segments[consumer.segment]
    end
```

src/lavinmq/amqp/stream/stream_reader.cr
snichme: It should be faster in the loop if we don't have to deref the `@store` var every time but the operations in the loop is this much slower so it probably doesn't do much. But thats the idea at least.

src/lavinmq/amqp/stream/stream_reader.cr
snichme: Awesome, I did some improvments here https://github.com/cloudamqp/lavinmq/pull/1236/commits/64d171f6846578e154a41909830e7ba55cb48048 also improved the specs

viktorerlingsson: Nice 👍

viktorerlingsson (approved):
LGTM!

Issue #891, opened 2024-12-17
Add human friendly name to exchange


kickster97:
**Is your feature request related to a problem? Please describe.**
![image](https://github.com/user-attachments/assets/a3dfbedd-fa67-4f81-abd0-0c6d5126e000)

**Describe the solution you'd like**
We should add a human friendly name to exchanges which we can expose in UI

**Describe alternatives you've considered**

kickster97:
@Bianka2112 solved it! 🥳

PR #1238, opened 2025-09-08
added human friendly exchange type names in controller and exchanges.js

Bianka2112:
Update Exchanges page to offer more readable Type selection categories. Addresses issue #891 [Add human friendly name when creating exchanges in the UI]

Look at Exchanges live view

### DIFF
--- a/src/lavinmq/http/controller/main.cr
+++ b/src/lavinmq/http/controller/main.cr
@@ -9,6 +9,7 @@ module LavinMQ
 
       OVERVIEW_STATS = {"ack", "deliver", "get", "deliver_get", "publish", "confirm", "redeliver", "reject"}
       EXCHANGE_TYPES = {"direct", "fanout", "topic", "headers", "x-federation-upstream", "x-consistent-hash"}
+      EXCHANGE_TYPES_HUMAN = {"Direct", "Fanout", "Topic", "Headers", "Federation Upstream", "Consistent Hash"}
       CHURN_STATS    = {"connection_created", "connection_closed", "channel_created", "channel_closed",
                         "queue_declared", "queue_deleted"}
 
@@ -106,6 +107,7 @@ module LavinMQ
             {% end %} } {% end %},
             listeners:      @amqp_server.listeners,
             exchange_types: EXCHANGE_TYPES.map { |name| {name: name} },
+            exchange_types_human: EXCHANGE_TYPES_HUMAN.map { |name| {name: name} },
           }.to_json(context.response)
           context
         end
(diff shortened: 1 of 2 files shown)


Bianka2112:

Bianka2112:
recheck

static/js/exchanges.js
```diff
@@ -6,7 +6,7 @@ import * as Table from './table.js'
 Helpers.addVhostOptions('addExchange')
 
 HTTP.request('GET', 'api/overview').then(function (response) {
-  const exchangeTypes = response.exchange_types
+  const exchangeTypes = response.exchange_types_human
```
spuun: This will probably break the create form. The exchange_types_human value can only be used as display text, the old value needs to be used as value of the option.

src/lavinmq/http/controller/main.cr
```diff
@@ -9,6 +9,7 @@ module LavinMQ
 
       OVERVIEW_STATS = {"ack", "deliver", "get", "deliver_get", "publish", "confirm", "redeliver", "reject"}
       EXCHANGE_TYPES = {"direct", "fanout", "topic", "headers", "x-federation-upstream", "x-consistent-hash"}
+      EXCHANGE_TYPES_HUMAN = {"Direct", "Fanout", "Topic", "Headers", "Federation Upstream", "Consistent Hash"}
```
spuun: I think it's better to change `EXCHANGE_TYPES` to a Hash of `exchange type => human friendly text`:
```suggestion
      EXCHANGE_TYPES = {
        "direct" => "Direct",
        "fanout" => "Fanout",
        "topic" => "Topic",
        "headers" => "Headers",
        "x-federation-upstream" => "Federation upstream",
        "x-consistent-hash" => "Consistent Hash"
       }
```
This should serialize nicely to a json object which can be used to create the `option` elements client side.

spuun: Oh, we must be compatible with rabbitmq, so it's better to do a tuple of named tuples:
```suggestion
      EXCHANGE_TYPES = {
        {name: "direct", human: "Direct"},
        {name: "fanout", human: "Fanout"},
        {name: "topic", human: "Topic"},
        {name: "headers", human: "Headers"},
        {name: "x-federation-upstream", human: "Federation upstream"},
        {name: "x-consistent-hash", human: "Consistent Hash"}
       }
```

Bianka2112: Thank you! I had assumed a solution in this direction was needed but I wasn't sure how it should be implemented. Learned something new with Tuples and Named Tuples. Will update with this model.

PR #1244, opened 2025-09-10
Prevent delayed exchanges to bind to its internal delayed queue

### LINKED ISSUE #1229: Binding a delayed exchange to its own internal delayed queue creates an infinite loop
**Describe the bug**
LavinMQ allows binding a delayed exchange to its automatically-created internal delayed queue (amq.delayed.{exchange_name}), which creates an infinite routing loop.

What happens: 
  1. Message routes to internal queue with TTL
  2. After expiration, dead-letters back to delayed exchange (without x-delay)
  3. Exchange routes message back to internal queue (due to binding)
  4. Message expires immediately (no TTL), dead-letters back to exchange
  5. Infinite loop continues

**How to reproduce**

  1. Create a delayed exchange (e.g., my-delayed)
  2. Bind the delayed exchange to its internal queue amq.delayed.my-delayed
  3. Publish a message with x-delay header

**Expected behavior**
  The system should prevent binding a delayed exchange to its corresponding delayed queue to avoid this scenario.

kickster97:
Fixes #1229 by adding a valdiation check when we bind an exchange, so its not allowed to bind to its own internal delayed queue. 

Run the spec that was added in this pull request.

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -188,6 +188,14 @@ module LavinMQ
         raise AccessRefused.new(self)
       end
 
+      protected def validate_delayed_binding(destination : AMQP::Destination) : Bool
+        return true unless delayed?
+        if destination.is_a?(Queue) && destination.name == "amq.delayed.#{@name}"
+          raise LavinMQ::Error::PreconditionFailed.new("Cannot bind delayed exchange '#{@name}' to its internal delayed queue '#{destination.name}'")
+        end
+        true
+      end
+
       abstract def type : String
       abstract def bind(destination : AMQP::Destination, routing_key : String, arguments : AMQP::Table?)
       abstract def unbind(destination : AMQP::Destination, routing_key : String, arguments : AMQP::Table?)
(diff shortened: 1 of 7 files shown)


src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -188,6 +188,14 @@ module LavinMQ
         raise AccessRefused.new(self)
       end
 
+      protected def validate_delayed_binding(destination : AMQP::Destination) : Bool
+        return true unless delayed?
+        if destination.is_a?(Queue) && destination.name == "amq.delayed.#{@name}"
```
spuun: I think a better check is
```suggestion
        if destination == @delayed_queue
```

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -249,7 +257,10 @@ module LavinMQ
         each_destination(routing_key, headers) do |d|
           case d
           in LavinMQ::Queue
-            queues.add(d)
+            # Prevent routing to own internal delayed queue to avoid infinite loops
+            unless delayed? && d.name == "amq.delayed.#{@name}"
```
spuun:
```suggestion
            unless delayed? && d == @delayed_queue
```

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -188,6 +188,14 @@ module LavinMQ
         raise AccessRefused.new(self)
       end
 
+      protected def validate_delayed_binding(destination : AMQP::Destination) : Bool
```
spuun: I think it should have an exclamation mark because of the raising.
```suggestion
      protected def validate_delayed_binding!(destination : AMQP::Destination) : Bool
```

spuun (changes requested):
See inline comments.

src/lavinmq/amqp/exchange/exchange.cr
kickster97: Ah ofc, yes! nice

PR #1245, opened 2025-09-10
update regex textbox to live update when typing

Bianka2112:
Updates the functionality of textbox filter for smoother UX. As the user types, the filtering results update directly. Previously only filtered after Enter; Backspace+Enter to change results. 
Tiny typo fix for renderSearch.

Begin typing into any rendered "Filter regex" textbox, that has a list of queues, policies, etc.
See filtered results based on user input. 
No need to press Enter or delete input+Enter to see results.

### DIFF
--- a/static/js/table.js
+++ b/static/js/table.js
@@ -76,8 +76,8 @@ function renderTable (id, options = {}, renderRow) {
             t.insertBefore(foundRow, currentRow)
           }
         } else {
-          // New item, create new row
-          const tr = t.insertRow(i)
+          // New item, create new row (clamp i to allowed maximum for logs)
+          const tr = t.insertRow(Math.min(i, t.rows.length))
           setKeyAttributes(tr, item)
           renderRow(tr, item, true)
         }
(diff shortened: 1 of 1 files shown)


static/js/table.js
```diff
@@ -101,24 +101,35 @@ function renderTable (id, options = {}, renderRow) {
...
+    
+    const apply = () => {
+      dataSource.searchTerm = filterInput.value
+      dataSource.page = 1
+      reload()
+    }
+    let liveType 
+    filterInput.addEventListener('input', () => {
+      clearTimeout(liveType)
+      liveType = setTimeout(apply, 120)
```
spuun, line 124: I think 120 may be a little bit to low. After some googling 500 may seems like a good value? WDYT?

static/js/table.js
```diff
@@ -101,24 +101,35 @@ function renderTable (id, options = {}, renderRow) {
...
     filterInput.classList.add('filter-table')
     filterInput.placeholder = 'Filter regex'
     filterInput.value = dataSource.searchTerm ?? ''
     form.appendChild(filterInput)
     container.insertBefore(form, container.children[0])
-    container.addEventListener('keyup', e => {
-      if (!e.target.classList.contains('filter-table')) return true
+    
+    const apply = () => {
+      dataSource.searchTerm = filterInput.value
```
spuun, line 117: We should probably clear the timeout in `apply` as well. Now I think `Enter` will apply once, and just after an timeout will occur and trigger apply again.

Bianka2112:
Renaming the branch for clarification of whats happening, which auto closed the PR.

PR #1263, opened 2025-09-12
Move initialization of some BoolChannel to prevent leaks

spuun:
A `BoolChannel` will spawn a fiber in it's initializer, meaning that if a `BoolChannel` is created in the "class body" `@i = BoolChannel.new(true)` and not in the initializer of the class using it, an exception will abort the initialization but the `BoolChannel` won't be closed.

We should make sure to initialize `BoolChannels` as late as possible in initializer to prevent unnecessary fiber spawning and cleanup.

This PR is the result of narrowing down a memory leak in `StreamConsumer`. The problem was that `@new_message_available` and `@has_capacity` (in `Consumer`) was initialized before `validate_preconditions` was called in the initializer of `StreamConsumer`. If `validate_preconditions` raised because of invalid frame we ended up with two dangling `BoolChannels`.

Uhm, manually by observing fibers before and after this PR...

### DIFF
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -20,7 +20,7 @@ module LavinMQ
       @flow : Bool
       @metadata : ::Log::Metadata
       @unacked = Atomic(UInt32).new(0_u32)
-      getter has_capacity = BoolChannel.new(true)
+      getter has_capacity : BoolChannel
 
       def initialize(@channel : AMQP::Channel, @queue : Queue, frame : AMQP::Frame::Basic::Consume)
         @tag = frame.consumer_tag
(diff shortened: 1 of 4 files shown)


viktorerlingsson:
Nice find! 

Could/should we do the same with `empty` in `MessageStore` and `paused` & `consumers_empty` in `Queue`?

spuun:
> Nice find!
> 
> Could/should we do the same with `empty` in `MessageStore` and `paused` & `consumers_empty` in `Queue`?

I looked through all `BoolChannel` usage and think we're fine at the moment. But I think we should a better overall solution, but that could be for another PR.

viktorerlingsson:
> > Nice find!
> > Could/should we do the same with `empty` in `MessageStore` and `paused` & `consumers_empty` in `Queue`?
> 
> I looked through all `BoolChannel` usage and think we're fine at the moment. But I think we should a better overall solution, but that could be for another PR.

Okay, nice!

PR #1268, opened 2025-09-12
Sidebar cleanup and icons added

bengtmagnus:
Adds icons to the sidebar navigation and divides the menu options into groups. Both for darkmode and lightmode.

See if it looks ok, the links are working as they should and the groups are divided correctly.

### DIFF
--- a/views/partials/header.ecr
+++ b/views/partials/header.ecr
@@ -55,40 +55,40 @@
       {
         title: "Cluster",
         items: [
-          { href: ".",           label: "Overview",   icon: "icon-overview" },
+          { href: ".", label: "Overview", icon: "icon-overview" },
           { href: "nodes", label: "Nodes", icon: "icon-nodes" },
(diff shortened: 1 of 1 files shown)


views/partials/header.ecr
```diff
@@ -51,24 +51,61 @@
...
-      "logs":         "Logs"
-      }.each do |href, label| -%>
-        <li<%= " class=\"active\"" if active_path?(href) %>><a href="<%=href%>"><%=label%></a></li>
-      <%-end-%>
-    <li><a href="docs/" target="_blank">HTTP API</a></li>
+    <% [
+      {
+        title: "Cluster",
+        items: [
+          { href: ".",           label: "Overview",   icon: "icon-overview" },
```
kickster97: Are the extra spaces here on pupose? feels like we should not have extra spaces after the comma?

bengtmagnus: Oops, my mistake. Not intended :)

baelter (approved):

PR #1274, opened 2025-09-15
Table filter: search as you type

Bianka2112:
- Takes suggested timeout of 500ms after user stops typing to update results. Add clearTimeout for apply().
- Updates search functionality with a native 'x' to clear and ESC key to clear. 
- Previously had a tiny typo fix and added small clamp fx for rendering Logs view with renderTable fx (separate branch update coming for Logs page).

- Type into any filter regex textbox to see 'x' for quick text clearing, or press ESC. 
- Test timeout feeling, for feedback on results delay.

### DIFF
--- a/static/js/table.js
+++ b/static/js/table.js
@@ -76,8 +76,8 @@ function renderTable (id, options = {}, renderRow) {
             t.insertBefore(foundRow, currentRow)
           }
         } else {
-          // New item, create new row (clamp i to allowed maximum for logs)
-          const tr = t.insertRow(Math.min(i, t.rows.length))
+          // New item, create new row
+          const tr = t.insertRow(i)
           setKeyAttributes(tr, item)
           renderRow(tr, item, true)
         }


spuun (approved):
LGMT! Nice thinking about the `search` event and escape!

Make sure to fix the JS linting. Running `npx standard static/js --fix` is probably enough to fix it!

Bianka2112:
lol.. what is LGMT? 😅 and do I push another commit or how do I pass the tests after updating npx?

spuun:
> lol.. what is LGMT? 😅

I misspelled (🤦 ) LGTM which means Looking Good To Me 😃 

>  and do I push another commit or how do I pass the tests after updating npx?

Yeah, just make a "lint" commit (message can be just "lint").

static/js/table.js
```diff
@@ -76,8 +76,8 @@ function renderTable (id, options = {}, renderRow) {
             t.insertBefore(foundRow, currentRow)
           }
         } else {
-          // New item, create new row
-          const tr = t.insertRow(i)
+          // New item, create new row (clamp i to allowed maximum for logs)
+          const tr = t.insertRow(Math.min(i, t.rows.length))
```
spuun: @Bianka2112 I was about to merge this, but took another look and wonder if this `Math.min` does anything? `i` should always be less than or equal to `r.rows.length`.
Was this change indented for this PR is it for #1279? Even if it is, I wonder if it will affect anything?

spuun: I noticed it's in the other PR too. Then I'll remove it from this PR.

PR #1275, opened 2025-09-15
Handle user permissions in MQTT `broker`

### LINKED ISSUE #1313: Handle user permissions in MQTT broker
Currently, the only permissions check for a user happens on connection with MQTT. 

I want write and read permissions to be checked on publish and subscribe

kickster97:
Adds permission checks for publishing and subscribing via MQTT, which are the two operation thats exposed to clients. 

Fixes #1313 

Run spec `spec/mqtt/permissions_spec.cr`

### DIFF
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -11,7 +11,6 @@ module LavinMQ
   module MQTT
     class Broker
       getter vhost, sessions
-      Log = LavinMQ::Log.for "mqtt.broker"
 
       # The `Broker` class acts as an intermediary between the `Server` and MQTT connections.
       # It is initialized by the `Server` and manages client connections, sessions, and message exchange.


spuun (commented):
LavinMQ only has "resource permissions", not topic permissions, and I interpret #1275 as a request for topic permissions.  However, I still think should add this.

src/lavinmq/mqtt/broker.cr
```diff
@@ -70,14 +70,20 @@ module LavinMQ
         @vhost.rm_connection(client)
       end
 
-      def publish(packet : MQTT::Publish)
+      def publish(user : Auth::User, packet : MQTT::Publish)
+        unless user.can_write?(@vhost.name, packet.topic)
```
spuun: This is wrong usage of `#can_write?`. The second argument should be the "resource name", in this case the exchange name.

src/lavinmq/mqtt/broker.cr
```diff
@@ -70,14 +70,20 @@ module LavinMQ
...
+          raise LavinMQ::Exchange::AccessRefused.new(@exchange)
+        end
         @exchange.publish(packet)
       end
 
       def subscribe(client, topics)
         session = sessions.declare(client)
         headers = AMQP::Table.new({RETAIN_HEADER => true})
         topics.map do |tf|
+          unless client.user.can_read?(@vhost.name, tf.topic)
```
spuun: This is wrong usage of `#can_read?`. For subscribe (bind) one must check both read on exchange and write on queue. But maybe this can work different with mqtt.

spuun (commented):
See inline comments

src/lavinmq/mqtt/broker.cr
kickster97: Yeah, the session(queue) "belongs" to the client, and client A does not really have access to client B's session.. so i think it can work a bit different for MQTT?

but definitely check read for the exchange, and check whether the user can subscribe to topics at all.

src/lavinmq/mqtt/broker.cr
```diff
@@ -70,14 +70,20 @@ module LavinMQ
...
         @exchange.publish(packet)
       end
 
       def subscribe(client, topics)
         session = sessions.declare(client)
         headers = AMQP::Table.new({RETAIN_HEADER => true})
         topics.map do |tf|
+          unless client.user.can_read?(@vhost.name, EXCHANGE)
+            raise LavinMQ::Exchange::AccessRefused.new(@exchange)
+          end
```
spuun: This can be moved to before loop since it's the same check every iteration.

spec/mqtt/permissions_spec.cr
```diff
@@ -0,0 +1,116 @@
...
+            topic_filter = MQTT::Protocol::Subscribe::TopicFilter.new("test/topic", 0u8)
+            subscribe(io, topic_filters: [topic_filter])
+
+            # Connection should still be alive
+            io.should_not be_closed
+          end
+        end
+      end
+
+      it "should enforce topic-specific write permissions" do
```
spuun: This spec can be removed since we have no topic permissions

spec/mqtt/permissions_spec.cr
```diff
@@ -0,0 +1,116 @@
...
+          with_client_io(server) do |io|
+            connect(io, username: "topic_user", password: "pass".to_slice)
+            # This should fail and close connection
+            publish(io, false, topic: "denied/topic", payload: "hello".to_slice)
+            io.should be_closed
+          end
+        end
+      end
+
+      it "should enforce topic-specific read permissions" do
```
spuun: This spec can be removed since we have no topic permissions

spec/mqtt/permissions_spec.cr
```diff
@@ -0,0 +1,116 @@
...
+            connect(io, username: "read_user", password: "pass".to_slice)
+            # This should fail and close connection
+            topic_filter = MQTT::Protocol::Subscribe::TopicFilter.new("denied/topic", 0u8)
+            subscribe(io, false, topic_filters: [topic_filter])
+            io.should be_closed
+          end
+        end
+      end
+
+      it "should block will messages when user lacks write permissions to will topic" do
```
spuun:
```suggestion
      it "should block will messages when user lacks write permissions" do
```
and adjust specs accordingly

spuun (changes requested):
Some thoughts!

snichme:
In order to keep the code simpler to read I would like to move the checks somewhere else, they add both more lines to the method but also adds an exception that we need to handle. 

How about doing 
```ruby 
if @broker.can_publish?(@user)
  @boker.publish....
else
  close_socket
end
``` 
To separate the concern for each method 

Another approach is to move the checks into a `check_permissions` method that we can use from `Broker#"publish` instead of having the `if`/`unless` checks in the method but then we still have the exception handling..

kickster97:
> In order to keep the code simpler to read I would like to move the checks somewhere else, they add both more lines to the method but also adds an exception that we need to handle. 
.... 

@snichme I agree with you, it would be nice to have them in `client.cr` (i think this is the way we do it for amqp?).. This would work for `publish()`, but the reason i put them in `broker.cr` is because the session might not exist on `subscribe()`. Since we need to check if we have access to the session we need to do the check after its declared. (we can discuss if its needed to check the permissions of a session, but this is how permissions work for amqp) 

I want the checks to be in the same file at least, so we avoid doing checks in different places in the code. Maybe breaking this out to a method is a way to go

src/lavinmq/mqtt/broker.cr
```diff
@@ -70,7 +71,7 @@ module LavinMQ
         @vhost.rm_connection(client)
       end
 
-      def publish(packet : MQTT::Publish)
+      def publish(user : Auth::User, packet : MQTT::Publish)
```
snichme: No need to pass user here

src/lavinmq/mqtt/client.cr
```diff
@@ -179,15 +190,17 @@ module LavinMQ
 
       private def publish_will
         if will = @will
-          @broker.publish MQTT::Publish.new(
+          @broker.publish(@user, MQTT::Publish.new(
```
snichme: Need to check write permissions for the vhost + exchange here as well and you can remove the rescue for AccessRefused

src/lavinmq/mqtt/broker.cr
kickster97: was a bit too fast with the changes this Friday 😅 thanks!

src/lavinmq/mqtt/broker.cr
```diff
@@ -11,6 +11,7 @@ module LavinMQ
   module MQTT
     class Broker
       getter vhost, sessions
+      Log = LavinMQ::Log.for "mqtt.broker"
```
snichme: Is this needed for this PR?

kickster97: ah no your right, this is probably a rest from when I had the logs in broker, so does no longer make sense

kickster97:
Waiting to merge untill 2.6.x

Issue #993, opened 2025-03-16
Logs page filter


jacob-84Codes:
**Summary**  
update the logs tab to include behavior similar to the queues, policy page etc.

**Proposed Solution (for improvements)**  
Implement regex filter textbox and relevant orderings eg: message (alphabetical), timestamp, severity (possibly just filter)

**Screenshots/Mockups**  
Current Logs page:

<img width="1395" alt="Image" src="https://github.com/user-attachments/assets/3886a573-583e-496a-89ff-bac43e12a057" />

Queues page with filters:

<img width="1398" alt="Image" src="https://github.com/user-attachments/assets/7056a23f-0fa3-44ee-905e-6381ea7981ca" />

ShemaleHelicopter:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #1279, opened 2025-09-15
logs page- added filter textbox and dynamic table

Bianka2112:
Renders: a filter regex textbox, logs in a searchable and sortable table, the columnSelector, a list count.

Have included the small insertRow edit again in table.js (shared in PR Feat/filter textbox #1274) so that it can render enough logs, otherwise it errors beyond 16rows. 

Included TODO comments to note where known fixes are needed. 

p.s. Would have tried to break this up in separate commits, but the way the rendering all comes from one fx, renderTable - it felt better to create a complete working table and then get the feedback. 😄 always happy to get workflow suggestions as well!

Try out Logs page view to see new features from this branch.

### DIFF
--- a/static/js/logs.js
+++ b/static/js/logs.js
@@ -1,7 +1,7 @@
 import * as Table from './table.js'
 
 class LiveLogDataSource {
-  constructor () {
+  constructor (streamUrl) {
     this.items = []
     this.totalCount = 0
     this.page = 1
@@ -18,10 +18,38 @@ class LiveLogDataSource {
     // Internal state:
     this.allLogs = []
     this._event = new EventTarget()
+
+    // Event Source
+    this._evtSource = new window.EventSource(streamUrl)
+    this._evtSource.onmessage = (event) => {
+    const timestamp = new Date(parseInt(event.lastEventId, 10))
+    const [severity, source, message] = JSON.parse(event.data)
+    this.pushLog({ timestamp, severity, source, message })
   }
+
+    this._evtSource.onerror = () => {
+      window.fetch('api/whoami')
+        .then(response => response.json())
+        .then(whoami => {
+          if (!whoami.tags.includes('administrator')) {
+            forbidden()
+          }
+        })
+    }
+
+    function forbidden () {
+      const tblError = document.getElementById('table-error')
+      tblError.textContent = 'Access denied, administator access required'
+      tblError.style.display = 'block'
+    }
+}
 
   on (eventName, handler) {
     this._event.addEventListener(eventName, handler)
+  }
+
+  close () {
+    this._evtSource?.close()
   }
 
   reload () {
@@ -99,10 +127,11 @@ function compareValues (a, b, direction) {
 }
 
 // Build table dynamically using table.js
-const logsDataSource = new LiveLogDataSource()
+const logsDataSource = new LiveLogDataSource('api/livelog')
+
 const tableOptions = {
   dataSource: logsDataSource,
-  keyColumns: ['timestamp', 'severity', 'source', 'message'],
+  keyColumns: ['timestamp'],
   pagination: false,
   columnSelector: true, // TODO: improve position/placement
   search: true,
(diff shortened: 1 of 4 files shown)


static/js/logs.js
```diff
@@ -1,28 +1,116 @@
...
+  pre.textContent = item.message
+  Table.renderCell(tr, 3, pre)
+})
+
+const evtSource = new window.EventSource('api/livelog')
+evtSource.onmessage = (event) => {
+  const timestampMs = parseInt(event.lastEventId, 10)
+  const [severity, source, message] = JSON.parse(event.data)
+  logsDataSource.pushLog({ timestamp: timestampMs, severity, source, message })
 }
```
spuun: I think this should be a part of the `LiveLogDataSource`. Maybe the `EventSource` can be passed to the constructor? Just just the URI and let `LiveLogDataSource` create the `EventSource` and setup all events?

static/js/logs.js
```diff
@@ -1,28 +1,116 @@
...
+    }
+
+    this.items = visible
+    this.totalCount = visible.length
+    this._event.dispatchEvent(new CustomEvent('update'))
+  }
+
+  pushLog (log) {
+    this.allLogs.push(log)
+    this.reload()
```
spuun: We need to have some kind of "debounce" of reload. If i navigate to the logs page it freezes for sometime because it's being served a lot of already existing log entries, and it will update the table for each of them.

Maybe it should wait 100ms before calling reload, and that timeout is cleared on every new entry. However, we should probably make sure to call `reload` after x entries (or y time without reload being called?) else we may end up in a state where no update is done if lavinmq is logging with a rate of one entry per 100ms or more.

It may be that `Table` isn't really suitable for this, since it's designed for views where rows will be updated, not only appended.

static/js/logs.js
```diff
@@ -1,28 +1,116 @@
...
+  Table.renderCell(tr, 1, item.severity)
+  Table.renderCell(tr, 2, item.source)
+  const pre = document.createElement('pre')
+  pre.textContent = item.message
+  Table.renderCell(tr, 3, pre)
+})
+
+const evtSource = new window.EventSource('api/livelog')
+evtSource.onmessage = (event) => {
+  const timestampMs = parseInt(event.lastEventId, 10)
```
spuun: It seems reasonable to create a `Date` object already here, and then just call `toISOString` in `joinFieldsForSearch` and `toLocaleString` in `Table.renderTable`, i.e. no need for `formatLocal` or `toMs`.

static/js/table.js
```diff
@@ -77,7 +77,7 @@ function renderTable (id, options = {}, renderRow) {
           }
         } else {
           // New item, create new row
-          const tr = t.insertRow(i)
+          const tr = t.insertRow(Math.min(i, t.rows.length))
```
spuun, line 80: What's the intention with this? I don't think this will have any affect?

static/js/logs.js
Bianka2112: Sounds good 👍

static/js/logs.js
Bianka2112: Great point! 😄 I’ll add debounce at ~100ms and force-refresh every 200 messages / 500ms, unless you recommend other values?
Can also cap logs at 5,000 ?

static/js/logs.js
Bianka2112: The biggest factor for change, was to allow the sorting logic to run comparison as numbers directly, stay timezone-neutral and then only converts once when rendering or building the ISOstring. The comparators support getting number or date in either case, so sticking with Date first to store the timestamp is no problem 👌

static/js/table.js
Bianka2112: After first rendering the table for logs, I was getting this error: Something went wrong: Failed to execute 'insertRow' on 'HTMLTableSectionElement': The provided index (48 is outside the range [-1, 15].

So the quick (& dirty 🥲) solution I found was allowing it to always append at the end, regardless of being out of the range and then insertBefore moves it to keep the final order correct. 

But I dug deeper and see now that, it's because of the keyColumns expecting unique identifiers for each row, and logs possibly having repeated keys that it was mismatching the indices and causing "out of range" ones. Will fix by giving logs a unique key. But up to you if we also keep the clamp (Math.min(i, t.rows.length)) as a safety net?

Bianka2112:
Thank you for your feedback, questions and suggestions! This is really helping me learn and amplify how I think about the code/solutions. 🙌 

I did notice that logs was a custom table originally and tried out some edits but since renderSearch only exists as part of renderTable and theres also already header stuff in there, I looked for ways to match the use for consistency and to solve the issue directly. I picked up: Logs page filter #993, thinking it was going to be straightforward and then found myself rewiring the entire thing- super interesting challenge, but if you feel this direction is not well-suited I can look at other options! 🤓

static/js/logs.js
spuun: Start with those values, it will probably fine. Let's wait with capping and try it out first :)

static/js/table.js
spuun: Strange with the index error. 🤔 At a first glance I'd think that `i` never could be greater than `t.rows.length`.

The `Table` is written to be "smart" when updating the view. E.g. when you're the Queues page you'll see a list of queues. This data is re-fetched every 5s, and then we don't want to clear and re-render the table, instead we want to update the existing rows, and this is what `keyColumns` are used for. Logs doesn't need this, so maybe you can omit `keyColumns`?

This is also why I think that maybe Table is wrong to use for logs. It's not made for "streamed data". You'd want to be able to just append data, not refresh the entire view.

Bianka2112: Okay, I think I get what you mean now. renderTable is overkill for the way it repaints existing data that might update, whereas logs are static data so just append-only (excuse any wrong terminology of mine 🤓). 
I finished the changes discussed here for completeness - but I am thinking I can close this PR and create a new branch where we keep logs build as it was, locally add the filter box and maybe use DocumentFragment to update results. 

What are your thoughts?

PR #1280, opened 2025-09-16
StreamQueue -> Stream

snichme:

Rename StreamQueue to Stream since it's not really a queue

Specs

### DIFF
--- a/src/lavinmq/amqp/queue/stream_queue_message_store.cr
+++ b/src/lavinmq/amqp/queue/stream_queue_message_store.cr
@@ -2,377 +2,383 @@ require "./stream_queue"
 require "../stream_consumer"
 
 module LavinMQ::AMQP
-  class StreamQueue < DurableQueue
-    class StreamQueueMessageStore < MessageStore
-      getter new_messages = ::Channel(Bool).new
-      property max_length : Int64?
-      property max_length_bytes : Int64?
-      property max_age : Time::Span | Time::MonthSpan | Nil
-      getter last_offset : Int64
-      @segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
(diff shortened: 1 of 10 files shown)


src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -2,377 +2,383 @@ require "./stream_queue"
 require "../stream_consumer"
 
 module LavinMQ::AMQP
-  class StreamQueue < DurableQueue
```
snichme, line 5: This diff is huge because I removed the `class StreamQueue < DurableQueue` "wrapper"

kickster97 (commented):
Wdyt about also naming the file `stream.cr `(not `stream_queue.cr`). Same with  `stream_queue_message_store.cr` -> `stream_message_store.cr`
I think its nice when the file structure mimics the code structure.
Maybe even the stream should be moved out of the queue folder into a stream folder under AMQP 
 
Maybe moving/renaming can be done in another pr, but it could also fit into this one.

src/lavinmq/amqp/queue/stream_queue_message_store.cr
```diff
@@ -2,377 +2,383 @@ require "./stream_queue"
...
+    def push(msg) : SegmentPosition
+      raise ClosedError.new if @closed
+      msg.properties.headers = add_offset_header(msg.properties.headers, @last_offset += 1)
+      sp = write_to_disk(msg)
+      @bytesize += sp.bytesize
+      @size += 1
+      @segment_last_ts[sp.segment] = msg.timestamp
+      # Consumer notification is handled by StreamQueue.publish directly
+      # to ensure all consumers are notified,
+      # not just one via the shared channel like other queue types
```
viktorerlingsson: We can probably remove this comment?

PR #1283, opened 2025-09-16
Fixes a bug where proxying from a follower that is performing full_sy…

### LINKED ISSUE #1282: Proxying from a follower does not work while the follower is performing full_sync
Trying to connect via proxy while the follower is performing it's initial sync yields errors on the leader: 

`server-01 lavinmq[3350]:  lmq.amqp.connection_factory[address: "x.x.x.x:#####"] Unexpected protocol "PROXY TC", closing socket`

Because `extract_conn_info` does not find a match in `followers.any? { |f| f.remote_address.address == remote_address.address }` [server.cr#L127](https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/server.cr#L127), because the follower is only added to the followers array after it has completed the full sync.

viktorerlingsson:
Fixes a bug where proxying from a follower that is performing full_sync does not work. This was because followers was only added to the followers array after the full sync was completed, meaning `extract_conn_info` could not find the follower when trying to proxy. That lead to the proxied connection to be treated as non-proxied, causing errors (`server-01 lavinmq[3350]:  lmq.amqp.connection_factory[address: "x.x.x.x:#####"] Unexpected protocol "PROXY TC", closing socket`)

This PR adds state to followers (`Synced/Syncing`). The state is then used in some helper methods to get either only synced, syncing or all followers. extract_conn_info can then look at a complete list of followers and allow the syncing followers to proxy their traffic to the leader. 

Fixes #1282 

### Future considerations
This means that we know which followers are in the process of syncing, meaning we could return that in the API response, and show it in the GUI. 

Run the added spec

### DIFF
--- a/src/lavinmq/clustering/replicator.cr
+++ b/src/lavinmq/clustering/replicator.cr
@@ -9,6 +9,8 @@ module LavinMQ
       abstract def append(path : String, obj)
       abstract def delete_file(path : String, wg : WaitGroup)
       abstract def followers : Array(Follower)
+      abstract def syncing_followers : Array(Follower)
+      abstract def all_followers : Array(Follower)
       abstract def close
       abstract def listen(server : TCPServer)
       abstract def clear
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -127,6 +127,18 @@ module LavinMQ
 
       def followers : Array(Follower)
         @lock.synchronize do
+          @followers.select(&.synced?).dup # for thread safety
+        end
+      end
+
+      def syncing_followers : Array(Follower)
+        @lock.synchronize do
+          @followers.select(&.syncing?).dup # for thread safety
+        end
+      end
+
+      def all_followers : Array(Follower)
+        @lock.synchronize do
           @followers.dup # for thread safety
         end
       end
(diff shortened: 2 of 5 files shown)


src/lavinmq/clustering/server.cr
```diff
@@ -161,14 +173,18 @@ module LavinMQ
...
+            stale_follower.close
+          end
+          @followers << follower # Starts in Syncing state
         end
         follower.full_sync # sync the bulk
         @lock.synchronize do
-          follower.full_sync # sync the last
-          @followers << follower
+          follower.full_sync    # sync the last
+          follower.mark_synced! # Change state to Synced
```
spuun, line 187: Wouldn't it make more sense if `#full_sync` changes state when it's done?

src/lavinmq/clustering/replicator.cr
```diff
@@ -9,6 +9,8 @@ module LavinMQ
       abstract def append(path : String, obj)
       abstract def delete_file(path : String, wg : WaitGroup)
       abstract def followers : Array(Follower)
+      abstract def syncing_followers : Array(Follower)
+      abstract def all_followers : Array(Follower)
```
spuun, line 13: Maybe not related to this PR, but I wonder if it's possible to change the return type of these to `Iterator(Follower)` to be able to return empty iterators instead of empty arrays in `NoopServer`? It would save us some allocations.

spuun (approved):
Ah, nice find!

src/lavinmq/clustering/server.cr
viktorerlingsson: Maybe, but we do `full_sync` twice, once for bulk syncing and then again to sync the last.

src/lavinmq/clustering/replicator.cr
viktorerlingsson: Yeah, maybe!

src/lavinmq/clustering/server.cr
spuun: Ah, right! Let's go with this for now then!

PR #1284, opened 2025-09-17
UI: Show stream consumer position in queue view

jage:
This is a demoe to show how this could work, not expecting this branch to be merged but rather as inspiration for a "real" implementation if we like this.

Background:

Since there's no concept of ready or unacked messages in streams, it is hard to see if the consumers are replaying old messages and/or not keeping up with the inflow and retention policy of a stream.

This is a rough draft how it could look like, besides listing consumers with position and throughput in a table, we also visualize the position by drawing a progress bar with all consumers.

Screenshot 2025-09-17 at 10 01 14

Video showing two consumers starting at the first stream position.

https://github.com/user-attachments/assets/baad72e0-dc94-446d-a59f-44969683d2e9

### DIFF
--- a/src/lavinmq/amqp/stream/stream_consumer.cr
+++ b/src/lavinmq/amqp/stream/stream_consumer.cr
@@ -7,7 +7,6 @@ require "./filters/gis"
 module LavinMQ
   module AMQP
     class StreamConsumer < Consumer
-      include SortableJSON
       property offset : Int64
       property segment : UInt32
       property pos : UInt32
@@ -165,6 +164,18 @@ module LavinMQ
         else
           @filters.all?(&.match?(headers))
         end
+      end
+
+      def details_tuple
+        tuple = super
+        tuple.merge({
+          stream_filters: {
+            # filters:          @filters,
+            match_type:       @filter_match_all ? "all" : "any",
+            match_unfiltered: @match_unfiltered,
+          },
+          stream_offset: @offset,
+        })
       end
     end
   end
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -9,6 +9,7 @@ module LavinMQ
   module AMQP
     class Consumer < LavinMQ::Client::Channel::Consumer
       include SortableJSON
+      include Stats
       Log = LavinMQ::Log.for "amqp.consumer"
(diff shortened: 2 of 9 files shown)


viktorerlingsson:
This is very cool! Would like to see us implementing something like this in the future 👍

snichme:
Might help this PR out https://github.com/cloudamqp/lavinmq/pull/1331 introduces a separate view for streams

src/lavinmq/amqp/stream/stream_consumer.cr
```diff
@@ -166,6 +165,18 @@ module LavinMQ
           @filters.all?(&.match?(headers))
         end
       end
+
+      def details_tuple
+        tuple = super
+        tuple.merge({
+          stream_filters: {
+            # filters:          @filters,
```
snichme, line 173: `@filters`  aren't no longer serializable to json since they are a list of `StreamFilter`. 
`StreamFilter` is a module, so we could introduce a 'to_json` either concrete or abstract. Or as a "view" wrapper around each of them which would be responsible to generate the JSON.

src/lavinmq/amqp/stream/stream_consumer.cr
```diff
@@ -166,6 +165,18 @@ module LavinMQ
...
+            match_type:       @filter_match_all ? "all" : "any",
+            match_unfiltered: @match_unfiltered,
+          },
+          stream_offset: @offset,
```
snichme, line 177: When I ran this locally I had a stream with 38 messages in it and for the consumers who had read the last message had `@offset` 39 so we rendered that as "Offset: 39 (103%)" which looks odd. 

`@offset` in this case means next to read and we should maybe keep them separate.

snichme (commented):
Rebased this onto main and move the position view to "stream.cr" the stream page which now is separated from queue. 


One thing I noticed and I think that we should have another round of discussions for is: We now show offset as the highest number a consumer has fetched, but not yet commited. 
If I start a consumer on offset=0 with prefetch=100 I will get Offset=100 in this view, but if the consumer crashes it will probably re-read some messages. 
If would be cool if we in the progress bar could show it as a range like 
```
[________|<--__consumer___-->|____________________]
```
So the beginning of the consumer range is last commited and the end of the range is offset.

Issue #1237, opened 2025-09-05
Queue `unacked_count` Arithmetic Underflow


kickster97:
**Describe the bug**
A race condition in queue's `unacked_count` tracking causes the counter to wrap around to UInt32::MAX when double-decremented (0-1 wraps to 4,294,967,295). This happens when message delivery fails and the counter is decreased in `mark_unacked`, but the client successfully receives and ACKs the message concurrently. The wrapped counter then causes `OverflowError` in `details_tuple` when calculating queue statistics.

**Describe your setup**
Lavinmq 2.5.0-rc.1

**How to reproduce**
Maybe not reproduce, but an explanation: 
 1. Message delivery starts via `mark_unacked` - `@unacked_count` incremented (queue.cr:796)
 2. During `yield` (message delivery), network/channel error occurs
 3. Rescue block executes - `@unacked_count` decremented (queue.cr:805)
 4. Concurrently, client successfully received message and sends ACK
 5. `ack` method processes ACK - `@unacked_count` decremented again (queue.cr:839)
 6. Double decrement: `0 - 1` wraps to `UInt32::MAX` (4,294,967,295)

**Expected behavior**
We should guard against decrementing unacked_count twice for the same message.

PR #1288, opened 2025-09-17
Introduce AcknowledgementTracker 

kickster97:
AcknowledgementTracker is a class that is responsible for tracking messages, marking and unmarking messages as acknowledged. 

This is the first step towards a more centralized acknowledgement flow. 
Thoguhts and idéas are very welcome 

There should be no functionality change, run spec

### DIFF
--- a/spec/api/queues_spec.cr
+++ b/spec/api/queues_spec.cr
@@ -128,8 +128,8 @@ describe LavinMQ::HTTP::QueuesController do
             2.times { q.publish "m1" }
             q.subscribe(no_ack: false) { }
 
-            wait_for { s.vhosts["/"].queues["unacked_q"].unacked_count == 2 }
-            s.vhosts["/"].queues["unacked_q"].basic_get_unacked.size.should eq 0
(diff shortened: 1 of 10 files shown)


carlhoerberg:
What's the idea here?

kickster97:
> What's the idea here?

to clear up the acknowledgement flow. It started out with he bug we have in #1237, we want some good way to track that a segment position has been marked unacknowledged, so we don't end up in a position where we unack a message twice. To do this we discussed adding even more trackers and it was all a bit spagetty-ish. if we break it out like this we can more easily implement safe tracking without adding even more stuff to queue (which is long as it is now). 
I also think it makes sense to break the tracking of the ack counters out of queue since we have it in channel as well, there might be some inconsistencies with how we mark as acked/unacked between the queue and the channel. It would be nice to have a more centralized handling.

wdyt 

edit: this is very much a draft, probably better off not to move the queues methods, rather have its own tracker methods so we don't need ot change the references everywhere

PR #1300, opened 2025-09-23
Topic exchange doesn't not route messages properly

### LINKED ISSUE #1299: Topic exchange fails with some routing keys
**Describe the bug**

If you bind a queue to `a.*` you would get messages published with `a.b.c` but the specification for `*` says that `*` matches one segment. So it should match `a.b` but not `a.b.c`

snichme:

Fixes #1299

Specs

### DIFF
--- a/src/lavinmq/amqp/exchange/topic.cr
+++ b/src/lavinmq/amqp/exchange/topic.cr
@@ -8,7 +8,7 @@ module LavinMQ
       def initialize(raw : Bytes)
         if first_dot = raw.index '.'.ord
           @value = raw[0, first_dot]
-          @raw = raw[(first_dot + 1)..]
+          @raw = raw[(first_dot + 1)..]? || Bytes.empty
         else
           @value = raw
           @raw = Bytes.empty


src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -2,8 +2,91 @@ require "./exchange"
...
+          else
+            v == @s && rest.empty?
+          end
+        end
+      end
+
+      @checker : Segment?
+
+      def initialize(@key : Array(String))
+        @checker = key.reverse.reduce(nil) do |prev, v|
```
carlhoerberg: avoid allocations 
```suggestion
        @checker = @key.reverse_each.reduce(nil) do |prev, v|
```

src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -2,8 +2,91 @@ require "./exchange"
...
+
+      def matches?(rk : Array(String)) : Bool
+        if checker = @checker
+          checker.match?(rk)
+        else
+          false
+        end
+      end
+
+      def as_fanout?
```
carlhoerberg:
```suggestion
      def is_fanout?
```

src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -2,8 +2,91 @@ require "./exchange"
...
+      end
+
+      class StringSegment < Segment
+        def initialize(@s : String, @next : Segment?)
+        end
+
+        def match?(rk : Array(String)) : Bool
+          v = rk.first?
+          return false unless v
+          rest = rk[1..]
```
carlhoerberg: Wondering if the rk also should be a linked list of strings, to avoid all  allocations on each match check

src/lavinmq/amqp/exchange/topic.cr
snichme: or `acts_as_fanout?` it's not a fanout but behaves as one

src/lavinmq/amqp/exchange/topic.cr
snichme: ahh, `first` produces a new string 
and looks like  `[1..]` produces a new array as well for each call. 

will test with another approach

src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -2,8 +2,100 @@ require "./exchange"
...
+            else
+              false
+            end
+          else
+            rk.next.nil?
+          end
+        end
+      end
+
+      class StringSegment < StarSegment
```
spuun: Any reason a `StringSegment` is a `StarSegment` and not a `Segment`?

snichme: Yes, `StringSegment::match?` uses the same logic as `StarSegment` it ju also needs to check that the string is equal. It can inherit directly from `Segment` but then we need to duplicate the logic form StarSegment. Not a big problem as I dont expect it to change very often.

spuun: Missed that `super` call. Anyway, it feels like an odd inheritance for some reason... Can't wrap my head around if String is extending Star, or if it's actually the other way around. 

Maybe 
```crystal
if check = @next
            if n = rk.next
              check.match?(n)
            else
              false
            end
          else
            rk.next.nil?
          end
```
should be a method on `Segment` that can be called from `match?`? Similar to `#call_next` on `HTTP::Handler`.

snichme: I agree that the inherit doesn't make much sense. I see no real problem with duplicate that small code snippet and have StringSegment inherit Segment directly. 

A risk with extracting it as a method in `Segment` is that it might expand for more use cases later.

src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -51,74 +155,17 @@ module LavinMQ
...
+          if bk.acts_as_fanout?
+            destinations.each do |destination, _binding_key|
+              yield destination
             end
+            return
           end
         end
 
-        rk_parts = routing_key.split(".")
+        rk = RkIterator.new(routing_key.to_slice)
```
snichme, line 166: @spuun need some 👀 on this. All specs passes but creating `rk` an instance of `RkIterator` would that be consumed on the first loop of bindings and then fail the next iteration? It doesn't, is that because it's a struct and we copy it into each `matches?` call?

spuun: Since it's copied and no state is shared it should work just fine.

src/lavinmq/amqp/exchange/topic.cr
```diff
@@ -2,8 +2,111 @@ require "./exchange"
 
 module LavinMQ
   module AMQP
+    struct RkIterator
+      getter value : Bytes
+
+      def initialize(raw : Bytes)
+        if first_dot = raw.index '.'.ord
+          @value = raw[0, first_dot]
+          @raw = raw[(first_dot + 1)..]
```
spuun: It should probably be
```suggestion
          @raw = raw[(first_dot + 1)..]? || Bytes.empty
```

PR #1307, opened 2025-09-25
Logs - with filtering and sticky headers

Bianka2112:
Addresses issue: Logs page filter #993, generally improving UX. 
Implements dynamic filtering from regex textbox, jump buttons to reach top and bottom of table quickly. 
Adjust download button to < a > tag with button css, vs button element wrapping link. 
Sticky toolbar so user can always access actions(filter, download) regardless of table location.
Minimal styling of toolbar and jump buttons, table head row now also sticky.

TODO: 
Timestamp sort
Severity sort - ?? maybe add icons or color?

Checkout branch, test logs page filter box, jump buttons and scroll.

### DIFF
--- a/static/js/logs.js
+++ b/static/js/logs.js
@@ -53,7 +53,7 @@ evtSource.onmessage = (event) => {
   buffer.push(log)
 
   if (matches(log)) {
-    const row = tbody.appendChild(buildRow(log))  
+    tbody.appendChild(buildRow(log))  
     if (shouldAutoScroll) livelog.scrollTop = livelog.scrollHeight
   }
 
@@ -100,6 +100,7 @@ filterInput.addEventListener('keydown', (e) => {
     e.preventDefault()
     rebuildFromBuffer()
   } else if (e.key === 'Escape' && filterInput.value) {
+    currentRegex = null
     filterInput.value = ''
     rebuildFromBuffer()
   }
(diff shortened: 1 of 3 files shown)


static/js/logs.js
```diff
@@ -1,46 +1,123 @@
...
+evtSource.onmessage = (event) => {
+  const timestamp = new Date(parseInt(event.lastEventId))
+  const [severity, source, message] = JSON.parse(event.data)
+  const log = { timestamp, severity, source, message }
+  buffer.push(log)
+
+  if (matches(log)) {
+    tbody.appendChild(buildRow(log))  
+    if (shouldAutoScroll) livelog.scrollTop = livelog.scrollHeight
+  }
```
spuun: I did some testing, and the page slows down and eventually freezes with tons of logs. I wonder if it would be good to buffer log entries in a second "batch" buffer, set a timeout with `requestAnimationFrame` (if not already set) and copy entries from the batch buffer to the original buffer and at the same time append to the DOM (maybe even use a fragment for the batch).

Also, I think we should limit the number of rows and entries in the buffer, let's say 2000 should be enough? I.e throw away old entries/rows.

static/js/logs.js
```diff
@@ -1,46 +1,123 @@
...
+          forbidden()
+        }
+      })
+  }
+
+  function forbidden () {
+    const tblError = document.getElementById('table-error')
+    tblError.textContent = 'Access denied, administator access required'
+    tblError.style.display = 'block'
+  }
```
spuun: Looks like this accidentally has been moved into the `onmessage` callback.

spuun (commented):
It works, but I think we should limit the number of rows. The page slows down and it's almost not possible to scroll when the table is too big.

PR #1311, opened 2025-09-26
Add Topology view

### LINKED ISSUE #1228: Topology Tool for Management Interface
It would be cool to have a view in the management interface to get an overview of the topology of exchanges, bindings, queues, connections/channels, clients etc. That way you could see message flows, more easily identify if there is anything wrong with the architecture in the broker.

kickster97:
Adds a new view to LavinMQ that renders a visualization showing broker architecture and message flow relationships. It displays message flow relationships between exchanges and queues. The tool provides a visual network graph showing exchanges as blue circles and queues as green circles, with animated flow indicators demonstrating real-time message throughput between components. Users can click nodes to view detailed information panels containing type, vhost, durability, policy settings, and binding relationships. 

The topology view integrates with existing AMQP API endpoints (/api/exchanges, /api/queues, /api/bindings).

image

image

Fixes #1228 

Compile LavinMQ and click around in the topology view.

### DIFF
--- a/static/js/topology.js
+++ b/static/js/topology.js
@@ -12,8 +12,7 @@ const CONFIG = {
   MARGINS: { DEFAULT: 80, VHOST: { X: 40, Y: 60 } },
   NODES: { EXCHANGE_RADIUS: 20, QUEUE_RADIUS: 15, NAME_MAX_LENGTH: 20 },
   LAYOUT: { START_SCALE_FACTOR: 1.8, BASE_X_OFFSET: 200 },
-  MAX_NODES_THRESHOLD: 200,
-  ANIMATION: { DASH_SPEED: 0.5, MAX_SPEED_MULTIPLIER: 2 }
+  MAX_NODES_THRESHOLD: 200
 }
 
 // Global state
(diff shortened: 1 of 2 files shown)


spuun:
I'm missing d7 js. I guess it must be added to `Makefile`?

kickster97:
> I'm missing d7 js. I guess it must be added to `Makefile`?

ah hmm i must have had an old make from when looking at the d3 pr.. anyway i added it to the make file so now you can try again :)

spuun (commented):
I think we should make a little bit simpler version which is just a "static snapshot" of the  topology, remove flows (since not "real" numbers anyway).

snichme:
I do like this feature but I think we should reconsider how to display the topology, more specifically the queues. 
I did a quick check and average queue count on CloudAMQP is ~260 queues per cluster, we do have some clusters with extreme queue count that I think we should consider as well. 
Based on this the view will be very long and you need to zoom out to see everything and then you dont see the names. 

One idea is that instead of showing everything directly you first need to input a queue or maybe even an exchange and it will show the topology around that resource you are looking for.

snichme:
This is probably improvments for later but include DLX routing, shovels and federations would be nice

kickster97:
> This is probably improvments for later but include DLX routing, shovels and federations would be nice

yes, this is out of scope for the first iteration

PR #1316, opened 2025-09-30
Fix Build failure on Alpine Linux

### LINKED ISSUE #1292: Build failure on Alpine Linux: undefined reference to 'statfs64' when linking lavinmq v2.4.4
Describe the bug
The build process fails during the linking stage with an undefined reference to statfs64 when compiling lavinmq v2.4.4 on Alpine Linux. The compilation completes successfully but fails at the final linking step.
Describe your setup

LavinMQ version: v2.4.4 (tag dd10ea3521f7698ee24cc85f145b2abebdebc65a, commit 3d7f775253617129466fc792b0b7ea0c9d2c325c)
Operating System: Alpine Linux
Compiler: GCC 15.2.0 (x86_64-alpine-linux-musl)
Crystal version: Used in build process
Build environment: Docker container (based on build output showing /tmp/lavinmq-build path)

Build info cannot be retrieved as the build fails before completion.

How to reproduce
Clone the repository: git clone https://github.com/cloudamqp/lavinmq.git -b v2.4.4
Install dependencies: shards install --production
Run make: make all MAKEFLAGS=-j2
Observe the linking error during the final compilation step

Expected behavior
The build should complete successfully and produce the bin/lavinmq executable without linking errors.

Additional context
The error occurs specifically in /tmp/lavinmq-build/src/stdlib/filesystem.cr:88 when trying to link against statfs64. This suggests a compatibility issue with the musl libc used in Alpine Linux, where statfs64 may not be available or may have a different name/signature compared to glibc systems.

Error details:
```
/usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: _main.o3.o: in function `info':
/tmp/lavinmq-build/src/stdlib/filesystem.cr:88:(.text+0x33b257): undefined reference to `statfs64'
collect2: error: ld returned 1 exit status
```

kickster97:
This PR applies the patch that was submitted in issue #1292 

Fixes #1292 

Build on alpine Linux and see that it works.

### DIFF
--- a/src/stdlib/filesystem.cr
+++ b/src/stdlib/filesystem.cr
@@ -17,11 +17,7 @@ lib LibC
       spare : StaticArray(Long, 4)
     end
 
-    {% if flag?(:musl) %}
-      fun statfs(file : Char*, buf : Statfs*) : Int
-    {% else %}
-      fun statfs64(file : Char*, buf : Statfs*) : Int
-    {% end %}
+    fun statfs64(file : Char*, buf : Statfs*) : Int
   {% elsif flag?(:darwin) %}
     struct Statfs
       bsize : UInt32
@@ -89,7 +85,7 @@ struct FilesystemInfo
   getter total : UInt64
 end
 
-{% if (flag?(:linux) && !flag?(:musl)) || flag?(:darwin) %}
+{% if flag?(:linux) || flag?(:darwin) %}
   module Filesystem
     def self.info(path)
       statfs = uninitialized LibC::Statfs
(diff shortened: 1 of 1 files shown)


carlhoerberg:
The diff looks like the opposite of what the patch does? Is it applied already and this is the revert?

kickster97:
> The diff looks like the opposite of what the patch does? Is it applied already and this is the revert?

Oh yeah this can be closed, looks like its already fixed in the main branch.

PR #1322, opened 2025-10-01
add broad rescue in consumer deliver loop to avoid silent failures

kickster97:
We should not let the consumers deliver loop fail silently, this pr adds a broad rescue and logs the error. 

raise an unspecified error in the deliver loop and see that it gets rescued and logged.


carlhoerberg:
An unhandled exception wont the swallowed silently, but the default exception handler will report it.

PR #1327, opened 2025-10-02
dev ui improvements

baelter:
- **Use non-standard livereload port to avoid conflicts**
- **Dont make in parallel on macos, causes invalid memory access**
- **Make sure build is not in release mode, required for dev-ui to work**


viktorerlingsson (approved):
LGTM

Issue #1233, opened 2025-09-05
Shovels on streams not possible


viktorerlingsson:
**Describe the bug**
When creating a shovel on a stream queue you get an error:
`WARN lmq.shovel[name: "sh1", vhost: "/"] 406 - PRECONDITION_FAILED - Stream consumers must not be exclusive`

Either shovel consumers for streams should be set as non-exclusive, or shovels should not be allowed to be created on streams. 

**Describe your setup**
LavinMQ v2.5.0-rc.1

markusgust:
some more info:

There are some code to set `exclusive: false` for stream queues: https://github.com/cloudamqp/lavinmq/blob/b998f34df79aadc9a158dba515e9f6a07ed6538d/src/lavinmq/shovel/shovel.cr#L179
and there's a spec for shovelling streams but the spec sets the x-stream-offset arg. [link](https://github.com/cloudamqp/lavinmq/blob/b998f34df79aadc9a158dba515e9f6a07ed6538d/spec/shovel_spec.cr#L528). Problems seems that if I create a shovel with a stream source, the x-stream-offset isn't set. 

if I set 
```
@args["x-stream-offset"] = 1
```
or  just `exclusive = false` instead of 

manually before the above piece of code (or just exclusive = false instead of) and build/run locally, shovel will connect and start consuming from the stream. however it will only shovel message once prefetch is full. So if prefetch for the shovel is 5, and queue has 4 messages unacked, they won't be shovelled until one more message is published.

snichme:
I think we should introduce Offset as a parameter if you consume from a stream that way the user can decide where to start shoveling messages from and we can handle the exclusive parts. 

There are some js things that needed to be done since I want to add Stream as a source option and only then will you see the Offset input field (same as for Exchange key when you select exchange). 
The tricky part is the edit button, it defaults to type=queue unless it has exchange-key set. So by the same idea I can check if we have consumer-args->x-stream-offet and then assume it's a Stream.

PR #1331, opened 2025-10-03
Stream/separate view

snichme:

Introduces a separate view for streams  

This allows easier handling for features that are different between queues and streams without having to add `if` statements everywhere. 
Background can be found in issue #1223 

This PR separates the views which allows different handling for the feature "Get messages" since the semantics between them are different. 

There are also some smaller changes to the view which make sense for a stream: 

* Remove "Move messages" should instead be "Copy message", and also the shovel for streams doesn't work #1233 
* Ready and Unacked message count is removed

Create a stream and publish 10 messages into that queue. In the "Get message" box enter offset=5 and count=5 and you should get 5 messages.


snichme:
Yes, the diff is massive, but `stream.ecr` and `stream.js` is copies from `queue.ecr` and `queue.js`  so if you instead compare the files it's easier to see the diffs :)

views/stream.ecr
```diff
@@ -0,0 +1,285 @@
...
+        <h3>Copy messages</h3>
+        <label>
+          <span>Destination queue</span>
+          <input type="text" name="shovel-destination" list="queue-list">
+          <datalist id="queue-list">
+          </datalist>
+        </label>
+        <button type="submit" class="btn btn-outlined">Move messages</button>
+      </form>
+      -->
```
viktorerlingsson, line 229: this can probably just be removed?

viktorerlingsson (approved):
Looks good to me 👍 

Would be pretty nice to not have to duplicate all that code though, both for the js and the view. Maybe we could have a shared js (can't think of a good name though 🤷 ) for things that are shared between the two? But I guess that can be worked on in the future.

src/lavinmq/http/controller/queues.cr
```diff
@@ -13,6 +13,13 @@ module LavinMQ
         not_found(context, "Not Found") unless q
         q
       end
+
+      private def stream(context, vhost, name)
+        q = @amqp_server.vhosts[vhost].queues[name]?
+        not_found(context, "Not Found") unless q
+        not_found(context, "Not Found") unless q.is_a?(LavinMQ::AMQP::Stream)
+        q.as(LavinMQ::AMQP::Stream)
```
carlhoerberg, line 21:
```suggestion
        @amqp_server.vhosts[vhost].queues[name]?.try(&.as?(LavinMQ::AMQP::Stream)) || not_found(context, "Not Found")
```

src/lavinmq/http/controller/queues.cr
```diff
@@ -13,6 +13,13 @@ module LavinMQ
         not_found(context, "Not Found") unless q
         q
       end
+
+      private def stream(context, vhost, name)
```
carlhoerberg: sounds like a verb, "to stream", but this returns a stream queue 
```suggestion
      private def fetch_stream(context, vhost, name)
```

src/lavinmq/http/controller/queues.cr
```diff
@@ -215,6 +222,63 @@ module LavinMQ
             end
           end
         end
+
+        post "/api/queues/:vhost/:name/read" do |context, params|
```
carlhoerberg: what about `/stream` as it only works with stream queues?
```suggestion
        post "/api/queues/:vhost/:name/stream" do |context, params|
```

snichme:
> Would be pretty nice to not have to duplicate all that code though, both for the js and the view. Maybe we could have a shared js (can't think of a good name though 🤷 ) for things that are shared between the two? But I guess that can be worked on in the future.

That was my idea, do the split now and the next time a change require us to fix something in both places, then we fix the shared parts

kickster97:
@snichme what is status on this one? Would be nice to merge this :)

PR #917, opened 2025-01-21
WIP: Rewrite config with annotations and macros

viktorerlingsson:
Work in progress / suggestion

### WHAT is this pull request doing?
Rewrites `config.cr` with the use of annotations and macros to specify how each property can be set. Each property can have multiple annotations to be available to configure in multiple ways. 

```
    @[CliOpt("-D DIRECTORY", "--data-dir=DIRECTORY", "Data directory")]
    @[IniOpt(section: "main")]
    @[EnvOpt("LAVINMQ_DATADIR")]
    property data_dir : String = "/var/lib/lavinmq"
```

`@[CliOpt("-D DIRECTORY", "--data-dir=DIRECTORY", "Data directory")]`
for options to be configured by command line

`@[IniOpt(section: "main")]`
for options to be configured by config file

`@[EnvOpt("LAVINMQ_DATADIR")]`
for options to be configured by env variables

### HOW can this pull request be tested?
No specs yet, however, it should be possible to cover this pretty well with specs.

PR #1332, opened 2025-10-03
Keep track of depricated configs 

kickster97:
Adds a deprecation module to track deprecated configuration options. Registers deprecations at compile-time and fails compilation if a deprecated feature reaches its removal version. 

If you add a depricated config var, it will log a warning during startup. 

Currently tracks two deprecated config options:
  - guest_only_loopback (deprecated in 2.2.0, will be removed in 3.0.0) - replaced by default_user_only_loopback
  - default_password (deprecated in 2.2.0, will be removed in 3.0.0) - replaced by default_password_hash

 Compile and see logs, or try to build `version=3.0.0 crystal build src/lavinmq.cr `

### DIFF
--- a/src/lavinmq/deprecation.cr
+++ b/src/lavinmq/deprecation.cr
@@ -8,9 +8,8 @@ module LavinMQ
     @@features = [] of NamedTuple(name: String, deprecated_in: String, remove_in: String, message: String)
 
     macro deprecated(name, deprecated_in, remove_in, message)
-      {% current_version = `[ -n "$version" ] && echo "$version" || git describe --tags 2>/dev/null || shards version`.chomp.stringify.gsub(/^v/, "") %}
-      {% if current_version.split(".")[0].to_i >= remove_in.split(".")[0].to_i %}
-        {% raise "DEPRECATION ERROR: '#{name.id}' was marked for removal in #{remove_in.id} but is still present in version #{current_version.id}. #{message.id}" %}
+      {% if compare_versions(LavinMQ::VERSION, remove_in) >= 0 %}
+        {% raise "DEPRECATION ERROR: '#{name.id}' was marked for removal in #{remove_in.id} but is still present in version #{LavinMQ::VERSION.id}. #{message.id}" %}
       {% end %}
 
       LavinMQ::Deprecation.add({{ name }}, {{ deprecated_in }}, {{ remove_in }}, {{ message }})
@@ -19,10 +18,6 @@ module LavinMQ
     def self.add(name : String, deprecated_in : String, remove_in : String, message : String)
       @@features << {name: name, deprecated_in: deprecated_in, remove_in: remove_in, message: message}
     end
-
-    # def self.warn(feature_name : String, deprecated_in : String, remove_in : String, message : String)
-    #   Log.warn { "DEPRECATED: '#{feature_name}' was deprecated in #{deprecated_in} and will be removed in #{remove_in}. #{message}" }
-    # end
 
     def self.log_startup_warnings
       return if @@features.empty?
--- a/src/lavinmq/deprecation.cr
+++ b/src/lavinmq/deprecation.cr
@@ -1,0 +1,35 @@
+require "./version"
+require "log"
+
+module LavinMQ
+  module Deprecation
+    Log = ::Log.for("deprecation")
+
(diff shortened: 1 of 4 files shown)


src/lavinmq/version.cr
```diff
@@ -1,5 +1,7 @@
 module LavinMQ
-  VERSION = {{ `[ -n "$version" ] && echo "$version" || git describe --tags 2>/dev/null || shards version`.chomp.stringify.gsub(/^v/, "") }}
```
kickster97, line 2: not sure if this has other implications? but when VERSION is defined with {{ }}, it seems that it can't be used in the macro expansion

snichme:
I like the idea, but I don't like that we have to keep track of what values are deprecated separably  from where it's registers as an option. This means that I might remove the option but forget to remove the deprecation warning as they are not tied together. 

Is it possible to have a macro like this instead: 

```crystal 

available_until("3.0.0") do 
        p.on("--default-password=PASSWORD-HASH", "(Deprecated) Hashed password for default user (default: '+pHuxkR9fCyrrwXjOD4BP4XbzO3l8LJr8YkThMgJ0yVHFRE+' (guest))") do |v|
          STDERR.puts "WARNING: 'default-password' is deprecated, use '--default-password-hash' instead"
          @default_password = v
        end
 end
```

where `available_until` is a macro that removes the code within the `do` block and also logs a WARN about it when compiling when version is >= 3.0.0. 
If compiling with < 3.0.0 is just adds the code in the `do` block and nothing happens. Then you ae free to show the deprecation warn as you like and the code disapears after a certain version.

src/lavinmq/version.cr
spuun: It's the order of macro evaluation, I guess. You can probably wrap the `deprecation` macro in a finished hook.

spuun:
> Is it possible to have a macro like this instead:

Should be doable, yes.

src/lavinmq/deprecation.cr
```diff
@@ -0,0 +1,30 @@
...
+      @@features << {name: name, deprecated_in: deprecated_in, remove_in: remove_in, message: message}
+    end
+
+    def self.log_startup_warnings
+      return if @@features.empty?
+      Log.warn { "This version contains #{@@features.size} deprecated feature(s):" }
+      @@features.each do |f|
+        Log.warn { "  - #{f[:name]} (deprecated in #{f[:deprecated_in]}, will be removed in #{f[:remove_in]}): #{f[:message]}" }
+      end
+    end
```
spuun, line 28: I'd probably do something like this, I think it's look cleaner and less like a mix of macro and crystal code.
```suggestion
     FEATURES = [] of NamedTuple(name: String, deprecated_in: String, remove_in: String, message: String)

    macro finished
      {% verbatim do %}
      macro deprecated(name, deprecated_in, remove_in, message)
        {%
          if compare_versions(LavinMQ::VERSION, remove_in) >= 0
            raise "DEPRECATION ERROR: '#{name.id}' was marked for removal in #{remove_in.id} but is still present in version #{LavinMQ::VERSION.id}. #{message.id}"
          end
          FEATURES << {name: name, deprecated_in: deprecated_in, remove_in: remove_in, message: message}
        %}
      end
      {% end %}
    end

    def self.log_startup_warnings
      return if FEATURES.empty?
      Log.warn { "This version contains #{FEATURES.size} deprecated feature(s):" }
      FEATURES.each do |f|
        Log.warn { "  - #{f[:name]} (deprecated in #{f[:deprecated_in]}, will be removed in #{f[:remove_in]}): #{f[:message]}" }
      end
    end
  end

```

spuun:
> I like the idea, but I don't like that we have to keep track of what values are deprecated separably from where it's registers as an option. This means that I might remove the option but forget to remove the deprecation warning as they are not tied together.

Agree.

kickster97:
> I like the idea, but I don't like that we have to keep track of what values are deprecated separably from where it's registers as an option. This means that I might remove the option but forget to remove the deprecation warning as they are not tied together...

Great input, thanks :)

snichme:
Status of this PR? 
Something to reconsider based on this  https://github.com/cloudamqp/lavinmq/pull/917 maybe?

kickster97:
> #917

Yes, I think so. I have left it as a draft more for remembrance sake, i still think its a good functionality to have

PR #1333, opened 2025-10-03
sticky header on logs

Bianka2112:
Sticky actions over table for better ux viewing logs stream.

Checkout branch and scroll to see sticky toolbar and table header over rows.

*Note: I am compartmentalizing into smaller branches for reading simplicity of all changes applied to new-logs. Working backwards from larger commits, so apologies if it's unclear or mismatched changes in individual branches.* 😅

### DIFF
--- a/views/logs.ecr
+++ b/views/logs.ecr
@@ -10,9 +10,13 @@
     <main class="main-grid">
       <div id="breadcrumbs" class="cols-12">
           <h2 class="page-title"><span><%=pagename%></span></h2>
+          <div class="tiny-badge" id="pagename-label"></div></h2>
       </div>
-      <section id="livelog" class="card livelog">
-        <button id="download-logs" class="btn btn-green"><a download href="api/logs">Download logs</a></button>
+      <section class="card">
+       <div class="log-toolbar-container">
+         <a id="download-logs" class="btn btn-green toolbar-action" download href="api/logs">Download logs</a>
+        </div>
+        <div id="livelog" class="livelog">
         <div id="table-error"></div>
         <table id="table" class="table">
           <thead>
@@ -25,6 +29,7 @@
           </thead>
           <tbody id="livelog-body"></tbody>
         </table>
+        </div>
       </section>
     </main>
     <% render "partials/footer" %>
--- a/static/main.css
+++ b/static/main.css
@@ -2165,22 +2165,33 @@ pre.arguments .inactive-argument:before {
   white-space: normal;
 }
 
-.btn-primary a,
-.btn-primary a:visited {
-  color: var(--color-text-inverse);
+#livelog {
+  max-height: 65vh;
+  overflow: auto;
+  overflow-anchor: none;
+  scrollbar-gutter: stable both-edges;
+  scroll-behavior: auto;
 }
 
-#download-logs {
-  margin: 5px 0 16px 0;
+#livelog table {
+  width: 100%;
+  border-collapse: separate;
+  border-spacing: 0;
+  table-layout: fixed;
 }
 
-#download-logs a {
-  text-decoration: none;
+#livelog thead th {
+  position: sticky;
+  top: 0;
+  background: var(--color-bg-secondary);
+  border-bottom: 2px solid rgba(0,0,0,0.08);
+  z-index: 2;
 }
 
 #livelog table td,
 #livelog table th {
-  padding: 0;
+  padding: 4px 8px;
+  text-align: left;
 }
 
 #livelog table thead .livelog-timestamp {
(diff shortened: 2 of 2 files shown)


views/logs.ecr
```diff
@@ -10,9 +10,13 @@
     <main class="main-grid">
       <div id="breadcrumbs" class="cols-12">
           <h2 class="page-title"><span><%=pagename%></span></h2>
+          <div class="tiny-badge" id="pagename-label"></div></h2>
```
kickster97:
```suggestion
          <div class="tiny-badge" id="pagename-label"></div>
```

I think maybe there is an h2 tag dangling here?

kickster97 (changes requested):

Bianka2112:
Made the changes 👆, but am thinking that the buttons (newest, oldest) are also best placed within this branch, given that you'd need the sticky top to see them when first navigating to logs. If I put them in their own branch, you'd be left scrolling up all the logs to reach & test them.

kickster97:
> Made the changes 👆, but am thinking that the buttons (newest, oldest) are also best placed within this branch, given that you'd need the sticky top to see them when first navigating to logs. If I put them in their own branch, you'd be left scrolling up all the logs to reach & test them.

Yes I agree, sounds good to add them here! 
If you create a commit with the buttons in your `new_logs` branch, you can cherry-pick that commit into this branch :)

Bianka2112:
I see the main Spec test is failing. Do you have a preference of Merge or Rebase to update this branch with the main changes, where I think this was fixed. Or let it be?

kickster97:
> I see the main Spec test is failing. Do you have a preference of Merge or Rebase to update this branch with the main changes, where I think this was fixed. Or let it be?

The specs are a bit flaky :/ I ran it again and it passed so it's all good. But rebasing in new changes from main is good practice, so I think you can always do that now and then when working on a branch.

static/js/logs.js
```diff
@@ -41,6 +43,21 @@ function forbidden () {
...
+  shouldAutoScroll = false
+})
+
+btnToBottom?.addEventListener('click', () => {
+  btnToBottom.setAttribute('aria-pressed', 'true')
+  btnToTop.setAttribute('aria-pressed', 'false')
+  livelog.scrollTop = livelog.scrollHeight
+  shouldAutoScroll = true
+})
+
```
kickster97, line 69: It would be nice if the buttons had an initial value, and if i click away form the page my chouce is persisted to the next time i want to view the logs, maybe something like this? 

```suggestion
function setScrollMode(toBottom) {
  shouldAutoScroll = toBottom
  localStorage.setItem('logScrollMode', toBottom ? 'bottom' : 'top')
  if (btnToBottom && btnToTop) {
    btnToBottom.setAttribute('aria-pressed', String(toBottom))
    btnToTop.setAttribute('aria-pressed', String(!toBottom))
  }
}

// Initialize from saved preference, default to newest
const savedMode = localStorage.getItem('logScrollMode')
const initialMode = savedMode ? savedMode === 'bottom' : true
setScrollMode(initialMode)

btnToTop?.addEventListener('click', () => {
  setScrollMode(false)
  livelog.scrollTop = 0
})

btnToBottom?.addEventListener('click', () => {
  setScrollMode(true)
  livelog.scrollTop = livelog.scrollHeight
})
```

views/logs.ecr
```diff
@@ -11,8 +11,13 @@
       <div id="breadcrumbs" class="cols-12">
           <h2 class="page-title"><span><%=pagename%></span></h2>
       </div>
-      <section id="livelog" class="card livelog">
-        <button id="download-logs" class="btn btn-green"><a download href="api/logs">Download logs</a></button>
+      <section class="card">
+       <div class="log-toolbar-container">
+        <button id="to-bottom" class="btn-outlined btn-toggle">⬇ Newest </button>
```
kickster97: We should use the arrow up/down images we have under `img/arrow-up.svg` and `img/arrow-down.svg`.

Bianka2112: Gotcha. I will look for an arrow-up.svg to add since there is only an arrow-down available

static/js/logs.js
```diff
@@ -41,6 +45,28 @@ function forbidden () {
   tblError.style.display = 'block'
 }
 
+// Scrolling
+function setScrollMode (toBottom) {
+  shouldAutoScroll = toBottom
+  localStorage.setItem('logScrollMode', toBottom ? 'bottom' : 'top')
+  if (btnToBottom && btnToTop) {
```
spuun, line 52: We should never end up in a situation that we need to check this? We should make sure to do DOM things when `DOMContentLoaded`.

static/js/logs.js
```diff
@@ -41,6 +45,28 @@ function forbidden () {
...
+    btnToBottom.setAttribute('aria-pressed', String(toBottom))
+    btnToTop.setAttribute('aria-pressed', String(!toBottom))
+  }
+}
+// Initialize from saved preference, default to newest
+const savedMode = localStorage.getItem('logScrollMode')
+const initialMode = savedMode ? savedMode === 'bottom' : true
+setScrollMode(initialMode)
+btnToTop?.addEventListener('click', () => {
```
spuun, line 61: No need for optional chaining if we make sure to wait for `DOMContentLoaded`?

static/js/logs.js
```diff
@@ -41,6 +45,28 @@ function forbidden () {
...
+  setScrollMode(false)
+  livelog.scrollTop = 0
+})
+btnToBottom?.addEventListener('click', () => {
```
spuun, line 65: Ditto.

static/js/logs.js
kickster97: nice, i have a commit for it here: https://github.com/cloudamqp/lavinmq/pull/1381

PR #1335, opened 2025-10-06
Don't raise in queue constructors

spuun:
This PR refactors queue creations and removes all(?) exceptions thrown from constructors because of invalid arguments.

The main reason for this was to get rid of a BoolChannel memory leak, but instead of doing a workaround I took the long path of refactoring more.

I'm not super happy with the code, but I think it's better than before.

It tries to move all argument validation into each queue type, instead of having some in the queue and some in the queue factory. 

Specs cover it, I think

### DIFF
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -21,9 +21,11 @@ describe LavinMQ::AMQP::Queue do
   it "Should dead letter expired messages" do
     with_amqp_server do |s|
       with_channel(s) do |ch|
+        puts "BEFORE QUEUE"
         q = ch.queue("ttl", args: AMQP::Client::Arguments.new(
           {"x-message-ttl" => 1, "x-dead-letter-exchange" => "", "x-dead-letter-routing-key" => "dlq"}
         ))
+        puts "AFTER QUEUE"
         dlq = ch.queue("dlq")
         x = ch.default_exchange
         x.publish_confirm("ttl", q.name)
--- a/src/lavinmq/amqp/argument_validator.cr
+++ b/src/lavinmq/amqp/argument_validator.cr
@@ -15,13 +15,11 @@ module LavinMQ
         raise LavinMQ::Error::PreconditionFailed.new(msg)
       end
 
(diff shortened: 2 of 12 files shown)


spuun:
Not fully done.. still work to do in Stream.

spec/queue_spec.cr
```diff
@@ -21,9 +21,11 @@ describe LavinMQ::AMQP::Queue do
   it "Should dead letter expired messages" do
     with_amqp_server do |s|
       with_channel(s) do |ch|
+        puts "BEFORE QUEUE"
```
snichme: ☝️

src/lavinmq/amqp/argument_validator.cr
```diff
@@ -0,0 +1,88 @@
...
+          value.as?(String) || raise_invalid!("#{header} header not a string")
+          nil
+        end
+      end
+
+      struct BoolValidator
+        include ArgumentValidator
+
+        def validate!(header : String, value : AMQP::Field) : Nil
+          return if value.nil?
```
snichme, line 64: can value be nil here? Based on the type it's `AMQP::Field` not `AMQP::Field?`

spuun: Yeah, `AMQP::Field` is a union of types including `Nil`.

spuun:
> I like the change, but I feel that I need to run some of the code, it's a big change to grasp just by looking at the diff.

Yeah, that's why I've added a lot of specs. There are probably more cases that should be added.

src/lavinmq/amqp/argument_validator.cr
snichme: Interesting...

spuun: But you got me wonder, so I had to check the code 🙃

snichme:
Looked at this change again. This makes the QueueFactory almost unnecessary, it's only used from Vhost and is basically a if/else that can maybe be made into a case/when for simplicity. 
WDYT about moving this to vhost.cr or the Queue super class?

PR #1337, opened 2025-10-07
Bugfix/1233

### LINKED ISSUE #1233: Shovels on streams not possible

snichme:

The bug was that if consumer arguments had `x-stream-offset` set it would disable the `exclusive` consumer flag and the API allowed to set this but there was no input to set `x-stream-offset` from the UI. 

This PR introduces a stream option in the source type  and when selected an additional input for offset is visible and that value is then added to consumer arguments as x-stream-offset. 
It also handles edit of a shovel with a stream source, here the type is based on if the `x-stream-offset` is in consumer args or not. 

Fixes #1233 

Manually

1.  Create two shovels one with queue as source and one with stream as source
2. Try edit both and see that the form is updated with correct values
3. Also not that both shovels are running since they are both now configured properly

### DIFF
--- a/static/js/shovels.js
+++ b/static/js/shovels.js
@@ -54,7 +54,7 @@ class ShovelsDataSource extends UrlDataSource {
 }
 const dataSource = new ShovelsDataSource(url, statusUrl)
 const tableOptions = { keyColumns: ['vhost', 'name'], columnSelector: true, dataSource }
-Table.renderTable('table', tableOptions, (tr, item, all) => {
+Table.renderTable('table', tableOptions, (tr, item, _all) => {
   Table.renderCell(tr, 0, item.vhost)
   Table.renderCell(tr, 1, item.name)
   if (Array.isArray(item.value['src-uri'])) {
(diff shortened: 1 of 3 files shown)


static/js/shovels.js
```diff
@@ -174,11 +182,28 @@ document.querySelector('#createShovel').addEventListener('submit', function (evt
       'ack-mode': data.get('ack-mode')
     }
   }
-  if (data.get('src-type') === 'queue') {
-    body.value['src-queue'] = data.get('src-endpoint')
-  } else {
-    body.value['src-exchange'] = data.get('src-endpoint')
-    body.value['src-exchange-key'] = data.get('src-exchange-key')
+  const srcType = data.get('src-type')
+  const offset = data.get('src-offset')
```
viktorerlingsson, line 186: maybe this should be moved inside `case 'stream':` ?  since it's not used for anything else

snichme: I thought so too but not the linter according to this rule: https://eslint.org/docs/latest/rules/no-case-declarations  :shrug:

viktorerlingsson: 😮 Thats a little surprising! But ok 👍

PR #1338, opened 2025-10-07
Bugfix: api endpoint should accept empty/no body

### LINKED ISSUE #1303: `Request body required` when trying to create a queue using HTTP API
```shell
$ curl -i -u guest:guest -X PUT http://localhost:15673/api/queues/%2F/foo
HTTP/1.1 400 Bad Request
Connection: keep-alive
Strict-Transport-Security: max-age=31536000
Content-Type: application/json
Transfer-Encoding: chunked

{"error":"bad_request","reason":"Request body required"}
```

Looks to me like we have defaults for everything so it doesn't make sense to complain like that 
https://github.com/cloudamqp/lavinmq/blob/b36840033c5e0d3dea61bec1d5fc9f3da9f3bcc7/src/lavinmq/http/controller/queues.cr#L61-L64

It also works without a body when using another broker:

```shell
$ curl -i -u guest:guest -X PUT http://localhost:15672/api/queues/%2F/foo
HTTP/1.1 201 Created
allow: HEAD, GET, PUT, DELETE, OPTIONS
cache-control: public, max-age=0, must-revalidate
content-length: 0
content-security-policy: script-src 'self' 'unsafe-eval' 'unsafe-inline'; object-src 'self'
date: Tue, 23 Sep 2025 21:56:16 GMT
server: Cowboy
vary: accept, accept-encoding, origin
```

spuun:
Fixes #1303

It turned it LavinMQ didn't support request without content-length header for PUT and POST. Some clients may omit setting content-length to 0 when there is no body.

Run spec

### DIFF
--- a/src/lavinmq/http/controller.cr
+++ b/src/lavinmq/http/controller.cr
@@ -142,14 +142,13 @@ module LavinMQ
       end
 
       private def parse_body(context) : JSON::Any
+        if context.request.content_length == 0
+          return JSON::Any.new(Hash(String, JSON::Any).new)
+        end
         if body = context.request.body
           ct = context.request.headers["Content-Type"]?
           if ct.nil? || ct.empty? || ct == "application/json"
-            json = if context.request.content_length == 0
-                     JSON::Any.new(Hash(String, JSON::Any).new)
-                   else
-                     JSON.parse(body)
-                   end
+            json = JSON.parse(body)
             if json.as_h?
               json
             else


src/lavinmq/http/controller.cr
```diff
@@ -142,14 +142,13 @@ module LavinMQ
       end
 
       private def parse_body(context) : JSON::Any
```
snichme, line 144: Or should this return `JSON::Any?`  so return nil if no body or content_length == 0?

spuun: I had that in mind, but that would require a lot of refactoring.

spuun: ["Models"](https://github.com/cloudamqp/lavinmq/compare/main...lab/router-json-models) would improve this.

PR #1339, opened 2025-10-07
Specs: fail on closed queues

### LINKED ISSUE #1315: Catch closed queues in specs
Because we rescue many exceptions in a queue/msg store and close the queue, it is possible that specs can silently fail unless we explicitly check that a queue is not closed. (In some cases that can also make it harder to find an actual error in a spec, since the spec itself will only catch any client errors that happen _after_ the queue has been closed, not the error that closed the queue.)

I propose that we either add something to `with_amqp_server` that verifies that queues aren't closed during a spec run, or that we add a flag or similar to queue that re-raises errors when running specs.

spuun:
Fixes #1315

Well, specs...

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -122,10 +122,10 @@ def with_amqp_server(tls = false, replicator = LavinMQ::Clustering::NoopServer.n
     # A closed queue indicates that something failed that shouldn't. However, the error may be
     # silent and not causing a spec to fail. It's not even sure that the failing queue is related
     # to what the spec tests, but by failing on closed queues we may find bugs.
-    # If a queue is suppsoed to be closed, it should be deleted in the end of the spec.
+    # If a queue is supposed to be closed, it should be deleted in the end of the spec.
 
     # We yield a few times (if necessary) to make sure things has really settled, e.g.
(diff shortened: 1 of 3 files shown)


viktorerlingsson (approved):

kickster97 (approved):
nice

spec/spec_helper.cr
```diff
@@ -118,14 +119,32 @@ def with_amqp_server(tls = false, replicator = LavinMQ::Clustering::NoopServer.n
...
     yield s
   ensure
+    # A closed queue indicates that something failed that shouldn't. However, the error may be
+    # silent and not causing a spec to fail. It's not even sure that the failing queue is related
+    # to what the spec tests, but by failing on closed queues we may find bugs.
+    # If a queue is supposed to be closed, it should be deleted in the end of the spec.
+
+    # We yield a few times (if necessary) to make sure things has really settled, e.g.
+    # everything has been cleaned up after a `with_channel` inside the `with_amqp_server`.
+    closed_queues = 3.times do
```
kickster97, line 129: I guess this will add some time to each spec using `with_amqp_server`?(which is most of them)

spuun: Yes, but probably not measurable. It's just a few fiber context switching (and only if any closed queue is found).

PR #1343, opened 2025-10-09
Rename message segment meta files to `meta.SEGMENT`

### LINKED ISSUE #1340: Rename msg meta files from `msg.SEGMENT.meta` to `meta.SEGMENT`
In 2.5 meta data files for segments will be introduced. Currently they are named `msg.SEGMENT.meta`. One problem with this is that it prevents downgrading from 2.5.x to 2.4.x, because `2.4.x` will treat any file starting with `msg.` as a message segment and therefore crash if any meta file exists.

By renaming the meta files to `meta.SEGMENT` we prevent this issue, and also follows the "naming pattern" for segments and ack files.

Maybe it's worth releasing a new 2.4 with a patch for this too?

spuun:
Fixes #1340

Specs

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -236,10 +236,10 @@ module LavinMQ
     end
 
     private def delete_file(file : MFile, including_meta = false)
-      File.delete?(metafile_from_mfile(file)) if including_meta
+      File.delete?(meta_file_name(file)) if including_meta
       file.delete(raise_on_missing: false)
       if replicator = @replicator
-        replicator.delete_file(metafile_from_mfile(file), WaitGroup.new) if including_meta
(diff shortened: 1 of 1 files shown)


src/lavinmq/message_store.cr
```diff
@@ -544,6 +546,24 @@ module LavinMQ
       end
     end
 
+    private def metafile_from_mfile(mfile : MFile) : String
+      metafile_from_path(mfile.path)
+    end
+
+    private def metafile_from_path(path : String) : String
```
viktorerlingsson: I think the naming here should be more clear that you're getting a meta file path, not the meta file itself

spuun: True!

PR #1345, opened 2025-10-09
Restart closed queues

### LINKED ISSUE #1290: Make it possible to restart a closed queue
Sometimes a queue is closed for any number of reasons, but there is no way of restarting it unless you restart LavinMQ.

Add an endpoint to the API and a button to the UI that makes it possible to restart a closed queue.

viktorerlingsson:
This adds functionality to restart closed queues/streams. 

Adds: 
- Restart method for queues/streams
- API endpoint
- UI button
- lavinmqctl cmd
- specs

Fixes #1290 

Run the new specs

### DIFF
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -248,10 +248,9 @@ describe LavinMQ::AMQP::Queue do
 
     it "should restart after corrupt data closes the queue" do
       with_amqp_server do |s|
-        vhost = s.vhosts.create("restart_vhost")
-        with_channel(s, vhost: vhost.name) do |ch|
+        with_channel(s) do |ch|
           q = ch.queue(q_name, durable: true)
-          queue = vhost.queues[q_name].as(LavinMQ::AMQP::DurableQueue)
+          queue = s.vhosts["/"].queues[q_name].as(LavinMQ::AMQP::DurableQueue)
           q.publish_confirm "test message"
           queue.message_count.should eq 1
 


carlhoerberg:
Are the claude code review comments valid?

viktorerlingsson:
> Are the claude code review comments valid?

Some of them where, some not! I've addressed the ones I think was relevant

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -190,13 +190,39 @@ module LavinMQ::AMQP
         @paused.set(true)
       end
       @msg_store = init_msg_store(@data_dir)
+      @empty = @msg_store.empty
+      start
+    end
+
+    private def start
       if @msg_store.closed
```
snichme, line 192: Why do we check if the msg_store is closed on start? From the code in MessageStore it looks like it's only set to closed = true on `close` and `delete`  so can this really happen here?

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -190,13 +190,39 @@ module LavinMQ::AMQP
...
+      reset_queue_state
+      @msg_store = init_msg_store(@data_dir)
       @empty = @msg_store.empty
-      handle_arguments
-      spawn queue_expire_loop, name: "Queue#queue_expire_loop #{@vhost.name}/#{@name}" if @expires
-      spawn message_expire_loop, name: "Queue#message_expire_loop #{@vhost.name}/#{@name}"
+      start
+    end
+
+    private def reset_queue_state
```
snichme, line 217: I feel that we already have a lot of state to keep track of, while this doesn't add more states it adds more places where we need to keep track of all the states for a queue. 
Not sure if there is a better solution for this at the moment. I just wanted to flag that the more state changes we add to more difficult it will be to keep everything in correct state.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -190,13 +190,39 @@ module LavinMQ::AMQP
...
       if @msg_store.closed
         close
+      else
+        handle_arguments
+        spawn queue_expire_loop, name: "Queue#queue_expire_loop #{@vhost.name}/#{@name}" if @expires
+        spawn message_expire_loop, name: "Queue#message_expire_loop #{@vhost.name}/#{@name}"
       end
+    end
+
+    def restart!
```
snichme: Do we know how the state ended up in closed state? Is there any thing to be aware of before trying to restart a queue or if this fails to restart a queue, how will the user know?

snichme (commented):
I do feel that this feature is valuable to have I'm just concerned if doing adds more complexity to the queue code that will be hard to handle in the future.

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: yes, `MessageStore#close` can be called if there are errors in `load_segments_from_disk` or `produce_metadata` during start, and if that happens there's no point in trying to start the queue

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: We can have an idea on why it's closed based on the log. It should most of the times be because of some kind of exception. We could maybe show that error somewhere?

I think it should be pretty safe to try to restart, but in many cases the queue probably won't start unless one has fixed the reason for why the queue was closed in the first place.

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: Yeah, there is a lot to keep track of. I guess another option would be to delete the queue object in VHost and recreate it, but I think that just moves the complexity to VHost instead.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -190,13 +190,39 @@ module LavinMQ::AMQP
...
+      start
+    end
+
+    private def reset_queue_state
+      @closed = false
+      @state = QueueState::Running
+      # Recreate channels that were closed
+      @queue_expiration_ttl_change = ::Channel(Nil).new
+      @message_ttl_change = ::Channel(Nil).new
+      @paused = BoolChannel.new(false)
```
snichme, line 223: Should we check if the queue was paused and if so reset into pasued state?

src/lavinmq/http/controller/queues.cr
```diff
@@ -114,6 +114,15 @@ module LavinMQ
           end
         end
 
+        put "/api/queues/:vhost/:name/restart" do |context, params|
+          with_vhost(context, params) do |vhost|
+            refuse_unless_management(context, user(context), vhost)
+            q = find_queue(context, params, vhost)
+            q.restart!
```
snichme: `restart!` is a no-up, just returns when the queue is running ( `unless @closed`) 

Maybe `recover` would be a better name? 

And also, have a response that say that the queue was restarted/recovered so you dont issue a PUT to this endpoint and you get nothing back but also nothing happens with the queue.

static/js/queue.js
```diff
@@ -341,6 +349,18 @@ resumeQueueForm.addEventListener('submit', function (evt) {
   }
 })
 
+restartQueueForm.addEventListener('submit', function (evt) {
+  evt.preventDefault()
+  const url = HTTP.url`api/queues/${vhost}/${queue}/restart`
+  if (window.confirm('Are you sure? This will restart the queue.')) {
+    HTTP.request('PUT', url)
+      .then(() => {
+        DOM.toast('Queue restarted!')
```
snichme, line 359: If we get a response text, also show here is the queue was or wasn't restarted

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: yeah, we can probably just move the paused state handling to start

src/lavinmq/http/controller/queues.cr
viktorerlingsson: Yeah , I took inspiration from `pause!` / `resume!`. But we can make it a little smarter so the user can know what happens.

snichme: I write the initial implementation on pause and resume, so even I should learn from my own comments. 
But that is for another PR :)

src/lavinmq/amqp/queue/queue.cr
snichme:
```suggestion
    def restart! : Bool
```

So we make sure we always get a bool response. If someone changes `start` to not return true this will always return nil which breaks the http controller logic.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -386,7 +387,7 @@ module LavinMQ::AMQP
 
     def pause!
       return unless @state.running?
-      @state = QueueState::Paused
+        @state = QueueState::Paused
```
snichme:
```suggestion
      @state = QueueState::Paused
```

snichme (approved):

PR #1347, opened 2025-10-10
filter for logs

Bianka2112:
- Add live regex filter and new render logic for logs view.
- Waits to render 1000 logs, lists up to 2000 total, with small batches on rerender if fire-hosed. 
- Adds loading status

- Checkout branch, try filter box at the top. 
- I tested locally with a fake logs burst, to check for smooth batching.  
- Happy to get feedback/recommendations on the UX + rendering, since I dont get much stream in my own dev env to see how it feels "live".

### DIFF
--- a/static/js/logs.js
+++ b/static/js/logs.js
@@ -2,6 +2,8 @@ const evtSource = new window.EventSource('api/livelog')
 const livelog = document.getElementById('livelog')
 const tbody = document.getElementById('livelog-body')
 const filterInput = document.getElementById('log-filter')
+const table = document.getElementById('table')
+const loading = document.getElementById('log-loading')
 
 const allLogs = []
 const pendingLogs = []
@@ -26,7 +28,8 @@ evtSource.onmessage = (event) => {
   const log = { timestamp, severity, source, message }
   pendingLogs.push(log)
   if (booting) {
-    tbody.style.visibility = 'hidden'
+    table.style.visibility = 'hidden'
+    loading.hidden = false
   }
   scheduleFlush()
 }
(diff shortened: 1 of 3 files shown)


static/js/logs.js
```diff
@@ -41,6 +50,149 @@ function forbidden () {
   tblError.style.display = 'block'
 }
 
+// Build Table
+const buildRow = (log) => {
```
carlhoerberg: prefer function declaration to assignment

```suggestion
function buildRow(log) {
```

static/js/logs.js
```diff
@@ -1,28 +1,37 @@
...
-  tdSrc.textContent = source
-  const preMsg = document.createElement('pre')
-  preMsg.textContent = message
-  const tdMsg = document.createElement('td')
-  tdMsg.appendChild(preMsg)
+const MAX_ROWS = 2000
+const INITIAL_ROWS = 1000
+const ROWS_PER_FLUSH = 200
+const BOOT_MIN_LOGS = INITIAL_ROWS
+const BOOT_MAX = 10000
```
carlhoerberg, line 16: We wait 10s or for 1000 lines to arrive before doing first rendering? I think that might be a bit too long.

carlhoerberg (commented):
Haven't tested but the code looks generally sound, except we prefer the `function` declarations instead of assignments.

static/js/logs.js
Bianka2112: The idea was a high enough threshold that would avoid too much flutter if theres a lot of backlog, but can absolutely bring it down. Would 500ms feel better?
In my testing, 500ms still allows the loading status to appear and renders batches without too much repainting. WDYT?

PR #1353, opened 2025-10-13
RingBuffer for stats

jage:

Speed up internal stats collection, a ring buffer should be faster for our use case.

While not the most common operation, for a broker with a lot of objects (queues, exchanges etc) we want to minimize any CPU time spent on just keeping track of the stats.

Improvements to the `to_a` method that's used in the UI/API, i.e. any time we refresh the web page with new stats (proposed to be once per second in #1195) should help a bit, even though the full JSON serialization is probably much more expensive. Getting around 9% improvements in the simulated http_benchmark.cr

See benchmark/ directory for all micro benchmarks (should be cleaned up before we merge this). Also kept the old Deque based code until we're ready to say this is good.

```
RingBuffer (steady-state) 355.76M (  2.81ns) (± 5.70%)  0.0B/op        fastest
     Deque (steady-state) 177.83M (  5.62ns) (± 2.41%)  0.0B/op   2.00× slower

RingBuffer (fill from empty) 285.64k (  3.50µs) (± 0.76%)  12.1kB/op        fastest
     Deque (fill from empty) 202.75k (  4.93µs) (± 1.55%)  12.1kB/op   1.41× slower

RingBuffer (to_a) 872.70k (  1.15µs) (± 0.84%)  12.1kB/op        fastest
     Deque (to_a) 146.77k (  6.81µs) (± 0.94%)  12.1kB/op   5.95× slower
```

---

Why RingBuffer is faster:

Bitwise Modulo Operations

- RingBuffer uses `& @mask` for wrap-around (single CPU instruction)
- Deque uses conditional branches (`if index >= @capacity`)
- Bitwise `AND` avoids branch misprediction penalties

No Memory Shifting

- RingBuffer: Simply overwrites oldest value when full
- Deque: Calls shift to remove first element, causing memory movement

Simpler Push Logic

- RingBuffer: Direct write to calculated position
- Deque: Check size → shift if needed → then push (3 operations vs 1)

CPU Cache Optimization

- RingBuffer: Predictable circular access pattern
- Deque: Shift operations invalidate cache lines

No Dynamic Memory Management

- RingBuffer: Fixed allocation, never grows
- Deque: Resize checks and potential reallocation overhead

Specs and looking at the stats manually, but we might want to improve specs.

### DIFF
--- a/src/lavinmq/ring_buffer.cr
+++ b/src/lavinmq/ring_buffer.cr
@@ -1,0 +1,76 @@
+require "json"
+
+module LavinMQ
+  class RingBuffer(T)
+    include Enumerable(T)
+
+    @buffer : Pointer(T)
+    @head = 0
+    @size = 0
+    @mask : Int32
+    @capacity : Int32
+
+    def initialize(capacity : Int32)
+      raise ArgumentError.new("Capacity must be at least 2") if capacity < 2
+      # Round up to next power of 2 for fast modulo via bitwise AND
+      @capacity = Math.pw2ceil(capacity)
+      @mask = @capacity - 1
+      @buffer = {% if T < Number %}
+                  GC.malloc_atomic(@capacity * sizeof(T)).as(Pointer(T))
+                {% else %}
+                  Pointer(T).malloc(@capacity)
+                {% end %}
+    end
+
+    def push(value : T) : Nil
+      tail = (@head + @size) & @mask
+      @buffer[tail] = value
+
+      if @size < @capacity
+        @size += 1
+      else
+        @head = (@head + 1) & @mask
+      end
+    end
+
+    def size : Int32
+      @size
+    end
+
+    def [](index : Int32) : T
+      raise IndexError.new if index >= @size || index < 0
+      actual_index = (@head + index) & @mask
+      @buffer[actual_index]
+    end
+
+    def each(& : T ->) : Nil
+      @size.times do |i|
+        yield @buffer[(@head + i) & @mask]
+      end
+    end
+
+    def to_json(json : JSON::Builder) : Nil
+      json.array do
+        each(&.to_json(json))
+      end
+    end
+
+    def to_a : Array(T)
+      return Array(T).new if @size == 0
+
+      result = Array(T).new(@size)
+      tail = (@head + @size) & @mask
+
+      if @size < @capacity || tail > @head
+        # Not full yet, or full but not wrapped: copy single segment
+        result.concat((@buffer + @head).to_slice(@size))
+      else
+        # Full and wrapped: copy two segments
+        first_chunk_size = @capacity - @head
+        result.concat((@buffer + @head).to_slice(first_chunk_size))
+        result.concat(@buffer.to_slice(tail))
+      end
+      result
+    end
+  end
+end
--- a/src/lavinmq/stats.cr
+++ b/src/lavinmq/stats.cr
@@ -1,3 +1,6 @@
+require "./config"
+require "./ring_buffer"
+
 module LavinMQ
   module Stats
     macro rate_stats(stats_keys, log_keys = %w[])
@@ -5,14 +8,16 @@ module LavinMQ
         @{{ name.id }}_count = Atomic(UInt64).new(0_u64)
         @{{ name.id }}_count_prev = 0_u64
         @{{ name.id }}_rate = 0_f64
-        @{{ name.id }}_log = Deque(Float64).new(Config.instance.stats_log_size)
+        @{{ name.id }}_log = RingBuffer(Float64).new(Config.instance.stats_log_size)
         def {{ name.id }}_count
           @{{ name.id }}_count.get(:relaxed)
         end
       {% end %}
       {% for name in log_keys %}
-        @{{ name.id }}_log = Deque(UInt32).new(Config.instance.stats_log_size)
-        getter {{ name.id }}_log
+        @{{ name.id }}_log = RingBuffer(UInt32).new(Config.instance.stats_log_size)
+        def {{ name.id }}_log
+          @{{ name.id }}_log
+        end
       {% end %}
 
       def stats_details
@@ -39,20 +44,13 @@ module LavinMQ
 
       def update_rates : Nil
         interval = Config.instance.stats_interval // 1000
-        log_size = Config.instance.stats_log_size
         {% for name in stats_keys %}
-          until @{{ name.id }}_log.size < log_size
-            @{{ name.id }}_log.shift
-          end
           {{ name.id }}_count = @{{ name.id }}_count.get(:relaxed)
           @{{ name.id }}_rate = (({{ name.id }}_count - @{{ name.id }}_count_prev) / interval).round(1)
           @{{ name.id }}_log.push @{{ name.id }}_rate
           @{{ name.id }}_count_prev = {{ name.id }}_count
         {% end %}
         {% for name in log_keys %}
-          until @{{ name.id }}_log.size < log_size
-            @{{ name.id }}_log.shift
-          end
           @{{ name.id }}_log.push {{ name.id }}
         {% end %}
       end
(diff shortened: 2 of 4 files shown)


src/lavinmq/ring_buffer.cr
```diff
@@ -0,0 +1,64 @@
...
+    @buffer : Pointer(T)
+    @head = 0
+    @size = 0
+    @mask : Int32
+    @capacity : Int32
+
+    def initialize(capacity : Int32)
+      raise ArgumentError.new("Capacity must be at least 2") if capacity < 2
+      # Round up to next power of 2 for fast modulo via bitwise AND
+      @capacity = next_power_of_two(capacity)
```
carlhoerberg:
```suggestion
      @capacity = Math.pw2ceil(capacity)
```

src/lavinmq/ring_buffer.cr
```diff
@@ -0,0 +1,64 @@
...
+        STDERR.puts "WARNING: RingBuffer capacity #{capacity} rounded up to #{@capacity} (must be power of 2)"
+      end
+      @mask = @capacity - 1
+      @buffer = Pointer(T).malloc(@capacity)
+    end
+
+    private def next_power_of_two(n : Int32) : Int32
+      2 ** Math.log2(n).ceil.to_i
+    end
+
```
carlhoerberg: replace with `Math.pw2ceil`
```suggestion
```

src/lavinmq/ring_buffer.cr
```diff
@@ -0,0 +1,64 @@
...
+    def initialize(capacity : Int32)
+      raise ArgumentError.new("Capacity must be at least 2") if capacity < 2
+      # Round up to next power of 2 for fast modulo via bitwise AND
+      @capacity = next_power_of_two(capacity)
+      if capacity != @capacity
+        STDERR.puts "WARNING: RingBuffer capacity #{capacity} rounded up to #{@capacity} (must be power of 2)"
+      end
+      @mask = @capacity - 1
+      @buffer = Pointer(T).malloc(@capacity)
```
carlhoerberg: Could maybe use `GC.malloc_atomic` instead (if `T` is limited to `Value`), for better GC performance (the memory isn't cleared and doesn't have to be scanned for pointers), but i guess it's a theoretical improvement.

carlhoerberg (commented):
Nice! Any reason the old `Stat` is kept?

jage:
> Nice! Any reason the old `Stat` is kept?

Nah just to nice to have the "original" left during development, but I'll add some cleanup commits and then this PR should be squash merged.

src/lavinmq/ring_buffer.cr
jage: Ah, yeah that sounds interesting here. 

Initial micro benchmarks show it's ~20% faster when we allocate a lot of buffers, so should help in a high churn environment. For channel churn it might be a measurable improvements, but yeah for connection/exchanges/queues I guess it's more theoretical.

If there's a huge amount of objects I guess the GC scan speedup is nice as well (haven't benchmarked).

Not sure of there's any downsides? Memory won't be cleared but that's implicitly already assumed by the design, so should be fine.

jage: 3d32b60

jage:
For version two of this we can re-architecture to use a "ring buffer group", e.g. we let a group of several ring buffers share the same `head`, `size`, `mask` etc, it would just be `@buffer` that's uniq to each stat. Since we update all values at the same time, this should save a few bytes and maybe be faster if we can get the CPU to do more at the same time.

This is relevant for "large" brokers, e.g. 100k exchanges would have 400k ring buffers (`publish_in`, `publish_out`, `unroutable`, `dedup`).

```
  RingBufferGroup (capacity=8) with 3 buffers:

  ┌──────────────────────────────────────────────────┐
  │ RingBufferGroup                                  │
  │   @head = 2                                      │
  │   @size = 6                                      │
  │   @mask = 7                                      │
  │   @capacity = 8                                  │
  └──────────────────────────────────────────────────┘
           │
           │ Shared State
           ├───────────┬───────────┬───────────┐
           ↓           ↓           ↓           ↓

  RingBufferView(Float64) - rates:
  ┌──────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┐
  │ 5.2  │ 6.1  │ 1.5  │ 2.3  │ 3.7  │ 4.9  │ 5.2  │ 6.1  │
  └──────┴──────┴──────┴──────┴──────┴──────┴──────┴──────┘
    0      1      2      3      4      5      6      7
                  ↑                                  ↑
                head=2                           tail=0 ((2+6)&7=0)

  RingBufferView(UInt32) - counts:
  ┌──────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┐
  │ 400  │ 500  │ 100  │ 200  │ 300  │ 400  │ 500  │ 600  │
  └──────┴──────┴──────┴──────┴──────┴──────┴──────┴──────┘
    0      1      2      3      4      5      6      7
                  ↑                                  ↑
                head=2                           tail=0

  RingBufferView(Int64) - timestamps:
  ┌──────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┐
  │ 1008 │ 1009 │ 1003 │ 1004 │ 1005 │ 1006 │ 1007 │ 1008 │
  └──────┴──────┴──────┴──────┴──────┴──────┴──────┴──────┘
    0      1      2      3      4      5      6      7
                  ↑                                  ↑
                head=2                           tail=0
```

src/lavinmq/stats.cr
```diff
@@ -7,14 +8,16 @@ module LavinMQ
...
           @{{name.id}}_count.get(:relaxed)
         end
       {% end %}
       {% for name in log_keys %}
-        @{{name.id}}_log = Deque(UInt32).new(Config.instance.stats_log_size)
-        getter {{name.id}}_log
+        @{{name.id}}_log = RingBuffer(UInt32).new(Config.instance.stats_log_size)
+        def {{name.id}}_log
+          @{{name.id}}_log.to_a
+        end
```
jage, line 20: Not 100% sure about this, if we really need to do this, should be handled at compile time if some code don't handle Ringbuffer.

src/lavinmq/stats.cr
```diff
@@ -23,7 +26,7 @@ module LavinMQ
             {{name.id}}: {{name.id}}_count,
             {{name.id}}_details: {
               rate: @{{name.id}}_rate,
-              log: @{{name.id}}_log
+              log: @{{name.id}}_log.to_a
```
jage: See above.

jage:
Need to take another round on this, missed how `add_logs!` works.

src/lavinmq/stats.cr
jage: Ah, because of `add_logs!` - should rewrite that.

spuun: You just need to add `#next` to `RingBuffer` to make it an `Iterator(T)` I think. Probably nice.

spuun:
> For version two of this we can re-architecture to use a "ring buffer group", e.g. we let a group of several ring buffers share the same `head`, `size`, `mask` etc, it would just be `@buffer` that's uniq to each stat. Since we update all values at the same time, this should save a few bytes and maybe be faster if we can get the CPU to do more at the same time.

And maybe it's worth making it a lock-free MT safe structure.

carlhoerberg:
Or, if we really want to support massive amount of entities, i think we should do away with the stat logs completely, only have counters and let the clients poll counters and calulate their own rates, a la. prometheus. 

The UI could poll the counters in the background and store them in the IndexedDB. The difficulty is of course how do with history of one queue among 100k. But maybe we can start naively and collect all counters from all entities and see how it performs..

jage:
This is pretty CPU efficient but not good enough memory management. We're wasting more memory with with the power of 2 setup, 128 slots -> 1024 bytes which Boehm slot allocates 1344 byte for. Got something better that will replace this.

PR #1358, opened 2025-10-15
Fix: Base lavinmq_global_messages_delivered_total on deliver_get to get all types of deliveries

jage:
Regardless how a messages is delivered, subscribed consumer or polling consumer, we should count it towards being delivered.

Covered by specs.

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -52,9 +52,9 @@ describe LavinMQ::HTTP::PrometheusController do
 
           # Deliver 2 messages via subscribe (push consumer)
           delivered_count = 0
-          q.subscribe(no_ack: false) do |msg|
+          q.subscribe(no_ack: false) do |delivery|
             delivered_count += 1
-            msg.ack
+            delivery.ack
           end
           wait_for { delivered_count == 2 }
 


jage:
Looks like some other spec is a bit flaky.

jage:
> I agree, but `deliver_get_count` you could say is a summary of `get_count` + `get_no_ack_count` + `deliver_count `+ `deliver_no_ack_count`
> 
> Maybe we could just swap `deliver` to `deliver_get`

Aha! Will adapt spec to trigger that and update code, thanks! :)

viktorerlingsson:
> > I agree, but `deliver_get_count` you could say is a summary of `get_count` + `get_no_ack_count` + `deliver_count `+ `deliver_no_ack_count`
> > Maybe we could just swap `deliver` to `deliver_get`
> 
> Aha! Will adapt spec to trigger that and update code, thanks! :)

Yes, it is based on rmq's message_stats objects:

```
deliver | Count of messages delivered in acknowledgement mode to consumers.
deliver_no_ack | Count of messages delivered in no-acknowledgement mode to consumers.
get | Count of messages delivered in acknowledgement mode in response to basic.get.
get_no_ack | Count of messages delivered in no-acknowledgement mode in response to basic.get.
deliver_get | Sum of all four of the above.

```

PR #1373, opened 2025-10-16
use set_blocking instead of constructor

viktorerlingsson:
Crystal 1.18.0 deprecation. 
Fixes build warning.

This is for 2.4.x!

Make sure builds & spec run

### DIFF
--- a/src/lavinmq/mfile.cr
+++ b/src/lavinmq/mfile.cr
@@ -168,7 +168,8 @@ class MFile < IO
   def copy_to(output : IO, size = @size) : Int64
     if unmapped? # don't remap unmapped files
       raise ClosedError.new if @fd < 0
-      io = IO::FileDescriptor.new(@fd, blocking: true, close_on_finalize: false)
+      io = IO::FileDescriptor.new(@fd, close_on_finalize: false)
+      IO::FileDescriptor.set_blocking(@fd, true)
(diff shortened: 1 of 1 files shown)


src/lavinmq/mfile.cr
```diff
@@ -168,7 +168,8 @@ class MFile < IO
   def copy_to(output : IO, size = @size) : Int64
     if unmapped? # don't remap unmapped files
       raise ClosedError.new if @fd < 0
-      io = IO::FileDescriptor.new(@fd, blocking: true, close_on_finalize: false)
+      io = IO::FileDescriptor.new(@fd, close_on_finalize: false)
+      IO::FileDescriptor.set_blocking(@fd, true)
```
spuun, line 172: I wonder if we should put `{% if flag?(:unix) %}` around this?
In the comment for the method it says `NOTE: Only implemented on UNIX targets. Raises on Windows.`. Dunno why that isn't in the api docs.

viktorerlingsson: Hmm, maybe.. It's very much not the only thing that doesn't work on Windows though 🙂

viktorerlingsson: (this is also only in 2.4.x, this copy_to method is gone in 2.5.x)

PR #1380, opened 2025-10-17
MetricsServer#close idempotent

viktorerlingsson:
Made `MetricsServer#close` idempotent by adding a `@closed` flag that prevents multiple close attempts

Fixes a bug introduced in #1367


viktorerlingsson:
> I wonder if we should consider changing all of our `closed` booleans to atomics and use `swap` just to make it thread safe. Even if we'd do it when not necessary it's so rarely touched so maybe it's worth to be better safe than sorry.

Yeah, maybe!

PR #1381, opened 2025-10-20
Wait for DOM to be loaded for logs page before JS execution

kickster97:
Waits for DOM to be loaded for logs page. 
adresses @spuun comments in #1333 

Compile and look at logs.

### DIFF
--- a/static/js/logs.js
+++ b/static/js/logs.js
@@ -49,20 +49,21 @@ function forbidden () {
 function setScrollMode (toBottom) {
   shouldAutoScroll = toBottom
   localStorage.setItem('logScrollMode', toBottom ? 'bottom' : 'top')
-  if (btnToBottom && btnToTop) {
-    btnToBottom.setAttribute('aria-pressed', String(toBottom))
-    btnToTop.setAttribute('aria-pressed', String(!toBottom))
-  }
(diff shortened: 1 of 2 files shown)


snichme:
As an alternative we could move all `<script>`  to bottom of the page, just before `</body>`  to have it loaded after all the HTML is parsed and in place or we can make use of `defer` https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/defer

Less code and additional events to handle in the js code.

spuun:
> As an alternative we could move all `<script>` to bottom of the page, just before `</body>` to have it loaded after all the HTML is parsed and in place or we can make use of `defer` https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/defer
> 
> Less code and additional events to handle in the js code.

I'd say we should go with `defer`. Totally forgot about that.

kickster97:
> As an alternative we could move all `<script>` to bottom of the page, just before `</body>` to have it loaded after all the HTML is parsed and in place or we can make use of `defer` https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/defer
> 
> Less code and additional events to handle in the js code.

TIL defer! thanks 🙏

carlhoerberg (approved):

snichme (approved):

PR #1390, opened 2025-10-23
Handle AMQPLAIN authentication without required keys

carlhoerberg:
AMQPLAIN is an uncommon alternative to passing credentials in the Start frame. There the response is an AMQP Table, that should include LOGIN and PASSWORD key/values.

Noticed that amqplib client doesn't set a default username/password, and in that case don't send a LOGIN/PASSWORD at all. This handles that case and doesn't raise, but still allow empty passwords.


viktorerlingsson (approved):
Looks good to me. 

I don't think we have any specs for AMQPLAIN auth? That would be nice to have (but I guess that can be added separately).

PR #1397, opened 2025-10-27
Adds GIS filtering to stream consumers

antondalgren:
This pull request adds support for advanced geographical (GIS) filters to stream consumers in the AMQP module, allowing messages to be filtered based on spatial criteria such as radius, bounding box, and polygon. The implementation introduces a new `GISFilter` module, updates the filter parsing logic to handle new filter types, and refactors how filters are constructed from consumer arguments.

**GIS filtering support:**

* Introduced the `GISFilter` module (`src/lavinmq/amqp/stream/filters/gis.cr`) with support for spatial filters: radius, bounding box, and polygon. This includes data structures for points, bounding boxes, and polygons, as well as filter classes that implement the `StreamFilter` interface.

**Filter parsing refactor:**

* Refactored the filter parsing logic by adding `StreamFilter.from_arguments`, which parses all supported filters (including GIS filters) from consumer arguments, replacing the previous `validate_stream_filter` method.
* Updated `StreamConsumer` to use the new `StreamFilter.from_arguments` method for constructing the list of filters from frame arguments, simplifying and centralizing filter handling. [[1]](diffhunk://#diff-7c1b4ac936d0e8545ea22bf434ba350572b7e89879951ec29200ee9f512da1b3L45-R47) [[2]](diffhunk://#diff-7c1b4ac936d0e8545ea22bf434ba350572b7e89879951ec29200ee9f512da1b3L71-L96)

Specs and by publishing data to a stream with GPS coordinates in `x-geo-lat` and `x-geo-lon` and utilizing the different filters.

### DIFF
--- a/src/lavinmq/amqp/stream/filters/gis.cr
+++ b/src/lavinmq/amqp/stream/filters/gis.cr
@@ -3,7 +3,7 @@ require "./filter"
 
 module LavinMQ::AMQP
   # GIS filtering module for geographical data in stream queues.
-  # Provides PostGIS-like spatial operations for filtering messages based on coordinates.
+  # Provides spatial operations for filtering messages based on coordinates.
   module GISFilter
     EARTH_RADIUS_KM = 6371.0
 


antondalgren:
> Redundant distance calculations (src/lavinmq/amqp/stream/filters/gis.cr:47-64)
>
> Haversine distance calculation performs expensive trigonometric operations on every message
> No early exit optimizations or bounding box pre-filtering for radius checks

```  ⎿  GIS Filter Benchmark: Current vs Optimized Radius Filter
     ======================================================================

     Correctness Check:
     ----------------------------------------------------------------------
     Center point              - Current: true, Optimized: true ✓
     Inside radius (~6km)      - Current: true, Optimized: true ✓
     Outside radius (~32km)    - Current: true, Optimized: true ✓
     Boundary (~11km)          - Current: true, Optimized: true ✓

     Performance Benchmarks (10,000 messages per scenario):
     ----------------------------------------------------------------------

     Scenario 1: Mixed Distribution (33% inside, 33% outside, 33% boundary)
       current  35.06  ( 28.53ms) (± 1.06%)  3.23MB/op        fastest
     optimized  35.05  ( 28.53ms) (± 1.06%)  3.23MB/op   1.00× slower

     Scenario 2: Mostly Outside (90% outside, 10% inside)
       current  22.96  ( 43.56ms) (± 0.98%)  4.44MB/op        fastest
     optimized  22.79  ( 43.87ms) (± 0.99%)  4.44MB/op   1.01× slower

     Scenario 3: Mostly Inside (90% inside, 10% outside)
       current   1.35k (738.24µs) (± 1.17%)  938kB/op   1.01× slower
     optimized   1.37k (730.11µs) (± 1.37%)  938kB/op        fastest

     Scenario 4: All Boundary Points (all require haversine calculation)
       current   1.34k (745.33µs) (± 1.19%)  938kB/op   1.01× slower
     optimized   1.35k (738.47µs) (± 1.15%)  938kB/op        fastest

     ======================================================================
     Summary:
     - The optimized version uses bounding box pre-filtering
     - Bounding box checks are simple lat/lon comparisons (very fast)
     - Haversine calculation only runs if point is inside bounding box
     - Best performance gain when most points are outside the radius
     - Minimal overhead when points are near the boundary

∴ Thinking…

  Interesting! The benchmark results show that the optimization doesn't actually provide a meaningful performance improvement. In fact, the current implementation is slightly faster in
  some cases. The differences are all within the margin of error (1.00-1.01× difference).

  This is surprising, but actually makes sense when you think about it:

  1. The Point.from_headers call is the same in both implementations
  2. The bounding box checks add 4 extra comparisons before the haversine calculation
  3. Modern CPUs are very good at trigonometric operations, and the haversine calculation is already quite optimized in Crystal's Math library
  4. The overhead of the extra bounding box checks might actually be similar to or greater than just doing the haversine calculation

  This is a great example of why benchmarking is important - the "obvious" optimization doesn't always pan out in practice.
  ```
  
  I would rate this comment from Claude as not applicable?

**Comment by antondalgren** [comment]
>Inefficient polygon point-in-polygon algorithm (src/lavinmq/amqp/stream/filters/gis.cr:148-167)
>
>The ray casting algorithm has a division by zero vulnerability when pj.lon == pi.lon
>Line 159: `(pj.lat - pi.lat) * (point.lon - pi.lon) / (pj.lon - pi.lon)` will crash with division by zero
>The algorithm lacks proper handling of edge cases like points exactly on polygon edges

I have a testcase with vertical lines. Seems not applicable

**Comment by antondalgren** [comment]
>Missing validation in polygon filter parsing (src/lavinmq/amqp/stream/filters/gis.cr:235-263)
>
>Line 252: No validation that parsed lat/lon values are finite numbers
>Malformed arrays with non-numeric values could cause runtime crashes

Handled in constructor of `Point`....

**Comment by antondalgren** [comment]
>Unhandled coordinate parsing edge cases (src/lavinmq/amqp/stream/filters/gis.cr:266-277)

Don't see where this happens

**Comment by antondalgren** [comment]
> Missing type safety (src/lavinmq/amqp/stream/filters/gis.cr:245-255)
>
>Array iteration assumes point_field[0] and point_field[1] exist without bounds checking
>Could cause IndexError for malformed polygon point arrays

It's checked on the line above.

File: src/lavinmq/amqp/stream/filters/gis.cr [hunk 1]
```diff
@@ -0,0 +1,279 @@
...
+      end
+    end
+
+    class PolygonFilter
+      include StreamFilter
+
+      getter polygon : Polygon
+
+      def initialize(@polygon : Polygon)
+      end
```
spuun, line 191: What about creating a `BoundingBox` here which then can be used as a rough filter in `PolygonFilter#match?`? If the box contains the point, continue to check the polgon.

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,279 @@
...
+          # Check if the edge crosses the ray from point to the right
+          # The first condition ensures pj.lon != pi.lon (no vertical edges)
+          if ((pi.lon > point.lon) != (pj.lon > point.lon)) &&
+             (point.lat < (pj.lat - pi.lat) * (point.lon - pi.lon) / (pj.lon - pi.lon) + pi.lat)
+            inside = !inside
+          end
+
+          j = i
+        end
+
```
kickster97: just nitpick, but it would be nice if we did not have this mny spaces between the code lines, its a bit harder to read 👼

src/lavinmq/amqp/stream/filters/filter.cr
```diff
@@ -1,5 +1,77 @@
...
+      filters = Array(StreamFilter).new
+
+      if stream_filter = arguments["x-stream-filter"]?
+        parse_stream_filter(stream_filter, filters)
+      end
+
+      parse_gis_filters(arguments, filters)
+
+      filters
+    end
```
kickster97, line 14: e.g. 😄 
```suggestion
    def self.from_arguments(arguments : AMQ::Protocol::Table) : Array(StreamFilter)
      filters = Array(StreamFilter).new
      if stream_filter = arguments["x-stream-filter"]?
        parse_stream_filter(stream_filter, filters)
      end
      
      parse_gis_filters(arguments, filters)
      filters
    end
```

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,279 @@
...
+      def initialize(@center : Point, @radius_km : Float64)
+        raise ArgumentError.new("Radius must be positive") if @radius_km <= 0
+      end
+
+      def match?(headers : AMQP::Table) : Bool
+        msg_point = Point.from_headers(headers)
+        return false unless msg_point
+
+        center.distance_to(msg_point) <= radius_km
```
viktorerlingsson: I think we most often try to stick to this kind of style when doing nil-checks
```suggestion
        if msg_point = Point.from_headers(headers)
          return center.distance_to(msg_point) <= radius_km
        end
        false
```

src/lavinmq/amqp/stream/filters/gis.cr
antondalgren: Did a benchmark and it seems that it starts to provide most value when the polygon consists of >10 vertices

     ================================================================================
     SUMMARY: Performance Improvement (% faster with bbox pre-filter)
     ================================================================================
     Vertices | Mostly Outside | Mixed Distribution | Mostly Inside
     --------------------------------------------------------------------------------
            3 |     +1.79%     |    -10.10%     |     -5.40%
            5 |     +1.21%     |     -8.74%     |    -10.10%
           10 |     +1.42%     |     -4.59%     |     +4.98%
           20 |     +1.21%     |     +5.73%     |     -2.57%
           50 |     +1.74%     |     +8.80%     |     +2.13%
          100 |     +2.36%     |    +17.34%     |     +6.51%
          200 |     +1.39%     |    +23.12%     |     +9.92%
          500 |     +2.28%     |    +26.37%     |     +8.96%
     ================================================================================

     Analysis:
     - Positive % means optimized version is faster
     - Negative % means current version is faster (bbox overhead not worth it)
     - Look for the vertex count where optimization becomes consistently positive
     - 'Mostly Outside' scenario should benefit most from bbox pre-filtering

antondalgren: Claude's conclusion seems that at around 50 points its worth adding a bounding box. So will go with that, but happy to alter it

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,291 @@
...
+      def to_s(io : IO)
+        io << "BoundingBox(lat=" << @min_lat << ".." << @max_lat
+        io << ", lon=" << @min_lon << ".." << @max_lon << ")"
+      end
+    end
+
+    # Represents a polygon as an array of points
+    struct Polygon
+      getter points : Array(Point)
+      getter bbox : BoundingBox | Nil
```
spuun:
```suggestion
      getter bbox : BoundingBox?
```

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,291 @@
...
+
+    # Helper to convert AMQ::Protocol::Field to Float64
+    def self.to_float64?(value : ::AMQ::Protocol::Field?) : Float64?
+      return nil unless value
+
+      case value
+      when Int32, Int64, Float32, Float64
+        value.to_f64
+      when String
+        value.to_f64?
```
spuun: Any reason to support `String`?

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,291 @@
...
+      bbox = BoundingBox.new(min_lat, max_lat, min_lon, max_lon)
+      BoundingBoxFilter.new(bbox)
+    rescue ArgumentError
+      nil
+    end
+
+    # Parse polygon filter from consumer arguments
+    # Expected format: {"points": [[lat1, lon1], [lat2, lon2], ...]}
+    def self.parse_polygon_filter(value : ::AMQ::Protocol::Field) : PolygonFilter?
+      return nil unless value.is_a?(::AMQ::Protocol::Table)
```
spuun: Dunno, but it feels odd to accept `Field` then validate that it's a `Table` and return nil for invalid type?
Why not type `value` as `Table` and force the caller to pass the correct type?

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,291 @@
...
+
+      points_field = value["points"]?
+      return nil unless points_field.is_a?(Array)
+
+      points = [] of Point
+      points_field.each do |point_field|
+        next unless point_field.is_a?(Array)
+        next unless point_field.size == 2
```
spuun: Here an invalid value is "ignored", but in other places we fail (like below if lat or lon is nil). Shouldn't we fail here too?

src/lavinmq/amqp/stream/filters/gis.cr
spuun: How did you (or claude) benchmark it? 

Shouldn't the "mostly outside" always perform better than "mixed"? 🤔

antondalgren:
> Any reason you chose to return nil instead of raising exceptions for invalid values?
> 
> If we raise instead of returning nil we can get rid of a lot of nil checks and code like
> 
> ```crystal
>         lat = headers["x-geo-lat"]?
>         lon = headers["x-geo-lon"]?
>         return nil unless lat && lon
> 
>         lat_f = GISFilter.to_float64?(lat)
>         lon_f = GISFilter.to_float64?(lon)
>         return nil unless lat_f && lon_f
> 
>         Point.new(lat_f, lon_f)
> ```
> 
> can probably be just (if we decide to not support strings)
> 
> ```crystal
>         lat_f = headers["x-geo-lat"].as(Number).to_f64
>         lon_f = headers["x-geo-lon"].as(Number).to_f64
> 
>         Point.new(lat_f, lon_f)
> ```
> 
> If we want to support strings we can probably create an alias and do something like
> 
> ```crystal
>        alias LatLongType = Int32 | Int64 | Float32 | Float64 | String
> 
>         lat_f = headers["x-geo-lat"].as(LatLongType).to_f64
>         lon_f = headers["x-geo-lon"].as(LatLongType).to_f64
> 
>         Point.new(lat_f, lon_f)
> ```
> 
> Rescuing these exceptions and pass them as cause to `PreconditionFailed` will give a nice backtrace all the way to where the initial exceptions was raised.

This would result in us raising errors if a message inside the stream doesn't have the `x-geo-lat` or `x-geo-lon`. I don't think that is a good solution! It's better to return nil. I would assume nil checks are cheaper than raises since we don't have to jump around in the callstack?

spuun:
> This would result in us raising errors if a message inside the stream doesn't have the `x-geo-lat` or `x-geo-lon`. I don't think that is a good solution! It's better to return nil. I would assume nil checks are cheaper than raises since we don't have to jump around in the callstack?

Right, this is headers from the message 🤦

src/lavinmq/amqp/stream/filters/filter.cr
```diff
@@ -1,5 +1,81 @@
...
+          parse_stream_filter(f, filters)
+        end
+      when Nil
+        # noop
+      else
+        raise LavinMQ::Error::PreconditionFailed.new("x-stream-filter must be a string, table, or array")
+      end
+    end
+
+    private def self.parse_gis_filters(arguments : AMQ::Protocol::Table, filters)
```
viktorerlingsson: I've been thinking about the naming/structure of filters, and I think these geo filters should be defined within the same `x-stream-filter` argument as existing filters. IMO that would make it easier to understand how things are connected, especially when one uses `x-filter-match-type`. If the geo filters are outside of `x-stream-filters`, I don't think it's very obvious what happens in this scenario for example. Like why do we filter on `x-stream-filter` and `x-geo-within-radius`, but not `x-foo`? 
```
{
	"x-stream-filter": {
        "app": "my-app"
    },
	"x-geo-within-radius": {
        "lat": lat,
        "lon": lon,
        "radius_km": radius,
     },
	"x-foo": "bar"
    "x-filter-match-type": "all"
}
```
That's more obvious if defined like this I think: 
```
{
	"x-stream-filter": {
        "app": "my-app",
		"geo-within-radius": {
			"lat": lat,
			"lon": lon,
			"radius_km": radius,
		},
    },
	"x-foo": "bar"
    "x-filter-match-type": "all"
}
```

antondalgren: Sounds like a good approach to me!

src/lavinmq/amqp/stream/filters/filter.cr
```diff
@@ -1,5 +1,68 @@
 module LavinMQ::AMQP
   module StreamFilter
     abstract def match?(headers : AMQP::Table) : Bool
+
+    # Parse all filters from consumer arguments
+    def self.from_arguments(arguments : AMQ::Protocol::Table) : Array(StreamFilter)
+      filters = Array(StreamFilter).new
```
viktorerlingsson, line 7: I think `@filters` on a stream_consumer should be nilable and default to nil, that would save an array allocation when no filters are used (which I think is the most used path). So this should then return nil if there are no filters.

antondalgren: I think that should be another issue and not a blocker for this.

viktorerlingsson: Sure!👍

src/lavinmq/amqp/stream/filters/gis.cr
```diff
@@ -0,0 +1,269 @@
+require "amq-protocol"
+require "./filter"
+
+module LavinMQ::AMQP
+  # GIS filtering module for geographical data in stream queues.
+  # Provides PostGIS-like spatial operations for filtering messages based on coordinates.
```
viktorerlingsson: Very nitpick-y but I don't think we should mention other products unless absolutely needed 🙂 

```suggestion
  # Provides spatial operations for filtering messages based on coordinates.
```

viktorerlingsson (approved):
LGTM! Nice job👍

PR #1398, opened 2025-10-27
automatically set --prerelease flag if tag includes alpha|beta|rc

viktorerlingsson:
Automatically tag releases as pre-release if they have `alpha`, `beta` or `rc` in the name, so we don't have to remember to do that manually. 

Create a new release? 😨


viktorerlingsson:
> The change make sense, but hard to test without creating a release :)

Yeah, guess we'll see next time we're creating a release! 🙂

PR #1406, opened 2025-10-28
add limit to the UI for how many messages you can get at once

### LINKED ISSUE #1371: Introduce a limit on "Get messages"
**Summary**  

Requesting too many messages can lead to issues.

**Current Behavior**  
Today a user can request an unlimited number of messages.

Image

**Desired Behavior**  
Limit to say 1000 messages or 10MiB, whichever is smaller.

kickster97:
Adds a limit of 1000 to the queue page when getting messages so users don't run into issues. 
fixes #1371 

We truncate the response to max 50kB, so when getting 1000 msgs we are max handling 50MiB. This takes a second to load in. An option could be to truncate the message to the first 10kB, then it will load a bit faster. 
 
start up LaivnMQ, enqueue many messages and try to get 0 or more than 1000 msgs

### DIFF
--- a/static/js/queue.js
+++ b/static/js/queue.js
@@ -217,15 +217,9 @@ document.querySelector('#publishMessage').addEventListener('submit', function (e
 document.querySelector('#getMessages').addEventListener('submit', function (evt) {
   evt.preventDefault()
   const data = new window.FormData(this)
-  const count = parseInt(data.get('messages'))
-  if (count < 1 || count > 1000) {
-    window.alert('Please enter a number between 1 and 1000 messages')
-    return
-  }
-
   const url = HTTP.url`api/queues/${vhost}/${queue}/get`
   const body = {
-    count: count,
+    count: parseInt(data.get('messages')),
     ack_mode: data.get('mode'),
     encoding: data.get('encoding'),
     truncate: 50000


views/queue.ecr
```diff
@@ -227,6 +227,7 @@
           <label>
             <span>Messages</span>
             <input type="number" name="messages" value=1>
```
viktorerlingsson: we could also set `max="100"` on the `<input>`

kickster97: I tried this but it did not work for some reason, must be some css overwriting happening.. 
Also, html can be edited so idk if its safe to just have the limit there?

viktorerlingsson: ah, okay 🤷 
I get this in a minimal example when trying to submit, and it prevents the form from being submitted 
image

kickster97: Ah yeah it works for me now aswell..  🤦 
but then again, i can just inspect the form and remove the limit. I think the real protection here is in the javascript

kickster97: but we could add it as an additional layer if its the form interaction you like 😄

viktorerlingsson: Neither really is protection, you can bypass the javascript validation pretty much as easily🙂 
Maybe we do both? 🤷

spuun:
> Shouldn't it be a limit in the backend aswell? here https://github.com/cloudamqp/lavinmq/blob/main/src/lavinmq/http/controller/queues.cr#L155

No, c1b8ec3 is enough to fix the backend to not crash. This UI fix is just to prevent users from breaking their browser.

views/queue.ecr
```diff
@@ -226,7 +226,8 @@
           </label>
           <label>
             <span>Messages</span>
-            <input type="number" name="messages" value=1>
+            <input type="number" name="messages" value=1 max="1000">
```
carlhoerberg: imho the JS validation is superfluous with this: 
```suggestion
            <input type="number" name="messages" value=1 min="1" max="1000">
```

carlhoerberg (commented):
Did you test with 1000 large messages too?

kickster97:
> Did you test with 1000 large messages too? 

Yes, its a bit slower but does not crash. What we could do is truncate at a lower value, like 10000? Then we only need to process 10 MiB (currently 50MiB)

views/queue.ecr
kickster97: ok!

PR #1410, opened 2025-10-29
Build for Debian 13 & Fedora 43. Remove builds for Fedora 40

viktorerlingsson:
✅  Adds builds for: 
- Debian 13 (Trixie)
- Fedora 43

❌  Removes builds for: 
- Fedora 40 (Out of support)

See that the builds works
Debian test-run here: https://github.com/cloudamqp/lavinmq/actions/runs/18913394201
RPM test-run here: https://github.com/cloudamqp/lavinmq/actions/runs/18913389885

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, ubuntu-22.04, debian-13, debian-12, debian-11]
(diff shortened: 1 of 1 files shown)


snichme:
Looked up usage on ubuntu versions and 20.04 appears to be quite heavily used still. Maybe keep them even though they are EOL?

viktorerlingsson:
> Looked up usage on ubuntu versions and 20.04 appears to be quite heavily used still. Maybe keep them even though they are EOL?

Sure!

snichme (approved):

PR #1413, opened 2025-10-29
get channels for all connections in a vhost, not just the first

### LINKED ISSUE #1414: Filtering by vhost on the channels page only shows channels for the first connection
**Describe the bug**
Filtering by vhost on the channels page only shows channels for the first connection, because we use `.find` instead of `.select`. 

Same goes for the API. 

**Describe your setup**
Any version of LavinMQ

**How to reproduce**
Open channels on more than one connection in a vhost, view the page filtered by your vhost and see that not all channels are displayed. 

**Expected behavior**
All channels for a specific vhost should be displayed

viktorerlingsson:
Use `@amqp_server.vhosts[vhost].connections.each` instead of  `.find` on the list of connections, to get channels for all connections in a vhost, not just the first. 

Fixes #1414 

Run spec

### DIFF
--- a/spec/api/channels_spec.cr
+++ b/spec/api/channels_spec.cr
@@ -41,6 +41,24 @@ describe LavinMQ::HTTP::ChannelsController do
         body.as_a.empty?.should be_true
       end
     end
+
+    # Filtering by vhost on the channels page only shows channels for the first connection
+    # https://github.com/cloudamqp/lavinmq/issues/1414
+    it "should return all channels for a vhost with multiple connections" do
+      with_http_server do |http, s|
+        s.vhosts.create("my-connection")
+        s.users.add_permission("guest", "my-connection", /.*/, /.*/, /.*/)
+        wg = WaitGroup.new(1)
+        2.times { spawn { with_channel(s, vhost: "my-connection") { wg.wait } } }
+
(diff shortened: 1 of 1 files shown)


src/lavinmq/http/controller/channels.cr
```diff
@@ -15,8 +15,8 @@ module LavinMQ
         get "/api/vhosts/:vhost/channels" do |context, params|
           with_vhost(context, params) do |vhost|
             refuse_unless_management(context, user(context), vhost)
-            c = @amqp_server.connections.find { |conn| conn.vhost.name == vhost }
-            channels = c.try(&.channels.each_value) || ([] of Client::Channel).each
+            conns = @amqp_server.connections.select { |conn| conn.vhost.name == vhost }
+            channels = Iterator(Client::Channel).chain(conns.map(&.channels.each_value))
```
spuun: What about instead of iterating through all connections:
```crystal
conns = @amqp_server.vhosts[vhost].connections.each
```
And the next line could probably be just
```crystal
channels = conns.flat_map(&.channels.each_value)
```
to get rid of array allocations.

viktorerlingsson: Yes, thanks!

PR #1421, opened 2025-11-01
Remove meta files after prio store migration

spuun:
Since the prio queue store migration was written we've introduced meta files which should be removed after migration.

Specs

### DIFF
--- a/src/lavinmq/message.cr
+++ b/src/lavinmq/message.cr
@@ -35,11 +35,6 @@ module LavinMQ
       nil
     end
 
-    def bytesize
-      sizeof(Int64) + 1 + @exchange_name.bytesize + 1 + @routing_key.bytesize +
-        @properties.bytesize + sizeof(UInt64) + @bodysize
-    end
-
     def to_io(io : IO, format = IO::ByteFormat::SystemEndian)
       io.write_bytes @timestamp, format
       io.write_bytes AMQ::Protocol::ShortString.new(@exchange_name), format


src/lavinmq/amqp/queue/priority_queue.cr
carlhoerberg, line 79: You can use `BytesMessage` here

carlhoerberg: Wait, what, add 8096 until the uint32 wraps around to exactly 0? that will take a lot of iterations.. 134,217,728 to be exact apparently.

src/lavinmq/amqp/queue/priority_queue.cr
```diff
@@ -95,8 +95,9 @@ module LavinMQ::AMQP
         old_store.close
         i = 0u32
         delete_wg = WaitGroup.new
+        pattern = %r{^(msgs|acks|meta)\.}
         Dir.each_child(@msg_dir) do |f|
-          if f.starts_with?("msgs.") || f.starts_with?("acks.")
+          if f.matches? pattern
```
carlhoerberg, line 92: regex is way slower than starts_with, but i guess it's ok here.

src/lavinmq/amqp/queue/priority_queue.cr
spuun: I have no idea, must have been some kind of meltdown or copilot completion.

src/lavinmq/amqp/queue/priority_queue.cr
spuun: Yes, that was my thinking.

src/lavinmq/amqp/queue/priority_queue.cr
spuun: Not without adding code to BytesMessage, but maybe worth it.

src/lavinmq/message.cr
```diff
@@ -35,6 +35,20 @@ module LavinMQ
       nil
     end
 
+    def bytesize
+      sizeof(Int64) + 1 + @exchange_name.bytesize + 1 + @routing_key.bytesize +
+        @properties.bytesize + sizeof(UInt64) + @bodysize
+    end
```
viktorerlingsson: This shouldn't be here, right? looks like it's just a copy of row 15-18

src/lavinmq/amqp/queue/priority_queue.cr
spuun: Actually, in this case regex seems to be 20x faster.
```bash
      regex  31.64M ( 31.61ns) (± 3.79%)  0.0B/op        fastest
starts_with   1.57M (636.45ns) (± 0.11%)  0.0B/op  20.14× slower
```

Code
https://gist.github.com/spuun/c39bce21c00db5bb2cb5f66d32d604f6

src/lavinmq/message.cr
spuun: Lol, yes. Sloppy reading from my side.

src/lavinmq/amqp/queue/priority_queue.cr
spuun: Oh, I didn't even do good test data. If I fix that I get a even bigger difference.

viktorerlingsson (approved):
LGTM 👍

PR #1427, opened 2025-11-03
Fix: Display error when binding to non-existent exchange/queue

### LINKED ISSUE #1426: UI: Add binding for missing exchange/queue does not show error message
**Describe the bug**

Image
Image

Developer console:

```
api/bindings/%2F/e/amq.topic/q/aoeu:1  Failed to load resource: the server responded with a status of 404 (Not Found)
http.js:32 Not found: undefined
standardErrorHandler @ http.js:32
exchange#vhost=%2F&name=amq.topic:1 Uncaught (in promise) Object
```

**Describe your setup**

Tested in latest main 315cabd7f4499593cec0fd026c750f5298210ccd locally, seen it in production as well.

**How to reproduce**

Try to bind an exchange to a queue/exchange that does not exist.

**Expected behavior**

It should show an error message that the queue/exchange does not exist and need to be explicitly created first.

jage:

When attempting to bind to a non-existent exchange or queue in the UI, a 404 error was logged to the console but no user-facing error message was shown.

Added catch handlers in exchange.js and queue.js to display a toast notification informing the user that the target exchange/queue does not exist and needs to be created first.

Screenshot 2025-11-03 at 16 18 15

Fixes #1426

Tested locally.

### DIFF
--- a/static/js/exchange.js
+++ b/static/js/exchange.js
@@ -109,6 +109,12 @@ document.querySelector('#addBinding').addEventListener('submit', function (evt) 
       bindingsTable.reload()
       evt.target.reset()
     })
+    .catch(e => {
+      if (e.status === 404) {
+        const type = t === 'q' ? 'Queue' : 'Exchange'
+        DOM.toast(`${type} '${d}' does not exist and needs to be created first.`, 'error')
+      }
+    })
 })
 
 document.querySelector('#publishMessage').addEventListener('submit', function (evt) {
--- a/static/js/queue.js
+++ b/static/js/queue.js
@@ -192,6 +192,11 @@ document.querySelector('#addBinding').addEventListener('submit', function (evt) 
       evt.target.reset()
       DOM.toast('Exchange ' + e + ' bound to queue')
     })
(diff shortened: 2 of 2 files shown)


jage:
> Do the changes in the ci workflow and in the `upstream_spec.cr` have something to do with the toast notifications? If not it would be nice to break them into another PR.

Frontend CI is broken without [81b1c58](https://github.com/cloudamqp/lavinmq/pull/1427/commits/81b1c58e5ca588cf5d302c926b152f4a6846093d) AFAIK, but removed it for now.

Upstream checks are unrelated but want CI to be green, but giving up on that so will just remove that commit.

jage:
> Other than that the javascript changes look good, don't know if we can get more detailed because now any 404 will display this error? :)

Just when adding bindings?

kickster97:
> > Other than that the javascript changes look good, don't know if we can get more detailed because now any 404 will display this error? :)
> 
> Just when adding bindings?

hmm yeah i think i hallucinated here.. i think we should raise a 400 but that is for another PR

PR #1428, opened 2025-11-05
Don't keep FD open for MFile

carlhoerberg:
There no longer a need to keep the FD around, it was before when
mmap:ing and unmap:ing was done dynamically.

MFile#read_at reads from the mmap, not using pread.

MFile#truncate now uses the truncate syscall which takes the path
as argument. It unmaps the truncated part. Only allows shrinking.
Growing might move the `@buffer` in which case all previous references
to it are invalidated.

Use stricter memory barriers for MFile closed variable,
to prevent instruction reordering

Thread safe `@deleted` variable

### DIFF
--- a/src/lavinmq/mfile.cr
+++ b/src/lavinmq/mfile.cr
@@ -11,6 +11,7 @@ lib LibC
   {% end %}
   fun munmap(addr : Void*, len : SizeT) : Int
   fun msync(addr : Void*, len : SizeT, flags : Int) : Int
+  fun truncate(path : Char*, length : OffT) : Int
 end
 
 # Memory mapped file
@@ -20,17 +21,18 @@ end
 # not `size` large, only on graceful close is the file truncated to its `size`.
 # The file does not expand further than initial `capacity`, unless manually expanded.
 class MFile < IO
+  private PAGE_SIZE = LibC.sysconf(LibC::SC_PAGESIZE)
+
   getter pos : Int64 = 0i64
   getter size : Int64 = 0i64
   getter capacity : Int64 = 0i64
   getter path : String
-  @fd = Atomic(Int32).new(-1)
   @buffer : Pointer(UInt8)
-  @deleted = false
-  getter? closed = false
+  @deleted = Atomic(Bool).new(false)
+  @closed = Atomic(Bool).new(false)
 
-  def fd
-    @fd.get(:relaxed)
+  def closed?
+    @closed.get(:acquire)
   end
 
   # Map a file, if no capacity is given the file must exists and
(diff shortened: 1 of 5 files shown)


src/lavinmq/mfile.cr
```diff
@@ -103,27 +108,20 @@ class MFile < IO
...
-      begin
-        if truncate_to_size && !@readonly && !@deleted
-          code = LibC.ftruncate(fd, @size)
-          raise File::Error.from_errno("Error truncating file", file: @path) if code < 0
-        end
-      ensure
-        code = LibC.close(fd)
-        raise File::Error.from_errno("Error closing file", file: @path) if code < 0
-      end
+    return if @closed.swap(true, :relaxed)
```
spuun: I think we need `:acquire_release` to be safe?

carlhoerberg: Why do you think that? Also in def closed?

spuun: The operation is still atomic, but the CPU and/or compiler may still reorganize operations to optimize things. 
In `closed?` we should probably have `:acquire`.

`:relaxed` is probably only viable for counters which no logic relies on.

carlhoerberg: ok!

src/lavinmq/mfile.cr
```diff
@@ -93,38 +102,65 @@ class MFile < IO
...
+        code = LibC.munmap(@buffer, old_capacity)
+        raise RuntimeError.from_errno("Error unmapping file") if code == -1
+
+        fd = open_fd
+        begin
+          @buffer = mmap(fd, new_capacity)
+        ensure
+          close_fd(fd)
+        end
+      {% end %}
```
viktorerlingsson: Maybe move this to a separate `def remap(new_capacity, old_capacity)` function to make it a little easier to read?

src/lavinmq/mfile.cr
```diff
@@ -93,38 +102,65 @@ class MFile < IO
...
+
+    # First truncate the file on disk
+    code = LibC.truncate(@path.check_no_null_byte, new_capacity)
     raise File::Error.from_errno("Error truncating file", file: @path) if code < 0
+
+    # Remap the memory region if capacity changed
+    if new_capacity != old_capacity
+      {% if flag?(:linux) %}
+        # Use mremap on Linux for efficient in-place resizing
+        new_buffer = LibC.mremap(@buffer, old_capacity, new_capacity, 0)
```
viktorerlingsson: reading up on mremap(), do we need the `MREMAP_MAYMOVE` flag? 
>By default, if there is not sufficient space to expand a
              mapping at its current location, then mremap() fails.  If
              this flag is specified, then the kernel is permitted to
              relocate the mapping to a new virtual address, if
              necessary.

carlhoerberg: Yes and no. There might not be space and then it needs to move, BUT, that will invalidate all current references to `@buffer` that might be used in other read/write operations. 

Ideally we shouldnt support growing the file at all, it's only used in specs anyway

src/lavinmq/mfile.cr
```diff
@@ -93,38 +104,61 @@ class MFile < IO
...
+    new_capacity = new_capacity.to_i64
+    old_capacity = @capacity
+    return if new_capacity == old_capacity # no change
+    raise ArgumentError.new("Cannot expand a MFile") if new_capacity > old_capacity
+
+    # First truncate the file on disk
+    code = LibC.truncate(@path.check_no_null_byte, new_capacity)
     raise File::Error.from_errno("Error truncating file", file: @path) if code < 0
+
+    # Remap the memory region if capacity changed
```
carlhoerberg: Maybe we shouldn't shrink the mmap, it adds overhead, and we already protect against writing past `@capacity` and reading past `@size` in `write` and `read`.

It does however free up some virtual memory, not sure how big of an impact it might be, maybe for very long queues.

PR #1430, opened 2025-11-05
Run OS X (macos) specs on Depot runners

jage:

Run OS X (macos) specs on Depot runners.
Hopefully faster and less flaky.

Look at the CI, green is good. :)


jage:
Aborting this for now, we need a new depot subscription for this to work.

jage:
Got it working, but not faster, closing.

PR #1431, opened 2025-11-05
Faster GHA macos runners

jage:
Newer macOS version, and bigger machine that's hopefully faster and
less flaky.

If CI is green, then maybe this helps. :)


jage:
It's a bit faster, but specs are just as flaky which was what I wanted to see if this could help with.

PR #1432, opened 2025-11-05
Bugfix: don't abort policy applying on invalid value

### LINKED ISSUE #1342: Unhandled exceptions when applying policies
One may add policies with invalid values causing `PreconditionFailed` to be raised when the policy is applied to a queue (in my case). This means that applying of the policy is aborted and no other arguments is applied, not even to any other entity because the exception bubbles up all the way to the fiber which crashes.

When LavinMQ then is restarted the error occurs during upstart.

Reproduce by creating a stream then adding a policy to queues with the definition `{"max-age":"abcd"}`.

spuun:
This move some of the policy code from the includers of `PolicyTarget` to `PolicyTarget`, centralizing the applying of policies to simplify the error handling.

Fixes #1342

Run specs

### DIFF
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -1,5 +1,19 @@
 require "./spec_helper"
 require "./../src/lavinmq/amqp/queue"
+require "./../src/lavinmq/policy"
+
+alias Policy = LavinMQ::Policy
+
+def with_queue(&)
+  with_amqp_server do |s|
+    vhost = s.vhosts["/"]
+    vhost.declare_queue("q", durable: true, auto_delete: false)
+    q = vhost.queues["q"]
+    yield q
+  ensure
+    q.try &.delete
+  end
+end
 
 describe LavinMQ::AMQP::Queue do
   it "should expire itself after last consumer disconnects" do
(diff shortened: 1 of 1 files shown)


src/lavinmq/policy.cr
```diff
@@ -3,10 +3,30 @@ require "./sortable_json"
...
+        # This log line isn't very good. Sometimes @log should be used, but we can't know that here. With the new L
+        # logging it's not as important.
+        Log.warn(exception: ex) { "Error applying policy argument #{key}=#{value}: #{ex.message}" }
+      end
+      @policy = policy
+      @operator_policy = operator_policy
+    end
+
+    def clear_policy
+    end
```
carlhoerberg, line 28: Should it really be empty? not better to make it abstract and do an empty where needed? more explicit..

carlhoerberg:
```suggestion
    def clear_policy
      @policy = nil
      @operator_policy = nil
    end
```

carlhoerberg: and then `super` where it needs to be overriden

spuun: No need for super, but it should be abstract!

spuun:
Since `PolicyTarget` is responsible for `@policy` and `@operator_policy` I made `#clear_policy` a method in `PolicyTarget` that clears these ivars and calls `#clear_policy_arguments` for the includer to handle what ever needs to be handled.

I could have made the includers override `#clear_policy` and call super, but that can be missed. By introducing the abstract `#clear_policy_arguments` instead the risk of mistake is lower.

PR #1435, opened 2025-11-05
Bugfix: remove meta files when removing segments for streams

### LINKED ISSUE #1434: meta. files aren't cleared for streams
**Describe the bug**
meta. files aren't cleared for streams when messages are dropped by `drop_overflow`

**Describe your setup**
LavinMQ v2.5.0

**How to reproduce**
Create a stream with max-length (or any other limiting factor), publish messages to make it overflow, see that segment files are cleared but not meta files. 

**Expected behavior**
Meta files should be cleared as well

viktorerlingsson:
Makes sure .`meta` files are also removed when segment file is being deleted by `drop_overflow` in streams. 

Fixes #1434 

Run spec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Boolean queue arguments validated as invalid if set to false [#1429](https://github.com/cloudamqp/lavinmq/pull/1429)
 - Use the correct path for meta files in streams [#1438](https://github.com/cloudamqp/lavinmq/pull/1438)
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -333,7 +333,7 @@ module LavinMQ::AMQP
         @offset_index.delete(seg_id)
         @timestamp_index.delete(seg_id)
         @bytesize -= mfile.size - 4
-        delete_file(mfile)
+        delete_file(mfile, including_meta: true)
```
carlhoerberg, line 336: `including_meta` should probably default to true

carlhoerberg: but let's go with this for now

carlhoerberg (approved):

PR #1440, opened 2025-11-05
Bugfix: Unable to consume from priority queues

### LINKED ISSUE #1439: No messages delivered on prio queues
**Describe the bug**
Consumers are unable to consume messages from prio queues

**Describe your setup**
LavinMQ v2.5.0

**How to reproduce**
Set up a prio queue, publish messages and try to consume. 

**Expected behavior**
Consumers should be able to consume from prio queues.

viktorerlingsson:
Fixes a bug where consumers where unable to consume from priority queues because they where stuck waiting on `empty?` to be false. It would never be false though, because it needs to look at all priority message stores. This adds an override `empty?` method that looks at all stores. 

Fixes #1439 

Manual.

### DIFF
--- a/src/lavinmq/amqp/queue/priority_queue.cr
+++ b/src/lavinmq/amqp/queue/priority_queue.cr
@@ -163,8 +163,10 @@ module LavinMQ::AMQP
       def shift?(consumer = nil) : Envelope?
         raise ClosedError.new if @closed
         @stores.reverse_each do |s|
-          envelope = s.shift?(consumer)
-          return envelope unless envelope.nil?
+          if envelope = s.shift?(consumer)
+            @empty.set true if size.zero?
+            return envelope
+          end
         end
       end
 


src/lavinmq/amqp/queue/priority_queue.cr
```diff
@@ -132,20 +132,24 @@ module LavinMQ::AMQP
...
+        size.zero?
+      end
+
       def push(msg) : SegmentPosition
         raise ClosedError.new if @closed
         prio = Math.min(msg.properties.priority || 0u8, @max_priority)
-        sp = store_for prio, &.push(msg)
         was_empty = size.zero?
         @empty.set false if was_empty
+        sp = store_for prio, &.push(msg)
```
spuun, line 143: These should change place. Don't signal before message is written to disk.
```suggestion
        sp = store_for prio, &.push(msg)
        @empty.set false if was_empty
```

src/lavinmq/amqp/queue/priority_queue.cr
```diff
@@ -132,20 +132,24 @@ module LavinMQ::AMQP
...
         sp
       end
 
       def requeue(sp : SegmentPosition)
         raise ClosedError.new if @closed
-        store_for sp, &.requeue(sp)
         was_empty = size.zero?
         @empty.set false if was_empty
+        store_for sp, &.requeue(sp)
```
spuun, line 151: Ditto
```suggestion
        store_for sp, &.requeue(sp)
        @empty.set false if was_empty
```

spuun (commented):
See inline comments

PR #1298, opened 2025-09-23
1235: Dont return empty msgs from first? and shift?

viktorerlingsson:
### WHAT is this pull request doing?
Raise an `IndexError` in `first?` and `shift?` if `msg.timestamp` is `zero?`. 
This should prevent expire_messages to ack an empty message if `mfile.size` for some reason is incorrect. (For this reason this also changes some specs to use 1 instead of 0 as timestamp)

This also adds protection against underflow of count and bytesize when reading metadata files, to protect against any errors written by earlier LavinMQ versions (where nr of acks maybe could be > msg_count) . 

Fixes #1235

### HOW can this pull request be tested?
Run the specs

PR #1441, opened 2025-11-05
Refactor: DRY meta file handling between MessageStore and StreamMessageStore

antondalgren:
Eliminates code duplication in `read_metadata_file` by extracting common logic to parent class with a template method pattern. This prevents bugs where fixes are only applied to one implementation.

Changes:
- Extract common meta file reading logic to MessageStore base class
- Add `read_extra_metadata_fields` hook for subclasses to extend
- StreamMessageStore overrides hook to read offset/timestamp indices
- Reduces code by ~25 lines

Fixes:
- StreamMessageStore now inherits underflow protection that was added to MessageStore in #1298 but was missing from StreamMessageStore
- Fix test file using old meta path format (line 116)

Related:
- Recent meta file path bug (#1438)
- This refactoring ensures future fixes automatically apply to both

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -463,9 +463,11 @@ module LavinMQ
 
     private def read_metadata_file(seg, mfile)
       metafile = meta_file_name(mfile)
-      count = File.open(metafile) do |f|
-        @replicator.try &.register_file(f)
-        f.read_bytes(UInt32)
+      count = File.open(metafile) do |file|
+        msg_count = file.read_bytes(UInt32)
+        @replicator.try &.register_file(file)
+        read_extra_metadata_fields(file, seg)
+        msg_count
       end
       @segment_msg_count[seg] = count
       bytesize = mfile.size - 4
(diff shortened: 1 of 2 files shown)


src/lavinmq/message_store.cr
```diff
@@ -463,9 +463,11 @@ module LavinMQ
 
     private def read_metadata_file(seg, mfile)
       metafile = meta_file_name(mfile)
-      count = File.open(metafile) do |f|
-        @replicator.try &.register_file(f)
-        f.read_bytes(UInt32)
+      count = File.open(metafile) do |file|
+        msg_count = file.read_bytes(UInt32)
+        @replicator.try &.register_file(file)
+        read_extra_metadata_fields(file, seg)
```
kickster97, line 469: It feels a little off to adapt the parent class to work for those inheriting it.. On the other hand I do like that you got rid of the duplicate code! maybe a comment in the empty `read_extra_metadata_fields` function to explain so we avoid getting confused about this in the future?

antondalgren: >It feels a little off to adapt the parent class to work for those inheriting it.. 

I agree, but it seems like its the best option given the current structure. What we could do is to make an abstract class or similar that does the basic shared stuff and then have the message_store and stream_message_store inherit from that and then have those classes decide what they want to do in the `read_extra_metadata_fields` method. But I don't think that is something we would want to do?

Can surely add a comment if we go with this approach. If we take the approach I mention here, we would probably not need the comment, but some extra work hehe :D

spuun: I think it's fine. It's some kind of template method pattern.

kickster97: Yeah sounds good, go with this! :)

PR #1442, opened 2025-11-06
Bugfix: signal empty after purge (and add specs)

spuun:
The intention was to add specs for priority queue message store, but adding specs exposed a bug which is handled in this PR: the `@empty` boolchannel wasn't set after `purge_all` or `purge`. It should however be set implicitly in `purge` since it's `shift?`ing messages, and `shift?` will signal empty. In `purge_all` it must be signal explicitly.

Run specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- Priority queue message store didn't signal empty after purge all. [#1442](https://github.com/cloudamqp/lavinmq/pull/1442)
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/queue/priority_queue.cr
```diff
@@ -189,11 +189,14 @@ module LavinMQ::AMQP
           break if count >= max_count
           Fiber.yield if (count % PURGE_YIELD_INTERVAL).zero?
         end
+        # Only signal if something was purged, hence count > 0
+        @empty.set true if empty? && count > 0
```
kickster97: If we implicitly signal empty in `shift?`, why is this needed here?

spuun: I don't like to rely on implicit events. What if that changes? But yeah, with specs I guess we should be fine to not set it explicit here.

kickster97: Oki, that make sense!

PR #1444, opened 2025-11-06
Refactor: move acl_cache to user.cr

### LINKED ISSUE #1395: Access refused for user if permissions changed after user connected
If a user's permissions are changed after the user has connected, the user will not get the new permissions. 

Here's a failing spec: 
https://github.com/cloudamqp/lavinmq/compare/update-user-while-connected?expand=1

kickster97:
Previous work: In commit [967de6c](https://github.com/cloudamqp/lavinmq/commit/967de6c6b6f20061d054a9641a1a78176811fb05), the ACL cache was moved from `User` to `Client` and all caching was removed except for write permission checks. By placing the cache in Client, only one fiber (the Client's read loop) accessed each cache instance.

This PR moves the cache handling back to the User and introduces the class `PermissionCache`. The cache instances live in `Client` (one per client connection), while the cache logic and validation live in `User` via revision tracking. This separation is deliberate: placing the cache in Client means only one fiber (the Client's read loop) accesses each cache instance, avoiding the need for locks that would be required if multiple clients shared a User's cache.

The User maintains a `@permission_revision` atomic counter that increments when permissions change. Clients pass their cache to `User#can_write?()`, which compares the cache's stored revision against the current revision. On mismatch, the cache is stale and gets cleared automatically. On match, cached entries are valid and new entries can be added without clearing existing ones.

This revision-based approach enables lock-free cache invalidation across all active clients. When permissions change (via `clear_permissions_cache`), the User's revision increments, and each client's cache detects the mismatch on its next check and self-invalidates. No direct cache clearing or coordination between clients is needed.

Only write checks are cached since they occur on every publish (high frequency), while read/config checks happen rarely and don't need caching.

This design also enables future OAuth implementations where User can check token expiration before returning cached results.

Fixes #1395
  Run the specs added in this PR: `make test SPEC=spec/users_spec.cr`

### DIFF
--- a/src/lavinmq/auth/user.cr
+++ b/src/lavinmq/auth/user.cr
@@ -150,7 +150,7 @@ module LavinMQ
         result = cache[{vhost, name}]?
         return result unless result.nil?
 
-        return cache[{vhost, name}] = can_write?(vhost, name)
+        cache[{vhost, name}] = can_write?(vhost, name)
       end
 
       def can_write?(vhost : String, name : String) : Bool


src/lavinmq/amqp/client.cr
```diff
@@ -845,15 +845,9 @@ module LavinMQ
...
-      @acl_cache = Hash({String, String}, Bool).new
-
       private def start_publish(frame)
         cache_key = {@vhost.name, frame.exchange}
-        allowed = @acl_cache[cache_key]?
-        if allowed.nil?
-          allowed = @acl_cache[cache_key] = @user.can_write?(*cache_key)
-        end
-        if allowed
+        if @user.can_write?(*cache_key, cache_key)
```
spuun: Let `User` be responsible for creating the cache key.

src/lavinmq/auth/user.cr
```diff
@@ -130,9 +132,16 @@ module LavinMQ
         }
       end
 
-      def can_write?(vhost, name) : Bool
+      def can_write?(vhost : String, name : String, cache_key : CacheKey? = nil) : Bool
```
spuun: There is no need for the caller to know that there is a cache (it maybe even isn't). Keep the old method signature and construct the cache key in `can_write?`. 
This will also simplify the entire method.

src/lavinmq/auth/user.cr
```diff
@@ -10,12 +10,14 @@ module LavinMQ
...
       property tags, plain_text_password
       alias Permissions = NamedTuple(config: Regex, read: Regex, write: Regex)
+      alias CacheKey = Tuple(String, String)
 
       @name : String
       @permissions = Hash(String, Permissions).new
       @password : Auth::Password? = nil
       @plain_text_password : String?
       @tags = Array(Tag).new
+      @acl_cache = Hash(CacheKey, Bool).new
```
spuun: I'd maybe rename it to `@acl_write_cache` since it's only handling write.

spuun (commented):
Some comments

src/lavinmq/auth/user.cr
```diff
@@ -130,9 +132,15 @@ module LavinMQ
...
+      def can_write?(vhost : String, name : String) : Bool
+        cache_key = {vhost, name}
+        cached = @acl_write_cache[cache_key]?
+        return cached unless cached.nil?
+
         perm = permissions[vhost]?
-        perm ? perm_match?(perm[:write], name) : false
+        result = perm ? perm_match?(perm[:write], name) : false
+        @acl_write_cache[cache_key] = result
+        result
```
spuun: The entire `can_write?` can probably be simplified to
```crystal
      cache_key = {vhost, name}
      @acl_write_cache.put_if_absent(cache_key) do
        perm = permissions[vhost]?
        perm ? perm_match?(perm[:write], name) : false
      end
```

carlhoerberg:
I looked this in another PR and I think i removed all caching expect for writes.

carlhoerberg:
Read/config auth stuff is done so rarely, write (at publish) is done so often though. But i think the cache has been to GC sometimes too, to not leak memory

spuun:
> Read/config auth stuff is done so rarely, write (at publish) is done so often though. But i think the cache has been to GC sometimes too, to not leak memory

That's what we figured!

src/lavinmq/auth/user.cr
```diff
@@ -10,12 +10,14 @@ module LavinMQ
...
       property tags, plain_text_password
       alias Permissions = NamedTuple(config: Regex, read: Regex, write: Regex)
+      alias CacheKey = Tuple(String, String)
 
       @name : String
       @permissions = Hash(String, Permissions).new
       @password : Auth::Password? = nil
       @plain_text_password : String?
       @tags = Array(Tag).new
+      @acl_write_cache : Atomic(Hash(CacheKey, Bool)) = Atomic.new(Hash(CacheKey, Bool).new)
```
carlhoerberg: Use copy on write instead

src/lavinmq/auth/user.cr
```diff
@@ -130,7 +140,19 @@ module LavinMQ
...
+      def can_write?(vhost : String, name : String, cache : PermissionCache) : Bool
+        permission_revision = @permission_revision.lazy_get
+        if permission_revision != cache.revision
+          cache.clear
+          cache.revision = permission_revision
+        else
+          result = cache[{vhost, name}]?
+          return result unless result.nil?
+        end
+        return cache[{vhost, name}] = can_write?(vhost, name)
```
spuun: Maybe this is a little bit cleaner
```suggestion
        if permission_revision == cache.revision
          result = cache[{vhost, name}]?
          return result unless result.nil?
        end
        cache.clear
        cache.revision = permission_revision
        return cache[{vhost, name}] = can_write?(vhost, name)
```

spuun:
I think the PR description should be more descriptive for future code archeologists to understand the changes. (And right now it's kind of false, the cache is still in `Client`, bit the logic is in `User`.)

src/lavinmq/auth/user.cr
kickster97: Now it looks like we clear the cache on every "miss", even if the revision does not change.. I feel like we should only be clearing the cache if the revision changes otherwise we loose all the cached entries?

carlhoerberg:
a) why is the cache instance in client and not user? it's common to have have multiple connections using the same user. 
b) what's up with the revision? why not just clear the cache on ACL updates? it happens extreamly rarely and we can pay the price for the cache invalidation

spuun:
> a) why is the cache instance in client and not user? it's common to have have multiple connections using the same user. 

To only make one fiber (Client read loop) access the cache. That's why it once was moved from User to Client.

>b) what's up with the revision? why not just clear the cache on ACL updates? it happens extreamly rarely and we can pay the price for the cache invalidation

That would require locks, which we tried to avoid.

src/lavinmq/auth/user.cr
spuun: > Now it looks like we clear the cache on every "miss", even if the revision does not change.. I feel like we should only be clearing the cache if the revision changes otherwise we loose all the cached entries?

That's true. My mistake!

carlhoerberg:
ok, now i see! smart. maybe document the general thinking in the code. 

Another think i found doing other stuff, is that we probably should only allow permission changes through the `UserStore`, because it handles saving the changes to file. Instead as now if we change permissions then manually have to call `UserStore#save`. Possibly another PR.

src/lavinmq/auth/user.cr
```diff
@@ -130,7 +140,20 @@ module LavinMQ
...
-      def can_write?(vhost, name) : Bool
+      def can_write?(vhost : String, name : String, cache : PermissionCache) : Bool
+        permission_revision = @permission_revision.lazy_get
+        if permission_revision != cache.revision
+          cache.clear
+          cache.revision = permission_revision
+        end
+
+        result = cache[{vhost, name}]?
+        return result unless result.nil?
```
carlhoerberg:
```suggestion
        if result = cache[{vhost, name}]?
          return result
        end
```

src/lavinmq/auth/user.cr
```diff
@@ -130,7 +140,20 @@ module LavinMQ
...
+
+        cache[{vhost, name}] = can_write?(vhost, name)
```
carlhoerberg: or
```suggestion
        if result = cache[{vhost, name}]?
          result
        else
          cache[{vhost, name}] = can_write?(vhost, name)
        end
```

kickster97: Sure!

kickster97: but hmm no, this means if we have permissions cached and you send in faulty permissions, we unnessecarily go through the regex check. we want it to skip that check if the sent in permission does not match the cached permission, since when a user changes the permission we will clear and update the cache.

spuun: I just want to add that `cache[{vhost, name}]?` may return `false` which is a cached value, therefore it must be an explicit nil check.

PR #1445, opened 2025-11-09
Write meta data without buffer

carlhoerberg:
Register file after it has been written doesn't make sense, the data won't be replicated then. So only doing the replace_file now, which does the right thing.

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -330,15 +330,10 @@ module LavinMQ
       metafile = meta_file_name(wfile)
       @log.debug { "Write message segment meta file #{metafile}" }
       File.open(metafile, "w") do |f|
-        f.buffer_size = 4096
-        write_metadata(f, seg)
-        @replicator.try &.register_file(f)
+        f.sync = true
+        f.write_bytes @segment_msg_count[seg]
       end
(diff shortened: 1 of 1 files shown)


spuun:
Isn't register required for the file to be synced to a new follower?

carlhoerberg:
Nope, `register_file` should only be used on startup, for existing files that are not modified: https://github.com/cloudamqp/lavinmq/blob/05955daa8a99ff25cd25729859510157a28d7af7/src/lavinmq/clustering/server.cr#L51-L66

carlhoerberg:
`replace_file` does what `register_file` does, but also sends it to the followers

spuun (approved):

src/lavinmq/message_store.cr
```diff
@@ -330,17 +330,12 @@ module LavinMQ
       metafile = meta_file_name(wfile)
       @log.debug { "Write message segment meta file #{metafile}" }
       File.open(metafile, "w") do |f|
-        f.buffer_size = 4096
-        write_metadata(f, seg)
```
viktorerlingsson, line 334: Removing this will break metadata writing for streams, no?

carlhoerberg: Ah, good point

PR #1447, opened 2025-11-11
Bugfix: Only render exchange info once

spuun:
The exchange view rendered info multiple times, and some of the rendering code appends data e.g. arguments:
image

This PR will make sure info is only rendered once as in queue view.

Manual

### DIFF
--- a/static/js/exchange.js
+++ b/static/js/exchange.js
@@ -13,43 +13,46 @@ const chart = Chart.render('chart', 'msgs/s')
 document.title = exchange + ' | LavinMQ'
 
 const exchangeUrl = HTTP.url`api/exchanges/${vhost}/${exchange}`
-function updateExchange () {
+function updateExchange (all) {
   HTTP.request('GET', exchangeUrl).then(item => {
     Chart.update(chart, item.message_stats)
-    let features = ''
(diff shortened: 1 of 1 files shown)


static/js/exchange.js
```diff
@@ -13,43 +13,45 @@ const chart = Chart.render('chart', 'msgs/s')
...
+    if (all) {
+      let features = ''
+      features += item.durable ? ' D' : ''
+      features += item.auto_delete ? ' AD' : ''
+      features += item.internal ? ' I' : ''
+      features += item.arguments['x-delayed-exchange'] ? ' d' : ''
+      document.getElementById('e-features').textContent = features
+      document.getElementById('e-type').textContent = item.type
+      document.querySelector('#pagename-label').textContent = exchange + ' in virtual host ' + item.vhost
+      const argList = document.createElement('div')
```
snichme, line 28: Instead of having an `all` argument to this function we could instead do `argList.textContent = ''` here 

Arguments shouldn't change for an exchange, but this is more that the "all" param for this functions is a little unclear

spuun: I agree, but this is how it's done in other views 🙃 Trying to stay consistent 🤷

snichme: Ok, makes sense .

PR #1454, opened 2025-11-11
Bugfix: navigate between exchanges

### LINKED ISSUE #1402: Link not working for exchange to exchange bindings
**Describe the bug**

I have a exchange that has some queue and also some exchanges bound to it, for me it's an amq.topic exchange but I'm pretty sure this affects all exchanges.  In the Bindings list when clicking on a link to an exchange nothing happens expect the URL is updated in the browser but i'm not directed to the exchange. When I click on a queue in the bindings list it works. 

LavinMQ version: 2.4.4. 

**How to reproduce**

Create a exchange to exchange binding and click on the exhcange name in the bindings table in the source exchange.

snichme:

Fixes #1402 

Specs and manual.

Create a exchange to exchange binding, before this patch, the link updates the hash but does not navigate. 
Do the same with this bugfix applied

### DIFF
--- a/spec/frontend/exchange.spec.js
+++ b/spec/frontend/exchange.spec.js
@@ -1,7 +1,7 @@
 import * as helpers from './helpers.js'
-import { test, expect } from './fixtures.js';
+import { test, expect } from './fixtures.js'
 
-test.describe("exchange", _ => {
+test.describe('exchange', _ => {
   const pagePath = '/exchange#vhost=%2F&name=amq.topic'
   const apiPath = 'api/exchanges/%2F/amq.topic'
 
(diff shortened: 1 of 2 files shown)


spec/frontend/exchange.spec.js
```diff
@@ -21,4 +21,24 @@ test.describe("exchange", _ => {
...
+    const apiNewExchangeRequest = helpers.waitForPathRequest(page, `api/exchanges/%2F/${targetExchange}`)
+
+    // Change the hash to simulate clicking an exchange binding link
+    await page.evaluate((hash) => {
+      window.location.hash = hash
+    }, newHash)
+
+    // Verify the new exchange is loaded
+    await expect(apiNewExchangeRequest).toBeRequested()
+    await expect(page).toHaveURL(new RegExp(`exchange${newHash.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}`))
```
spuun: That replace is to escape regexp?

```suggestion
    await expect(page).toHaveURL(new RegExp(`exchange${RegExp.escape(newHash)}`))
```

PR #1456, opened 2025-11-11
deleted files should be closed directly when @clustering = Clustering::NoopServer

viktorerlingsson:
Previously `delete_file` would do some unnecessary work when deleting files when not in a cluster because` @clustering` would be `Clustering::NoopServer` and not `nil`. This changes the behavior so single nodes will just close the files directly. 

Manual


viktorerlingsson:
Not relevant because of #1457

PR #1461, opened 2025-11-12
Rename delay exchange queue

### LINKED ISSUE #1458: Rename the delayed exchanges' backing queue
Rename the backing queue from `amq.delayed.<exchange name>` to `amq.delayed-<exchange-name>` to follow other reserved name patterns such as `amq.ctag-<random>` and `amq.gen-<random>`.

Make sure to be backward compatible and pick up old queues

spuun:
Before this PR a delayed exchange's backing queue was named `amq.delayed.<exchange name>` which doesn't follow other reserved entities naming pattern.

This PR will change the name to be `amq.delayed-<exchange name>`. Existing queues won't be renamed/migrated to enable smooth downgrades.

Fixes #1458

Run specs

### DIFF
--- a/src/lavinmq/amqp/queue/delayed_exchange_queue.cr
+++ b/src/lavinmq/amqp/queue/delayed_exchange_queue.cr
@@ -42,9 +42,6 @@ module LavinMQ::AMQP
       DelayedMessageStore.new(data_dir, replicator, durable?, metadata: @metadata)
     end
 
-    private def migrate_data_dir
-    end
-
     private def expire_at(msg : BytesMessage) : Int64?
       msg.timestamp + (msg.delay || 0u32)
     end


src/lavinmq/amqp/queue/delayed_exchange_queue.cr
```diff
@@ -16,6 +42,9 @@ module LavinMQ::AMQP
       DelayedMessageStore.new(data_dir, replicator, durable?, metadata: @metadata)
     end
 
+    private def migrate_data_dir
+    end
+
```
viktorerlingsson: This shouldn't be here, right?

spuun: Ah, no! I was thinking along those terms at first!

PR #1467, opened 2025-11-12
Multi-threading: One ExecutionContext per VHost 

carlhoerberg:
### Single-threaded ExecutionContext per VHost
- Each VHost gets its own single-threaded ExecutionContext for fiber isolation
- Client connections are moved to their VHost's ExecutionContext after connecting
- All VHost operations (queue management, client read loops, policies) run in VHost's EC
- Provides isolation between VHosts while keeping each VHost single-threaded

### Multi-threaded Clustering
- Clustering server uses a multi-threaded ExecutionContext
- Each cluster client connection gets its own thread
- Separates cluster replication workload from user-facing VHost operations

### Benefits
- Thread isolation prevents cross-VHost interference
- Single-threaded VHosts avoid complex synchronization within a VHost
- Multi-threaded clustering maximizes replication throughput
- Clear execution context boundaries improve debugging and performance analysis

### Safe Concurrent Access
- Queue: Atomic @closed variable using Atomic(Bool)
- UserStore: Mutex protection for user/permission operations
- VHost: Thread-safe policy and parameter management
- MessageStore: Proper closed state tracking for safe shutdown

### Safe Shutdown Handling
- Check closed state before writing to definitions file
- Don't expire or dead-letter messages if queue/vhost is closed
- Don't raise when acking/rejecting after queue deletion
- Queue expiration calls vhost.delete_queue instead of direct close
- Removed @deleted field, use closed? method consistently

### Fixed 7 Flaky Tests
1. Federation test - Handle race between consumer unsubscribe and message delivery
2. Stream queue test - Wait for async policy application
3. Shovel test - Reliable under load
4. Queue expiration tests (2) - Polling instead of fixed sleeps
5. MQTT unsubscribe test - Time for server to close connection
6. HTTP API test - Synchronize before assertions

### Test Robustness Patterns Applied
- Replace fixed sleeps with wait_for/should_eventually polling
- Use realistic timeouts (50ms+) instead of 1-5ms
- Relax strict timing assertions (correctness over precision)
- Handle race conditions by accepting multiple valid outcomes
- Synchronize on observable state changes

### Key Test Changes
- Queue expiration: 1ms→50ms timeout, poll for deletion
- Timing tests: Strict ranges (±50ms)→upper bounds (<500ms)
- Federation: Handle both race outcomes gracefully
- Stream policies: Wait for spawned policy fibers
- Message availability: Wait before API calls

These changes establish clear execution context boundaries while making LavinMQ more robust and the test suite reliable under load.

### DIFF
--- a/src/lavinmq/auth/user_store.cr
+++ b/src/lavinmq/auth/user_store.cr
@@ -51,6 +51,7 @@ module LavinMQ
       end
 
       def each_value : Iterator(User)
+        # Hash#values returns a new array, safe to iterate outside lock
         @lock.synchronize do
           @users.values.each
         end
--- a/src/lavinmq/vhost.cr
+++ b/src/lavinmq/vhost.cr
@@ -444,9 +444,10 @@ module LavinMQ
       end
       close_done.close
       # then force close the remaining (close tcp socket)
-      @connections_lock.synchronize do
-        @connections.each &.force_close
-      end
+      # Copy connections before releasing lock to avoid deadlock
+      # when force_close triggers callbacks that call rm_connection
+      connections = @connections_lock.synchronize { @connections.dup }
+      connections.each &.force_close
       Fiber.yield # yield so that Client read_loops can shutdown
       @queues.each_value &.close
       Fiber.yield


src/lavinmq/auth/user_store.cr
```diff
@@ -14,85 +14,130 @@ module LavinMQ
...
+            yield v
+          end
+        end
+      end
+
+      def each_value : Iterator(User)
+        @lock.synchronize do
+          @users.values.each
         end
       end
```
spuun, line 58: As Claude says, this isn't safe. Either rework code that depends on this, or maybe add some kind of `LockingIterator` that wraps an iterator and locks on first `next` and unlocks on `Stop`: `@users.values.each.locked(@lock)`

carlhoerberg: Absolutely. Cool idea with the iterator, but i think just a dup is better, we dont know how long the lock might be held for, potentially a long time if it's a very slow http client that lists the vhosts.

carlhoerberg: Ah, wait, `Hash#values` already creates a new Array on every call:
https://github.com/crystal-lang/crystal/blob/31039a545/src/hash.cr#L1516
https://github.com/crystal-lang/crystal/blob/31039a545/src/hash.cr#L2104

carlhoerberg:
```
crystal eval 'h = {"a" => 1, "b" => 2}; v1 = h.values; v2 = h.values; puts v1.object_id; puts v2.object_id; puts v1.object_id == v2.object_id'
``` 

``` 
139890634682272
139890634682048
false
```

PR #1468, opened 2025-11-12
Thread safe MessageStore

carlhoerberg:
Moved msg_store_lock into MessageStore and updated dependent classes (AMQP::Queue, AMQP::Stream, MQTT::Session) to use the now thread-safe MessageStore methods without external synchronization.

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -26,7 +26,7 @@ module LavinMQ
     getter bytesize = 0u64
     getter size = 0u32
     getter empty = BoolChannel.new(true)
-    @lock = Mutex.new(:reentrant)
+    @lock = Mutex.new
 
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -499,31 +492,27 @@ module LavinMQ::AMQP
...
-            expire_msg(env, :maxlen)
-            counter &+= 1
-            if counter >= 16 * 1024
-              Fiber.yield
-              counter = 0
-            end
+        while @msg_store.size > ml
+          env = @msg_store.shift? || break
+          @log.debug { "Overflow drop head sp=#{env.segment_position}" }
+          expire_msg(env, :maxlen)
```
viktorerlingsson, line 498: Couldn't another thread do stuff with the msg_store in between the `shift?` and `expire_msg` here now, possibly causing problems? (Or between checking `@msg_store.size > ml` and `shift?`)

PR #1477, opened 2025-11-14
Allow max_packet_size for MQTT connections

snichme:

Introduce a configurable max payload size for mqtt packets

### DIFF
--- a/src/lavinmq/mqtt/connection_factory.cr
+++ b/src/lavinmq/mqtt/connection_factory.cr
@@ -20,15 +20,15 @@ module LavinMQ
         metadata = ::Log::Metadata.build({address: connection_info.remote_address.to_s})
         logger = Logger.new(Log, metadata)
         begin
-          io = MQTT::IO.new(socket)
-          if packet = Packet.from_io(socket).as?(Connect)
+          io = MQTT::IO.new(socket, @config.mqtt_max_packet_size)
+          if packet = io.read_packet.as?(Connect)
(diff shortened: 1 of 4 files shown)


shard.lock
```diff
@@ -18,7 +18,7 @@ shards:
 
   mqtt-protocol:
     git: https://github.com/84codes/mqtt-protocol.cr.git
-    version: 0.2.0+git.commit.5cf504b32313b3b63eecde22e2a136b9b9935b8b
+    version: 0.2.0+git.commit.739428e1cd4ea9dab1bb6cf101fb18ea54a099e5
```
spuun: Maybe we should have updated the version of mqtt-protocol to 0.3.0? Maybe start using tags too...

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -20,15 +20,15 @@ module LavinMQ
         metadata = ::Log::Metadata.build({address: connection_info.remote_address.to_s})
         logger = Logger.new(Log, metadata)
         begin
-          io = MQTT::IO.new(socket)
-          if packet = Packet.from_io(socket).as?(Connect)
+          io = MQTT::IO.new(socket, max_frame_size = @config.mqtt_max_packet_size)
```
spuun:
```suggestion
          io = MQTT::IO.new(socket, @config.mqtt_max_packet_size)
```

PR #1479, opened 2025-11-14
Stream: Store offset only for first message per segment

viktorerlingsson:
Optimize stream message storage by only storing offsets for the first message in each segment  instead of every message. The offset for subsequent messages is calculated on-demand based on the  segment's first offset and message position in the segment. 

  - Removes `x-stream-offset` header from messages at write time, reducing disk I/O and message size
  - Calculates offsets dynamically when consuming messages or looking up positions

Increases write-performance by around ~30% with small messages in my tests. Doesn't have a big impact on read performance. 

Works with streams created before this change as well, but messages published with the changes in this PR can not be read by older lavinmq versions. 

Should be well covered by existing specs.

### DIFF
--- a/src/lavinmq/amqp/stream/stream_message_store.cr
+++ b/src/lavinmq/amqp/stream/stream_message_store.cr
@@ -383,14 +383,14 @@ module LavinMQ::AMQP
 
       # Validate and fix (possibly) incorrect offsets from existing metadata
       @segment_first_offset[seg] = if seg == 1u32
-                         1i64 # First segment always starts at offset 1
-                       elsif prev_first = @segment_first_offset[seg - 1]?
-                         # Calculate based on previous segment
-                         prev_count = @segment_msg_count[seg - 1]? || 0i64
-                         prev_first + prev_count
-                       else
-                         stored_offset # No previous segment info, use stored value
-                       end
+                                     1i64 # First segment always starts at offset 1
+                                   elsif prev_first = @segment_first_offset[seg - 1]?
+                                     # Calculate based on previous segment
+                                     prev_count = @segment_msg_count[seg - 1]? || 0i64
+                                     prev_first + prev_count
+                                   else
+                                     stored_offset # No previous segment info, use stored value
+                                   end
     end
 
     class OffsetError < Exception


src/lavinmq/amqp/stream/stream_message_store.cr
carlhoerberg, line 13: How about this for clarity? 
```suggestion
@segment_last_ts = Hash(UInt32, Int64).new(0i64) # used for max-age
@segment_first_offset = Hash(UInt32, Int64).new          # segment_id => offset of first msg
@segment_first_ts = Hash(UInt32, Int64).new       # segment_id => ts of first msg
```

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -86,8 +80,15 @@ module LavinMQ::AMQP
     private def offset_at(seg, pos, retried = false) : Tuple(Int64, UInt32, UInt32)
       return {@last_offset, seg, pos} if @size.zero?
       mfile = @segments[seg]
-      msg = BytesMessage.from_bytes(mfile.to_slice + pos)
-      offset = offset_from_headers(msg.properties.headers)
+      offset = @offset_index[seg]
+      mfile.pos = 4
+      loop do
+        BytesMessage.skip(mfile)
+        break if mfile.pos >= pos
```
carlhoerberg: no need to read the last message (eg. looking for the offset of message as `pos` 4). 
```suggestion
      while mfile.pos < pos
        BytesMessage.skip(mfile)
```

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -86,8 +80,15 @@ module LavinMQ::AMQP
...
+        offset += 1
+      rescue IO::Error
+        break
```
carlhoerberg: shouldn't really happen, we haven't really found the offset if this happens. 

```suggestion
```

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -279,18 +285,19 @@ module LavinMQ::AMQP
 
     def push(msg) : SegmentPosition
       raise ClosedError.new if @closed
-      msg.properties.headers = add_offset_header(msg.properties.headers, @last_offset += 1)
+      @last_offset += 1
       sp = write_to_disk(msg)
       @bytesize += sp.bytesize
       @size += 1
       @segment_last_ts[sp.segment] = msg.timestamp
+      @offset_index[sp.segment] = @last_offset if (@offset_index[sp.segment]? || 0i64).zero?
```
carlhoerberg:
```suggestion
      @offset_index[sp.segment] = @last_offset unless @offset_index.has_key?(sp.segment)
```

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -279,18 +285,19 @@ module LavinMQ::AMQP
...
       sp
     end
 
     private def open_new_segment(next_msg_size = 0) : MFile
       super.tap do
         drop_overflow
-        @offset_index[@segments.last_key] = @last_offset + 1
+        @offset_index[@segments.last_key] = @last_offset
```
carlhoerberg:
```suggestion
        @offset_index[@segments.last_key] = @last_offset unless @last_offset.zero?
```

src/lavinmq/amqp/stream/stream_message_store.cr
carlhoerberg: What kind of `IndexError` can happen here? And is this the correct remedy? Not `@last_offset + 1`? Or just leave it empty?

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -101,24 +102,26 @@ module LavinMQ::AMQP
...
       pos = 4u32
-      msg_offset = 0i64
+      msg_offset = @offset_index[segment] || 0i64
       loop do
         rfile = @segments[segment]?
         if rfile.nil? || pos == rfile.size
           if segment = @segments.each_key.find { |sid| sid > segment }
             rfile = @segments[segment]
-            pos = 4_u32
+            rfile.pos = 4u32
```
carlhoerberg: We're using a local `pos` variable? 
```suggestion
```

src/lavinmq/amqp/stream/stream_message_store.cr
viktorerlingsson: I think that only happens if the msg store is empty and we try to get the first message here. We can handle that with `if @empty` instead of rescue  (and it should be `@last_offset + 1` 👍).

viktorerlingsson: Yes, sounds good! 👍

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -86,8 +80,12 @@ module LavinMQ::AMQP
     private def offset_at(seg, pos, retried = false) : Tuple(Int64, UInt32, UInt32)
       return {@last_offset, seg, pos} if @size.zero?
       mfile = @segments[seg]
-      msg = BytesMessage.from_bytes(mfile.to_slice + pos)
-      offset = offset_from_headers(msg.properties.headers)
+      offset = @segment_first_offset[seg]
+      mfile.pos = 4
```
carlhoerberg, line 84: This is not introduced, or only a problem here, but using the internal MFile pos wont be thread safe. Doing a `.dup` of the MFile is one way to solve it, there are probably more efficent ways.

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -230,7 +229,7 @@ module LavinMQ::AMQP
     def shift?(consumer : AMQP::StreamConsumer) : Envelope?
       raise ClosedError.new if @closed
 
-      if env = shift_requeued(consumer.requeued)
+      if env = shift_requeued(consumer)
```
carlhoerberg:
```suggestion
      if env = shift_requeued(consumer.requeued)
```

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -252,11 +252,13 @@ module LavinMQ::AMQP
       end
     end
 
-    private def shift_requeued(requeued) : Envelope?
-      while sp = requeued.shift?
+    private def shift_requeued(consumer) : Envelope?
+      while sp = consumer.requeued.shift?
```
carlhoerberg: Shouldn't have to add cyclomatic complexity, right? 
```suggestion
    private def shift_requeued(requeued) : Envelope?
      while sp = requeued.shift?
```

carlhoerberg (commented):
How backwards compatible is this? Looks like the new metadata is not added to existing segments.

carlhoerberg:
I think the offset will be different for existing messages in existing stream queues.

viktorerlingsson:
> How backwards compatible is this? Looks like the new metadata is not added to existing segments.

It should work with existing streams, but I did find a bug where the last segment (with no metadata file) would not get the correct value because we used `@empty` instead of `empty?`. Fixed in the latest commit. 
(Also, downgrading to an earlier version after writing messages with this version will not work)

>I think the offset will be different for existing messages in existing stream queues. 

Yeah, seems to be off by one, I think it's a bug in existing code. We seem to write the wrong offsets to the metadata files. First segment starts with 0 according to metadata, but should start with 1. And following segments starts one above the offset of the first message in that segment. Made a fix in `read_extra_metadata_fields` that corrects those numbers. (And I think we are writing the correct number to metadata files with the other changes in this branch)

PR #1482, opened 2025-11-18
Ascii art

carlhoerberg:
Display the LavinMQ logo as ascii art on boot


snichme:
It's a good looking logo but I think there should also be a config option to not show this. 
It can default to true (show it) but if you want to disable it you can

carlhoerberg:
The logo is not shown if logging to journald or file, only for console output.

PR #1172, opened 2025-06-25
Oauth2 Support

kickster97:
### WHAT is this pull request doing?
Add OAuth2/OpenID Connect authentication to LavinMQ. Clients can now authenticate using JWT tokens (RS256) as the password field. The authenticator validates tokens against the OAuth provider's JWKS endpoint, checks issuer/audience/expiration, and maps scopes to LavinMQ permissions. Supports token refresh via UpdateSecret frame without reconnecting. JWKS public keys are cached with configurable TTL.

Refactors the user system into an abstract User base class with LocalUser (password-based) and OAuthUser (token-based with expiration tracking) implementations. Authentication chain tries local auth first, then OAuth.

fixes #857

### HOW can this pull request be tested?
specs to come once we are satisfied with the general logic and structure

PR #1483, opened 2025-11-18
Rename BasicAuthenticator to LocalAuthenticator

kickster97:
Renames the BasicAuthenticator to LocalAuthenticator. 
`Local` better describes the nature of the authentication. ([See comment](https://github.com/cloudamqp/lavinmq/pull/1172#discussion_r2532344571))

This makes the diff in #1172 slightly smaller

run specs, no functionality should be changed.

### DIFF
--- a/spec/auth_spec.cr
+++ b/spec/auth_spec.cr
@@ -5,7 +5,7 @@ require "../src/lavinmq/auth/authenticators/local"
 describe LavinMQ::Auth::Chain do
   it "Creates a default authentication chain if not configured" do
     with_amqp_server do |s|
-      chain = LavinMQ::Auth::Chain.create(s.@config, s.@users)
+      chain = LavinMQ::Auth::Chain.create(s.@users)
       chain.@backends.should be_a Array(LavinMQ::Auth::Authenticator)
       chain.@backends.size.should eq 1
     end
@@ -13,7 +13,7 @@ describe LavinMQ::Auth::Chain do
 
   it "Successfully authenticates and returns a local user" do
     with_amqp_server do |s|
-      chain = LavinMQ::Auth::Chain.create(s.@config, s.@users)
+      chain = LavinMQ::Auth::Chain.create(s.@users)
       user = chain.authenticate("guest", "guest")
       user.should_not be_nil
     end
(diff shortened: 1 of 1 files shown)


src/lavinmq/config.cr
```diff
@@ -73,7 +73,7 @@ module LavinMQ
     property default_consumer_prefetch = UInt16::MAX
     property yield_each_received_bytes = 131_072    # max number of bytes to read from a client connection without letting other tasks in the server do any work
     property yield_each_delivered_bytes = 1_048_576 # max number of bytes sent to a client without tending to other tasks in the server
-    property auth_backends : Array(String) = ["basic"]
+    property auth_backends : Array(String) = ["local"]
```
snichme, line 76: `auth_backends` has that been possible to change before? Like from cli args on config file? 
If so, it means that someone might have set basic as value for this property and changing this breaks backwards compatibility. 

If it has not been possible to change before, that this is fine

kickster97: No this value is only set here. It is not parsed by any of the parsers meaning it will not be picked up from a config.ini, and there is no cli argument for it.

src/lavinmq/auth/chain.cr
```diff
@@ -14,12 +14,12 @@ module LavinMQ
         backends = config.auth_backends
         authenticators = Array(Authenticator).new
         if backends.nil? || backends.empty?
```
carlhoerberg: it can't be nil
```suggestion
        if backends.empty?
```

src/lavinmq/auth/chain.cr
```diff
@@ -14,12 +14,12 @@ module LavinMQ
...
-          authenticators << BasicAuthenticator.new(users)
+          authenticators << LocalAuthenticator.new(users)
         else
           backends.each do |backend|
             case backend
-            when "basic"
-              authenticators << BasicAuthenticator.new(users)
+            when "local"
```
carlhoerberg: should it be possible to disable the local authenticator? or change the order of it?

kickster97: No probably should not be.. In #1172 the local authenticator will always try first. if we add more authenticators, then i think we should be able to order those. 

The check for a local user is so small so there is no big cost of always trying it first

PR #1484, opened 2025-11-19
Memory pressure monitor

carlhoerberg:
SystemD can notify processes when it's under memory pressure, it gives the process a chance to recover memory.

The integration is done here: https://github.com/84codes/systemd.cr/blob/main/src/memory_pressure.cr

Background: https://systemd.io/MEMORY_PRESSURE/


carlhoerberg:
Should try to do more when under pressure, for instance compact all hashes/arrays.

PR #1487, opened 2025-11-19
Improved default config

carlhoerberg:
Cleaned up config that is deliviered with deb/rpm packages.

Correct default config values in `--help` banner. 

Enable amqps/mqtts/https by default if tls_cert/key is supplied (before the port also had to be configured, not it has to explicilty disabled with `-1`)

### DIFF
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -1,24 +1,29 @@
 [main]
 data_dir = /var/lib/lavinmq
-default_user_only_loopback = true
 log_level = info
-;log_file = /var/log/lavinmq.log
-;socket_buffer_size = 16384
+;default_user_only_loopback = true
 ;tls_cert = /etc/lavinmq/cert.pem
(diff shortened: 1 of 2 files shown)


carlhoerberg:
New default deb/rpm config: 
```ini
[main]
data_dir = /var/lib/lavinmq
log_level = info
;default_user_only_loopback = true
;tls_cert = /etc/lavinmq/cert.pem
;tls_key = /etc/lavinmq/key.pem

[mgmt]
bind = ::
;port = 15672
;tls_port = 15671

[amqp]
bind = ::
;port = 5672
;tls_port = 5671

[mqtt]
bind = ::
;default_vhost = /
;port = 1883
;tls_port = 8883

[clustering]
enabled = false
bind = ::
;port = 5679
;advertised_uri = tcp://hostname.local:5679
;etcd_endpoints = localhost:2379
```

PR #1489, opened 2025-11-19
Overridden effective arguments not effective

spuun:
If an e.g. a queue argument is overridden by a policy it shouldn't be marked as active in the argument list,

This PR will not highlight override arguments:

image

`x-message-ttl` overridden, but not `x-max-length`

This PRs also prepares for "effective policy arguments", but visualizing that is for another PR.

Some specs added


carlhoerberg:
claude says that the same changes are missing for mqtt sessions, is that a valid concern or doesn't apply?

spuun:
> claude says that the same changes are missing for mqtt sessions, is that a valid concern or doesn't apply?

Doesn't apply because no arguments for mqtt sessions. 

My intention is to move "Effective policy definition" from the Details table to a list under the Arguments instead under a "effective policy arguments" heading (or similar).

src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -39,16 +39,30 @@ module LavinMQ
...
 
-      private def apply_policy_argument(key : String, value : JSON::Any)
+      private def apply_policy_argument(key : String, value : JSON::Any) : Bool
         case key
         when "alternate-exchange"
-          @alternate_exchange ||= value.as_s?
+          if @alternate_exchange.nil?
+            @alternate_exchange = value.as_s?
+            @effective_args.delete("x-alternate-exchange")
+            @effective_args.delete("alternate-exchange")
```
kickster97, line 48: OT: why do we accept both these arguments?

src/lavinmq/policy.cr
```diff
@@ -5,30 +5,41 @@ module LavinMQ
...
     end
 
     def clear_policy
       clear_policy_arguments
+      @effective_policy_args.clear
       @policy = nil
       @operator_policy = nil
     end
 
+    def after_policy_applied
```
kickster97, line 39: Maybe a comment here to explain why it is empty?

src/lavinmq/amqp/exchange/exchange.cr
spuun: Dunno.

PR #1491, opened 2025-11-20
Add shell command hooks for cluster leader transitions

carlhoerberg:
## Summary

Adds two new configuration options for executing shell commands during cluster leader election events:
- `clustering_on_leader_elected` - Executed when node becomes cluster leader
- `clustering_on_leader_lost` - Executed when node loses leadership

Commands are executed asynchronously in background fibers with full shell support (pipes, redirections, etc.). Output and errors are logged through LavinMQ's logging system.

## Implementation

**Modified files:**
- `src/lavinmq/config.cr` - Added config properties and parsers for CLI and INI
- `src/lavinmq/clustering/controller.cr` - Added shell execution logic and hooks

**Key features:**
- Commands execute with shell context (supports pipes, redirections)
- Asynchronous execution in background fibers to avoid blocking the controller
- Full error handling and logging (INFO on success, WARN/ERROR on failure)
- Empty/unset commands are silently ignored

## Configuration Examples

**Via command line:**
```bash
lavinmq --clustering \
  --clustering-on-leader-elected="echo 1 | sudo tee /sys/class/leds/PWR/brightness" \
  --clustering-on-leader-lost="echo 0 | sudo tee /sys/class/leds/PWR/brightness"
```

**Via config file:**
```ini
[clustering]
enabled = true
etcd_endpoints = localhost:2379
on_leader_elected = echo 1 | sudo tee /sys/class/leds/PWR/brightness
on_leader_lost = echo 0 | sudo tee /sys/class/leds/PWR/brightness
```

## Use Cases

- **Visual indicators**: Control LEDs on Raspberry Pi to show leader status
- **Service coordination**: Start/stop dependent services based on leader role
- **Monitoring integration**: Send notifications to external monitoring systems
- **Resource management**: Acquire/release shared resources during transitions
- **Logging/auditing**: Trigger external logging for leadership changes

## Test Plan

- [ ] Verify commands execute when node becomes leader
- [ ] Verify commands execute when node loses leadership (lease lost)
- [ ] Test with empty/unset configuration (should be no-op)
- [ ] Test command failure handling (non-zero exit code)
- [ ] Test command with shell features (pipes, redirections)
- [ ] Verify logging output for success and failure cases
- [ ] Test in multi-node cluster with leader failover

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -135,17 +135,19 @@ class LavinMQ::Clustering::Controller
   private def execute_shell_command(command : String, event : String)
     return if command.empty?
 
-    Log.info { "Executing #{event} hook: #{command}" }
+    Log.info { "Executing #{event} hook in background: #{command}" }
 
-    begin
-      status = Process.run(command, shell: true, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
-      if status.success?
-        Log.info { "#{event} hook completed successfully" }
-      else
-        Log.warn { "#{event} hook failed with exit code #{status.exit_code}" }
+    spawn name: "#{event} hook" do
+      begin
+        status = Process.run(command, shell: true, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
+        if status.success?
+          Log.info { "#{event} hook completed successfully" }
+        else
+          Log.warn { "#{event} hook failed with exit code #{status.exit_code}" }
+        end
+      rescue ex
+        Log.error(exception: ex) { "Failed to execute #{event} hook" }
       end
-    rescue ex
-      Log.error(exception: ex) { "Failed to execute #{event} hook" }
     end
   end
 


src/lavinmq/clustering/controller.cr
```diff
@@ -31,6 +31,7 @@ class LavinMQ::Clustering::Controller
     wait_to_be_insync
     @etcd.election_campaign("#{@config.clustering_etcd_prefix}/leader", @advertised_uri, lease: @id) # blocks until becoming leader
     @is_leader.set(true)
+    execute_shell_command(@config.clustering_on_leader_elected, "leader_elected")
```
lukas8219, line 34:
```suggestion
    execute_event_hook(@config.clustering_on_leader_elected, "leader_elected")
```

src/lavinmq/clustering/controller.cr
```diff
@@ -130,5 +132,22 @@ class LavinMQ::Clustering::Controller
     end
   end
 
+  private def execute_shell_command(command : String, event : String)
```
snichme, line 135: We should have some timeout here, a short timeout, so the external script doesn't impact lavinmq runtime

carlhoerberg: good point, or rather, run them in the background

PR #1492, opened 2025-11-20
Federation refactor/improvements

spuun:
In an attempt on fixing flaky upstream specs I think I discovered some things that could be improved in queue federation.

This PR will improve the QueueLink's handling of consumers on the downstream queue. Now an unsubscribe from the upstream will happen instead of an exception being raised that probably was poorly handled, causing at least specs to be flaky.

This PR also adds some spec assertions for Channel and BoolChannel in `spec/support/channel.cr` and `spec/support/bool_channel.cr` which we probably want to make use of in other specs.

Run specs

### DIFF
--- a/src/lavinmqctl.cr
+++ b/src/lavinmqctl.cr
@@ -894,7 +894,7 @@ class LavinMQCtl
 
     # Set default values if not provided
     @args["prefetch-count"] ||= JSON::Any.new(LavinMQ::Federation::DEFAULT_PREFETCH.to_i64)
-    @args["reconnect-delay"] ||= JSON::Any.new(LavinMQ::Federation::DEFAULT_RECONNECT_DELAY.to_i64)
+    @args["reconnect-delay"] ||= JSON::Any.new(LavinMQ::Federation::DEFAULT_RECONNECT_DELAY.total_seconds)
     @args["ack-mode"] ||= JSON::Any.new(LavinMQ::Federation::DEFAULT_ACK_MODE.to_s.underscore.gsub("_", "-"))
     @args["max-hops"] ||= JSON::Any.new(LavinMQ::Federation::DEFAULT_MAX_HOPS)
 


src/lavinmq/federation/link.cr
```diff
@@ -83,14 +86,24 @@ module LavinMQ
             start_link
             break if @state.terminating?
             state(State::Stopped)
-            sleep @upstream.reconnect_delay.seconds
+            select
+            when timeout @upstream.reconnect_delay.seconds
+            when @running.receive?
+              break
+            end
```
lukas8219: may be method `wait_for_upstream_connection_or_retry`?

src/lavinmq/federation/link.cr
```diff
@@ -184,14 +197,55 @@ module LavinMQ
...
 
+        @consumer_available = Channel(Nil).new
+
         def initialize(@upstream : Upstream, @federated_q : Queue, @upstream_q : String)
-          @federated_q.register_observer(self)
-          consumer_available if @federated_q.immediate_delivery?
           super(@upstream)
           @metadata = @metadata.extend({link: @federated_q.name})
+
+          spawn(monitor_consumers, name: "consumer monitor")
```
lukas8219: Would it be interesting to have the Federated Queue name on the logs?

src/lavinmq/bool_channel.cr
```diff
@@ -13,10 +13,19 @@ class BoolChannel
...
   end
 
+  def value : Bool
+    select
+    when when_true.receive
+      false
+    when when_false.receive
+      true
+    end
+  end
```
lukas8219, line 27: No need to handle timeouts?

carlhoerberg: timeout should not happen, so add a `else` statement that raises? 

and please explain in a comment why the results are inverted. `when_true` => `false` is not intuitive

src/lavinmq/bool_channel.cr
```diff
@@ -3,7 +3,7 @@
 class BoolChannel
   getter when_true = Channel(Nil).new
   getter when_false = Channel(Nil).new
-  @value = Channel(Bool).new
+  @value = ::Channel(Bool).new
```
carlhoerberg: not needed, same for all `Channel` => `::Channel` changes in this file.

src/lavinmq/bool_channel.cr
spuun: > and please explain in a comment why the results are inverted. `when_true` => `false` is not intuitive

Lol, it should not be `when_true` => `false`. This is a sloppy copy-paste error from when i moved this select statement from `QueueLink` to here. The "negation" should of course happen in `QueueLink`.

> No need to handle timeouts?

> timeout should not happen, so add a else statement that raises?

Could raise in an else statement for good measures.

src/lavinmq/federation/link.cr
spuun: Yes, I've been thinking about that!

src/lavinmq/bool_channel.cr
spuun: Actually, there is a timing problem so `else` may be triggered. Maybe a high timeout is better.

spuun: But if this property is used "often" somewhere we'll add and remove a lot of unnecessary events to the event loop, to prevent something that "never" can occur. Hm.

src/lavinmq/federation/link.cr
```diff
@@ -450,31 +432,32 @@ module LavinMQ
...
+                headers, received_from = received_from_header(msg)
+                received_from << ::AMQP::Client::Arguments.new({
+                  "uri"         => @scrubbed_uri,
+                  "exchange"    => @upstream_exchange,
+                  "redelivered" => msg.redelivered,
+                })
+                headers["x-received-from"] = received_from
+                msg.properties.headers = headers
+                federate(msg, @federated_ex.name, msg.routing_key)
+              else
```
lukas8219, line 467: The `unless` was a bit more idiomatic wasn't it?

spuun: I think I had a plan for why i changed... hm. 
I'll actually make this PR a draft, too much issues going on. The main goal was to fix flaky specs xD

src/lavinmq/vhost.cr
```diff
@@ -470,11 +470,15 @@ module LavinMQ
...
-        apply_policies
+      has_parameters = !@parameters.empty?
+      has_policies = !@policies.empty? || !@operator_policies.empty?
+      if has_parameters || has_policies
+        spawn(name: "Load parameters") do
+          sleep 10.milliseconds
+          next if @closed
+          apply_parameters
+          apply_policies
+        end
```
spuun, line 481: This is mostly stop annoying issues in specs.

src/lavinmq/federation/upstream.cr
```diff
@@ -11,7 +11,7 @@ module LavinMQ
 
     class Upstream
       DEFAULT_PREFETCH        = 1000_u16
-      DEFAULT_RECONNECT_DELAY =        1
+      DEFAULT_RECONNECT_DELAY = 1.second
```
spuun: This implies that `@reconnect_delay` also is a `Time::Span`. This makes it possible to pass a low value (e.g. `1.millisecond`)  to the constructor. This is useful in specs to get rid of unnecessary long delays.

spuun:
🤞 that upstream specs won't be flaky now

Federation also works better now. Less reconnects, it will instead unsubscribe and subscribe to upstream depending on downstream consumer state.

src/lavinmq/bool_channel.cr
```diff
@@ -13,10 +13,19 @@ class BoolChannel
...
   end
 
+  def value : Bool
+    select
+    when when_true.receive
+      true
+    when when_false.receive
+      false
+    end
+  end
```
carlhoerberg, line 27: Or is it more efficient to let `value` be a getter, and rename current `@value` to `@value_channel` and in `send_loop` assign the new `@value : Bool` on `@value_channel.receive`?

carlhoerberg:
@spuun let me know what you think of https://github.com/cloudamqp/lavinmq/pull/1492/commits/23507002a50f868a868ea7b1aa41da37de3ca43f

spuun:
> @spuun let me know what you think of [2350700](https://github.com/cloudamqp/lavinmq/commit/23507002a50f868a868ea7b1aa41da37de3ca43f)

For how `#value` is used in this PR it should be fine (but also the `Channel` solution since no hot path). Dunno if it could cause any problems in future usage?

PR #1497, opened 2025-11-22
Linting

carlhoerberg:
Basically `ameba --fix && ameba --gen-config`


carlhoerberg:
Oh, you're right, yes, used main/master of ameba. I went back to the one we specify in the shards file.

PR #1498, opened 2025-11-24
 Fix OpenAPI schema for bcrypt hashing_algorithm name

kickster97:
The OpenAPI schema listed simple algorithm names (Bcrypt, SHA256, etc.) as enum values for hashing_algorithm, but the API returns normalized values like lavinmq_bcrypt and rabbit_password_hashing_sha256. This causes Terraform to detect configuration drift when the inserted value didn't match the retrieved value.

Updated users.yaml to document the normalized values the API actually returns and explain the flexible input format it accepts. The API still accepts simplified algorithm names (case-insensitive)

Look at http api users endpoints.

### DIFF
--- a/static/docs/schemas/users.yaml
+++ b/static/docs/schemas/users.yaml
@@ -17,12 +17,17 @@ user-properties:
       - impersonator
     hashing_algorithm:
       type: string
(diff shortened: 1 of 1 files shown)


carlhoerberg:
isn't it the same for the others? prefix with either lavinmq_ or rabbitmq_?

PR #1499, opened 2025-11-24
Add TROUBLESHOOTING.md

### LINKED ISSUE #1446: Add support dump feature to lavinmqctl
When issues occur in production LavinMQ instances, the immediate priority is to get the system back up and running. However, by the time the system is stabilized and it's safe to investigate, crucial diagnostic information (state at the time of the incident, metrics, etc.) has often been lost or overwritten. 

It would be great if we can add a support-dump command to lavinmqctl that collects diagnostic information into a single archive file for troubleshooting purposes.

kickster97:
This markdown guides users through their troubleshooting process, and aims to help them solve problems with their lavinmq instances, or provide important details in bug reports. 

kind of Fixes #1446 

preview as markdown

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -264,9 +264,3 @@ These tools are useful for diagnosing LavinMQ issues. Refer to each tool's docum
 - **strace** (Linux) / **dtruss** (macOS) - Trace system calls for low-level debugging
 - **tcpdump** - Capture network packets for protocol-level debugging
 - **Wireshark** - Analyze captured AMQP/MQTT traffic
-
-## Getting Help
-
-- **GitHub Issues**: https://github.com/cloudamqp/lavinmq/issues
-- **Documentation**: https://lavinmq.com/documentation
-- **Community**: Start discussions on GitHub, or join our [slack channel](https://join.slack.com/t/lavinmq/shared_invite/zt-1kzmlb85z-J_7QcOjMRyhECLTZ7Gprhg)


TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+- **Garbage Collector statistics** - heap size, free bytes, unmapped bytes, GC cycle number
+- **Fiber list** - all running fibers and their states (useful for detecting deadlocks or stuck operations)
+- **Complete server state**:
+  - Users and their permissions
+  - VHosts and their resources
+  - Exchanges (size and capacity)
+  - Queues (name, durability, arguments, message counts, segments)
+  - Connections and channels
+  - Consumers and unacked messages
+  - ACL caches
```
carlhoerberg: We've removed the `Reporter` because it wasn't threadsafe.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+**Where to find the output:**
+
+The output goes to `stdout`, which means it's captured by `journald`:
```
carlhoerberg: journald only in linux, and only if systemd is managing it (which is probably our most common deployment). Other scenarios such as docker/kubernetes, OSX, BSD etc are possible. 
```suggestion
The output goes to `stdout`, which if the LavinMQ is managed by SystemD the output is captured by `journald`:
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+```sh
+# Follow logs and then trigger USR1
+journalctl -fu lavinmq
+# In another terminal: kill -USR1 $(pidof lavinmq)
```
carlhoerberg: Probably easer to not `follow` in this case
```suggestion
kill -USR1 $(pidof lavinmq)
journalctl -eu lavinmq
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+# View recent logs
+journalctl -u lavinmq --since "1 hour ago"
+
+# View logs from a specific time range
+journalctl -u lavinmq --since "2025-01-20 10:00:00" --until "2025-01-20 11:00:00"
+
+# Search for specific errors
+journalctl -u lavinmq | grep -i error
+journalctl -u lavinmq | grep -i exception
```
carlhoerberg: `journalctl` has a built in `--grep` flag, that is case insenstive if the argument is all lower case. 

```suggestion
journalctl -u lavinmq --grep "error|exception"
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+# Show only errors and warnings
+journalctl -u lavinmq -p err
+
+# Export logs to a file for sharing
+journalctl -u lavinmq --since "1 hour ago" > lavinmq_logs.txt
```
carlhoerberg: can show of another `--since` argument here:
```suggestion
journalctl -u lavinmq --since today > lavinmq_logs.txt
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+### Log Navigation
+
+When viewing logs with `journalctl`:
```
carlhoerberg: i'm not sure we should have a detailed journalctl guide here. it is a very common logging framework for us, but not for everybody (think kubernetes ppl)

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+**Before trying to reproduce a segfault:**
+
+```sh
+# Check if core dumps are currently enabled
+ulimit -c
+# If it shows 0, core dumps are disabled
+```
+
+**To capture a core dump when debugging LavinMQ:**
```
carlhoerberg: how do to do it if systemd is starting lavinmq?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+./bin/lavinmq -D /path/to/data
+
+# When it crashes, the core dump will be captured
+```
+
+If LavinMQ is running as a systemd service, temporarily enable core dumps:
+
+```sh
+# Set core dump limit for this session
+sudo sh -c 'ulimit -c 2097152 && systemctl restart lavinmq'
```
carlhoerberg: pretty sure this won't work, `ulimit` only applies to the current shell session

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+**After debugging:** Reset the limit to avoid filling disk:
+
+```sh
+ulimit -c 0
+```
```
carlhoerberg: not sure about this

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+```sh
+gzip lavinmq_core.dump
+```
+
+## System Resource Checks
+
+### Disk Space
+
+LavinMQ stores all messages on disk. A full disk will prevent LavinMQ from accepting new messages and can cause data corruption.
```
carlhoerberg: it should not cause data corruption
```suggestion
LavinMQ stores all messages on disk. A full disk will prevent LavinMQ from accepting new messages.
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+du -sh /var/lib/lavinmq/*
+
+# Check disk space
+df -h /var/lib/lavinmq
+
+# Find largest queue directories
+du -sh /var/lib/lavinmq/*/* | sort -h | tail -20
+
+# Check for large message segment files
+find /var/lib/lavinmq -name "msgs.*" -exec ls -lh {} \; | sort -k5 -h | tail -20
```
carlhoerberg: `sort` can't sort if size is human readable, 
```suggestion
find /var/lib/lavinmq -name "msgs.*" -exec ls -l {} \; | sort -k5 -n | tail -20
```

carlhoerberg: and `ls` can do recusrive listing so:
```suggestion
ls -lR /var/lib/lavinmq | sort -k5 -n | tail -20

```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+cat /proc/$(pidof lavinmq)/status | grep -i vmsize
+cat /proc/$(pidof lavinmq)/status | grep -i vmrss
+
+# Memory maps (can show memory leaks)
+sudo pmap -x $(pidof lavinmq)
+```
+
+### File Descriptors
+
+LavinMQ needs 1 file descriptor per connection + 2 per durable queue + overhead. Running out of FDs prevents accepting new connections or creating queues.
```
carlhoerberg: it does not use 2 per durable queue anymore

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+## Cluster Debugging
+
+For clustered LavinMQ instances:
+
+```sh
+# Check cluster status via API
+curl -u guest:guest http://localhost:15672/api/cluster
+
+# Check etcd ISR (In-Sync Replicas)
+sudo etcdctl get lavinmq/isr
```
carlhoerberg: you normally don't need `sudo` for `etcdctl`
```suggestion
etcdctl get lavinmq/isr
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+```sh
+# Check FD limit
+cat /proc/$(pidof lavinmq)/limits | grep "open files"
+
+# Count currently open FDs
+sudo ls -l /proc/$(pidof lavinmq)/fd | wc -l
+
+# Show what FDs are open
+sudo ls -l /proc/$(pidof lavinmq)/fd
+```
```
carlhoerberg: `lsof` is probably better, as it shows filenames etc too

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+# Or use massif-visualizer for GUI
+massif-visualizer massif.out.*
+```
+
+### Trace System Calls
+
+```sh
+# Trace system calls made by LavinMQ
+sudo strace -p $(pidof lavinmq) -f -e trace=open,close,read,write
```
carlhoerberg: we use many more, should at least add `mmap`, `munmap`, `mremap`

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+# Trace specific syscalls
+sudo strace -p $(pidof lavinmq) -e trace=network
+
+# Count syscalls
+sudo strace -c -p $(pidof lavinmq)
```
carlhoerberg: this works, but `perf trace` adds way less overhead for the same result

```suggestion
# Count syscalls
sudo perf trace --summary --pid $(pidof lavinmq)
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+### Network Packet Capture
+
+```sh
+# Capture AMQP traffic
+sudo tcpdump -i any -w lavinmq_traffic.pcap port 5672
```
carlhoerberg: isn't `-i any` implicit?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+# Capture and display
+sudo tcpdump -i any -A port 5672
+
+# Analyze with Wireshark
+wireshark lavinmq_traffic.pcap
```
carlhoerberg: Maybe also show how to use wireshark on a remote machine: 

```
ssh lavinmq.example.com sudo tcpdump -U -s0 'port 5672' -i ens5 -w - | wireshark -k -i -
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+### Queue Message Buildup
+
+```sh
+# Dump state to see consumer status
+kill -USR1 $(pidof lavinmq)
+journalctl -u lavinmq | grep -A 20 "Queue <queue-name>"
+
+# Check if consumers are active and prefetch settings
+# Check for unacked messages
+```
```
carlhoerberg: Drop this, not working without `Report`, which we have remved

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+```sh
+# Verify LavinMQ is listening
+sudo lsof -i :5672 -i :15672
+
+# Check for connection errors in logs
+journalctl -u lavinmq | grep -i "connection\|authentication"
+
+# Check firewall
+sudo iptables -L -n | grep 5672
```
carlhoerberg: Many use `nftables` nowadays. Or something completly different on OSX/BSD machines.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+### High Memory Usage
+
+```sh
+# Dump state to see what's using memory
+kill -USR1 $(pidof lavinmq)
+
+# Force garbage collection
+kill -USR2 $(pidof lavinmq)
```
carlhoerberg: wasn't this already described earlier?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+## Common Issues and Quick Fixes
+
+### LavinMQ Won't Start
+
+```sh
+# Check logs for errors
+journalctl -u lavinmq --since "5 minutes ago"
+
+# Check data directory lock
+ls -la /var/lib/lavinmq/.lock
```
carlhoerberg: what to check about it? I beliive we use a file lock on that file, but just the existence of it doesn't do any difference.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,604 @@
...
+
+### Additional Context
+
+- Configuration files (`/etc/lavinmq/lavinmq.ini`)
+- Queue/exchange definitions
+- Client library and version
+- Load characteristics (messages/sec, message sizes)
+- Whether issue is reproducible
+
+### Example Issue Template
```
carlhoerberg: probably shouldn't be here, but in the github issue template. if we like the template.

TROUBLESHOOTING.md
kickster97: If you don't specify `-i`, tcpdump listens on the first available network interface and you might not know if clients are connecting locally or remotely

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+killall -USR1 lavinmq
+```
+
+**What USR1 dumps:**
+- **Garbage Collector statistics** - heap size, free/unmapped bytes, allocated memory, GC cycle counts, and reclaimed bytes (useful for identifying memory leaks or unusual growth patterns)
+- **Fiber list** - all running fibers with their object IDs and names (useful for counting active fibers and detecting fiber leaks)
+
+**Where to find the output:**
+
+The output goes to `stdout`, which if the LavinMQ is managed by SystemD the output is captured by `journald`:
```
baelter:
```suggestion
The output goes to `stdout`. If LavinMQ is managed by SystemD this is captured by `journald`:
```

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+
+```sh
+# Follow logs in real-time
+journalctl --follow --unit lavinmq
+
+# View recent logs
+journalctl --unit lavinmq --since "1 hour ago"
+
+# Export logs to a file for sharing
+journalctl --unit lavinmq --since today > lavinmq_logs.txt
```
viktorerlingsson: IMO we shouldn't tell people how to use journalctl, I think it's enough with just `journalctl --unit lavinmq`

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+
+### Using perf top (Linux)
+
+`perf top` shows real-time CPU usage by function, helping identify performance bottlenecks and infinite loops.
+
+**Installation (not installed by default on Linux):**
+
+```sh
+sudo apt install linux-tools-common
+sudo apt install linux-tools-aws  # or linux-tools-generic depending on kernel
```
baelter: Maybe the other way around? aws is the special case.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+
+```sh
+# If running as launchd service, check system logs
+log stream --predicate 'process == "lavinmq"' --level debug
+
+# Or run LavinMQ in foreground with output redirection
+./lavinmq -D /path/to/data 2>&1 | tee lavinmq.log
+
+# View existing log file
+tail -f lavinmq.log
```
viktorerlingsson: is running as launchd service default? if so, I don't think we need this

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+sudo lsof -p $(pidof lavinmq) -i | awk '{print $8}' | sort | uniq -c
+
+# Show TCP connections
+sudo lsof -p $(pidof lavinmq) -i TCP
+
+# Show connections to specific port
+sudo lsof -p $(pidof lavinmq) -i :5672
+
+# Count total file descriptors
+sudo lsof -p $(pidof lavinmq) | wc -l
```
viktorerlingsson: IMO this is to much.

TROUBLESHOOTING.md
viktorerlingsson: Do we even need to tell people how to install this? IMO it's enough to say that you can use `perf` to view or record data (and what to look for when you do)

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,542 @@
...
+
+```sh
+# Check cluster status via API
+curl -u guest:guest http://localhost:15672/api/cluster
+
+# Check etcd ISR (In-Sync Replicas)
+etcdctl get lavinmq/isr
+```
+
+## Common Issues and Quick Fixes
```
viktorerlingsson: I think this section is great! Maybe it should be earlier in the file?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,266 @@
...
+# Check disk I/O
+iostat -x 1
+
+# Check for disk-related errors
+journalctl --unit lavinmq | grep --ignore-case "disk\|i/o"
+```
+
+## Gathering Debug Information
+
+### Dump Server State (USR1 Signal)
```
annieblomgren, line 74: coredumps end up in /var/lib/systemd/coredump/

coredumpctl --help
coredumpctl list

Issue #1486, opened 2025-11-19
Instance repeatedly losing etcd leadership


viktorerlingsson:
**Describe the bug**

A one node cluster repeatedly lost leadership and restarted. No super obvious reason as to why (no spikes in resource use, no obvious logs preceeding this error).
No error logs from etcd. 

```
2025-11-13 09:49:00.293 error lmq.etcd Lost leadership
2025-11-13 11:56:45.495 error lmq.etcd Lost leadership
2025-11-13 12:36:28.453 error lmq.etcd Lost leadership
2025-11-13 13:28:04.724 error lmq.etcd Lost leadership
2025-11-13 14:20:58.110 error lmq.etcd Lost leadership
2025-11-13 17:29:42.652 error lmq.etcd Lost leadership
2025-11-13 17:57:59.854 error lmq.etcd Lost leadership
2025-11-13 19:14:00.556 error lmq.etcd Lost leadership
2025-11-13 19:47:06.650 error lmq.etcd Lost leadership
2025-11-13 20:27:01.791 error lmq.etcd Lost leadership
```

```
2025-11-13 20:27:01.791	 Lost leadership (LavinMQ::Etcd::Error)
2025-11-13 20:27:01.791	 lmq.etcd Lost leadership
2025-11-13 20:27:01.791	 lmq.clustering.controller Lost cluster leadership
2025-11-13 20:27:01.791	   from ???
2025-11-13 20:27:01.791	   from /lib/aarch64-linux-gnu/libc.so.6 in '??'
2025-11-13 20:27:01.791	   from /usr/bin/lavinmq in 'GC_call_with_stack_base'
2025-11-13 20:27:01.791	   from /usr/bin/lavinmq in 'GC_inner_start_routine'
2025-11-13 20:27:01.791	   from usr/share/crystal/src/crystal/system/thread.cr:246:13 in '->'
2025-11-13 20:27:01.791	   from usr/share/crystal/src/fiber/execution_context/isolated.cr:202:13 in '->'
2025-11-13 20:27:01.791	   from usr/share/crystal/src/gc/boehm.cr:193:7 in '->'
2025-11-13 20:27:01.793	  from ???
2025-11-13 20:27:01.793	  from usr/share/crystal/src/fiber.cr:170:11 in 'run'
2025-11-13 20:27:01.793	  from usr/share/crystal/src/log/log.cr:41:3 in '->'
[...]

PR #1503, opened 2025-11-25
only rescue `Etcd::Error` in `keepalive_loop`

viktorerlingsson:
This is a step in trying to find the cause of the `Lost leadership` issues in #1486 
Don't rescue all errors in `keepalive_loop`, only `Etcd::Error`. 
Also log `ex.message` when leadership is lost. 

.

### DIFF
--- a/src/lavinmq/etcd/lease.cr
+++ b/src/lavinmq/etcd/lease.cr
@@ -33,8 +33,8 @@ module LavinMQ
           sleep (ttl * 0.7).seconds
           ttl = @etcd.lease_keepalive(@id)
         end
-      rescue ex
-        Log.error(exception: ex) { "Lost leadership" } unless @lost_leadership.closed?
+      rescue ex : Etcd::Error # only rescue etcd errors
+        Log.error(exception: ex) { "Lost leadership because of #{ex.message}" } unless @lost_leadership.closed?
       ensure
         @lost_leadership.close
       end


viktorerlingsson:
> So what happens with other exceptions? We still close the leadership channel

There _shouldn't_ be any other exceptions here, unless we've missed something in `Etcd`, and if that's the case then we should handle those as well IMO. 

And maybe the `ex.message` logging can give us a better idea of what causes leadership to be lost.

carlhoerberg:
Or, it's the log line that raises, because the log channel is closed

viktorerlingsson:
> Or, it's the log line that raises, because the log channel is closed

I don't think that's the problem. I think the log channel is _sometimes_ closed because we exit in `Clustering::Controller` while trying to log in another fiber. But I think that's a separate error that happens _after_ lavinmq loses leadership, not the other way around. See my latest comment here https://github.com/cloudamqp/lavinmq/issues/1486#issuecomment-3571953427
We always got the "Lost leadership" log, so the log channel can't be closed before that runs. 

For example, this _sometimes_ causes a `Channel::ClosedError`
```
require "log"

spawn do # this could be any fiber in LavinMQ
  loop do
    Log.info {"This is a log message."}
    Fiber.yield
  end
end

sleep 0.1.seconds
exit # this could be lavinmq exiting because of lost leadership
```

PR #1506, opened 2025-11-26
Add spec to verify #1500 (error truncating deleted file)

spuun:
This PR will add a spec to verify at least one case which #1500 fixes.

Also changed so most message store specs are run against both a durable and non-durable message store.

Run specs


spuun:
@viktorerlingsson i modified the spec file so most specs run against durable AND non-durable stores. It turns out the bug (#1500) happened in `#first?` and `#shift?` too.

PR #1511, opened 2025-11-27
Fix: Skip truncate on deleted files to prevent File::NotFoundError

baelter:
## Summary

Fixes a race condition where `MFile#truncate` attempts to truncate files that have already been deleted, causing `File::NotFoundError` and broker crashes.

## Problem

When processing high volumes of messages with rapid republishing (e.g., stress tests with 10k+ messages), LavinMQ 2.5.x encounters a race condition in the message store:

1. `MessageStore#open_new_segment` calls `@wfile.truncate(@wfile.size)` to truncate the old segment file
2. For transient queues, files are marked as deleted immediately after creation
3. The file may already be deleted when truncate is called, causing:
   ```
   File::NotFoundError: Error truncating file: '.../msgs.0000000001': 
   No such file or directory
   ```

This error occurs at:
- `src/lavinmq/mfile.cr:149` in 'truncate'
- `src/lavinmq/message_store.cr:314` in 'open_new_segment'

## Solution

Added a check to `MFile#truncate` to skip truncation if the file has been deleted:
```crystal
return if @deleted.get(:acquire) # don't truncate deleted files
```

This matches the existing pattern in `MFile#close` (line 119), which already performs this check before truncating.

## Testing

- Tested against the amqp-client.js test suite, specifically the "can republish in consume block without race condition" test
- The test publishes 10,000 messages with each message triggering 2 new publishes, creating high load
- Previously failed with connection resets due to broker crashes
- Now passes consistently

## Impact

- Prevents broker crashes during high-volume message processing
- Particularly affects transient/non-durable queues
- No functional changes to normal operation
- Defensive check that prevents attempting filesystem operations on deleted files

### DIFF
--- a/src/lavinmq/mfile.cr
+++ b/src/lavinmq/mfile.cr
@@ -131,6 +131,7 @@ class MFile < IO
     new_capacity = new_capacity.to_i64
     old_capacity = @capacity
     return if new_capacity == old_capacity # no change
+    return if @deleted.get(:acquire) # don't truncate deleted files
     raise ArgumentError.new("Cannot expand a MFile") if new_capacity > old_capacity
 
     # First truncate the file on disk


baelter:
Closing this PR as main branch already contains a more comprehensive fix for this issue (see src/lavinmq/mfile.cr:135-148). The fix includes both the deleted? check and graceful ENOENT error handling.

PR #1517, opened 2025-12-01
Enhance memory pressure handling with collection compaction

carlhoerberg:
## Summary

Enhances systemd memory pressure monitoring to proactively compact collections and reduce memory usage beyond just calling GC. This addresses the issue where Crystal collections (Array, Hash, Deque) never shrink their internal buffers, leading to wasted memory after temporary spikes.

## Problem

Currently, LavinMQ only calls `GC.collect` when systemd detects memory pressure. However, Crystal's collections retain their maximum capacity even after elements are removed. A queue that grew to 10,000 messages keeps that buffer size even when drained to 10 messages, wasting memory.

## Solution

Implements a cascading `compact_collections` pattern throughout the entire object hierarchy:

### Core Components

- **Array monkey patch** - Adds `trim_to_size` method for in-place buffer compaction
- **Hash/Deque** - Uses `.dup` reassignment to create right-sized copies
- **IO::Memory** - Assigns new instances when empty
- **MFiles** - Calls `dontneed` on memory-mapped files to release physical pages

### Compaction Cascade

```
Server
├── UserStore (users hash)
├── ParameterStore (parameters hash)
├── MQTT::Brokers (brokers hash)
│   └── Broker (clients hash)
│       └── Session (unacked hash)
└── VHostStore (vhosts hash)
    └── VHost (queues, exchanges, consumers hashes + connections array)
        ├── Queue (deliveries, consumers, unacked)
        │   └── MessageStore (deleted, segments, requeued + MFile dontneed)
        ├── Exchange (bindings - all 5 types)
        ├── Client (channels hash)
        │   └── Channel (unacked, tx arrays, consumers, IO::Memory)
        ├── ShovelStore (shovels hash)
        └── UpstreamStore (upstreams, upstream_sets hashes)
```

### Compaction Threshold

Collections are compacted when `capacity > size * 2`, providing aggressive memory reclamation during pressure events while avoiding excessive compaction overhead during normal operation.

## Changes

- Add `Array#trim_to_size` with `rewind` and `ensure_capacity` helpers
- Implement `compact_collections` in 21 files across the codebase
- Update launcher to run compaction before GC with timing logs
- Cascade pattern ensures all collections are compacted systematically

## Expected Benefits

- **10-50 MB memory savings** during typical pressure events
- **Faster GC cycles** - fewer large objects to scan
- **Better peak handling** - system recovers from memory spikes
- **Reduced swap usage** on memory-constrained systems

## Test Plan

- [ ] Build and run server: `make bin/lavinmq CRYSTAL_FLAGS=`
- [ ] Trigger memory pressure with high message throughput
- [ ] Verify compaction logs appear: "Memory pressure detected, running compaction"
- [ ] Monitor RSS reduction after compaction
- [ ] Run test suite: `make test`
- [ ] Check for performance regressions under normal load

## Notes

- Compaction is only triggered during systemd memory pressure events
- Uses 2x threshold to balance memory savings vs compaction overhead
- All collection types handled: Array, Hash, Deque, Set, IO::Memory
- MFile `dontneed` is safe - kernel can reclaim physical pages without breaking mmap

### DIFF
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -821,6 +821,23 @@ module LavinMQ
         @client.flush
       end
 
+      def compact_collections
+        # Deque uses dup
+        @unacked = @unacked.dup if @unacked.capacity > @unacked.size * 2
+
+        # Array uses trim_to_size
+        @tx_publishes.trim_to_size
+        @tx_acks.trim_to_size
+        @consumers.trim_to_size
+
+        # IO::Memory - reset if empty
+        @next_msg_body_tmp = IO::Memory.new if @next_msg_body_tmp.size == 0 && @next_msg_body_tmp.capacity > 0
+
+        # Clear temporary sets (they should be empty between publishes)
+        @visited.clear if @visited.size > 0
+        @found_queues.clear if @found_queues.size > 0
+      end
+
       class ClosedError < Error; end
     end
   end
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -896,6 +896,12 @@ module LavinMQ
           @socket.flush
         end
       end
+
+      def compact_collections
+        # Hash uses dup
+        @channels = @channels.dup if @channels.capacity > @channels.size * 2
+        @channels.each_value(&.compact_collections)
+      end
     end
   end
 end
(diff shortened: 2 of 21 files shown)


spuun:
How will this play with multi-threading?

carlhoerberg:
> How will this play with multi-threading?

as long as all structures are protected using `Sync::Shared`/`Mutex` etc it shouldn't be a problem.

PR #1519, opened 2025-12-01
Handle "lease not found" etcd error without infinite retry (#1518)

### LINKED ISSUE #1518: Infinite retry loop on "lease not found" etcd error
## Describe the bug

When etcd returns "lease not found", LavinMQ retries infinitely instead of failing fast.

Noticed on a 2.4.1 server, verified still an issue in `main`.

    2025-12-01 05:48:50.215 warn lmq.etcd Service Unavailable at localhost:2379, etcdserver: requested lease not found, retrying

## Root cause

In `src/lavinmq/etcd.cr`, the `with_tcp` method catches generic `Error` and retries. The "lease not found" error isn't handled specifically, so it triggers infinite retry.

https://github.com/cloudamqp/lavinmq/blob/f0f0bcecd50d679e808b6e84efde8df8a9b302ac/src/lavinmq/etcd.cr#L289-L306

## Expected behavior

Raise `LeaseNotFound` exception immediately. Controller catches it and exits (like `Lease::Lost`), allowing systemd to restart with a fresh lease.

jage:

When etcd returns "lease not found", LavinMQ retried indefinitely, spamming logs and preventing recovery. By failing fast and exiting, systemd can restart the process with a fresh lease.

The etcd API returns errors in two formats depending on endpoint:

- `{"error": "..."}` or `{"error": {"message": "..."}}`
- `{"code": N, "message": "..."}`

The existing code only checked the `error` field, so errors from endpoints like `/v3/election/campaign` were not handled.

Adds `LeaseNotFound` exception that propagates immediately (no retry), causing the clustering controller to exit with code 3.

Close #1518

Added specs, guess one could try to simulate this in a real cluster by manipulating etcd (wiping it), I have not done that.

### DIFF
--- a/spec/etcd_spec.cr
+++ b/spec/etcd_spec.cr
@@ -126,7 +126,7 @@ describe LavinMQ::Etcd do
       etcd = LavinMQ::Etcd.new(cluster.endpoints)
       expect_raises(LavinMQ::Etcd::LeaseNotFound) do
         etcd.election_campaign("test/leader", "node1", lease: 999999i64)
-     end
+      end
     end
   end
 


spec/etcd_spec.cr
```diff
@@ -120,6 +120,21 @@ describe LavinMQ::Etcd do
...
+    cluster.run do
+      etcd = LavinMQ::Etcd.new(cluster.endpoints)
+      begin
+        etcd.election_campaign("test/leader", "node1", lease: 999999i64)
+        fail "Expected exception to be raised"
+      rescue ex : LavinMQ::Etcd::LeaseNotFound
+        # expected
+      rescue ex
+        fail "Expected LeaseNotFound but got #{ex.class}: #{ex.message}"
+      end
```
spuun: What about
```suggestion
      expect_raises(LavinMQ::Etcd::LeaseNotFound) do
        etcd.election_campaign("test/leader", "node1", lease: 999999i64)
     end
```

src/lavinmq/etcd.cr
```diff
@@ -377,24 +379,22 @@ module LavinMQ
...
-        when "etcdserver: no leader"
-          raise NoLeader.new(error_msg)
-        when "etcdserver: lease already exists"
-          raise LeaseAlreadyExists.new
-        else
-          raise Error.new error_msg
-        end
+      error_msg = json.dig?("error", "message").try(&.as_s) ||
+                  json["error"]?.try(&.as_s?) ||
+                  json["message"]?.try(&.as_s)
```
snichme: Any risk of a response with `{"message": "success"}`  or similiar, that will in this case generate an exception even though it was success.

jage: Yes, let's see how we can tighten this.

jage: API reference https://etcd.io/docs/v3.6/dev-guide/api_reference_v3/

jage: https://grpc.io/docs/guides/status-codes/

jage: Looks like message isn't in the top object for other responses, but let's look at >0 status code as well.

snichme: https://etcd.io/docs/v3.6/dev-guide/api_concurrency_reference_v3/#service-election-serveretcdserverapiv3electionv3electionpbv3electionproto

Cannot find the specific response.

jage:
```Sh
$ curl -X POST http://localhost:2399/v3/election/campaign \
    -d '{"name":"dGVzdA==","value":"bm9kZTE=","lease":"999999"}'
{"code":2,"message":"etcdserver: requested lease not found"}
```

Put a research agent on it:

Why code 2 (UNKNOWN) instead of 5 (NOT_FOUND):

  The election server at https://github.com/etcd-io/etcd/blob/release-3.6/server/etcdserver/api/v3election/election.go#L38 is not a standard gRPC server - it uses the etcd client library internally (clientv3). When
  concurrency.NewSession() fails:

  1. The client's KeepAlive returns rpctypes.ErrLeaseNotFound (a client-side wrapper)
  2. This error bubbles up to the election server's Campaign method
  3. The error is returned to the gRPC-gateway
  4. Crucially: The error goes through https://github.com/grpc/grpc-go/blob/master/status/status.go#L44-L49 in grpc-gateway's https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/errors.go
  5. Since rpctypes.ErrLeaseNotFound (the client error) may not implement GRPCStatus() properly in this flow, status.Convert() falls back to codes.Unknown (2)

  The server-side code path through /v3/lease/ endpoints correctly maps lease.ErrLeaseNotFound → rpctypes.ErrGRPCLeaseNotFound (code 5) via
  https://github.com/etcd-io/etcd/blob/release-3.6/server/etcdserver/api/v3rpc/util.go#L70. But the election API uses the client library, which has a different error wrapping path.

  Conclusion: The code handles this correctly by matching on the message string "etcdserver: requested lease not found" rather than the unreliable gRPC code.

PR #1521, opened 2025-12-01
Enforce authorization for metrics on mgmt port

### LINKED ISSUE #1520: On mgmt port, only serve metrics on vhost the user has access to
**Describe the bug**

Regression introduced here: https://github.com/cloudamqp/lavinmq/commit/e3b4f4cb56bca125848e7aead06dae005c018aa2#diff-15e2d3e66d909518080eae38a9cc1bf342352885ce7e352a65d055d2042babb2R182 

A metrics_server was introduced to serve prometheus metrics without auth and to keep backwards compatibility the endpoint was kept on the mgmt port (default 15672) but user permission checks was removed since no user was available on the new port. 
The result is that if authed, you get all metrics for all vhosts even if no access was given.  

No data could be access, only names of vhosts and queues was visible for a users without permissions to that vhost.

snichme:

Fixes #1520

Specs

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -3,13 +3,13 @@ require "string_scanner"
 
 describe LavinMQ::HTTP::PrometheusController do
   describe "authentication" do
-    it "should allow guest access to metrics on metrics port" do
+    it "should allow unauthenticated access to metrics on metrics port" do
       with_metrics_server do |http, _|
         response = http.get("/metrics")
         response.status_code.should eq 200
       end
     end
-    it "should allow not guest access to metrics on mgmt port" do
+    it "should require authenticated access to metrics on mgmt port" do
       with_http_server do |http, _|
         response = HTTP::Client.get("http://#{http.addr}/metrics")
         response.status_code.should eq 401
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -176,16 +176,16 @@ module LavinMQ
     class PrometheusController < Controller
       include Prometheus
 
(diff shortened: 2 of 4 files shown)


spec/api/prometheus_spec.cr
```diff
@@ -2,6 +2,20 @@ require "../spec_helper"
...
 
 describe LavinMQ::HTTP::PrometheusController do
+  describe "authentication" do
+    it "should allow guest access to metrics on metrics port" do
+      with_metrics_server do |http, _|
+        response = http.get("/metrics")
+        response.status_code.should eq 200
+      end
+    end
+    it "should allow not guest access to metrics on mgmt port" do
```
carlhoerberg: what is "not guest", anonymous? without any credentials? 
```suggestion

    it "should allow anonymous access to metrics on mgmt port" do
```

src/lavinmq/http/controller/prometheus.cr
```diff
@@ -176,8 +176,18 @@ module LavinMQ
     class PrometheusController < Controller
       include Prometheus
 
+      def initialize(amqp_server : LavinMQ::Server, @allow_guest : Bool)
```
carlhoerberg: by "guest" you mean unauthenticated?  
```suggestion
      def initialize(amqp_server : LavinMQ::Server, @allow_unauthenticated : Bool)
```

or: 

```suggestion
      def initialize(amqp_server : LavinMQ::Server, @require_authentication : Bool)
```

carlhoerberg: i'm afraid "guest" could otherwise be interpreted as the default guest user.

snichme: agree, fixed now

PR #1524, opened 2025-12-02
Download chartjs only once in build

viktorerlingsson:
Currently we download chart.js twice, and check sha256sum twice. Since we run make multi-threaded, it is possible that the file is 
deleted before the step is completed, meaning the package can end up without the correct files. 

This changes that to only download chart.js once, to make sure the file can not be deleted between build steps

This also adds an exit if the sha256sum fails. 

run `make js`, verify output

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,17 +63,20 @@ lib: shard.yml shard.lock
 bin static/js/lib man1 static/js/lib/chunks:
 	mkdir -p $@
 
-static/js/lib/chart.js: | static/js/lib
+.PHONY: chart.js.tgz
+chart.js.tgz:
 	curl --fail --retry 5 -sL -o chart.js.tgz https://github.com/chartjs/Chart.js/releases/download/v4.0.1/chart.js-4.0.1.tgz && \
-		echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c - && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,17 +63,20 @@ lib: shard.yml shard.lock
...
-		echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c - && \
-		tar -zxOf chart.js.tgz package/dist/chart.js > $@ ; \
-		(rm -f chart.js.tgz && echo "removed chart.js.tgz")
+	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c - || (rm -f chart.js.tgz && exit 1)
 
-static/js/lib/chunks/helpers.segment.js: | static/js/lib/chunks
-	curl --fail --retry 5 -sL -o chart.js.tgz https://github.com/chartjs/Chart.js/releases/download/v4.0.1/chart.js-4.0.1.tgz && \
-		echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c - && \
-		tar -zxOf chart.js.tgz package/dist/chunks/helpers.segment.js > $@ ; \
-		(rm -f chart.js.tgz && echo "removed chart.js.tgz")
```
carlhoerberg, line 76: the solution is to use another download name

PR #1525, opened 2025-12-02
Use intermediate targets to download files on demand

spuun:
This is an improvement of JS dependency handling.

Each js file has its own intermediate target that will handle the downloading and checksum verification of the file. The real js target will have the intermediate target as a dependency, meaning that the file will only be downloaded when needed. 

Manually by running `make js` etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ bin static/js/lib man1 static/js/lib/chunks:
 .INTERMEDIATE: downloaded_chart.js.tgz
 downloaded_chart.js.tgz:
 	curl --fail --retry 5 -sL -o $@ https://github.com/chartjs/Chart.js/releases/download/v4.0.1/chart.js-4.0.1.tgz && \
-	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c -
+	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f $@" | sha256sum -c -
 
 .INTERMEDIATE: downloaded_luxon.js
 downloaded_luxon.js: 


Makefile
```diff
@@ -63,41 +63,51 @@ lib: shard.yml shard.lock
...
 	mkdir -p $@
 
-.PHONY: chart.js.tgz
-chart.js.tgz:
-	curl --fail --retry 5 -sL -o chart.js.tgz https://github.com/chartjs/Chart.js/releases/download/v4.0.1/chart.js-4.0.1.tgz && \
-	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c - || (rm -f chart.js.tgz && exit 1)
+.INTERMEDIATE: downloaded_chart.js.tgz
+downloaded_chart.js.tgz:
+	curl --fail --retry 5 -sL -o $@ https://github.com/chartjs/Chart.js/releases/download/v4.0.1/chart.js-4.0.1.tgz && \
+	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f chart.js.tgz" | sha256sum -c -
```
viktorerlingsson: Shouldn't use `chart.js.tgz` here, right? 
```suggestion
	echo "461dae2edc0eda7beeb16c7030ab630ab5129aedd3fc6de9a036f6dfe488556f $@" | sha256sum -c -
```

spuun: Oh!

carlhoerberg:
This doesnt make sense at all to me. Why are the files not targets as they used to and as Makefiles are supposed to work?

spuun:
> This doesnt make sense at all to me. Why are the files not targets as they used to and as Makefiles are supposed to work?

There was a few reasons (one being a race condition), but piping curl to tar in combination with `.DELETE_ON_ERROR`(as in #1527) is better. Totally missed that.

carlhoerberg:
i didn't know about `.INTERMEDIATE`, and on the phone it looked scary, but cool thing. But yes, i think `.DELETE_ON_ERROR` is even better.

PR #1526, opened 2025-12-02
Clear permissions on connected user if revoked/deleted

kickster97:
A publishing client can use stale permissions to publish if already connected with those permissions. 
This pr adds that permissions are actually revoked during the publish flow if you remove permissions for a user, or if you delete a user. 

try to publish with a user that you revoke permissions from or that you delete.

### DIFF
--- a/src/lavinmq/auth/user_store.cr
+++ b/src/lavinmq/auth/user_store.cr
@@ -72,14 +72,6 @@ module LavinMQ
         save!
       end
 
-      def clear_permissions(user : String)
-        if u = @users[user]?
-          u.permissions.clear
-          u.clear_permissions_cache
-          Log.info { "Cleared all permissions for user=#{user}" }
-        end
-      end
-
       def delete(name, save = true) : User?
         return if name == DIRECT_USER
         if user = @users.delete name


src/lavinmq/http/controller/users.cr
```diff
@@ -59,6 +59,7 @@ module LavinMQ
             unless u.as_s?
               bad_request(context, "Field 'users' must be array of user names")
             end
+            @amqp_server.users.clear_permissions(u.as_s)
             @amqp_server.users.delete(u.as_s, false)
```
spuun: Instead of remember to call `User#clear_permissions` before delete, why not call `User#clear_permissions_cache` cache from `UserStore#delete`? That would make `UserStore#clear_permissions` obsolete for now.

kickster97: I tested this, and when I delete a user the permissions are still set on a user if I try and use `clear_permissions_cache`, becasue that mehod does not actually clear the permissions.. 
If it should actually clear the permissions is another question, maybe? but maybe also permissions vs permissions cache is good to keep separate

Issue #1510, opened 2025-11-27
SSLKEYLOGFILE support


carlhoerberg:
**Is your feature request related to a problem? Please describe.**

To monitor/decrypt TLS traffic for debugging, the server needs to log the TLS session keys, describe here: https://datatracker.ietf.org/doc/html/draft-ietf-tls-keylogfile

You can then import the server private key and the SSLKEYLOGFILE into wireshark and view the decrypted TCP traffic. 

**Describe the solution you'd like**

Should be able to enable a config so that the server starts writing to the SSLKEYLOGFILE.

kickster97:
Fixed in #1531

PR #1531, opened 2025-12-03
 SSLKEYLOGFILE support 

carlhoerberg:
## Summary
When set, TLS session keys will be logged to a file in NSS Key Log Format, which can be used to decrypt TLS traffic with tools like Wireshark.

### Configuration options
- `tls_keylog_file` in `[main]` section - applies to all TLS contexts
- `SSLKEYLOGFILE` environment variable - fallback when config is not set
- Per-SNI-host configuration with `tls_keylog_file` and protocol-specific overrides (`amqp_tls_keylog_file`, `mqtt_tls_keylog_file`, `http_tls_keylog_file`)

### Implementation details
- Uses OpenSSL's `SSL_CTX_set_keylog_callback` (requires OpenSSL 1.1.1+ or LibreSSL 3.4.1+)
- Atomic writes to prevent interleaving when multiple contexts write to the same file
- SNI contexts independently configurable, with `SSLKEYLOGFILE` env as fallback

## Test plan
- [x] Added spec for basic keylog functionality
- [x] Added spec for concurrent SNI connections writing to same file without interleaving

### DIFF
--- a/spec/stdlib/openssl_keylog_spec.cr
+++ b/spec/stdlib/openssl_keylog_spec.cr
@@ -1,5 +1,6 @@
 require "spec"
 require "../../src/stdlib/openssl_keylog"
+require "../../src/stdlib/openssl_sni"
 
 describe OpenSSL::SSL::Context::Server do
   describe "#keylog_file=" do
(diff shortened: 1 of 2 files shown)


carlhoerberg:
Fixes #1510

PR #1537, opened 2025-12-07
Improved cluster replica tracking and management

### LINKED ISSUE #1425: OSR(out of sync replicas) for better nodes interface
Today we do not save what nodes have been connected or part of ISR before, so its not really part of the UI. You cant see if a node is disconnected, or if a node that should be part of the cluster is missing etc. 

save what nodes have gotten out of sync in OSR.

carlhoerberg:
## Summary

- Track replicas using individual etcd keys instead of comma-separated ISR list
- Show all known replicas in Nodes page with Connected/In-sync status columns  
- Add ability to forget disconnected replicas via UI and CLI
- Generate sequential cluster node IDs using atomic etcd transactions

### Use individual etcd keys for ISR tracking
Instead of maintaining a comma-separated list at a single "isr" key, each replica now has its own key at `{prefix}/replica/{id}/insync` with value "1" for in-sync and "0" for out-of-sync.

### Show all known replicas in nodes page
- Add `known_replicas` method to Clustering::Server that queries etcd
- Merge etcd replicas with connected followers in NodesController
- Add Connected and In-sync columns to followers table
- Show leader role in replicas table, sorted first

### Add forget button for disconnected replicas
- Allow removing disconnected replicas via Nodes page UI or `DELETE /api/nodes/:id`
- Prevent forgetting leader node
- Add `lavinmqctl forget_cluster_node <id>` CLI command

### Generate sequential cluster node IDs
- New nodes get sequential IDs (1, 2, 3, ...) instead of random IDs
- Uses atomic etcd transactions to ensure IDs don't collide
- Handles manually created nodes by checking for existing replica keys

## Test plan
- [x] Start a 3-node cluster and verify nodes get IDs 1, 2, 3
- [x] Verify Nodes page shows all replicas with correct Connected/In-sync status
- [x] Stop a follower and verify it shows as disconnected
- [x] Use Forget button to remove disconnected replica
- [x] Verify `lavinmqctl forget_cluster_node` works
- [x] Restart forgotten node and verify it gets a new sequential ID

### DIFF
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -67,12 +67,24 @@ class LavinMQ::Clustering::Controller
     begin
       id = File.read(id_file_path).to_i(36)
     rescue File::NotFoundError
-      id = rand(Int32::MAX)
+      id = next_cluster_node_id
       Dir.mkdir_p @config.data_dir
       File.write(id_file_path, id.to_s(36))
-      Log.info { "Generated new clustering ID" }
+      Log.info { "Generated new clustering ID: #{id.to_s(36)}" }
     end
     id
+  end
+
+  # Generate a unique sequential cluster node ID
+  # Tries to create replica/<id>/insync key, incrementing until finding an unused ID
+  private def next_cluster_node_id : Int32
+    prefix = @config.clustering_etcd_prefix
+    id = 1
+    loop do
+      key = "#{prefix}/replica/#{id.to_s(36)}/insync"
+      return id if @etcd.put_new(key, "0")
+      id += 1
+    end
   end
 
   # Replicate from the leader
--- a/src/lavinmq/clustering/controller.cr
+++ b/src/lavinmq/clustering/controller.cr
@@ -31,6 +31,7 @@ class LavinMQ::Clustering::Controller
     wait_to_be_insync
     @etcd.election_campaign("#{@config.clustering_etcd_prefix}/leader", @advertised_uri, lease: @id) # blocks until becoming leader
     @is_leader.set(true)
+    @etcd.del("#{@config.clustering_etcd_prefix}/isr") # delete legacy ISR key (used up until v2.6.x)
     execute_shell_command(@config.clustering_on_leader_elected, "leader_elected")
     @repli_client.try &.close
     # TODO: make sure we still are in the ISR set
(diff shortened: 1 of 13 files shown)


carlhoerberg:
Fixes #1425

src/lavinmq/http/controller/nodes.cr
```diff
@@ -79,10 +79,60 @@ module LavinMQ
           proc_used:          Fiber.count,
           run_queue:          0,
           sockets_used:       @amqp_server.vhosts.sum { |_, v| v.connections.size },
-          followers:          @amqp_server.followers,
+          followers:          merged_replicas,
```
snichme: Does this mean that we change the response? Meaning the keys for each follower is not the same? If so this is a breaking change. We could handle it like that or add `merged_replicas` as additional data in the response like so 


```suggestion
          followers:          @amqp_server.followers,
          merged_replicas:    merged_replicas
```

src/lavinmq/http/controller/nodes.cr
```diff
@@ -79,10 +79,60 @@ module LavinMQ
...
+              role:           is_leader ? "leader" : "follower",
+              insync:         insync,
+              remote_address: nil,
+              sent_bytes:     nil,
+              acked_bytes:    nil,
+            }
+          end
+        end
+
+        # Add any connected followers not yet in etcd (shouldn't happen but be safe)
```
snichme: How can a connected follower not be in etcd?

src/lavinmq/clustering/server.cr
```diff
@@ -207,19 +226,28 @@ module LavinMQ
...
-        ids = String.build do |str|
-          @followers.each do |f|
-            next unless f.synced?
-            f.id.to_s(str, 36)
-            str << ","
-          end
-          @id.to_s(str, 36)
+      private def mark_insync(id : Int32)
+        key = "#{@config.clustering_etcd_prefix}/replica/#{id.to_s(36)}/insync"
+        @etcd.put(key, "1")
```
snichme, line 234: We should rescue `Etcd::Error` and re-raise with something more specific here to improve the error handling.

src/lavinmq/clustering/controller.cr
```diff
@@ -123,16 +136,55 @@ class LavinMQ::Clustering::Controller
...
+    if @etcd.get(new_key) == "1"
+      return
+    end
+
+    # Check if already in sync via old comma-separated format (used up until v2.6.x)
+    legacy_isr_exists = false
+    if isr = @etcd.get(old_key)
+      legacy_isr_exists = true
+      if isr.split(",").map(&.to_i(36)).includes?(@id)
+        Log.info { "In sync via legacy ISR key" }
```
snichme: If we are insync vid legacy ISR key, shouldn't we write that we are insync in with the new key as well?

snichme (commented):
I looked at the code in `Clustering::Server` but haven't worked with that code so cannot give a good review there.  Will run the code locally as well. 

Possible to add some more specs, feel like there is a lot of logic added but not much specs around that

PR #1544, opened 2025-12-09
Use parallel execution context for clustering followers

carlhoerberg:
Run follower connection handlers in a dedicated thread pool instead of spawning fibers on the main context.


carlhoerberg:
23% throughput performance in a three node cluster. Going from 420.000 msgs/s to 520.000 msgs/s.

PR #1545, opened 2025-12-10
Retry pushing message 5 times, then reject the message

### LINKED ISSUE #1357: Overflow mode `reject-publish` not respected by shovel
**Describe the bug**
Sending messages with a shovel to a queue that is full (`x-max-length` has been reached) with `x-overflow: "reject-publish"` makes the shovel drain the upstream queue. 

**Describe your setup**
LavinMQ 2.5.rc.2, 3-node cluster

**How to reproduce**
Run attached script, Fill `q1` with 100k messages. The shovel should fill q2 with 100 messages and the rest should be kept in q2. Instead `q1` is drained into the void.

[lmq_reject-overflow.sh](https://github.com/user-attachments/files/22912898/lmq_reject-overflow.sh)

snichme:

Fixes #1357

Checks return value of publish_confirm and if was not publish retries 5 times before rejecting the message. 

There is a script attached to the issue, run that before and after applying this fix. 
Before, q1 is just emptied, messages disappears. After this fix the messages are kept in the queue.

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -93,6 +93,13 @@ module LavinMQ
         end
       end
 
+      def reject(delivery_tag)
+        if ch = @ch
+          return if ch.closed?
+          ch.basic_reject(delivery_tag, requeue: true)
+        end
+      end
+
       def started? : Bool
         !@q.nil? && !@conn.try &.closed?
       end
@@ -183,7 +190,7 @@ module LavinMQ
 
       abstract def stop
 
-      abstract def push(msg, source)
+      abstract def push(msg, source) : Bool
 
       abstract def started? : Bool
     end
@@ -207,8 +214,12 @@ module LavinMQ
         @current_dest = nil
       end
 
-      def push(msg, source)
-        @current_dest.try &.push(msg, source)
+      def push(msg, source) : Bool
+        if dest = @current_dest
+          dest.push(msg, source)
+        else
+          false
+        end
       end
 
       def started? : Bool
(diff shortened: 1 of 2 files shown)


src/lavinmq/shovel/shovel.cr
```diff
@@ -272,21 +283,26 @@ module LavinMQ
...
-      def push(msg, source)
+      def push(msg, source) : Bool
         raise "Not started" unless started?
         ch = @ch.not_nil!
         ex = @exchange || msg.exchange
         rk = @exchange_key || msg.routing_key
         case @ack_mode
         in AckMode::OnConfirm
-          ch.basic_publish(msg.body_io, ex, rk, props: msg.properties) do
+          if ch.basic_publish_confirm(msg.body_io, ex, rk, props: msg.properties)
```
carlhoerberg, line 293: fyi, the block to `basic_publish` takes a boolean too, the result from the confirm (ack/nack). 
```suggestion
          ch.basic_publish(msg.body_io, ex, rk, props: msg.properties) do |ok|
```
The upside/downside with this approach is that this fiber is blocked until an ack/nack is recievied from the server, hampering the throughput, but also limits the number of messages not confirmed (to 1)

src/lavinmq/shovel/shovel.cr
```diff
@@ -382,7 +399,15 @@ module LavinMQ
           @retries = 0
           @source.each do |msg|
             @message_count += 1
-            @destination.push(msg, @source)
+            push_retries = 0
+            until @destination.push(msg, @source)
+              if push_retries >= 5
+                @source.reject(msg.delivery_tag)
```
carlhoerberg, line 405: Wasn't the message rejected with requeue? that means the message will be retried soon again?

src/lavinmq/shovel/shovel.cr
carlhoerberg: you could reject in the block instead, with requeue if the `x-delivery-count` is less than 5, and without requeue if > 5. Or even use the `delivery-limit` queue argument/policy somehow, instead of this hardcoded one.

carlhoerberg (commented):
Have you considered a `delivery-limit` policy? or at least look at the `x-delivery-count` header?

src/lavinmq/shovel/shovel.cr
snichme: The idea is to try 5 times and then put the message back in the queue so it could be removed or consumed by another consumer. 
Don't want to reject without requeue since the message will disappear if no DLX is configured and the issue related to this PR #1357 was just about disappearing messages.

src/lavinmq/shovel/shovel.cr
snichme: I like the idea of using deliver limit. But then again, the message will be removed from the queue and "disappear" if there is no DLX configured. 

What I wanted to achieve with this PR is that the shovel basically just waits until the destination is available again, as a default behavior.

PR #1549, opened 2025-12-12
Add buffer pool for socket buffers to reduce memory for idle connections

carlhoerberg:
## Summary

- Introduce a buffer pool for `IO::Buffered` read/write buffers to save memory when handling many concurrent connections
- Most connections are idle most of the time, so releasing buffers back to a shared pool when not actively in use allows memory to be reused across connections
- Write buffers are released after flush, read buffers released on close

## Test plan

- [x] All existing tests pass (1218 examples, 0 failures)
- [ ] Manual testing with many concurrent idle connections to verify memory reduction

### DIFF
--- a/src/lavinmq/bool_channel.cr
+++ b/src/lavinmq/bool_channel.cr
@@ -1,46 +1,69 @@
 # A class which has two channels, one for which a state is true
 # and the other for when the state if false
 class BoolChannel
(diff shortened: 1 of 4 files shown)


carlhoerberg:
Dropping memory usage for 10.000 connections, each with one channel, from 637MiB to 413MiB, a 35% reduction.

Issue #1550, opened 2025-12-12
`BCC` is removed to early, should be removed on delivery


spuun:
The `BCC` header is handled incorrectly. Now it's removed by the exchange after it's been used in the routing. This is wrong, it should be preserved. It should be used in some cases of dead lettering.

This line should be removed:
https://github.com/cloudamqp/lavinmq/blob/d5a064a356aa49216b276bc0e971232615cb1346/src/lavinmq/amqp/exchange/exchange.cr#L271

However, the `BCC` should be removed before deliver and needs to be added.

PR #1552, opened 2025-12-12
Bugfix and refactor dead lettering

spuun:
The rabbitmq-java-client tests are failing because of a bug in dead lettering, causing an infinite dead lettering loop which makes lavinmq crash.

Fixed/added/changed:
- Dead letter code has been extracted from Queue to its own file to reduce the size of Queue. This part also prepares some for more code to be extracted, hence the `ArgumentTarget` and `add_argument_validator`.
- Death cycle detection. Instead of publishing to an exchange the queue itself will handling the routing, meaning that it can route the message to all destination except itself.
- Messages are routed to BCC and CC if no dead lettering routing key is set
- More dead lettering specs

Run specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ This release introduces the ability to restart closed queues, enhanced TLS capab
 - BCC remove on delivery [#1557](https://github.com/cloudamqp/lavinmq/pull/1557)
 
 ### Fixed
+- Dead lettering refactord to have proper routing and death cycle detection. [#1552](https://github.com/cloudamqp/lavinmq/pull/1552)
+
+### Fixed
 - Prefetch button icons in CSS [#1556](https://github.com/cloudamqp/lavinmq/pull/1556)
 
 ## [2.6.1] - 2025-12-07


spuun:
Failing specs is because #1550

spec/support/channel.cr
kickster97: this is good, but why is it in this pr?

shard.lock
```diff
@@ -6,7 +6,7 @@ shards:
 
   amq-protocol:
     git: https://github.com/cloudamqp/amq-protocol.cr.git
-    version: 1.1.20
+    version: 1.1.22
```
snichme, line 9: Is this version change needed for this PR? If not, I think we should do this as a separate commit

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -312,14 +302,14 @@ module LavinMQ::AMQP
...
+        if @dead_letter.dlx.nil?
+          @dead_letter.dlx ||= value.as_s
           @effective_args.delete("x-dead-letter-exchange")
           return true
         end
       when "dead-letter-routing-key"
-        if @dlrk.nil?
-          @dlrk ||= value.as_s
+        if @dead_letter.dlrk.nil?
+          @dead_letter.dlrk ||= value.as_s
```
snichme, line 312: 1) Can you setup dead lettering with only one of these arguments or do you need to setup both? 
Or rather what should happen if you only set one of them? I dont have the answer right now but open to a discussion if needed. 


2) Can we instead of checking `@dead_letter.dlx.nil?` do something like `@dead_letter.active?` I like that as an API insteda

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -685,111 +675,10 @@ module LavinMQ::AMQP
...
-        xd.merge!({
-          count:          count + 1,
-          time:           RoughTime.utc,
-          "routing-keys": routing_keys,
-        })
-        xd["original-expiration"] = expiration if expiration
-        found_at = idx
-        break
-      end
+      @dead_letter.maybe_publish(msg, reason)
```
snichme: Would it be possible to have something like this instead


```suggestion
      @dead_letter.publish(msg, reason) if @dead_letter.should_dl=
```

A method that is called `maybe_` you cannot tell what might happen

shard.lock
spuun: Yes!
It's needed for `DeadLetteringValidator` that will read `arguments` while `arguments` is being iterated by `validate_arguments!`.

src/lavinmq/amqp/queue/queue.cr
spuun: DLX is always needed. DLRK alone will fail argument validation.

src/lavinmq/amqp/queue/queue.cr
spuun: Yeah, not happy with that naming, but I want the condition in the method.

`@dead_letter.dead_letter(msg, reason)` felt weird...

shard.lock
snichme: Ok

src/lavinmq/amqp/queue/queue.cr
snichme: How about `@dead_letter.republish` and it returns an Int which is how many queues it was published to. Then the `maybe` part isn't needed since you can check how many it actually published to

src/lavinmq/amqp/queue/queue.cr
snichme: Ok. WDYT of a method to check if it's active instead of checking if the string dlx is nil?

src/lavinmq/amqp/queue/queue.cr
spuun: Preferably i'd like to avoid `publish` in the name, since it's not really a publish. Maybe `reroute`? `death_route`? Or `route`?

src/lavinmq/amqp/queue/queue.cr
spuun: Yeah, but maybe that can be done when breaking out other features to separate classes? 

Maybe each "argument module" should handle policy and queue arguments in the same way as `DeadLettering` is doing with argument validators?

src/lavinmq/amqp/queue/queue.cr
snichme: I like `route` in this case.

PR #1553, opened 2025-12-12
Collapsible menu

bengtmagnus:
Makes it possible to switch between expanded and collapsed sidebar menu. When collapsed, only the icons are visible with a clear hover indicator on which page is applicable.

Skärmavbild 2025-12-12 kl  23 24 44

A variable is saved in LocalStorage to see if the menu should be collapsed or expanded when the page loads. This check is done directly in the head to avoid jumping. The collapsed/expanded class is set on the html tag.

Since we want overflow auto in the y-direction but hidden in the x-direction on the menu, the tooltips for the menu options need to be fixed. That's why I calculate the position with javascript.

- Manually switch between the different states
- Some playwright tests

### DIFF
--- a/src/lavinmq/http/controller/views.cr
+++ b/src/lavinmq/http/controller/views.cr
@@ -64,7 +64,7 @@ module LavinMQ
             context.response.headers.add("X-Frame-Options", "SAMEORIGIN")
             context.response.headers.add("Referrer-Policy", "same-origin")
             context.response.headers.add("Content-Security-Policy", "default-src 'none'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; script-src 'self' 'sha256-7WqJMeRnYJHbEWj9TgKVXAh9Vz/3wErO1WSGhQ2LTf4='")
-           {{ block.body if block }}
+            {{ block.body if block }}
             render {{ view }}
           end
           context


carlhoerberg:
Looks great! But clicking the username, that's only temporary?

bengtmagnus:
> Looks great! But clicking the username, that's only temporary?

Ah! yes :) Will add a proper button for it. Just wanted to test the toggle on click.

static/main.css
```diff
@@ -992,6 +1027,60 @@ footer .logo {
   opacity: 1;
 }
 
+@media (min-width: 1000px) {
+  .menu-collapsed {
+    #menu-content {
+      .menu-group {
+        display: none;
+      }
```
spuun, line 1079: WDYT about
```suggestion
      .menu-group {
        margin-bottom: 0;
        span {
          display: none;
        }
      }
```

I kind of missed the grouping when the menu is collapsed.

bengtmagnus: Sure thing! Fixing

jage:
There's some issue with the two different "types" of collapsed menu, I can trigger this view:

Screenshot 2026-01-22 at 15 15 22

Maybe easiest to view this video:

https://github.com/user-attachments/assets/64a116a2-d290-422c-9bb6-befec2312a06

Ping @bengtmagnus

viktorerlingsson:
Not sure how we want this to work, but going backwards/forwards while recently having changed the collapsed state makes for a pretty jumpy experience

This is a short clip of me going backwards and forwards a few times: 
https://github.com/user-attachments/assets/079d7136-4387-404c-871c-676a734bfe1c


Reproduceable by being one page with the menu expanded, navigate to another page and collapsing the menu, then pressing back.

bengtmagnus:
> Not sure how we want this to work, but going backwards/forwards while recently having changed the collapsed state makes for a pretty jumpy experience
> 
> This is a short clip of me going backwards and forwards a few times: https://github.com/user-attachments/assets/079d7136-4387-404c-871c-676a734bfe1c
> 
> Reproduceable by being one page with the menu expanded, navigate to another page and collapsing the menu, then pressing back.

I can't seem to recreate this. Strange.

bengtmagnus:
> There's some issue with the two different "types" of collapsed menu, I can trigger this view:
> 
> Screenshot 2026-01-22 at 15 15 22
> Maybe easiest to view this video:
> 
>  Screen.Recording.2026-01-22.at.15.13.16.mov 
> Ping @bengtmagnus

Thank you, this should be fixed now.

viktorerlingsson:
> I can't seem to recreate this. Strange.

Weird! Maybe the behavior is different in different browsers 🤷 Anyway, I don't think it's blocking!

bengtmagnus:
> > I can't seem to recreate this. Strange.
> 
> Weird! Maybe the behavior is different in different browsers 🤷 Anyway, I don't think it's blocking!

What browser do you use?

viktorerlingsson:
> > > I can't seem to recreate this. Strange.
> > 
> > 
> > Weird! Maybe the behavior is different in different browsers 🤷 Anyway, I don't think it's blocking!
> 
> What browser do you use?

Firefox

jage:
Took this for another spin, think it looks good! (We had some discussions around icons, but that's for a later PR)

PR #1554, opened 2025-12-15
Cache checksums on followers while streaming changes

### LINKED ISSUE #1452: Cache checksums on followers
If followers always calculates checksums of all files and keeps them in memory, fail-over to a new leader should be even quicker.

This could be achieved in many ways, either by sending the already cached checksums from the leader, or by calculating them on the followers. Probably safer to calculate them on the follower?

viktorerlingsson:
Implements checksum caching on follower nodes to accelerate failover when promoted to leader. Checksums are now calculated incrementally during file streaming, eliminating the need to recalculate them during promotion.

  Fixes #1452

  ### Implementation

**Incremental Hashing**
  - Checksums calculated in real-time as data is streamed to followers
  - New `stream_with_checksum` helper reads, hashes, and writes data in a single pass
  - Uses 64KB buffer for efficient streaming
  - SHA1 state maintained per-file in memory, finalized on close

  **Simplified Architecture**
  - Removed background checksum worker (~150 lines)
  - Eliminated idle file tracking and timing logic
  - No configuration needed for thresholds or worker intervals
  - Checksums always current with file state

  ### Changes
  - `src/lavinmq/clustering/client.cr`: Replace worker with incremental hashing
  - `src/lavinmq/clustering/checksums.cr`: No changes needed (already thread-safe)
  - `spec/clustering_spec.cr`: Updated tests for immediate checksumming behavior

  ### Testing
  - 3 tests covering checksum persistence, streaming behavior, and file deletion
  - All tests pass, faster execution (8.7s vs 15s previously)
  - Linter clean

  ### Performance
  - Minimal overhead: SHA1 updates happen during streaming (CPU cost negligible vs I/O)
  - Better for promotion use case: checksums immediately available, no waiting
  - Simpler code is easier to maintain and debug

  All tests pass, linter clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This release introduces the ability to restart closed queues, enhanced TLS capab
 - TLS SNI and mTLS support [#1516](https://github.com/cloudamqp/lavinmq/pull/1516)
 - Shell command hooks for cluster leader transitions [#1491](https://github.com/cloudamqp/lavinmq/pull/1491)
 - Shovel and federation management commands to lavinmqctl [#1277](https://github.com/cloudamqp/lavinmq/pull/1277)
+- Cache checksums on followers for faster failover [#1554](https://github.com/cloudamqp/lavinmq/pull/1554)
 
 ### Changed
 - Convert queued message graph to stacked line graph [#1565](https://github.com/cloudamqp/lavinmq/pull/1565)


carlhoerberg:
ideally, shouldn't the leader transfer the checksum after a file is complete? now there could theoretically be a discrepancy between what the leader and the follower got?

viktorerlingsson:
> is the mutex required? aren't all files processed serially?

No, the lock shouldn't be needed any longer.

viktorerlingsson:
> ideally, shouldn't the leader transfer the checksum after a file is complete? now there could theoretically be a discrepancy between what the leader and the follower got?

Hmm, yeah maybe! But then we'd need to calculate hash for every file on the leader, add that to the "replication protocol" so it can be sent to followers, and handle on the followers what happens if that hash doesn't match. I think that could be another PR?

viktorerlingsson:
> > is the mutex required? aren't all files processed serially?
> 
> No, the lock shouldn't be needed any longer.

Removed the mutex in `Checksums`. `@file_digests_lock` is still probably needed since `close` can be called from another fiber while `stream_changes` is running.

carlhoerberg:
> > > is the mutex required? aren't all files processed serially?
> > 
> > No, the lock shouldn't be needed any longer.
> 
> Removed the mutex in `Checksums`. `@file_digests_lock` is still probably needed since `close` can be called from another fiber while `stream_changes` is running. 

I think we should wait for the follower loop to exit

viktorerlingsson:
> I think we should wait for the follower loop to exit

Yeah, that's cleaner 👍

src/lavinmq/clustering/client.cr
```diff
@@ -280,20 +284,44 @@ module LavinMQ
...
         end
       end
 
+      # Read from lz4, update SHA1, and write to file incrementally
+      private def stream_with_checksum(filename : String, lz4 : IO, file : IO, length : Int64) : Nil
+        # Get or create SHA1 digest for this file
+        sha1 = @file_digests[filename] ||= Digest::SHA1.new
+
+        # Read, hash, and write incrementally
+        buffer = uninitialized UInt8[65536]
```
carlhoerberg: any reason you chose 64KiB over 32? 
```suggestion
        buffer = uninitialized UInt8[IO::DEFAULT_BUFFER_SIZE]
```

viktorerlingsson: No, it's just what Claude suggested

viktorerlingsson:
I did set this to draft earlier because the specs hanged a couple of times, but now I've re-run them several times and they pass every time, and haven't been able to get them to hang when running locally. So not entirely sure why that happened. 
(This https://github.com/cloudamqp/lavinmq/pull/1554/commits/a4c5af1f3591880dbdd7b29789d99b56dac83f3e fixed one reason the specs could hang, but after that I'm not sure)

carlhoerberg (approved):

src/lavinmq/clustering/client.cr
```diff
@@ -280,20 +284,44 @@ module LavinMQ
...
+        # Get or create SHA1 digest for this file
+        sha1 = @file_digests[filename] ||= Digest::SHA1.new
+
+        # Read, hash, and write incrementally
+        buffer = uninitialized UInt8[IO::DEFAULT_BUFFER_SIZE]
+        remaining = length
+        while remaining > 0
+          read_len = Math.min(remaining, buffer.size)
+          lz4.read_fully(buffer.to_slice[0, read_len])
+          bytes = buffer.to_slice[0, read_len]
```
carlhoerberg:
```suggestion
          bytes = buffer.to_slice[0, read_len]
          lz4.read_fully(bytes)
```

PR #1555, opened 2025-12-15
Add Sparkplug 3.0 Aware Server support

### LINKED ISSUE #1546: Add Sparkplug B Awareness for MQTT
  LavinMQ currently implements MQTT 3.1.1 and can transport Sparkplug B messages, but it is not Sparkplug-aware. Missing automatic birth certificate management: Sparkplug-aware brokers must intercept `NBIRTH` and `DBIRTH` messages and republish them as retained messages on the `$sparkplug/certificates/#` topic hierarchy. Currently, LavinMQ treats these as regular MQTT messages, requiring clients to manually manage state through standard retained messages.

1. Birth Certificate Detection & Storage
    - Automatically detect `NBIRTH` (Node Birth) and `DBIRTH` (Device Birth) messages matching the Sparkplug B topic structure: `spBv1.0/{namespace}/{group_id}/NBIRTH/{edge_node_id}` and
  `spBv1.0/{namespace}/{group_id}/DBIRTH/{edge_node_id}/{device_id}`

    - Store these birth certificates independently of client-set retain flags
    - Persist birth certificates to disk (similar to existing retained message storage)

2. Certificate Topic Publishing
    - Automatically publish stored birth certificates to the `$sparkplug/certificates/{namespace}/{group_id}/NBIRTH/{edge_node_id}` topic structure as retained messages
    - Make these certificates immediately available to new subscribers
    - Support wildcard subscriptions to `$sparkplug/certificates/#`

3. State Cleanup
    - Remove birth certificates when corresponding `NDEATH`/`DDEATH` messages are received
    - Handle cleanup on edge node disconnection

4. Configuration
    - Add `sparkplug_aware` configuration option (`default: false` for backward compatibility)
    - Allow enabling/disabling per vhost if needed

carlhoerberg:
## Summary

Implements opt-in **Sparkplug 3.0 Aware Server** support for LavinMQ's MQTT broker, per the Eclipse Sparkplug specification for Industrial IoT.

### Key Features

- ✅ **Topic validation** — validates `spBv3.0/{group_id}/{message_type}/{edge_node_id}/{device_id?}`, plus the host STATE certificate form `spBv3.0/STATE/{host_id}`
- ✅ **Message types** — NBIRTH, NDEATH, DBIRTH, DDEATH, NDATA, DDATA, NCMD, DCMD, STATE
- ✅ **Protobuf payload validation** — checks required fields (timestamp/metrics/seq/bdSeq) per message type; payloads parsed without copying
- ✅ **Auto-retention** — BIRTH messages (NBIRTH/DBIRTH) are forced retained
- ✅ **Certificate access** — `$sparkplug/certificates/#` (and group/node-scoped variants) expand to the matching NBIRTH/DBIRTH topics
- ✅ **Strict validation** — invalid or malformed topics in the reserved namespace close the connection (logged); publishing to `$sparkplug/certificates/...` is rejected
- ✅ **Opt-in per vhost** — `sparkplug-aware` flag (default disabled), persisted in `limits.json`
- ✅ **Backward compatible** — zero overhead when disabled

### Implementation

**New modules** (`src/lavinmq/mqtt/sparkplug/`):
- `validator.cr` — topic format validation and parsing
- `certificate_mapper.cr` — certificate subscription expansion (yields expanded topics)
- `protobuf_validator.cr` — minimal protobuf wire-format payload validation

**Modified files**:
- `vhost.cr` — `sparkplug_aware` flag with persistence
- `mqtt/client.cr` — publish validation + certificate subscription expansion

**Configuration** (persisted in vhost `limits.json`):
```json
{ "sparkplug-aware": true }
```

Enable via HTTP API:
```bash
curl -u guest:guest -X PUT \
  http://localhost:15672/api/vhosts/%2F/limits \
  -H "Content-Type: application/json" \
  -d '{"sparkplug-aware": true}'
```

### Test Coverage

- Validator, certificate mapper, protobuf validator unit specs
- End-to-end integration specs (validation, auto-retain, certificate access, backward compat)
- All existing MQTT specs continue to pass

### Performance Considerations

- Fast-path `starts_with?("spBv3.0/")` check before any validation
- Zero overhead when `sparkplug-aware` is disabled
- Topic split once per publish; protobuf metrics validated via slices (no per-metric copy)

### References

- [Sparkplug 3.0 Specification](https://www.eclipse.org/tahu/spec/sparkplug_spec.pdf)
- [Eclipse Sparkplug Working Group](https://sparkplug.eclipse.org/specification/)

Fixes #1546


carlhoerberg:
Fixes #1546

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
```diff
@@ -0,0 +1,139 @@
...
+        end
+
+        # Check if topic is a certificate access topic
+        def self.certificate_topic?(topic : String) : Bool
+          topic.starts_with?(CERTIFICATE_PREFIX)
+        end
+
+        # Parse certificate topic into components
+        # Returns: {group_id?, message_type?, edge_node_id?, device_id?}
+        def self.parse_certificate_topic(topic : String) : Tuple(String?, String?, String?, String?)?
```
spuun: Only used in specs?

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
```diff
@@ -0,0 +1,139 @@
...
+        #
+        # Examples:
+        #   "$sparkplug/certificates/#" -> ["spBv3.0/+/NBIRTH/+", "spBv3.0/+/DBIRTH/+/+"]
+        #   "$sparkplug/certificates/group1/#" -> ["spBv3.0/group1/NBIRTH/+", "spBv3.0/group1/DBIRTH/+/+"]
+        #   "$sparkplug/certificates/group1/NBIRTH/#" -> ["spBv3.0/group1/NBIRTH/+"]
+        #   "$sparkplug/certificates/group1/NBIRTH/node1" -> ["spBv3.0/group1/NBIRTH/node1"]
+        #   "$sparkplug/certificates/group1/DBIRTH/#" -> ["spBv3.0/group1/DBIRTH/+/+"]
+        #   "$sparkplug/certificates/group1/DBIRTH/node1/#" -> ["spBv3.0/group1/DBIRTH/node1/+"]
+        # ameba:disable Metrics/CyclomaticComplexity
+        def self.expand_certificate_subscription(filter : String) : Array(String)
```
spuun: Why not let this method yield values instead of allocating arrays?

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
```diff
@@ -0,0 +1,139 @@
...
+          # Remove certificate prefix
+          return [filter] unless filter.starts_with?(CERTIFICATE_PREFIX)
+
+          suffix = filter[CERTIFICATE_PREFIX.size..]
+          segments = suffix.split('/')
```
spuun: It seems like we expect this to be at most 4 segments? I think this will require 2 allocations (well, one reallocation) for the array, so it's probably better to do
```suggestion
segments = Array(String).new(4)
suffix.split('/') { |segment| segments << segment }
```

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
```diff
@@ -0,0 +1,139 @@
...
+            if segments[0] == "#"
+              # "$sparkplug/certificates/#" - all BIRTH certificates
+              [
+                "spBv3.0/+/NBIRTH/+",
+                "spBv3.0/+/DBIRTH/+/+",
+              ]
+            else
+              # "$sparkplug/certificates/group1" - invalid, need more segments
+              [] of String
+            end
```
spuun: With yield this would be
```suggestion
          when 0
            # "$sparkplug/certificates/" - invalid, return empty
          when 1
            if segments[0] == "#"
              # "$sparkplug/certificates/#" - all BIRTH certificates
               yield "spBv3.0/+/NBIRTH/+"
               yield "spBv3.0/+/DBIRTH/+/+"
            else
              # "$sparkplug/certificates/group1" - invalid, need more segments
            end
```

src/lavinmq/mqtt/broker.cr
```diff
@@ -71,6 +77,25 @@ module LavinMQ
       end
 
       def publish(packet : MQTT::Publish)
+        # Track Sparkplug BIRTH/DEATH messages if state tracker is enabled
+        if tracker = @sparkplug_state_tracker
+          begin
+            if msg_type = Sparkplug::Validator.validate_topic(packet.topic)
+              # Parse topic to extract group_id and edge_node_id
+              if parts = Sparkplug::Validator.parse_topic(packet.topic)
```
spuun: `#validate_topic` calls `#parse_topic` which will split the topic, so it's done twice. Maybe `#validate_topic` should take the parts instead?
```suggestion
            if parts = Sparkplug::Validator.parse_topic(packet.topic)
              if msg_type = Sparkplug::Validator.validate_topic(parts)
```
But should a validate method return a message type, not a boolean?

src/lavinmq/mqtt/client.cr
```diff
@@ -153,7 +188,26 @@ module LavinMQ
...
+        topic_filters = packet.topic_filters
+        if @broker.vhost.sparkplug_aware?
+          expanded_filters = [] of MQTT::Subscribe::TopicFilter
+          topic_filters.each do |tf|
+            if Sparkplug::CertificateMapper.certificate_topic?(tf.topic)
+              # Expand to actual BIRTH topics
+              actual_topics = Sparkplug::CertificateMapper.expand_certificate_subscription(tf.topic)
+              actual_topics.each do |actual_topic|
+                expanded_filters << MQTT::Subscribe::TopicFilter.new(actual_topic, tf.qos)
+              end
```
spuun: Let `#expand_certificate_subscription` yield values to get rid of allocations:
```suggestion
              Sparkplug::CertificateMapper.expand_certificate_subscription(tf.topic) do |actual_topic|
                expanded_filters << MQTT::Subscribe::TopicFilter.new(actual_topic, tf.qos)
              end
```

src/lavinmq/mqtt/client.cr
```diff
@@ -153,7 +188,26 @@ module LavinMQ
             return
           end
         end
-        qos = @broker.subscribe(self, packet.topic_filters)
+
+        # Expand certificate topics if Sparkplug aware
+        topic_filters = packet.topic_filters
+        if @broker.vhost.sparkplug_aware?
+          expanded_filters = [] of MQTT::Subscribe::TopicFilter
```
spuun: No need for reallocation when populating array
```suggestion
          # Times 2  because each topic may be expanded into at most 2 sparkplug topics
          expanded_filters = Array(MQTT::Subscribe::TopicFilter).new(topic_filters.size * 2) 
```

src/lavinmq/mqtt/sparkplug/validator.cr
```diff
@@ -0,0 +1,153 @@
...
+          end
+
+          parts.message_type
+        end
+
+        # Parse Sparkplug topic into components
+        # Returns: TopicParts if valid format, nil if not a Sparkplug topic
+        def self.parse_topic(topic : String) : TopicParts?
+          # Expected format: spBv3.0/{group_id}/{message_type}/{edge_node_id}/{device_id?}
+          segments = topic.split('/')
```
spuun: Pre-allocate an array with capacity 5 populate with yielding split?

src/lavinmq/mqtt/sparkplug/state_tracker.cr
```diff
@@ -0,0 +1,119 @@
...
+        last_birth_timestamp : Int64,
+        last_death_timestamp : Int64? do
+        # Edge node is online if it has never died, or if the last birth is more recent than last death
+        def online? : Bool
+          death_ts = last_death_timestamp
+          death_ts.nil? || last_birth_timestamp > death_ts
+        end
+      end
+
+      class StateTracker
```
spuun: I guess `StateTracker` will be used to more than `#track_birth` and `#track_death`? If not it has a lot of dead weight.

spuun (commented):
I guess it's claude that loves allocations? :)

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
carlhoerberg: Right, it was only used in specs and duplicated `Validator.parse_topic`. Removed it (and its specs).

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
carlhoerberg: Done — `expand_certificate_subscription` now yields each topic instead of allocating arrays.

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
carlhoerberg: Done — using `Array(String).new(4)` with a yielding split.

src/lavinmq/mqtt/sparkplug/certificate_mapper.cr
carlhoerberg: Done — rewrote the whole `case` to yield values.

src/lavinmq/mqtt/broker.cr
carlhoerberg: Good catch. `validate_topic` now takes the parsed `TopicParts` (so the topic is split once) and returns the `MessageType`. The broker-side call is gone entirely since the StateTracker was removed.

src/lavinmq/mqtt/client.cr
carlhoerberg: Done — `expand_certificate_subscription` yields now and the client builds the filters in the block.

src/lavinmq/mqtt/client.cr
carlhoerberg: Done — pre-allocating `Array.new(topic_filters.size * 2)`.

src/lavinmq/mqtt/sparkplug/validator.cr
carlhoerberg: Done — `parse_topic` now pre-allocates `Array(String).new(5)` and populates it with a yielding split.

src/lavinmq/mqtt/sparkplug/state_tracker.cr
carlhoerberg: Agreed — nothing read the tracked state back yet, so I removed StateTracker and the broker integration entirely. We can reintroduce it alongside the management endpoint that actually surfaces the state.

carlhoerberg:
Rebased on `main` and addressed the review feedback:

- **Allocations:** `expand_certificate_subscription` and `parse_topic`/subscription expansion now yield / pre-allocate instead of building intermediate arrays.
- **Double split:** `validate_topic` now takes the parsed `TopicParts` and returns the `MessageType`, so the topic is only split once.
- **Dead weight:** removed `StateTracker` and its broker integration entirely — nothing read the tracked state back yet (this also resolves the bot's swallowed-`ValidationError` and unbounded-state-growth notes). Can reintroduce it together with a management endpoint that actually surfaces the state.
- Removed the spec-only `CertificateMapper#parse_certificate_topic`.

All specs and `make lint` pass.

carlhoerberg:
To make this more useful we should probably make a UI for it too

snichme (approved):
I think a UI would be helpful but can be added in a separate PR.

PR #1556, opened 2025-12-16
[CSS] Fix prefetch button icons

ThomasSarlin:
The buttons allocated to reset and save the prefetch count has their icons embedded in the background of the button. 
image

The structure of the `.btn-icon` class (which is only used in this context) gets an alteration to move the icon/background to the `::before` pseuedo class instead of applying it directly to the button:
image

:eyes:

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -1837,6 +1837,7 @@ pre.arguments .inactive-argument:before {
 }
 
 .btn-icon {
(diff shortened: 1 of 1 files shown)


ThomasSarlin:

bengtmagnus:
Nice!

PR #1557, opened 2025-12-16
Bcc remove on delivery

### LINKED ISSUE #1550: `BCC` is removed to early, should be removed on delivery

spuun:

BCC should not be exposed to clients. Before this PR BCC was removed during routing, which is too early since BCC may also be used later in e.g. dead lettering. Instead BCC should be removed on delivery.

Fixes #1550

Run specs

### DIFF
--- a/spec/message_routing_spec.cr
+++ b/spec/message_routing_spec.cr
@@ -431,22 +431,6 @@ module MessageRoutingSpec
       end
     end
 
-    # This is false, remove?
-    pending "should drop BCC from header" do
-      with_amqp_server do |s|
-        vhost = s.vhosts.create("x")
-        q1 = LavinMQ::QueueFactory.make(vhost, "q1")
-        q2 = LavinMQ::QueueFactory.make(vhost, "q2")
-        x = LavinMQ::AMQP::DirectExchange.new(vhost, "")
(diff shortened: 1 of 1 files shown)


spec/message_routing_spec.cr
```diff
@@ -431,7 +431,8 @@ module MessageRoutingSpec
       end
     end
 
-    it "should drop BCC from header" do
+    # This is false, remove?
```
kickster97: If its false, yeas.. but why is it false? is that not what we do?

spuun: We don't want to drop it during routing, which is what's being tested here.

kickster97: Ok!

carlhoerberg:
Did it have an impact on delivery rates?

PR #1559, opened 2025-12-16
Add ECR views as dependencies for bin/lavinmq

baelter:
## Summary
- Views (ECR templates) are compiled into the binary but weren't listed as Makefile dependencies
- Changes to `views/*.ecr` or `views/partials/*.ecr` didn't trigger recompilation of `bin/lavinmq`

## Test plan
- [x] Modify a view file (e.g., `views/shovels.ecr`)
- [x] Run `make bin/lavinmq` and verify it recompiles

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,13 @@ all: $(BINS)
 bin/%: src/%.cr $(SOURCES) lib $(JS) $(DOCS) | bin
 	crystal build $< -o $@ $(CRYSTAL_FLAGS)
 
+bin/lavinmq: src/lavinmq.cr $(SOURCES) $(VIEW_SOURCES) $(VIEW_PARTIALS) lib $(JS) $(DOCS) | bin
+	crystal build $< -o $@ $(CRYSTAL_FLAGS)
+
 bin/%-debug: src/%.cr $(SOURCES) lib $(JS) $(DOCS) | bin
+	crystal build $< -o $@ --debug $(CRYSTAL_FLAGS)
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,9 +52,15 @@ all: $(BINS)
...
 
+bin/lavinmq: src/lavinmq.cr $(SOURCES) $(VIEW_SOURCES) $(VIEW_PARTIALS) lib $(JS) $(DOCS) | bin
+	crystal build $< -o $@ $(CRYSTAL_FLAGS)
+
 bin/%-debug: src/%.cr $(SOURCES) lib $(JS) $(DOCS) | bin
 	crystal build $< -o $@ --debug $(CRYSTAL_FLAGS)
 
+bin/lavinmq-debug: src/lavinmq.cr $(SOURCES) $(VIEW_SOURCES) $(VIEW_PARTIALS) lib $(JS) $(DOCS) | bin
+	crystal build $< -o $@ --debug $(CRYSTAL_FLAGS)
+
```
viktorerlingsson, line 63: do we need the -debug stuff? is it used anywhere? 
looks like the only difference is adding `--debug`, I'd say one can just as easily just easily do `make bin/lavinmq --debug` as `make bin/lavinmq-debug`?

baelter: Not sure what the rationale for it is, but that change can go in another PR :)

viktorerlingsson: Yes! Just noticed it when looking at this 😊

PR #1560, opened 2025-12-16
Upgrade to claude github action v1

carlhoerberg:
https://code.claude.com/docs/en/github-actions

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,9 +2,9 @@ name: Claude Code Review
 on:
   pull_request:
     types: [opened, synchronize]
-    # paths: # Only run on specific file changes
-    #   - "src/**/*.cr"
-    #   - "static/**/*.js"
+    paths: # Only run on specific file changes
+      - "src/**/*.cr"
+      - "static/**/*.js"
 jobs:
   claude-review:
     runs-on: ubuntu-latest


.github/workflows/claude-code-review.yml
```diff
@@ -1,82 +1,24 @@
...
-    # Optional: Only run on specific file changes
-    # paths:
-    #   - "src/**/*.ts"
-    #   - "src/**/*.tsx"
-    #   - "src/**/*.js"
-    #   - "src/**/*.jsx"
-
+    # paths: # Only run on specific file changes
+    #   - "src/**/*.cr"
+    #   - "static/**/*.js"
```
viktorerlingsson: should this be commented out?

.github/workflows/claude-code-review.yml
```diff
@@ -1,82 +1,24 @@
...
-          #   'Please provide a thorough code review focusing on our coding standards and best practices.' }}
-
-          # Optional: Add specific tools for running tests or linting
-          # allowed_tools: "Bash(npm run test),Bash(npm run lint),Bash(npm run typecheck)"
-
-          # Optional: Skip review for certain conditions
-          # if: |
-          #   !contains(github.event.pull_request.title, '[skip-review]') &&
-          #   !contains(github.event.pull_request.title, '[WIP]')
+          prompt: "/review"
```
viktorerlingsson, line 22: Do we not want to keep any of our old prompt?

.github/workflows/claude-code-review.yml
carlhoerberg: No, but wanted to try the action before merging, but apparently the workflow cant be edited in an PR, or is forbidden to run then. I guess to protect against external PR attacks.

.github/workflows/claude-code-review.yml
carlhoerberg: Wanted to see what we got without the old prompt. Normally less is more, not sure how the system prompt might have changed since we last tried without a custom prompt.

viktorerlingsson: Ok! 👍

viktorerlingsson (approved):

.github/workflows/claude-code-review.yml
carlhoerberg:
```suggestion
    paths: # Only run on specific file changes
      - "src/**/*.cr"
      - "static/**/*.js"
```

PR #1563, opened 2025-12-17
Refactor lavinmqctl & lavinmqperf for testability and add comprehensive specs

### LINKED ISSUE #1540: Add specs for lavinmqperf & lavinmqctl
Would help make sure we don't break anything. 

Capture output from lavinmqctl/lavinmqperf and make sure it matches what we want it to match. 
Can probably copy some code or at least take inspiration from how io was captured in this commit https://github.com/cloudamqp/lavinmq/pull/1330/commits/b24096ceae833eae1616a8cbf6a035855df7f35a

viktorerlingsson:
This pull request refactors both `lavinmqctl` and `lavinmqperf` to make them testable, then adds comprehensive test coverage

### lavinmqctl

- Moved `LavinMQCtl` class to `src/lavinmqctl/cli.cr` so it can be required in specs without running the app. 
- Added capturable `@io` parameter to `LavinMQCtl` class to enable test output capture
- **37 specs** covering user management, queue operations, VHost management, exchanges, policies, connections, definition import/export, and error scenarios

### lavinmqperf 
- Added capturable `@io` parameter to `LavinMQPerf::Perf` class to enable test output capture
- **8 specs** validating AMQP/MQTT output, IO parameters, invalid arguments, and throughput calculations

### Additional Improvements

- Fixed bug where `--internal` flag was incorrectly setting the durable flag instead of internal flag

Fixes #1540

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -11,32 +11,34 @@ require "./stdlib/*"
 Signal::SEGV.reset # Let the OS generate a coredump
 Log.setup_from_env
 
-# Default to 'amqp' if the first argument is not a protocol or an option
-protocols = {"amqp", "mqtt"}
-protocol = protocols.includes?(ARGV[0]?) ? ARGV.shift : "amqp"
-case protocol
-when "amqp"
-  mode = ARGV.shift?
-  case mode
-  when "throughput"       then LavinMQPerf::AMQP::Throughput.new.run
-  when "bind-churn"       then LavinMQPerf::AMQP::BindChurn.new.run
-  when "queue-churn"      then LavinMQPerf::AMQP::QueueChurn.new.run
-  when "connection-churn" then LavinMQPerf::AMQP::ConnectionChurn.new.run
-  when "channel-churn"    then LavinMQPerf::AMQP::ChannelChurn.new.run
-  when "consumer-churn"   then LavinMQPerf::AMQP::ConsumerChurn.new.run
-  when "connection-count" then LavinMQPerf::AMQP::ConnectionCount.new.run
-  when "queue-count"      then LavinMQPerf::AMQP::QueueCount.new.run
-  when /^.+$/             then LavinMQPerf::Perf.new.run([mode.not_nil!])
-  else                         abort LavinMQPerf::Perf.new.amqp_banner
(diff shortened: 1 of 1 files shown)


spec/lavinmqctl_spec.cr
```diff
@@ -0,0 +1,384 @@
...
+class LavinMQCtl
+  def puts(*args, **options)
+    @io.puts(*args, **options)
+  end
+
+  def print(*args)
+    @io.print(*args)
+  end
+end
+
```
carlhoerberg: looks like we're doing `@io.puts` in LavinMQCtl everywhere already? or could this go into the actual LavinMQCtl? 
```suggestion
```

spec/lavinmqctl_spec.cr
```diff
@@ -0,0 +1,384 @@
...
+# Helper to run lavinmqctl commands against test server
+def run_lavinmqctl(http_addr : String, argv : Array(String))
+  stdout_capture = IO::Memory.new
+  stderr = ""
+  exit_code = 0
+
+  original_argv = ARGV.dup
+  begin
+    ARGV.clear
+    ARGV.concat(["--uri", "http://#{http_addr}", "--user", "guest", "--password", "guest"] + argv)
```
carlhoerberg:
```suggestion
    ARGV.concat(["--uri", "http://#{http_addr}", "--user", "guest", "--password", "guest"]).concat(argv)
```

carlhoerberg:
Make it clearer in the title and description that this is about refactoring and splitting ctl into multiple files.

spec/lavinmqctl_spec.cr
viktorerlingsson: You're right, I don't think we need this

viktorerlingsson:
> Make it clearer in the title and description that this is about refactoring and splitting ctl into multiple files.

Yes 👍

spec/lavinmqctl_spec.cr
carlhoerberg:
```suggestion
    ARGV.concat(["--uri", "http://#{http_addr}", "--user", "guest", "--password", "guest"]).concat(argv)
```

Issue #1558, opened 2025-12-16
Change "Queues messages" chart to lines


spuun:
<img width="649" height="264" alt="Image" src="https://github.com/user-attachments/assets/00cf374f-a2f3-410f-85d1-8d09021a1765" />

Unacked are hidden behind Ready until Unacked > Ready with the current chart. If it had been lines both will be visible as long as the values isn't equal.

carlhoerberg:
they should be stacked, something seems to have happened..

PR #1565, opened 2025-12-17
Convert queued message graph to stacked line graph

ThomasSarlin:
Convert Queued message graph to a stacked line graph, looks bad with overlapping ack/queued messages.

- Properly name the options to fill to correspond with chart.js structure
- Add an option to reverse the stack (needed to keep the unacked messages at the bottom of the graph)

Close https://github.com/cloudamqp/lavinmq/issues/1558

image

### DIFF
--- a/static/js/chart.js
+++ b/static/js/chart.js
@@ -37,6 +37,18 @@ function render (id, unit, fill = false, stacked = false, reverseStack = false) 
       responsive: true,
       maintainAspectRatio: false,
       aspectRatio: 1.3,
+      plugins: {
+        legend: {
+          labels: {
+            generateLabels: function (chart) {
+              return Chart.defaults.plugins.legend.labels.generateLabels(chart)
+                .map(function (label) {
+                  return { ...label, fillStyle: label.strokeStyle }
+                })
+            }
+          }
+        }
+      },
       tooltips: {
         mode: 'x',
         intersect: false,


carlhoerberg:
I'm not sure, I think this chart is one that actually would benefit from being stacked? if we want to go with line charts, i think total (instead of ready) + unacked would make more sense.

carlhoerberg:
I wonder what `'origin'` came from, should be true or false, right?

ThomasSarlin:
> I wonder what `'origin'` came from, should be true or false, right? 

Yes, I think true/false is the correct. But `'origin'` might be "truthy" enough.

ThomasSarlin:
> I'm not sure, I think this chart is one that actually would benefit from being stacked? if we want to go with line charts, i think total (instead of ready) + unacked would make more sense. 

Stacked makes way more sense, having a bottom of blue (unacked) with green (ready) on top and a total count. Good suggestion, I'll look into it tomorrow 👍

static/js/chart.js
```diff
@@ -108,12 +111,13 @@ function value (data) {
   return (data.rate === undefined) ? data : data.rate
 }
 
-function createDataset (key, color, fill) {
+function createDataset (key, color, fill, order = 0) {
```
kickster97, line 126: Should we set the background colorto be more translucent? 
something like 

```
backgroundColor: color + '66'
```

looks a lot more refined 
Screenshot 2025-12-18 at 08 24 26

kickster97 (changes requested):

static/js/chart.js
ThomasSarlin: Yeah, that looks better! Just need to make sure it only applies to the filled graphs tho (I see it gets applied to Message rate selectors as well)

ThomasSarlin: https://github.com/cloudamqp/lavinmq/pull/1565/commits/775ef495c37129cffce1f132f6411bb2e1538848

Makes sure other graphs are unaffected:
Screenshot 2025-12-18 at 08 51 07

kickster97: Cool, but in my (very quick and maybe not accurate) test, the unfilled graphs do not get filled.. did it do that for you? 
The only thing i'm thinking is that now the graph ledgends look incohesive.

ThomasSarlin: > Cool, but in my (very quick and maybe not accurate) test, the unfilled graphs do not get filled.. did it do that for you? 

What do you mean with "the unfilled graphs do not get filled"? 

> The only thing i'm thinking is that now the graph ledgends look incohesive.

Checking documentation to see if you can set the legends to not have the transparency.

kickster97: like this: graphs that are not supposed to be filled (message rates, data rates) were not filled 
image

ThomasSarlin: Yes, this is the current result after (https://github.com/cloudamqp/lavinmq/pull/1565/commits/39f1f81d1f1a2fb54cf48a2261b61b7b4eb9e9f6)

All legends are now filled, regardless on the state/color of the graph. 
Screenshot 2025-12-18 at 09 48 56

ThomasSarlin: And with some data:

Screenshot 2025-12-18 at 09 51 19

PR #1576, opened 2025-12-19
Less macro for deprecated properties

spuun:
I think this is a better and more future safe forwarding of deprecated values.
It's more explicit and adds the possibility to convert the value if needed to suite the new value. Also less macro.

One good thing is that there is no getter for the old property, so it can't be used by accident.

Added some specs

### DIFF
--- a/src/lavinmq/auth/user_store.cr
+++ b/src/lavinmq/auth/user_store.cr
@@ -132,7 +132,7 @@ module LavinMQ
       end
 
       private def create_default_user
-        add(Config.instance.default_user, Config.instance.default_password.to_s, "SHA256", tags: [Tag::Administrator], save: false)
+        add(Config.instance.default_user, Config.instance.default_password_hash.to_s, "SHA256", tags: [Tag::Administrator], save: false)
         add_permission(Config.instance.default_user, "/", /.*/, /.*/, /.*/)
         save!
       end


src/lavinmq/config/options.cr
```diff
@@ -215,9 +215,15 @@ module LavinMQ
...
+      @[CliOpt("", "--default-password=PASSWORD-HASH",
+        "(Deprecated) Hashed password for default user (default: '+pHuxkR9fCyrrwXjOD4BP4XbzO3l8LJr8YkThMgJ0yVHFRE+' (guest))",
+        deprecated: "--default-password is deprecated, use --default-password-hash", section: "options")]
       @[IniOpt(section: "main", deprecated: "default_password_hash")]
-      property default_password : Auth::Password::SHA256Password = DEFAULT_PASSWORD_HASH # Hashed password for default user
+      @default_password : Auth::Password::SHA256Password = DEFAULT_PASSWORD_HASH # Hashed password for default user
+      def default_password=(value)
+        # Forward value to the new property
+        @default_password_hash = value
+      end
```
spuun, line 226: If `default_password_hash` has another type than `default_password` it would have been possible to convert it here.

viktorerlingsson (approved):

PR #1578, opened 2025-12-22
Move loopback check to authenticator

spuun:
Currently the "default user loopback" check is done in multiple places. I think this should be part of the authenticator.

This PR will introduce `Auth::Context` and change the signature of `Authenticator#authenticate` to require a `Auth::Context` instead of username and password. By doing this it's easy to extend what can be passed to the authenticator to enable more checks than just username and password.

Initially the `Auth::Context` also has the property `loopback?` which tells if the request is coming via loopback.

This will also change  the http auth handler to use the authentication chain instead of accessing the user store direct, meaning that mgmt authentication is done against all authenticators.

Run pecs

### DIFF
--- a/spec/auth_spec.cr
+++ b/spec/auth_spec.cr
@@ -22,8 +22,9 @@ describe LavinMQ::Auth::Chain do
 
   it "does not authenticate when given invalid credentials" do
     with_amqp_server do |s|
+      s.@users.create("foo", "bar")
       chain = LavinMQ::Auth::Chain.create(s.@users)
-      ctx = LavinMQ::Auth::Context.new("guest", "invalid".to_slice, loopback: true)
+      ctx = LavinMQ::Auth::Context.new("bar", "baz".to_slice, loopback: true)
       user = chain.authenticate(ctx)
       user.should be_nil
     end


src/lavinmq/http/handler/auth_handler.cr
```diff
@@ -56,17 +56,9 @@ module LavinMQ
 
       private def valid_auth?(username, password, remote_address) : Bool
         return false if password.empty?
-        if user = @server.users[username]?
-          if user_password = user.password
-            if user_password.verify(password)
-              if default_user_only_loopback?(remote_address, username)
-                return false if user.tags.empty?
```
kickster97, line 63: With these changes we remove the `return false if user.tags.empty?` check, is that intentional? We don't do this check for amqp/mqtt flow i think

kickster97 (commented):
nice!

src/lavinmq/auth/authenticators/local.cr
```diff
@@ -7,17 +7,21 @@ module LavinMQ
...
+        return unless user = @users[context.username]?
+        return unless default_user_only_loopback?(context)
+        return unless passwd = user.password
+        return unless passwd.verify(context.password)
+        user
       rescue ex : Exception
         Log.error { "Local authentication failed: #{ex.message}" }
       end
+
+      private def default_user_only_loopback?(context) : Bool
```
snichme, line 20: There is a lot of `unless` and boolean logic here which makes it hard to follow for which cases it will return a `User` in the `authenticate` method. 

How about inlineing this method would that make it easier?

spec/auth_spec.cr
```diff
@@ -3,26 +3,38 @@ require "../src/lavinmq/auth/chain"
...
       chain = LavinMQ::Auth::Chain.create(s.@users)
-      user = chain.authenticate("guest", "guest")
+      ctx = LavinMQ::Auth::Context.new("guest", "guest".to_slice, loopback: true)
+      user = chain.authenticate(ctx)
       user.should_not be_nil
     end
   end
 
-  it "Does not authenticate when given invalid credentials" do
+  it "does not authenticate when given invalid credentials" do
```
snichme, line 23: Maybe out of scope for this PR but saw it now because of your change. 
For this test to pass we need to know that "invalid" is in fact an invalid password but it's not stated anyway. 
I guess it's because we setup the server with the default user `guest:guest` but it's very hidden and if for some reason we set the default password to "invalid" this spec will fail for the wrong reason.  

Can we enforce that the passwords for all the users in the UserStore (from `s.@users`)  have a different password than the one used in the the spec?

spuun: The test should probably add a user with a given password, then use this user and an invalid password!

src/lavinmq/auth/authenticators/local.cr
spuun: As long as all conditions are `unless` and there's no negations I think it's pretty readable But yeah, maybe it should be written in another way. I just wanted to get rid of the nested if statements.

src/lavinmq/http/handler/auth_handler.cr
spuun: Tags are only used in management interface, so it's intentional to not check tags in the authenticator. I don't recall how I reasoned, but all api endpoints (should) have `refuse_unless_management` or such. Maybe we should add back the tags check here... hm.

src/lavinmq/auth/authenticators/local.cr
spuun: I see now that I misinterpreted your comment. Yeah, this is a little bit hard to wrap your head around.

kickster97 (approved):

src/lavinmq/auth/authenticators/local.cr
spuun: I'll go with this. Maybe `default_user_only_loopback?` should be renamed or something, but dunno if it would make things any clearer.

PR #1579, opened 2025-12-22
Refactor some user handling to use User instead of username

spuun:
Pass around user object instead of string

Run specs (still passes)

### DIFF
--- a/src/lavinmq/auth/user_store.cr
+++ b/src/lavinmq/auth/user_store.cr
@@ -5,8 +5,12 @@ module LavinMQ
   module Auth
     class UserStore
       include Enumerable({String, User})
-      DIRECT_USER = "__direct"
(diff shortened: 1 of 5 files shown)


snichme (approved):

src/lavinmq/auth/user_store.cr
```diff
@@ -5,9 +5,13 @@ module LavinMQ
   module Auth
     class UserStore
       include Enumerable({String, User})
-      DIRECT_USER = "__direct"
+      private DIRECT_USER = "__direct"
       Log         = LavinMQ::Log.for "user_store"
 
+      def direct_user
+        @users[DIRECT_USER]
+      end
```
walro, line 13: https://github.com/cloudamqp/lavinmq/pull/1864 :-)

Issue #1336, opened 2025-10-06
Same color on all lines in the logs views


snichme:
**Summary**  

The lines in the log view alternate between two colors, it's nice and distinguished two lines form each other but when there are higher volumnes of logs, a lot of logs produced, this makes it really hard to read since the page flickers in colors while reading.  

**Desired Behavior**  

Use the same color for each line, still easy to read

**Steps to Reproduce (if reporting an issue)**  

Run lavinmq with verbose logging and consume some messages from a queue

dentarg:
Solved by @LoisSotoLopez and https://github.com/cloudamqp/lavinmq/pull/1581 I think

PR #1581, opened 2025-12-23
Add plain-table class wout alternating row bg

LoisSotoLopez:
Add plain-table class without alternating row background color

Issue #1336

Adds a `plain-table` CSS class which only purpose is to override the `table` class styling for odd and even rows.

Manual.

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -1169,6 +1169,10 @@ footer .logo {
   background-color: var(--color-table-row-even);
 }
 
+.table.plain-table tbody tr {
+  background-color: var(--color-table-row-even);
+}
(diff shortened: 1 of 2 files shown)


LoisSotoLopez:

LoisSotoLopez:
recheck

kickster97 (commented):
Thanks for looking at this :) 
looks good to me, @bengtmagnus are these changes in line with your thoughts on the redesign?

bengtmagnus:
Yes, looks good. The plan is to remove the odd/even different background style on all tables.

PR #1583, opened 2025-12-29
Apply ssl verify mode to the new sni context

tuupola:
Apparently with SNI the call to `LibSSL.ssl_set_ssl_ctx()` does not carry the `verify_mode` from `new_context` to the `ssl` object. This causes connections with bad client cert always succeed even if mTLS is configured.

This PR fixes the problem by calling `LibSSL.ssl_set_verify()` on the SSL connection to set the verify mode after changing the context.

Not sure if there is a better way to do this but atleast the fix works for me.

## Configuration

I have the following config (domain changed to example.com):

```
[main]
data_dir = /var/lib/lavinmq
tls_cert = /etc/lavinmq/certs/server.crt
tls_key = /etc/lavinmq/certs/server.key
log_level = debug

[mqtt]
bind = 0.0.0.0
tls_port = 8883

[sni:mqtt.example.com]
tls_cert = /etc/lavinmq/certs/server.crt
tls_key = /etc/lavinmq/certs/server.key
mqtt_tls_verify_peer = true
mqtt_tls_ca_cert = /etc/lavinmq/certs/ca.crt
```

## Behavior before patching

Before the patch connection succeeds both without client cert and with a bad client cert.

```
$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
^C

$ openssl req -x509 -newkey rsa:2048 -keyout bad.key -out bad.crt -days 365 -nodes -subj "/CN=baaad-mmmkay"
$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -cert bad.crt -key bad.key -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
^C
```

Connection succeeds with the correct client cert as expected.

```
$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -cert client.crt -key client.key -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
^C
```

## Behavior after patching

After applying this patch connection without client cert or with bad client cert fail as expected.

```
$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
8042B0F1F67F0000:error:0A00045C:SSL routines:ssl3_read_bytes:tlsv13 alert certificate required:ssl/record/rec_layer_s3.c:909:SSL alert number 116

$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -cert bad.crt -key bad.key -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
80E27D14757F0000:error:0A000418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:ssl/record/rec_layer_s3.c:909:SSL alert number 48
```

Connections with client cert still work as expected.

```
$ openssl s_client -connect mqtt.example.com:8883 -servername mqtt.example.com -CAfile ca.crt -cert client.crt -key client.key -quiet
Connecting to xx.xxx.28.46
depth=1 CN=Testing Root CA
verify return:1
depth=0 CN=mqtt.example.com
verify return:1
^C
```

### DIFF
--- a/spec/sni_spec.cr
+++ b/spec/sni_spec.cr
@@ -314,8 +314,7 @@ describe "SNI end-to-end" do
     server_done.receive
   end
 
-  it "copies verify_mode from the SNI selected context" do
-    # Setup SNI manager with mTLS enabled host
+  it "rejects connection without client cert when mTLS is enabled via SNI" do
     sni_manager = LavinMQ::SNIManager.new
     mtls_host = LavinMQ::SNIHost.new("mtls.localhost")
     mtls_host.tls_cert = "spec/resources/server_certificate.pem"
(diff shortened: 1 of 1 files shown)


tuupola:

dentarg:
Thanks for reporting this and writing up a fix! Do you think it is possible to cover this scenario [in the specs](https://github.com/cloudamqp/lavinmq/blob/67893de8f4d75ecce3ddc0e23d2600cc8bfbfe44/spec/sni_spec.cr)?

tuupola:
I am not too familiar with Crystal nor the codebase but I will give it a try. After all proper PR should contain tests :)

tuupola:
I did not find any easier way to check. Tests are mostly copied from `mtls_spec.cr`.

dentarg:
Thanks for trying. The team will take a look after the holidays. :)

Happy new year! 🎆

src/stdlib/openssl_sni.cr
```diff
@@ -62,6 +63,8 @@ class OpenSSL::SSL::Context::Server
         if new_context
           # Switch to the new SSL_CTX for this connection
           LibSSL.ssl_set_ssl_ctx(ssl, new_context.to_unsafe)
+          verify_mode = LibSSL.ssl_ctx_get_verify_mode(new_context.to_unsafe).to_i
+          LibSSL.ssl_set_verify(ssl, verify_mode, Pointer(Void).null)
```
viktorerlingsson: Kind of nitpicking, but instead of using `Pointer(Void).null`, I think it's cleaner to just pass `nil`

```suggestion
          LibSSL.ssl_set_verify(ssl, verify_mode, nil)
```

From crystal [docs](https://crystal-lang.org/api/1.18.2/Pointer.html): 
> When calling a C function that expects a pointer you can also pass nil instead of using [Pointer.null](https://crystal-lang.org/api/1.18.2/Pointer.html#null-class-method) to construct a null pointer.

spec/sni_spec.cr
```diff
@@ -313,4 +313,82 @@ describe "SNI end-to-end" do
...
+      # If handshake succeeds, try to read too
+      ssl_client2.gets
+      ssl_client2.close
+    ensure
+      tcp_client2.close
+    end
+
+    tcp_server.close
+    server_done.receive
+  end
```
viktorerlingsson, line 422: I think the two tests in this test should be split up to separate tests, to make sure both scenarios always run and keep each test a little shorter

viktorerlingsson:
Added a couple of comments, besides that I think it looks good! 👍

src/stdlib/openssl_sni.cr
tuupola: Sorry I was away for awhile. Will make this change soon(ish).

spec/sni_spec.cr
tuupola: Yeah I agree. I just don't know Crystal well enough to make informed choices. I copied the code from other tests in `sni_spec.cr` and altered it slightly. I think it was copied from this: https://github.com/cloudamqp/lavinmq/blob/26ded76c2e5e6fe13169fe65422291b66fde4508/spec/sni_spec.cr#L222-L315

viktorerlingsson: 👍 
Yeah, that test should also be changed to 3 separate tests at some point IMO 🙂

carlhoerberg:
Looks like there are more options that should be copied: 

```
// 1. Verify mode and callback (you already know this)
SSL_set_verify(ssl, SSL_CTX_get_verify_mode(new_ctx),
               SSL_CTX_get_verify_callback(new_ctx));

// 2. Verify depth
SSL_set_verify_depth(ssl, SSL_CTX_get_verify_depth(new_ctx));

// 3. Client CA list (for requesting client certificates)
SSL_set_client_CA_list(ssl, SSL_dup_CA_list(SSL_CTX_get_client_CA_list(new_ctx)));

// 4. Options flags
SSL_clear_options(ssl, SSL_get_options(ssl));
SSL_set_options(ssl, SSL_CTX_get_options(new_ctx));

// 5. ALPN (Application-Layer Protocol Negotiation)
// Usually set via callback, but may need attention depending on your setup

PR #1589, opened 2026-01-08
Initial LavinMQ TUI

dentarg:
Built with https://github.com/omarluq/termisu (and with help from Gemini & Codex)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,16 @@ make
 sudo make install # optional
 ```
 
+Development helpers:
+
+```sh
+crystal run extras/amqp_seed.cr -- --messages 200
+crystal run extras/amqp_seed.cr -- --live
+extras/tui_inspect.sh
+```
+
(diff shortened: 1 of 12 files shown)


dentarg:
So currently it is not possible to use both https://github.com/omarluq/termisu and https://github.com/84codes/systemd.cr and in the same project as they both try to add missing functionality to `LibC` ([src](https://github.com/omarluq/termisu/blob/1381a1547737daaf54ef0d08369d85421927bde4/src/termisu/event/poller/poll.cr#L27-L43), [src](https://github.com/84codes/systemd.cr/blob/1616dd568d717653768703ad750561b5e40924a1/src/memory_pressure.cr#L4-L15))

omarluq:
@dentarg Hey I would like to enable this I think I can probably drop the `libc` patch from termisu or at least try to find a way to skip it on compile time if its provided by the project instead

dentarg:
@omarluq sounds interesting

omarluq:
hey @dentarg `termisu` 0.4.5 should work fine with `systemd.cr` I resolved the compatibility issue between the 2 libs

dentarg:
@omarluq Oh cool, I will try to find time to try it out :)

PR #1591, opened 2026-01-08
Split shovel into muliple files

snichme:

Separate each component of the shovel feature into separate files. No functionality changes just moved the code into files to make it easier to browse it.

### DIFF
--- a/src/lavinmq/shovel/shovel.cr
+++ b/src/lavinmq/shovel/shovel.cr
@@ -4,9 +4,3 @@ require "amqp-client"
 require "http/client"
 require "wait_group"
 require "./constants"
-
-# module LavinMQ
-#   module Shovel
-#     Log = LavinMQ::Log.for "shovel"
-#   end
-# end


src/lavinmq/shovel/shovel.cr
```diff
@@ -5,483 +5,8 @@ require "http/client"
...
-        @state.running?
-      end
-    end
-  end
-end
+# module LavinMQ
+#   module Shovel
+#     Log = LavinMQ::Log.for "shovel"
+#   end
+# end
```
viktorerlingsson: should this be commented out?

src/lavinmq/shovel/source.cr
```diff
@@ -0,0 +1,6 @@
+module LavinMQ
+  module Shovel
+    abstract class Source
+    end
+  end
+end
```
viktorerlingsson, line 6: should this just be an empty abstract class?

viktorerlingsson:
There's some lint errors (`not_nil!` usage) that already existed prior to this refactor, but maybe they should be addressed?

src/lavinmq/shovel/shovel.cr
snichme: No, it should be removed.

src/lavinmq/shovel/source.cr
snichme: Yes, the idea is that the `Shovel::Runner` takes a `Source` and a `Destination` the destination has abstract methods and it works since we already have three destinations implemented. 

For `Source` we only have one implementation so we dont know what might be a good interface, but we can still use the abstract class in the runner and rely on duck typing.

snichme:
> There's some lint errors (`not_nil!` usage) that already existed prior to this refactor, but maybe they should be addressed?

I want this PR to just be a split of files. Addressing the `not_nil!` should be done but at some other change

PR #1592, opened 2026-01-08
HTTP Source for Shovels

snichme:

Introduces a HTTP source for the shovel feature. It will start a HTTP server that will forward the payload of the request to the shovel. 

It allows to have an endpoint in the broker that is already  configured where the message should end up. 
Using the endpoint in the api like this on https://docs.lavinmq.com/docs/#/operations/PostExchangePublish to publish to an exchange is already available but to use this you need to send the information on where the message should end up for every message. 
Posting a message into a shovel then the shovel has the information already so you only need to publish the payload. 

This is also a PoC in how different sources and destinations could work in a shovel to extend it from a traditional AMQP->AMQP shovel to something that can do even more "shoveling". 

Based on this https://github.com/cloudamqp/lavinmq/pull/1591 

No specs or testing done as this still is a PoC


spuun:
Should it really start it's own http server?

> It allows to have an endpoint in the broker that is already configured where the message should end up.
Using the endpoint in the api like this on https://docs.lavinmq.com/docs/#/operations/PostExchangePublish to publish to an exchange is already available but to use this you need to send the information on where the message should end up for every message.
Posting a message into a shovel then the shovel has the information already so you only need to publish the payload.

I'm not seeing the real benefit over a fanout exchange + shovel? I won't have to set properites (which can be empty), routing_key (which can be empty) or payload_encoding, but besides that?

PR #1593, opened 2026-01-08
Auto-detect proxy protocol version instead of configuration

antondalgren:
Remove the tcp_proxy_protocol and unix_proxy_protocol config options. Instead, automatically detect proxy protocol by peeking at the first bytes of incoming connections:
- V1: starts with "PROXY " text
- V2: starts with 12-byte binary signature

This simplifies configuration and allows mixed V1/V2 connections without requiring users to know which version their load balancer uses.

Specs

### DIFF
--- a/spec/follower_proxy_during_sync_spec.cr
+++ b/spec/follower_proxy_during_sync_spec.cr
@@ -113,32 +113,6 @@ describe "extract_conn_info during full_sync with syncing_followers", tags: "etc
     end
   end
 
-  it "should reject PROXY protocol when no followers are syncing or synced" do
-    with_amqp_server do |server|
-      server.@config.clustering = true
-
-      server_port = amqp_port(server)
-      client_socket = TCPSocket.new("localhost", server_port)
-
-      begin
-        proxy_header = "PROXY TCP4 192.168.1.100 127.0.0.1 54321 #{server_port}\r\n"
-        client_socket.write(proxy_header.to_slice)
-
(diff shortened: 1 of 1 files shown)


dentarg:
Is removing the config options graceful, e.g. nothing happens if you still have them in the config with this?

Should we warn about unknown options in the config? So users can be aware and clean up cruft if they want to (you should also be able to silence such warnings)

viktorerlingsson:
> Is removing the config options graceful, e.g. nothing happens if you still have them in the config with this?
> 
> Should we warn about unknown options in the config? So users can be aware and clean up cruft if they want to (you should also be able to silence such warnings)

Yes, removing options is OK. You get a warning when starting LavinMQ if you have them in your config: `WARNING: Unrecognized configuration 'amqp/tcp_proxy_protocol'`.

viktorerlingsson:
I'm thinking there might be some security concerns with always allowing PROXY protocol? Maybe we should keep the `tcp_proxy_protocol` & `unix_proxy_protocol` settings, but just as true/false. And use the automatic detection if it's V1/V2 from this PR.

carlhoerberg:
Yes, like Viktor says, proxy protocol should not be auto detected as per spec: https://github.com/haproxy/haproxy/blob/dbe52cc23e7fa15c6621e4b46896858b55f351fd/doc/proxy-protocol.txt#L865

We dont do IP based authentication much, but we do with the guest loopback thing, and could potentially add something more in the future and then forget about this.

antondalgren:
Some changes was picked into #1601

PR #1594, opened 2026-01-09
Bugfix: Removing to many bindings in Consistent hash exchange on unbind 

### LINKED ISSUE #1567: Consistent Hash Exchange: unbinding duplicate binding removes queue from hash ring (routing stops)
Description:

When a queue is bound multiple times to a consistent hash exchange with the same routing key (weight) but different binding arguments, unbinding any one of these bindings removes the queue from the hash ring entirely, even though other bindings still exist, i.e. routing stops. Found in 2.6.1.

Steps to reproduce:

1. Create a consistent hash exchange
2. Bind a queue with routing key "30" and no arguments
3. Bind the same queue again with routing key "30" and arguments `{"x-hash-on" => "cluster"}`
4. Both bindings are accepted and tracked in `@bindings`
5. Unbind the binding with arguments
6. Queue no longer receives messages, despite the first binding still existing

Expected behavior:

Queue should continue receiving messages as long as at least one binding exists.

Actual behavior:

Queue stops receiving messages after unbinding either duplicate binding.

Root cause:

https://github.com/cloudamqp/lavinmq/blob/33cc53c87460035622f325ab65c093f6c64865d6/src/lavinmq/amqp/exchange/consistent_hash.cr#L37-L47

The `@bindings` `Set` tracks bindings by `{Destination, BindingKey}` where `BindingKey` includes arguments - so duplicate bindings with different arguments are distinct entries.

However, `@hasher.remove(destination.name, w)` removes all hash ring entries for that destination name, regardless of how many bindings remain.

https://github.com/cloudamqp/lavinmq/blob/33cc53c87460035622f325ab65c093f6c64865d6/src/lavinmq/consistent_hasher.cr#L22-L27

Failing spec:

```Crystal
it "should still route after unbinding duplicate binding with different arguments" do
  with_amqp_server do |s|
    with_channel(s) do |ch|
      x = ch.exchange("con-hash", "x-consistent-hash")
      q = ch.queue("my-queue")

      q.bind(x.name, "3")
      q.bind(x.name, "3", args: AMQP::Client::Arguments.new({"x-hash-on" => "cluster"}))

      x.publish("test message 1", "rk")
      q.get(no_ack: true).try(&.body_io.to_s).should eq("test message 1")

      q.unbind(x.name, "3", args: AMQP::Client::Arguments.new({"x-hash-on" => "cluster"}))

      x.publish("test message 2", "rk")
      q.get(no_ack: true).try(&.body_io.to_s).should eq("test message 2")
    end
  end
end
```

snichme:

Fixes #1567

specs

### DIFF
--- a/spec/consistent_hash_exchange_spec.cr
+++ b/spec/consistent_hash_exchange_spec.cr
@@ -184,6 +184,26 @@ describe LavinMQ::AMQP::ConsistentHashExchange do
       end
     end
 
+    it "should still route after unbinding duplicate binding with different arguments" do
+      with_amqp_server do |s|
+        with_channel(s) do |ch|
+          x = ch.exchange(x_name, "x-consistent-hash")
+          q = ch.queue("my-queue")
+
+          q.bind(x.name, "3")
+          q.bind(x.name, "3", args: AMQP::Client::Arguments.new({"x-hash-on" => "cluster"}))
+
+          x.publish("test message 1", "rk")
+          q.get(no_ack: true).try(&.body_io.to_s).should eq("test message 1")
+
+          q.unbind(x.name, "3", args: AMQP::Client::Arguments.new({"x-hash-on" => "cluster"}))
+
+          x.publish("test message 2", "rk")
+          q.get(no_ack: true).try(&.body_io.to_s).should eq("test message 2")
+        end
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


viktorerlingsson (approved):
Changes LGTM 👍 but I think the PR description can be a little more descriptive :)

PR #1601, opened 2026-01-09
Add proxy protocol trusted sources verification

viktorerlingsson:
## Summary
 - Add `proxy_protocol_trusted_sources` config option supporting individual IPs and CIDR notation (e.g., `10.0.0.1, 192.168.0.0/24, 2001:db8::/32`)
  - Change `tcp_proxy_protocol` from version number (`0`/`1`/`2`) to boolean - protocol version is now auto-detected
  - Only parse PROXY protocol headers from trusted sources or cluster followers
  - Connections from untrusted sources sending PROXY headers will fail (protocol mismatch)
  - Remove `unix_proxy_protocol` config option - unix sockets now always auto-detect proxy protocol if present
  - Add warning at startup if proxy protocol is enabled without trusted sources configured

  ## "Breaking" Changes

  - `tcp_proxy_protocol` now accepts boolean values (`true`/`false`/`yes`/`no`) or `1`/`2` (both treated as enabled).     
  Setting `0` disables it.
  - `unix_proxy_protocol` config option is removed. Unix sockets always accept proxy protocol headers if present.

  ## Configuration

  ```ini
  [amqp]
  tcp_proxy_protocol = true
  proxy_protocol_trusted_sources = 10.0.0.1, 192.168.0.0/24, 2001:db8::/32
```
  If proxy_protocol_trusted_sources is empty and tcp_proxy_protocol is enabled, proxy headers are accepted from all  sources (with a warning logged).

  ## Security

  This follows the HAProxy proxy protocol specification recommendation to verify that proxy protocol traffic comes from trusted load balancers, preventing IP spoofing from untrusted clients.

### DIFF
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -16,7 +16,7 @@ bind = ::
 ;port = 5672
 ;tls_port = 5671
 ;tcp_proxy_protocol = 0
-;proxy_protocol_trusted_sources = 10.0.0.1, 10.0.0.2
+;proxy_protocol_trusted_sources = 10.0.0.1, 192.168.0.0/24, 2001:db8::/32
 
 [mqtt]
 bind = ::
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -15,6 +15,8 @@ bind = ::
 bind = ::
 ;port = 5672
 ;tls_port = 5671
+;tcp_proxy_protocol = 0
+;proxy_protocol_trusted_sources = 10.0.0.1, 10.0.0.2
 
 [mqtt]
 bind = ::
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -15,7 +15,7 @@ bind = ::
 bind = ::
 ;port = 5672
 ;tls_port = 5671
-;tcp_proxy_protocol = 0
+;tcp_proxy_protocol = false
 ;proxy_protocol_trusted_sources = 10.0.0.1, 192.168.0.0/24, 2001:db8::/32
 
 [mqtt]
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -78,22 +78,8 @@ module LavinMQ
       @[IniOpt(section: "amqp", transform: ->(v : String) { true?(v) || v.to_u8? == 2 })]
       property? tcp_proxy_protocol = false
 
-      @[IniOpt(section: "amqp", transform: ->(v : String) { parse_trusted_sources(v) })]
+      @[IniOpt(section: "amqp", transform: ->IPMatcher.parse_list(String))]
       property proxy_protocol_trusted_sources = Array(IPMatcher).new
-
-      def parse_trusted_sources(sources : String) : Array(IPMatcher)
-        sources.split(',')
-          .map(&.strip)
-          .reject(&.empty?)
-          .compact_map do |source|
-            begin
-              IPMatcher.parse(source)
-            rescue ex : Socket::Error | ArgumentError
-              STDERR.puts "WARNING: Invalid IP/CIDR in proxy_protocol_trusted_sources: #{source} - #{ex.message}"
-              nil
-            end
-          end
-      end
 
       @[CliOpt("", "--http-bind=BIND", "IP address that the HTTP server will listen on (default: 127.0.0.1)", section: "bindings")]
       @[IniOpt(ini_name: bind, section: "mgmt")]
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -76,7 +76,7 @@ module LavinMQ
       property mqtt_unix_path = ""
 
       @[IniOpt(section: "amqp", transform: ->(v : String) { true?(v) || v.to_u8? == 2 })]
-      property tcp_proxy_protocol = false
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/lavinmq/server.cr [hunk 1]
```diff
@@ -48,6 +49,10 @@ module LavinMQ
       @mqtt_brokers = MQTT::Brokers.new(@vhosts, @replicator)
       @parameters = ParameterStore(Parameter).new(@data_dir, "parameters.json", @replicator)
       @authenticator = Auth::Chain.create(@users)
+      @trusted_proxy_sources = @config.proxy_protocol_trusted_sources.split(',').map(&.strip).reject(&.empty?)
```
spuun: I think we should parse the values to `IPAddress`, probably already in the config. Also i think it would be nice if one can use netmasks, like  `proxy_protocol_trusted_sources = 192.168.1.0/24, 10.2.0.0/16`.

viktorerlingsson: Okay, agreed. Fixed in the last commit!

carlhoerberg:
Good! But maybe then even combine this with @antondalgren PR, where we auto detect the proxy protocol, if there is a trusted source, we only do it for those connections, but can auto detect v1 and v2. Then we we can remove one config option.

src/lavinmq/config/options.cr
```diff
@@ -66,11 +68,25 @@ module LavinMQ
...
+          .reject(&.empty?)
+          .compact_map do |source|
+            begin
+              IPMatcher.parse(source)
+            rescue ex : Socket::Error | ArgumentError
+              STDERR.puts "WARNING: Invalid IP/CIDR in proxy_protocol_trusted_sources: #{source} - #{ex.message}"
+              nil
+            end
+          end
+      end
```
viktorerlingsson: @antondalgren WDYT about putting this as a class method in IPMatcher instead? 🤔 Would keep the options file cleaner!

antondalgren: Yeah I guess we could go with that!

src/lavinmq/config/options.cr
```diff
@@ -66,11 +68,11 @@ module LavinMQ
       @[IniOpt(ini_name: unix_path, section: "mqtt")]
       property mqtt_unix_path = ""
 
-      @[IniOpt(section: "amqp", transform: ->(v : String) { true?(v) ? 1u8 : v.to_u8? || 0u8 })]
-      property unix_proxy_protocol = 1_u8 # PROXY protocol version on unix domain socket connections
+      @[IniOpt(section: "amqp", transform: ->(v : String) { true?(v) || v.to_u8? == 2 })]
+      property? tcp_proxy_protocol = false
```
carlhoerberg, line 79: do we even need that now? shouldn't we always auto deteced if there is any trusted source?

carlhoerberg: ah, for backward compability?

viktorerlingsson: Yes, for backwards compatibility.

src/lavinmq/config/options.cr
```diff
@@ -74,11 +75,11 @@ module LavinMQ
       @[IniOpt(ini_name: unix_path, section: "mqtt")]
       property mqtt_unix_path = ""
 
-      @[IniOpt(section: "amqp", transform: ->(v : String) { true?(v) ? 1u8 : v.to_u8? || 0u8 })]
-      property unix_proxy_protocol = 1_u8 # PROXY protocol version on unix domain socket connections
```
kickster97, line 78: no need to deprecate this first?

extras/lavinmq.ini
```diff
@@ -15,6 +15,8 @@ bind = ::
 bind = ::
 ;port = 5672
 ;tls_port = 5671
+;tcp_proxy_protocol = 0
```
kickster97: should be `false` now, no?

src/lavinmq/ip_matcher.cr
```diff
@@ -0,0 +1,173 @@
...
+            nil
+          end
+        end
+    end
+
+    # Check if an IP address matches this matcher
+    def matches?(address : String) : Bool
+      case @type
+      when Type::ExactIPv4, Type::ExactIPv6
+        @address_string == address
```
kickster97: Will this (`@address_string == address`) work for IPv6? I'm thinking since ::1 and 0:0:0:0:0:0:0:1 are the same address but different strings. 

You could normalize on parse or store as bytes and byte-compare.

kickster97 (approved):

src/lavinmq/config/options.cr
viktorerlingsson: Unix proxy protocol is handled automatically in this PR, so the setting isn't needed anymore. I guess we could deprecate the setting instead of removing it, but it won't make any difference, just show "deprecated" instead of "unknown setting"

extras/lavinmq.ini
viktorerlingsson: yes!

src/lavinmq/ip_matcher.cr
viktorerlingsson: You're right! I did a little rewrite, removing ExactIPv4 and ExactIPv6, and added a bunch of specs to cover more cases (like this)

PR #1604, opened 2026-01-13
Introduce Jump consistent hash as new algorithm for Consistent Hash Exchange

### LINKED ISSUE #1087: x-consistent-hash exhange consistently sending to the same queue
**Describe the bug**

I have followed this example precisely from the official docs, but changed the routing key for binding queues to the exchange to 1 for both.
https://lavinmq.com/documentation/consistent-hash-exchange

```python
channel.queue_bind(exchange="ce", queue='q1', routing_key="1")
channel.queue_bind(exchange="ce", queue='q2', routing_key="1")
```

And when I test it by publishing messages with various keys, they almost always go to q2.
I tried the following Routing Keys when publishing messages to the exchange:

```
user123
user1234
user12345
user123456
user1234567
1user123456
2user123456
3user123456
4user123456
5user123456
```

Every single one of the messages went to q2.

I have also tested it with 4 instances and it still preferentially sends to the 2nd queue that was bound.

**Describe your setup**
I am using the current cloudamqp/lavinmq docker image.
But I can confirm that I am getting the same result on V2.1.0 on a Loyal Lemming instance, hosted on cloudAMQP.

I have used the latest `amqplib` npm library and gotten these results, as well as using the Dashboard interface to set up the exchange, queues and binding as well as to publish the messages.

**How to reproduce**
Follow the example in the documentation (but with equal weight given to each queue at the time of binding) and change the message publish Routing Key randomly when producing messages and observe that the second bound queue will be picked preferentially to send messages to.

```python
import pika

connection = pika.BlockingConnection(pika.ConnectionParameters('localhost'))
channel = connection.channel()

channel.queue_declare(queue='q1', durable=True)
channel.queue_declare(queue='q2', durable=True)

channel.exchange_declare(exchange="ce", exchange_type="x-consistent-hash", durable=True)

channel.queue_bind(exchange="ce", queue='q1', routing_key="1")
channel.queue_bind(exchange="ce", queue='q2', routing_key="1")

channel.basic_publish(
    exchange="ce",
    routing_key="user123",  # swap out the routing key with random values to test
    body="Message for user123",
)

connection.close()
```

**Expected behavior**
One would expect that a number of messages published to an exchange using random Routing Key values, would result in a more or less even distribution of messages between the 2 queues bound to the exchange, if they are both bound with the same weight (same Routing key used when binding).

snichme:

Introduces Jump consistent hash as new algorithm for Consistent Hash Exchange. 

Article for Jump Consistent Hash: https://arxiv.org/pdf/1406.2294 

This algorithm is better at distributing messages equal between targets while still keeping the consistency between targets which means that when you add a queue to a consistent hash exchange messages will be also be routed to that new queue but all hashes aren't recalculated so messages that was routed to q1 might be routed to the new queue but never to another existing queue. 
Also when the new queue is removed, the existing queues will get their default hash back so it will get the same messages as before the new queue was even added. These scenarios are available as specs. 

The existing algorithm Ring consistency hash is still available and is the default, you can change to the new algorithm from config.ini. 

Fixes #1087 

New specs are introduced, all the old ones still passes. 

Here is a small python script to produce a bunch of messages into a CHE to see the differences (taken and extended from the issue #1087):  

```python
import pika

connection = pika.BlockingConnection(pika.ConnectionParameters('localhost'))
channel = connection.channel()

channel.queue_declare(queue='q1', durable=True)
channel.queue_declare(queue='q2', durable=True)

channel.exchange_declare(exchange="ce", exchange_type="x-consistent-hash", durable=True)

channel.queue_bind(exchange="ce", queue='q1', routing_key="1")
channel.queue_bind(exchange="ce", queue='q2', routing_key="1")

routing_keys = [
    "user123",
    "user1234",
    "user12345",
    "user123456",
    "user1234567",
    "1user123456",
    "2user123456",
    "3user123456",
    "4user123456",
    "5user123456",
    "alice",
    "bob",
    "charlie",
    "diana",
    "eve",
    "frank",
    "grace",
    "henry",
    "ivy",
    "jack",
]

for key in routing_keys:
    channel.basic_publish(
        exchange="ce",
        routing_key=key,
        body=f"Message for {key}",
    )
    print(f"Published message with routing_key={key}")

connection.close()
print(f"\nPublished {len(routing_keys)} messages to exchange 'ce'")
```

Run this using Ring hash first, stop lavinmq and change config to using Jump hash instead and see the different how many messages each queue gets. 
Add a queue, run the script, remove the queue and run the script again, q1 and q2 gets the same messages as it did the first time.

### DIFF
--- a/spec/consistent_hash_exchange_spec.cr
+++ b/spec/consistent_hash_exchange_spec.cr
@@ -256,7 +256,7 @@ describe LavinMQ::AMQP::ConsistentHashExchange do
       with_amqp_server do |s|
         with_channel(s) do |ch|
           x_args = AMQP::Client::Arguments.new({"x-algorithm" => "jump"})
-          x = ch.exchange(x_name, "x-consistent-hash", args: x_args)
+          ch.exchange(x_name, "x-consistent-hash", args: x_args)
           ex = s.vhosts["/"].exchanges[x_name].as(LavinMQ::AMQP::ConsistentHashExchange)
           ex.@hasher.class.should eq JumpConsistentHasher(LavinMQ::AMQP::Exchange | LavinMQ::AMQP::Queue)
         end
@@ -266,7 +266,7 @@ describe LavinMQ::AMQP::ConsistentHashExchange do
       with_amqp_server do |s|
         with_channel(s) do |ch|
           x_args = AMQP::Client::Arguments.new({"x-algorithm" => "ring"})
-          x = ch.exchange(x_name, "x-consistent-hash", args: x_args)
+          ch.exchange(x_name, "x-consistent-hash", args: x_args)
           ex = s.vhosts["/"].exchanges[x_name].as(LavinMQ::AMQP::ConsistentHashExchange)
           ex.@hasher.class.should eq RingConsistentHasher(LavinMQ::AMQP::Exchange | LavinMQ::AMQP::Queue)
         end
@@ -276,7 +276,7 @@ describe LavinMQ::AMQP::ConsistentHashExchange do
       with_amqp_server do |s|
         with_channel(s) do |ch|
           LavinMQ::Config.instance.default_consistent_hash_algorithm.should eq LavinMQ::ConsistentHashAlgorithm::Ring
-          x = ch.exchange(x_name, "x-consistent-hash")
+          ch.exchange(x_name, "x-consistent-hash")
           ex = s.vhosts["/"].exchanges[x_name].as(LavinMQ::AMQP::ConsistentHashExchange)
           ex.@hasher.class.should eq RingConsistentHasher(LavinMQ::AMQP::Exchange | LavinMQ::AMQP::Queue)
         end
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/exchange/consistent_hash.cr
```diff
@@ -1,13 +1,30 @@
...
+
   module AMQP
     class ConsistentHashExchange < Exchange
-      @hasher = ConsistentHasher(AMQP::Destination).new
+      @hasher : Hasher(AMQP::Destination)
       @bindings = Set({Destination, BindingKey}).new
 
+      def initialize(*args, **kwargs)
+        super(*args, **kwargs)
+        @hasher = case Config.instance.consistent_hash_algorithm
```
spuun: Add argument `x-algorithm` too?

snichme: Thats a good idea. Should we only have the option to set this as argument to avoid the situations where config says one thing and some exchanges has the argument set so it's really difficult to understand what the status is. 

We could have Ring as default, until next major release, and if you want to use Jump you set that as an argument

baelter: If we go with a config value as well, maybe name it `default_consistent_hash_algorithm` to hint that argument/policy always wins.

snichme: Suggestion: https://github.com/cloudamqp/lavinmq/pull/1611

snichme: @baelter and @spuun wdyt of #1611?

PR #1606, opened 2026-01-13
Resume expiration after consuming non-expiring msg

### LINKED ISSUE #1596: Message expiration stops after non-expiring message has been first in queue
Message expiration stops if a non-expiring message is published in the queue.

How to reproduce:

1. Publish a message without expiration 
2. Publish a message with expiration
3. Consume one message (without expiration)

Now the second message is never expired. It will start working again if queue is drained.

The problem is that the message expire loop gets stuck in
https://github.com/cloudamqp/lavinmq/blob/26ded76c2e5e6fe13169fe65422291b66fde4508/src/lavinmq/amqp/queue/queue.cr#L132-L140

Solution is probably to signal `message_ttl_change` after a message has been consumed/purged and it didn't have an `expiration` set.

kickster97:
Fixes #1596 
Message expiration loop was stuck when a non-expiring message was at the head of the queue. After consuming that message, the loop wouldn't wake up to check if the next message had an expiration.

Now it signals the expire loop when consuming/purging a non-expiring message if the next message has an expiration, waking the loop only when needed.

Run added spec

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -718,6 +718,7 @@ module LavinMQ::AMQP
     # yield the next message in the ready queue
     # returns true if a message was deliviered, false otherwise
     # if we encouncer an unrecoverable ReadError, close queue
+    # ameba:disable Metrics/CyclomaticComplexity
     private def get(no_ack : Bool, & : Envelope -> Nil) : Bool
       raise ClosedError.new if @closed
       loop do # retry if msg expired or deliver limit hit


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -744,6 +745,11 @@ module LavinMQ::AMQP
           yield env # deliver the message
           # requeuing of failed delivery is up to the consumer
         end
+        # Signal expire loop if we consumed a non-expiring message and the next message has expiration
```
kickster97: I'm thinking we need to check the next message here to see if it makes sense to wake the fiber, otheriwse we could have a queue `[exp, no_exp, ..1M no_exp.., exp] `and we would recieve 1M times on `@message_ttl_change`

spuun: Not sure how you mean, but I think we should signal for every message. 
This was not in the issue, but i think this is also a scenario:

1. Publish a message with ttl 1 minute
2. Publish a message with ttl 10 seconds
3. Instantly consume one message (1 minute ttl)

The  message_expire_loop will still wait for the ttl of the consumed message, then check the next message's ttl. It means that the second message will expire ~1 minute after publish, but should expire after ~10s.

spec/queue_spec.cr
```diff
@@ -48,6 +48,30 @@ describe LavinMQ::AMQP::Queue do
...
+        # Create a queue with dead letter exchange so we can verify expiration happened
+        q = ch.queue("exp_test", args: AMQP::Client::Arguments.new(
+          {"x-dead-letter-exchange" => "", "x-dead-letter-routing-key" => "dlq"}
+        ))
+        dlq = ch.queue("dlq")
+        q.publish_confirm("no expiration")
+        q.publish_confirm("with expiration", props: AMQP::Client::Properties.new(expiration: "100"))
+
+        msg1 = q.get(no_ack: true)
+        msg1.not_nil!.body_io.to_s.should eq "no expiration"
```
spuun: I think this is a nicer way of handling it in specs:
```suggestion
        msg1 = q.get(no_ack: true).should_not be_nil
        msg1.body_io.to_s.should eq "no expiration"
```

src/lavinmq/amqp/queue/queue.cr
kickster97: If there are 1M messages with no TTL, we signal to wake the fiber 1M times.. I'm thinking it should be cheaper to just check if the next message has a TTL or not, to avoid the fiber switch?

spuun: Ah, yeah. But you should probably not have the `had_no_expiration` check.

PR #1608, opened 2026-01-14
Fix: Topic exchange # wildcard now matches zero segments

### LINKED ISSUE #1607: Topic exchange: binding with # does not match zero segments (inconsistent with AMQP / RabbitMQ)
## Description

In LavinMQ **v2.6.1**, topic exchange bindings using `#` **do not match routing keys with zero trailing segments**, which is inconsistent with AMQP semantics and RabbitMQ behavior.

According to AMQP and RabbitMQ documentation, the `#` wildcard must match **zero or more words** (segments separated by `.`).
However, in LavinMQ, `#` appears to require **at least one segment** to match.

This breaks compatibility with RabbitMQ and causes unexpected routing failures when migrating workloads.

---

## Environment

* **LavinMQ version:** 2.6.1
* **Exchange type:** `topic`
* **Consumers:** none (queue without consumers)
* **Test performed on a fresh exchange and queue**

---

## Reproduction steps (minimal)

1. Create a topic exchange:

   ```
   testing-topic-exchange
   ```

2. Create a queue and bind it with:

   ```
   a.*.c.#
   ```

3. Publish messages:

   **Routing key:**

   ```
   a.b.c
   ```

   **Result:** ❌ No match (message not routed)

   **Routing key:**

   ```
   a.b.c.d
   ```

   **Result:** ✅ Match (message routed correctly)

---

## Expected behavior

The routing key:

```
a.b.c
```

**SHOULD match** the binding:

```
a.*.c.#
```

Because `#` must match **zero or more segments**, meaning the binding should accept routing keys that end exactly at `c`.

This is the documented and observed behavior in RabbitMQ.

---

## Actual behavior

* `a.b.c.#` only matches when **at least one segment exists after `c`**
* `#` behaves as **one-or-more segments**, not zero-or-more

This makes the following bindings **non-equivalent** in LavinMQ, even though they should be logically equivalent for suffix-less routing keys:

```
a.*.c
a.*.c.#
```

---

## Impact

* Breaks compatibility with RabbitMQ topic semantics
* Causes silent message loss when migrating bindings
* Forces users to add duplicate bindings as a workaround:

  ```
  a.*.c
  a.*.c.#
  ```
* Increases operational complexity and risk of misrouting

reb-felipe:
Fixes #1607

The `#` wildcard in topic exchange bindings was not matching zero segments, which violates AMQP specification and differs from RabbitMQ behavior.

This PR fixes a bug in the topic exchange routing logic where the `#` wildcard was not matching zero segments, which is required by the AMQP specification.

According to AMQP spec, `#` must match **zero or more** words (segments separated by `.`), but LavinMQ was requiring at least one segment. This caused bindings like `a.*.c.#` to incorrectly reject routing keys like `a.b.c` (where `#` should match zero segments after `c`).

**Changes:**
- Modified `HashSegment.match?` to test the zero-segment case first before attempting to consume segments
- Modified `StarSegment` and `StringSegment` to properly handle `nil` routing keys
- Added test case `"# should match zero segments"` with multiple scenarios

**Example:**
- Binding: `a.*.c.#`
- Before: `a.b.c` ❌ (incorrectly rejected)
- After: `a.b.c` ✅ (correctly accepted, matching RabbitMQ behavior)

Run the message routing specs:
```bash
make test SPEC=spec/message_routing_spec.cr
```

The new test case `"# should match zero segments (AMQP spec compliance)"` verifies that a binding with `#` at the end correctly matches routing keys with zero, one, and multiple trailing segments.

### DIFF
--- a/spec/message_routing_spec.cr
+++ b/spec/message_routing_spec.cr
@@ -160,6 +160,15 @@ module MessageRoutingSpec
         x.unbind(q11, "#.a.#")
       end
 
+      it "# should match zero segments" do
+        q = LavinMQ::QueueFactory.make(vhost, "q_zero_segments")
+        x.bind(q, "a.*.c.#")
+        matches(x, "a.b.c").should eq(Set{q})
+        matches(x, "a.b.c.d").should eq(Set{q})
+        matches(x, "a.b.c.d.e").should eq(Set{q})
+        x.unbind(q, "a.*.c.#")
+      end
+
       it "should match double star-wildcards" do
         q12 = LavinMQ::QueueFactory.make(vhost, "q12")
         x.bind(q12, "c.*.*")
--- a/src/lavinmq/amqp/exchange/topic.cr
+++ b/src/lavinmq/amqp/exchange/topic.cr
@@ -31,10 +31,13 @@ module LavinMQ
 
         def match?(rk) : Bool
           if n = @next
+            return true if n.match?(rk)
+            return false unless rk
+
             loop do
-              return true if n.match?(rk)
               rk = rk.next
               break unless rk
+              return true if n.match?(rk)
             end
             return false
           end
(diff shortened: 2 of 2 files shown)


reb-felipe:

PR #1609, opened 2026-01-14
Refactor http api auth to set user earlier

spuun:

Refactors HTTP API authentication to separate authentication (setting the user) from authorization (requiring the user). This moves user authentication earlier in the middleware chain, making the authenticated user available to more handlers (e.g. websocket).

Run specs

### DIFF
--- a/src/lavinmq/http/http_server.cr
+++ b/src/lavinmq/http/http_server.cr
@@ -20,9 +20,9 @@ module LavinMQ
       def initialize(@amqp_server : LavinMQ::Server)
         handlers = [
           StrictTransportSecurity.new,
+          StaticController.new,
           AuthHandler.new(@amqp_server.authenticator, @amqp_server.users.direct_user),
           WebsocketProxy.new(@amqp_server),
-          StaticController.new,
           ViewsController.new,
           ApiErrorHandler.new,
           RequireUserHandler.new,


baelter:
Hm, now we run auth, which does some cryptography that take some cpu cycles, even for static assets. Can we move that outside this middleware?

spuun:
> Hm, now we run auth, which does some cryptography that take some cpu cycles, even for static assets. Can we move that outside this middleware?

Ah, true! It should be possible to fix, of course!

spuun:
Instead of the `AuthHandler` we can make `#user` lazy on `Context` and do the auth first time it's requested... hmhm
I'll think about that, but that will be for another PR.

spuun:
> Instead of the `AuthHandler` we can make `#user` lazy on `Context` and do the auth first time it's requested... hmhm I'll think about that, but that will be for another PR.

Scratch that, it will be to messy. The order of handlers will do for now.

snichme:
Is there a background to why this change is needed?

carlhoerberg (approved):

spuun:
> Is there a background to why this change is needed?

1. Be able to access the user in e.g. websocket handler. E.g. in #1585 that would be needed to be able to filter entities on permissions.
2. Be able to access the user in views controller to be able to move some stuff from client side to server side.

PR #1611, opened 2026-01-15
allow x-algorithm to set algorithm for the consistent hash exchange

snichme:

Adressing https://github.com/cloudamqp/lavinmq/pull/1604#discussion_r2689324407

image

image

### DIFF
--- a/src/lavinmq/amqp/exchange/consistent_hash.cr
+++ b/src/lavinmq/amqp/exchange/consistent_hash.cr
@@ -39,9 +39,9 @@ module LavinMQ
 
       private def select_hasher(option : ConsistentHashAlgorithm)
         case option
-        when .jump?
+        in .jump?
           JumpConsistentHasher(AMQP::Destination).new
-        else
+        in .ring?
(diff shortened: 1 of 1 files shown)


src/lavinmq/amqp/exchange/consistent_hash.cr
```diff
@@ -31,9 +26,26 @@ module LavinMQ
...
         @effective_args << "x-hash-on" if @arguments["x-hash-on"]?
       end
 
+      private def select_hasher(option : ConsistentHashAlgorithm)
+        case option
+        when .jump?
+          JumpConsistentHasher(AMQP::Destination).new
+        else
+          RingConsistentHasher(AMQP::Destination).new
+        end
```
spuun, line 46: Use [exhaustive case](https://crystal-lang.org/reference/1.19/syntax_and_semantics/case.html#exhaustive-case) when possible:
```suggestion
        case option
        in .jump?
          JumpConsistentHasher(AMQP::Destination).new
        in .ring?
          RingConsistentHasher(AMQP::Destination).new
        end
```

PR #1614, opened 2026-01-15
Idle fiber management for queue message expiration

### LINKED ISSUE #1595: Possible segfault after `Cannot allocate memory` error
**Describe the bug**
When creating many consumers (and potentially queues, channels, or connections), fiber allocation can fail if the system hits resource limits like `vm.max_map_count` (because each fiber requires a memory mapping). This can lead to a segfault due to orphaned fibers accessing uninitialized instance variables.

The issue _probably_ occurs when a constructor spawns a fiber before all instance variables (BoolChannels) are initialized. If a subsequent operation fails (e.g., spawning another fiber in a BoolChannel), the constructor raises but the already-spawned fiber remains in the run queue. When the scheduler eventually runs that fiber, it accesses uninitialized instance variables, causing a segfault.

Example: Consumer (confirmed reproducible)

  In src/lavinmq/amqp/consumer.cr:35-37:
```  
  @queue.deliver_loop_wg.spawn(name: fiber_name) { deliver_loop }  # ✓ Succeeds, fiber queued
  @flow_change = BoolChannel.new(@flow)              # ✗ BoolChannel.new spawns fiber, FAILS
  @has_capacity = BoolChannel.new(true)              # Never reached
```

  The orphaned deliver_loop fiber later calls @has_capacity.when_true.receive, but @has_capacity was never initialized → SEGFAULT.

Stack trace:
```
Unexpected error, while reading: Cannot allocate new fiber stack: Cannot allocate memory (RuntimeError)
    from usr/share/crystal/src/atomic.cr:307:29 in 'checkout'
    from usr/share/crystal/src/fiber.cr:105:3 in '??'
    from usr/src/lavinmq_2.6.1/src/lavinmq/bool_channel.cr:8:3 in '??'
    from usr/src/lavinmq_2.6.1/src/lavinmq/amqp/consumer.cr:36:9 in 'new'
    from usr/src/lavinmq_2.6.1/src/lavinmq/amqp/channel.cr:382:17 in 'process_frame'
    from usr/src/lavinmq_2.6.1/src/lavinmq/amqp/client.cr:143:15 in 'read_loop'
```

**Describe your setup**
Reproduced in LavinMQ v.2.5.5 & v2.6.1. 
`max_map_count = 1048576`

**How to reproduce**
Run `lavinmqperf amqp connection-count -C 1000` and let it create 200k+ consumers.
Another example is to lower max_map_count to like 10k and run `lavinmqperf amqp connection-count -C 10`. 

**Expected behavior**
LavinMQ should not crash with a segfault when creating many consumers. 

**Workarund**
Increase the system's memory map limit:
  sysctl -w vm.max_map_count=1048576  # or higher as needed

viktorerlingsson:
## Summary

Queue `message_expire_loop` fibers are now started on demand and stop when there's nothing to expire, instead of running for the lifetime of every queue.

## When the fiber runs

- **Started** when a queue gains expirable messages — head of queue has an `expiration`, or queue-level `message-ttl` is set. Triggered on publish, `basic_get`, and TTL configuration changes.
- **Stopped** after `EXPIRE_FIBER_IDLE_THRESHOLD_MS` (30s) of waiting on any of: consumers empty, message store empty, no message at head needing expiry. Also exits cleanly on queue close via `Channel::ClosedError`.
- `DelayedExchangeQueue` always runs the fiber, since its messages are only delivered via expiration.

## Implementation

A single `Atomic(Bool)` per queue (`@message_expire_fiber_active`) tracks whether the fiber is running, with release/acquire ordering so the fiber observes a consistent view of queue state. `expiration: "0"` is treated as a valid expiration (immediate expiry).

## Performance
Idle fiber management has no measurable impact on queue churn throughput, but cuts per-queue memory use by ~20-30% at steady state. Each idle queue no longer reserves a Crystal fiber stack, which saves ~9 KB of resident memory and ~8 MB of virtual memory per queue.

  | Queues | RSS (main) | RSS (branch) | Δ RSS         | VSZ (main) | VSZ (branch) |
  |-------:|-----------:|-------------:|--------------:|-----------:|-------------:|
  |  1,000 |      43 MB |        34 MB |  -9 MB (-20%) |      16 GB |         8 GB |
  |  5,000 |     157 MB |       113 MB | -44 MB (-28%) |      79 GB |        40 GB |
  | 10,000 |     299 MB |       208 MB | -91 MB (-30%) |     157 GB |        79 GB |

### Throughput: `on-demand-expire-fibers` vs `main`

  `extras/benchmark.sh` (10s/scenario). Branch = **1 run**; main = avg of 2 runs - so treat small deltas as noise. Metric is
  **consume rate** except *Publish only* (publish rate). Δ = (branch − main) / main.

  | Scenario | Branch (msgs/s) | main (msgs/s) | Δ |
  |---|--:|--:|--:|
  | Baseline 1→1 (16 B) | 1,029,649 | 1,040,073 | −1.0% |
  | 1→1 with properties | 884,827 | 770,814 | +14.8% |
  | Publish only* | 1,858,591 | 1,982,339 | −6.2% |
  | Consume only (drain) | 1,690,995 | 1,786,710 | −5.4% |
  | Fanout 1→4 | 1,427,176 | 1,427,592 | 0.0% |
  | Large messages (100 KB) | 8,587 | 8,505 | +1.0% |
  | Prefetch 1000, ack 100 | 95,764 | 101,632 | −5.8% † |
  | Publish confirm | 33,804 | 33,875 | −0.2% ‡ |
  | Publish in tx | 9,629 | 9,834 | −2.1% ‡ |
  | Ack in tx | 870 | 865 | +0.6% |
  | Stream 1→1 | 87,231 | 92,404 | −5.6% |
  | Stream + properties | 106,872 | 122,661 | −12.9% |
  | MT fan-in 64→1 | 192,562 | 200,011 | −3.7% |
  | MT competing 1→64 | 698,741 | 698,453 | 0.0% |
  | MT symmetric 64→64 | 884,815 | 903,472 | −2.1% |
  | MT sharded 64→64 | 842,629 | 803,785 | +4.8% |
  | MT stream multi 1→64 | 1,003,001 | 941,862 | +6.5% |

  \* publish rate (no consumer).
  † high run-to-run variance, treat as noise.
  ‡ **unreliable** in `benchmark.sh` - single 10s run swings widely on all builds; ignore.

  **No regression.** Branch tracks `main` within run-to-run noise across every scenario (the −5 to −13% on drain / stream-props are within single-run variance, not a consistent signal). The +14.8% on *1→1 properties* mirrors what the deliver-loop  branch shows against the same `main` binary, so it's a shared-base/build artifact rather than an effect of this change.
  Single-run sample - rerun before drawing firm conclusions.

## Test plan

- [x] Empty queue does not start the fiber
- [x] Queue with no-TTL messages does not start the fiber
- [x] Queue-level TTL: fiber starts on publish and expires messages
- [x] Per-message TTL: fiber starts on publish and expires messages

## Related

Helps mitigate #1595 by reducing fiber count for idle queues with message TTL.

### DIFF
--- a/src/lavinmq/amqp/queue/delayed_exchange_queue.cr
+++ b/src/lavinmq/amqp/queue/delayed_exchange_queue.cr
@@ -85,7 +85,13 @@ module LavinMQ::AMQP
       end
     rescue ::Channel::ClosedError
     ensure
+      @message_expire_fiber_active.set(false, :release)
       @log.debug { "message_expire_loop stopped" }
+    end
+
+    # Delayed exchange queues always need their expire fiber running
+    private def should_start_expire_fiber? : Bool
+      true
     end
 
     def expire_messages
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -115,7 +115,7 @@ module LavinMQ::AMQP
       @message_expire_fiber_active.get(:relaxed)
     end
 
-    private EXPIRE_FIBER_IDLE_THRESHOLD_MS = 30_000
+    private EXPIRE_FIBER_IDLE_THRESHOLD = 30.seconds
 
     getter? internal = false
 
@@ -144,7 +144,7 @@ module LavinMQ::AMQP
         select
         when @consumers_empty.when_true.receive
           @log.debug { "Consumers empty" }
-        when timeout EXPIRE_FIBER_IDLE_THRESHOLD_MS.milliseconds
+        when timeout EXPIRE_FIBER_IDLE_THRESHOLD
           @log.debug { "Idle timeout while waiting for consumers empty, stopping fiber" }
           break
         end
@@ -152,7 +152,7 @@ module LavinMQ::AMQP
         select
         when @msg_store.empty.when_false.receive
           @log.debug { "Message store not empty" }
-        when timeout EXPIRE_FIBER_IDLE_THRESHOLD_MS.milliseconds
+        when timeout EXPIRE_FIBER_IDLE_THRESHOLD
           @log.debug { "Idle timeout while waiting for messages, stopping fiber" }
           break
         end
@@ -179,7 +179,7 @@ module LavinMQ::AMQP
             @log.debug { "Message TTL changed" }
           when @msg_store.empty.when_true.receive
             @log.debug { "Msg store is empty" }
-          when timeout EXPIRE_FIBER_IDLE_THRESHOLD_MS.milliseconds
+          when timeout EXPIRE_FIBER_IDLE_THRESHOLD
             @log.debug { "Idle timeout while no messages need expiring, stopping fiber" }
             break
           end
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -108,6 +108,15 @@ module LavinMQ::AMQP
     @queue_expiration_ttl_change = ::Channel(Nil).new
     @effective_args = Array(String).new
 
+    # Idle fiber management
+    @message_expire_fiber_active = Atomic(Bool).new(false)
+
+    def message_expire_fiber_active?
+      @message_expire_fiber_active.get(:relaxed)
+    end
+
+    EXPIRE_FIBER_IDLE_THRESHOLD_MS = 30_000
+
     getter? internal = false
 
     private def queue_expire_loop
@@ -132,10 +141,22 @@ module LavinMQ::AMQP
     private def message_expire_loop
       @vhost.closed.when_false.receive?
       loop do
-        @consumers_empty.when_true.receive
-        @log.debug { "Consumers empty" }
-        @msg_store.empty.when_false.receive
-        @log.debug { "Message store not empty" }
+        select
+        when @consumers_empty.when_true.receive
+          @log.debug { "Consumers empty" }
+        when timeout EXPIRE_FIBER_IDLE_THRESHOLD_MS.milliseconds
+          @log.debug { "Idle timeout while waiting for consumers empty, stopping fiber" }
+          break
+        end
+
+        select
+        when @msg_store.empty.when_false.receive
+          @log.debug { "Message store not empty" }
+        when timeout EXPIRE_FIBER_IDLE_THRESHOLD_MS.milliseconds
+          @log.debug { "Idle timeout while waiting for messages, stopping fiber" }
+          break
+        end
+
         next unless @consumers.empty?
         if ttl = time_to_message_expiration
           @log.debug { "Next message TTL: #{ttl}" }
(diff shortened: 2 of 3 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -206,9 +232,36 @@ module LavinMQ::AMQP
       start
     end
 
+    # Start message expire loop fiber
```
kickster97: I think we should avoid comments like this, since the function name says the same thing

kickster97 (commented):
Could it be a problem that we're not calling `ensure_expire_fiber` after `basic.get`? I'm thinking about scenarios where someone is using basic.get (no consumers) and they consume the first message. If there's a TTL message behind it, the fiber won't start up again?

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: agreed! Claude leftover 🙈

viktorerlingsson:
> Could it be a problem that we're not calling `ensure_expire_fiber` after `basic.get`? I'm thinking about scenarios where someone is using basic.get (no consumers) and they consume the first message. If there's a TTL message behind it, the fiber won't start up again?

Yes, you're right. Fixed in the last commit!

src/lavinmq/amqp/queue/delayed_exchange_queue.cr
```diff
@@ -85,9 +85,15 @@ module LavinMQ::AMQP
       end
     rescue ::Channel::ClosedError
     ensure
+      @message_expire_fiber_active.set(false, :relaxed)
```
carlhoerberg: are you sure we can be `relaxed` here?

carlhoerberg (commented):
as claude said, i think we have to be more granual with the ordering throughout

src/lavinmq/amqp/queue/delayed_exchange_queue.cr
viktorerlingsson: Hmm, yeah, you're probably right. I guess that could lead to the fiber not being started. Let's go with `:release` and `:aquire` instead!

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -79,6 +79,15 @@ module LavinMQ::AMQP
...
     @effective_args = Array(String).new
 
+    # Idle fiber management
+    @message_expire_fiber_active = Atomic(Bool).new(false)
+
+    def message_expire_fiber_active?
+      @message_expire_fiber_active.get(:relaxed)
+    end
+
+    private EXPIRE_FIBER_IDLE_THRESHOLD_MS = 30_000
```
carlhoerberg:
```suggestion
    private EXPIRE_FIBER_IDLE_THRESHOLD = 30.seconds
```

src/lavinmq/amqp/queue/queue.cr
kickster97, line 1024: Could this have the same issue as `basic.get` had? If the expire fiber has stopped because the head message has no TTL, a partial `purge(count)` can remove that head message and expose a TTL message behind it. Since `try_send?` has no receiver when the fiber is stopped, nothing restarts expiration until some later trigger.

viktorerlingsson: Yes, I think you're right about that. Changed so `ensure_expire_fiber` runs after purge

kickster97 (approved):

Issue #1616, opened 2026-01-16
Crystal 1.19.0 compatibility


jage:
From our Homebrew tap on macOS https://github.com/cloudamqp/homebrew-cloudamqp

```
==> make bin/lavinmq bin/lavinmqctl bin/lavinmqperf DOCS=
Last 15 lines from /Users/robin/Library/Logs/Homebrew/lavinmq/01.make.log:
In src/lavinmq/server.cr:34:19

 34 | @start = Time.monotonic
                    ^--------
Warning: Deprecated Time.monotonic. Use `Time.instant` instead.

In src/lavinmq/rough_time.cr:4:22

 4 | @@monotonic = Time.monotonic
                        ^--------
Warning: Deprecated Time.monotonic. Use `Time.instant` instead.

A total of 9 warnings were found.
make: *** [bin/lavinmq] Error 1
make: *** Deleting file `bin/lavinmq'
```

Noted by @walro

jage:
Noted this in main today:

```
$ make
crystal build src/lavinmq.cr -o bin/lavinmq --release --stats --error-on-warnings -Dpreview_mt -Dexecution_context --link-flags="-pie"
Parse:                             00:00:00.000053584 (   1.17MB)
Showing last frame. Use --error-trace for full trace.

In src/stdlib/openssl_sni.cr:13:3

 13 | SSL_CTRL_SET_TLSEXT_SERVERNAME_CB  = 53
      ^--------------------------------
Error: already initialized constant LibSSL::SSL_CTRL_SET_TLSEXT_SERVERNAME_CB
make: *** [bin/lavinmq] Error 1
$ crystal version
Crystal 1.19.0 (2026-01-15)
```

Which is due to https://github.com/crystal-lang/crystal/releases/tag/1.19.0 now includes https://github.com/crystal-lang/crystal/pull/16452 which makes our implementation redundant.

jage:
Working on getting a branch with 1.19.0 compatibility https://github.com/cloudamqp/lavinmq/tree/crystal-1.19.0

We want to test it and verify everything is working as expected.

jage:
Done #1620 merged into main.

PR #1617, opened 2026-01-16
Don't use --error-on-warnings on release build

jage:
It completely stops a build that "just" has warnings, we do want it in CI though.

Related to #1616


viktorerlingsson (approved):
LGTM! 
Note that we won't get --error-on-warnings when building locally with make now either.

PR #1620, opened 2026-01-16
Crystal 1.19 compatibility

jage:
See individual commits, only tested with specs so far.

Here I think we should run a bit of manual testing (set up a machine, test with lavinmqperf, maybe some Snowstorm)

- Updated codebase for Crystal 1.19.0 compatibility by addressing deprecated APIs                                        
- Replaced `Time.monotonic` with `Time.instant` (returns `Time::Instant` instead of `Time::Span`)                        
- Replaced deprecated `Random::DEFAULT` with `Random::Secure` for security-sensitive code and instance `Random.new` for performance tools                                                                                                       
- Updated to stdlib SNI support (`on_server_name` callback), removing custom extension                                   
- Fix `exit` function signature for Crystal 1.19's `Process::Status` type

### DIFF
--- a/spec/support/rough_time.cr
+++ b/spec/support/rough_time.cr
@@ -1,7 +1,7 @@
 module RoughTime
   @@paused_utc = Time.utc
   @@paused_unix_ms : Int64 = @@paused_utc.to_unix_ms // 100 * 100
-  @@paused_monotonic = Time.instant
+  @@paused_instant = Time.instant
   @@paused = false
 
   def self.utc : Time
@@ -18,9 +18,9 @@ module RoughTime
     previous_def
   end
 
-  def self.monotonic : Time::Instant
+  def self.instant : Time::Instant
     if @@paused
-      return @@paused_monotonic
+      return @@paused_instant
     end
     previous_def
   end
@@ -29,13 +29,13 @@ module RoughTime
     @@paused = true
     @@paused_utc = Time.utc
     @@paused_unix_ms = @@paused_utc.to_unix_ms // 100 * 100
-    @@paused_monotonic = Time.instant
+    @@paused_instant = Time.instant
   end
 
   def self.travel(time : Time::Span)
     @@paused_utc += time
     @@paused_unix_ms += @@paused_utc.to_unix_ms // 100 * 100
-    @@paused_monotonic += time
+    @@paused_instant += time
   end
 
   def self.resume
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -413,7 +413,7 @@ module LavinMQ
             ok = q.basic_get(frame.no_ack) do |env|
               delivery_tag = next_delivery_tag(q, env.segment_position, frame.no_ack, nil)
               unless frame.no_ack # track unacked messages
-                q.basic_get_unacked << UnackedMessage.new(self, delivery_tag, RoughTime.monotonic)
+                q.basic_get_unacked << UnackedMessage.new(self, delivery_tag, RoughTime.instant)
               end
               get_ok = AMQP::Frame::Basic::GetOk.new(frame.channel, delivery_tag,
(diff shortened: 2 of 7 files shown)


src/lavinmqperf/amqp/throughput.cr
```diff
@@ -348,22 +349,22 @@ module LavinMQPerf
...
           queue_idx = 0
           until @stopped
             if @measure_latency
               # Write timestamp at the beginning of the message
-              IO::ByteFormat::LittleEndian.encode(Time.monotonic.total_nanoseconds.to_i64, data)
+              IO::ByteFormat::LittleEndian.encode((Time.instant - BASE_INSTANT).total_nanoseconds.to_i64, data)
               # Fill the rest with random or pattern data
               if @random_bodies
-                Random::DEFAULT.random_bytes(data[8..])
+                Random::Secure.random_bytes(data[8..])
```
carlhoerberg: i think it's more appropriate to instantiate a new `Random` higher up and use that. `Random::Secure` is way slower.

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -204,12 +204,12 @@ module LavinMQPerf
...
         Fiber.yield
 
-        start = Time.monotonic
+        start = Time.instant
         pubs_this_second = 0
         packet_id_generator = (1_u16..).each
         wait_until_all_are_connected(connected)
         until @stopped
-          Random::DEFAULT.random_bytes(data) if @random_bodies
+          Random::Secure.random_bytes(data) if @random_bodies
```
carlhoerberg: Should not need cryptographic level of randomness here.

carlhoerberg (commented):
Should probably remove the file `src/stdlib/openssl_sni.cr` too.

I haven't followed but i suspect `Random::DEFAULT` was removed because it's not thread safe. Note that `Random::SECURE` is many times slower than the PRG32. I cases where it matters (lavinmqperf etc), I think we should probably create a new `Random` instance and use that.

src/lavinmqperf/amqp/throughput.cr
```diff
@@ -323,7 +324,7 @@ module LavinMQPerf
             # Replace random element with probability RESERVOIR_SIZE / count
             # Standard reservoir sampling: pick random position in [0, count),
             # and if it's within reservoir size, replace that element
-            j = Random::DEFAULT.rand(@latencies_count)
+            j = Random::Secure.rand(@latencies_count)
```
carlhoerberg:
```suggestion
            j = Random.rand(@latencies_count)
```

jage:
Adressed comment, still haven't tested this on any server.
There's rumours 1.19.1 is about to be released as well.

jage:
Feel free to merge and get https://github.com/cloudamqp/lavinmq/tree/crystal-1.19.0 branch into main, I'm pretty much done for the day now.

src/lavinmqperf/amqp/throughput.cr
```diff
@@ -323,7 +325,7 @@ module LavinMQPerf
             # Replace random element with probability RESERVOIR_SIZE / count
             # Standard reservoir sampling: pick random position in [0, count),
             # and if it's within reservoir size, replace that element
-            j = Random::DEFAULT.rand(@latencies_count)
+            j = Random.rand(@latencies_count)
```
viktorerlingsson: Is there a reason we use `Random.rand` instead of `@random.rand` here?

jage: Oops, missed. :)

jage:
Hm, `RoughTime` still calls it `monotonic`, keep that or change while we're at it? :)

viktorerlingsson:
> Hm, `RoughTime` still calls it `monotonic`, keep that or change while we're at it? :)

It doesn't really matter since that's just our own internal implementation, but I think we should change it to `instant` to better match stdlib

PR #1624, opened 2026-01-16
Oauth2: Use jwt.cr shard instead of implementing it natively

kickster97:

Extract the JWT logic to its own shard and refactor to use that instead

run updated specs


kickster97:
Have decided against this change, it will be easier to maintain if kept natively

PR #1627, opened 2026-01-19
Add Prometheus metrics for mfiles

jage:
Track open MFile instances with atomic counter and expose via Prometheus:

* `lavinmq_mfile_count` Number of MFile memory-mapped files

Documentation need to be updated once released https://lavinmq.com/documentation/prometheus

### DIFF
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -428,9 +428,9 @@ module LavinMQ
                       type:  "gauge",
                       help:  "Number of MFile memory-mapped files"})
         {% if flag?(:linux) %}
-          if content = File.read?("/proc/self/maps")
+          if count = count_proc_maps_lines
             writer.write({name:  "process_mmap_count",
-                          value: content.count('\n').to_i64,
+                          value: count,
                           type:  "gauge",
                           help:  "Total number of virtual memory areas in the process"})
           end
@@ -630,6 +630,21 @@ module LavinMQ
             end
           end
         end
+      end
+
+      # Count lines in /proc/self/maps without allocating the entire file content.
+      # Reads in chunks and counts newlines incrementally to reduce memory usage.
+      private def count_proc_maps_lines : Int64?
+        File.open("/proc/self/maps", "r") do |file|
+          count = 0i64
+          buffer = uninitialized UInt8[4096]
+          while (bytes_read = file.read(buffer.to_slice)) > 0
+            count += buffer.to_slice[0, bytes_read].count('\n'.ord.to_u8)
+          end
+          count
+        end
+      rescue
+        nil
       end
     end
   end
--- a/src/lavinmq/http/controller/prometheus.cr
+++ b/src/lavinmq/http/controller/prometheus.cr
@@ -427,20 +427,6 @@ module LavinMQ
                       value: MFile.mmap_count,
                       type:  "gauge",
                       help:  "Number of MFile memory-mapped files"})
-        {% if flag?(:linux) %}
-          if count = count_proc_maps_lines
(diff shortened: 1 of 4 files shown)


jage:
Draft for now, need to do some real world testing of this.

src/lavinmq/http/controller/prometheus.cr
```diff
@@ -405,6 +405,24 @@ module LavinMQ
                         type:   "gauge",
                         help:   "Bytes that hasn't been synchronized with the follower yet"})
         end
+        writer.write({name:  "mfile_count",
+                      value: MFile.mmap_count,
+                      type:  "gauge",
+                      help:  "Number of MFile memory-mapped files"})
+        {% if flag?(:linux) %}
+          if content = File.read?("/proc/self/maps")
```
carlhoerberg: reading the 16th element in `/proc/self/stat` is more efficient. `/proc/self/maps` could be huge if the process has a lot of mmaps. Keeping the file open is another allocation win, especially if you reuse the read buffer. 

note that the second element `/proc/self/stat` can include spaces, so you have to account for that. 

Gemini gave me this: 
```crystal
def mmap_count : Int32
  stat = File.read("/proc/self/stat")
  pos = stat.rindex(')') + 2 # Skip ") "
  
  # We need to skip 13 spaces to reach the 16th field
  13.times do
    pos = stat.index(' ', pos).not_nil! + 1
  end
  
  # Now find the end of that 16th field
  end_pos = stat.index(' ', pos).not_nil!
  
  # Convert the slice directly to an integer
  stat[pos...end_pos].to_i
end
```

And even an allocation free one, that didn't allocate a single `String`. Not sure that's necessary here though.

carlhoerberg: Keeping the `File` open, in an instance variable, and protect it with a mutex, rewind and read, will also save some syscalls and cpu cycles

jage: You sure `/proc/$(pgrep lavinmq)/stat` holds that info? I can't find it in  https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html

carlhoerberg: Youre right, AI slop. It now blames its mistake that there has been kernel patches for it, but never merged into mainline

src/lavinmq/http/controller/prometheus.cr
```diff
@@ -569,6 +587,21 @@ module LavinMQ
           end
         end
       end
+
+      # Count lines in /proc/self/maps without allocating the entire file content.
+      # Reads in chunks and counts newlines incrementally to reduce memory usage.
+      private def count_proc_maps_lines : Int64?
+        File.open("/proc/self/maps", "r") do |file|
```
carlhoerberg: I'm afraid this output is very heavy for the kernel to generate when the number of mmaps are large. We should benchmark it at least.

jage: Good point, I'll try to benchmark and look at the kernel parts as well (user space time shouldn't be a huge issue IMO).

In reality I think this will scrape at most every 15s, probably more like every 60s.

jage: There's a `mmap_lock` (read lock) for the entire read — confirmed in kernel 6.8 fs/proc/task_mmu.c (`m_start`/`m_stop`). Will remove this from LavinMQ.

Will look into monitoring this from otel-collector+eBPF instead, think we can read `mm->map_count` from the kernel (it already tracks this to be able to enforce the max values).

jage:
Think PR is good, but we should have one last round of "thinking" if we need this (or can assume certain usage based on other metrics).

snichme (approved):

PR #1629, opened 2026-01-19
Crystal 1.19.0 compatibility

viktorerlingsson:
Crystal 1.19.0 compatibility

- Upgrade to Crystal 1.19.0
- Switch to Crystal stdlib SNI support (`on_server_name`), removing custom `openssl_sni.cr` extension        
- Unpin Crystal version in CI workflows and Dockerfiles (use `latest`)
- Move `--error-on-warnings` from release builds to CI compile job only
- Replace deprecated `Time.monotonic` with `Time.instant`
- Replace deprecated `Random::DEFAULT` with `Random::Secure` for security-sensitive code and instance `Random.new` for performance tools
- Update `exit` function signature for Crystal 1.19's `Process::Status` type

Maybe rebase merge this to get better history?

Run CI - make sure specs and builds work and use the correct crystal version. 
Should also be properly manually tested before being released, but that can probably be done during the RC stage of LavinMQ v2.7.0

### DIFF
--- a/src/lavinmqperf/amqp/bind_churn.cr
+++ b/src/lavinmqperf/amqp/bind_churn.cr
@@ -8,7 +8,7 @@ module LavinMQPerf
       def run(args = ARGV)
         super(args)
 
-        r = Random::DEFAULT
+        r = Random.new
         ::AMQP::Client.start(@uri) do |c|
           ch = c.channel
           temp_q = ch.queue
--- a/src/lavinmqperf/amqp/queue_churn.cr
+++ b/src/lavinmqperf/amqp/queue_churn.cr
@@ -8,7 +8,7 @@ module LavinMQPerf
       def run(args = ARGV)
         super(args)
 
-        r = Random::DEFAULT
+        r = Random.new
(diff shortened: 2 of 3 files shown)


src/lavinmqperf/amqp/connection_count.cr
```diff
@@ -76,7 +77,7 @@ module LavinMQPerf
 
       private def client : ::AMQP::Client
         client = @client ||= ::AMQP::Client.new(@uri)
-        client.host = "127.0.#{Random.rand(UInt8)}.#{Random.rand(UInt8)}" if @random_localhost
+        client.host = "127.0.#{@random.rand(UInt8)}.#{@random.rand(UInt8)}" if @random_localhost
```
carlhoerberg, line 80: fyi, `rand` and `Random.rand` still exists and are not deprecated, however, it lazy loads a thread local instance of `Random`. This is fine.

jage (approved):
Found a few extra deprecated `Random::DEFAULT`, but think we're good now.

PR #1631, opened 2026-01-20
Fix: Apply stream max-length and max-length-bytes from policy

### LINKED ISSUE #1613: Stream max-length-bytes and max-length policies not enforced (regression in v2.6.0)
The `max-length-bytes` and `max-length` policies are not enforced on stream queues because the values are never propagated to the `StreamMessageStore`.

Introduced in commit b66052e29fab94f704d6e909f13fde66f372917c (Nov 14, 2025) - "Bugfix: don't abort policy applying on invalid value (#1432)"

First released in v2.6.0, version 2.5.5 and earlier are not affected.

## Root Cause

In `src/lavinmq/amqp/stream/stream.cr`, the `apply_policy_argument` method only propagates `max_length_bytes` to the stream message store when `key == "max-age"`:

https://github.com/cloudamqp/lavinmq/blob/80009dcebac7a4e8627f62170ed0f94cdc96c996/src/lavinmq/amqp/stream/stream.cr#L49-L71

When `key == "max-length-bytes"` or `key == "max-length"`:

1. `super(key, value)` is called
2. Parent Queue class sets `@max_length_bytes` / `@max_length` at queue level
3. `stream_msg_store.max_length_bytes` / `stream_msg_store.max_length` remains `nil`
4. `StreamMessageStore#drop_overflow` checks `nil` and skips enforcement

## Impact

- Stream queues grow unbounded despite having `max-length-bytes` or `max-length` policies
- Disk fills up until `free_disk_min` watermark is hit
- No way to reduce stream size via policy changes

## Steps to Reproduce

1. Create a stream queue
2. Publish messages filling multiple segments
3. Apply a `max-length-bytes` or `max-length` policy
4. Observe that segments are NOT removed

## Expected Behavior

Applying a `max-length-bytes` or `max-length` policy should immediately trigger segment cleanup, just like `max-age` does.

jage:
After the refactor in b66052e29fab94f704d6e909f13fde66f372917c, Stream relied on the parent Queue class to handle max-length policies. However, Queue calls drop_overflow which was a no-op in Stream, so limits were never synced to StreamMessageStore and cleanup never triggered.

Handle all three policy types (`max-age`, `max-length`, `max-length-bytes`) directly in Stream, syncing to StreamMessageStore immediately.

Tested this PR on a 2.6.1 instance that got hit by this bug (disk filled up), now it it works as expected.

Screenshot 2026-01-20 at 10 12 18

Fixes #1613

### DIFF
--- a/spec/stream_queue_spec.cr
+++ b/spec/stream_queue_spec.cr
@@ -290,6 +290,38 @@ describe LavinMQ::AMQP::Stream do
       end
     end
 
+    it "removes segments when max-length-bytes policy is applied" do
+      with_amqp_server do |s|
+        with_channel(s) do |ch|
+          args = {"x-queue-type": "stream"}
+          q = ch.queue("stream-max-length-bytes-policy", args: AMQP::Client::Arguments.new(args))
+          data = Bytes.new(LavinMQ::Config.instance.segment_size)
+          2.times { q.publish_confirm data }
+          q.message_count.should eq 2
+          # Apply policy - should immediately trigger cleanup
+          s.vhosts["/"].add_policy("mlb", "stream-max-length-bytes-policy", "queues",
+            {"max-length-bytes" => JSON::Any.new(1_i64)}, 0i8)
+          q.message_count.should eq 1
+        end
+      end
+    end
+
+    it "removes segments when max-length policy is applied" do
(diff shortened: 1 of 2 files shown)


jage:
This should go out in a 2.6.x patch release.

PR #1632, opened 2026-01-20
OAuth2 support

### LINKED ISSUE #857: Oauth2 Support
The lack of OAuth2 support makes integration with modern identity providers like Okta, Auth0, or Azure AD cumbersome. 

LavinMQ should natively support OAuth2 for authentication and authorization. This would allow clients to authenticate seamlessly using industry-standard protocols and integrate directly with OAuth2 providers.

snichme:

Add OAuth2/OpenID Connect authentication to LavinMQ. Clients can now authenticate using JWT tokens (RS256) as the password field. The authenticator validates tokens against the OAuth provider's JWKS endpoint, checks issuer/audience/expiration, and maps scopes to LavinMQ permissions. Supports token refresh via UpdateSecret frame without reconnecting. JWKS public keys are cached with configurable TTL.

Refactors the user system into an abstract BaseUser class with User (password-based) and OAuthUser (token-based with expiration tracking) implementations. 

The authenticator chain is now configurable on this authenticators to chain. 

fixes #857

specs

### DIFF
--- a/spec/auth_spec.cr
+++ b/spec/auth_spec.cr
@@ -261,38 +261,6 @@ describe LavinMQ::Auth::Chain do
         end
       end
 
-      # Fails at the moment, but the code around the expiration loop needs to be refactor anyway.
-      it "resets expiration timer when token_updated receives" do
-        exp = Time.utc + 300.milliseconds
-        verifier = MockVerifier.new(LavinMQ::Config.new, "testuser", nil, nil, exp)
-
-        user = LavinMQ::Auth::OAuthUser.new(
-          "testuser",
-          [] of LavinMQ::Tag,
-          {} of String => LavinMQ::Auth::BaseUser::Permissions,
-          Time.utc + 300.milliseconds,
-          verifier
-        )
-
-        callback_called = false
-
-        user.on_expiration do
-          callback_called = true
-        end
-
-        # Send token update before expiration
-        sleep 70.milliseconds
-
-        user.@token_updated.send nil
-
-        # Wait past the original expiration time
-        sleep 250.milliseconds
-
-        # Callback should not have been called yet since timer was reset
-        # but it will be called eventually (negative token_lifetime triggers immediately)
-        callback_called.should be_false
-      end
-
       it "does not call block before expiration" do
         config = LavinMQ::Config.new
         config.oauth_issuer_url = URI.parse("https://auth.example.com")


src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,205 @@
...
+      # This allows components to access the fetcher without passing it through layers.
+      @@jwks_fetcher : JWKSFetcher?
+
+      def self.jwks_fetcher=(fetcher : JWKSFetcher?)
+        @@jwks_fetcher = fetcher
+      end
+
+      def self.jwks_fetcher : JWKSFetcher?
+        @@jwks_fetcher
+      end
```
viktorerlingsson: These aren't used anywhere as far as I can tell?

kickster97: You're right! fixed that now

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,193 @@
...
+              retry_delay = {retry_delay * 2, max_retry_delay}.min
+            end
+          end
+        end
+
+        def fetch_and_update
+          result = fetch_jwks
+          @public_keys.update(result.keys, result.ttl)
+          Log.info { "Refreshed JWKS with #{result.keys.size} key(s), TTL=#{result.ttl}" }
+        end
```
viktorerlingsson: This happens synchronously on startup if oauth is configured, right? If the configured endpoint is slow or down, that would make startup slower. Could it be done async instead? But maybe it doesn't matter, since you won't be able to auth before this has finished anyway?

src/lavinmq/auth/authenticators/oauth.cr
```diff
@@ -0,0 +1,31 @@
...
+        OAuthUser.new(claims.username, claims.tags, claims.permissions, claims.expires_at, @token_verifier)
+      rescue ex : JWT::PasswordFormatError
+        Log.debug { "skipping authentication for user \"#{context.username}\": " \
+                    "password is not a JWT token" }
+      rescue ex : JWT::DecodeError
+        Log.debug { "authentication failed for user \"#{context.username}\": " \
+                    "Could not decode token - #{ex.message}" }
+      rescue ex : JWT::VerificationError
+        Log.debug { "authentication failed for user \"#{context.username}\": " \
+                    "Token verification failed - #{ex.message}" }
```
spuun, line 29: Should these be just debug? Hard for an operator to answer anything about why auth failed if this is hidden down in debug level.

Or should we change the behavior of authenticators to raise instead, so the `ConnectionFactory` can rescue and log? Or change the value of `#authenticate` to `Tuple(User?, String?)` where String is an error message? Or something else? But refactoring can be for another PR.

src/lavinmq/auth/chain.cr
```diff
@@ -1,23 +1,48 @@
...
+        authenticators = [] of Authenticator
+        config.auth_backends.each do |backend|
+          case backend
+          when "local"
+            authenticators << LocalAuthenticator.new(users)
+          when "oauth"
+            begin
+              # We want to make sure the Oauth server is alive before creating the authenticator
+              jwks_fetcher.fetch_and_update
+              spawn jwks_fetcher.refresh_loop, name: "JWKS refresh"
```
spuun: How can this fiber be stopped?

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,193 @@
...
+          @refresh_trigger = Channel(Nil).new
+        end
+
+        def refresh_loop
+          retry_delay = 5.seconds
+          max_retry_delay = 5.minutes
+
+          loop do
+            begin
+              retry_delay = 5.seconds
```
spuun: Shouldn't this be moved to after `fetch_and_update`? Only reset the retry delay after a successful fetch?

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,193 @@
...
+          @public_keys.update(result.keys, result.ttl)
+          Log.info { "Refreshed JWKS with #{result.keys.size} key(s), TTL=#{result.ttl}" }
+        end
+
+        private def calculate_wait_time : Time::Span
+          if expires_at = @public_keys.expires_at
+            remaining = expires_at - RoughTime.utc
+            return remaining if remaining > 0.seconds
+          end
+          5.seconds # Default minimum wait if keys expired or not set
```
spuun: Maybe `expires_at` is always set, but this feels very low? If it for some reason isn't set, should we really fetch new ever 5 seconds?

src/lavinmq/auth/jwt/token_verifier.cr
```diff
@@ -0,0 +1,123 @@
...
+          raise JWT::DecodeError.new("Missing or invalid iss claim in token") unless issuer
+
+          if issuer.chomp("/") != @expected_issuer
+            raise JWT::VerificationError.new("Token issuer does not match the expected issuer")
+          end
+        end
+
+        private def validate_audience(payload)
+          aud = payload["aud"]?
+          return unless aud # No audience in token, nothing to validate
```
viktorerlingsson: If `oauth_verify_aud` is set, but `payload["aud"]` isn't set, should we really return here?

src/lavinmq/auth/jwt/jwt.cr
```diff
@@ -0,0 +1,145 @@
...
+          payload_json = base64url_decode(parts[1])
+          signature = base64url_decode_bytes(parts[2])
+
+          header = JSON.parse(header_json)
+          payload = JSON.parse(payload_json)
+
+          # Verify algorithm
+          alg = header["alg"]?.try(&.as_s)
+          raise DecodeError.new("Missing algorithm in header") unless alg
+          raise DecodeError.new("Expected RS256, got #{alg}") unless alg == "RS256"
```
viktorerlingsson, line 132: This alg header is checked both here and in `prevalidate_token`, do we need to do that twice?

src/lavinmq/auth/oauth_user.cr
```diff
@@ -0,0 +1,57 @@
...
+      def initialize(@name : String, @tags : Array(Tag), @permissions : Hash(String, Permissions),
+                     @expires_at : Time, @verifier : JWT::TokenVerifier)
+      end
+
+      def token_lifetime : Time::Span
+        @expires_at - RoughTime.utc
+      end
+
+      def on_expiration(&block)
+        spawn(name: "OAuthUser#on_expiration") do
```
viktorerlingsson, line 22: this will use an extra memory map for each connection, have we considered any other alternatives? (Like one central fiber doing expiration for all oauth connections or something like that?)

src/lavinmq/auth/jwt/token_verifier.cr
kickster97: I now added so it will fail with verification error, as it should!

src/lavinmq/auth/oauth_user.cr
```diff
@@ -0,0 +1,57 @@
...
+        if claims.username != @name
+          raise JWT::VerificationError.new("Token username mismatch: expected '#{@name}', got '#{claims.username}'")
+        end
+
+        # Update authorization and expiration (trust new token)
+        @tags = claims.tags
+        @permissions = claims.permissions
+        @expires_at = claims.expires_at
+        clear_permissions_cache
+        @token_updated.send nil
```
viktorerlingsson, line 54: these updates aren't atomic, but since there's no yielding going on it should be safe. Could be an issue if/when we go multi-threaded though? We maybe don't need to do anything about it now, but good to know for the future.

src/lavinmq/amqp/client.cr
```diff
@@ -66,6 +66,12 @@ module LavinMQ
         @vhost.add_connection(self)
         @log.info { "Connection established for user=#{@user.name}" }
         spawn read_loop, name: "Client#read_loop #{@connection_info.remote_address}"
+        case user = @user
+        when Auth::OAuthUser
+          user.on_expiration do
+            close_connection(nil, ConnectionReplyCode::CONNECTION_FORCED, "token expired")
+          end
+        end
```
viktorerlingsson, line 74: should this exist for mqtt clients as well?

viktorerlingsson:
Added some comments and questions. In general I think this looks great though!👍

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -119,8 +119,6 @@ module LavinMQ
         user = @authenticator.authenticate(context)
         return user if user
 
-        log.warn { "Authentication failure for user \"#{username}\"" }
```
spuun, line 122: No? Not even as `info`?

src/lavinmq/auth/jwt/jwks_fetcher.cr
spuun: We could just trigger `fetch_and_update` before the loop in `#refresh_loop` and it will done async.

snichme: The reason we chose to have this synchronously is to make sure that the keys have been downloaded and `@public_keys` was updated before the authenticator started. 
If we do it async and the server is slow or not responding the authenticator will start up in an invalid state and not allowing anyone to authenticate and no logs about why. 

I'm not sure which is best/worst but it's something we should make a decision on

src/lavinmq/amqp/client.cr
snichme: Can an mqtt client expire?

src/lavinmq/auth/oauth_user.cr
snichme: How can we make it atomic?

src/lavinmq/auth/oauth_user.cr
snichme: Good idea, I will do an experiment

src/lavinmq/auth/jwt/jwt.cr
snichme: @kickster97 do you know?

kickster97: Its not needed, but it is convenient for the specs to have it like this.

viktorerlingsson: okay 👍

src/lavinmq/auth/oauth_user.cr
viktorerlingsson: we could use a mutex and lock both here and anywhere we get these values. But maybe not worth it? As stated it shouldn't be an issue now since we only run one thread and there's no yield.

src/lavinmq/amqp/client.cr
viktorerlingsson: mqtt clients can be pretty long-lived I guess, so if they are able to auth with OAuth, then they should also expire when a token expires I think. Otherwise they could stay connected with an expired token, and that kinda defeats the purpose?

snichme: We do use the same authenticator for amqp and mqtt so yes, they should expire as well. 
Unless we change to only use the local authenticator for mqtt, if that would make more sense.

snichme: @spuun WDYT about connecting mqtt clients using oauth2, something you heard being requested?

src/lavinmq/amqp/connection_factory.cr
kickster97: yes maybe should be info atleast

src/lavinmq/auth/jwt/jwks_fetcher.cr
snichme: And should we only return `remaining` is that is also above 5 seconds, so we have a common fastest time. Now `remaining` can be anyone greater than 0.

src/lavinmq/auth/authenticators/oauth.cr
snichme: This comes from how the Authenticator chain works, if the first one, which commonly is the `LocalAuthenticator`, fails it will try this. 
So if you try to authenticate with a local user but with a wrong password you will end up in any of these. 

Suggestion is to make the `LocalAuthenticator` fail if username matches a user but not the password. And than that would "break the chain". 
Agree that it could be nice to not need debug log to see this but dont want to log these on any login attempts.

kickster97: In irl discussion we landed on logging info level for authentication failures and debug on the rest. I added back the info log line in connection_factory, so now we log on that level, and debug here.

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,229 @@
...
+            loop do
+              begin
+                break if @stopped
+                result = fetch_jwks
+                @public_keys.update(result.keys, result.ttl)
+                Log.info { "Refreshed JWKS with #{result.keys.size} key(s), TTL=#{result.ttl}" }
+                retry_delay = 5.seconds
+                wait_time = calculate_wait_time
+                select
+                when @refresh_trigger.receive
```
kickster97: Do we ever send on this channel?

src/lavinmq/auth/jwt/jwks_fetcher.cr
kickster97: No, i agree, I think 1 hour would be more resonable.. And if we don't require the expiration to be set, this maybe should be configurable?

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,229 @@
...
+        end
+
+        def cleanup
+          @stopped = true
+        end
+
+        def start_refresh_loop
+          retry_delay = 5.seconds
+          max_retry_delay = 5.minutes
+          spawn do
```
spuun, line 65: This fiber won't stop until a new refresh is happening, no?

Maybe `@stopped` can be a `BoolChannel` which then also be used in the select statements in the loop to instantly stop the loop?

snichme: Makes sense. I also see that the `@refresh_trigger` channel isn't used. Where did the usage of that go.  Do you remember why we added that  one @kickster97 ?

src/lavinmq/amqp/client.cr
carlhoerberg: i don't think there's a mechanism for MQTT clients to reauthenticate without disconnecting and reconnecting

src/lavinmq/amqp/client.cr
```diff
@@ -447,6 +470,10 @@ module LavinMQ
         @exclusive_queues.each(&.close)
         @exclusive_queues.clear
         @vhost.rm_connection(self)
+        case user = @user
+        when Auth::OAuthUser
+          user.cleanup
+        end
```
carlhoerberg, line 482:
```suggestion
        @user.as?(Auth::OAuthUser).try &.cleanup
```

src/lavinmq/server.cr
```diff
@@ -91,10 +95,12 @@ module LavinMQ
       Dir.mkdir_p @data_dir
       Schema.migrate(@data_dir, @replicator)
       @users = Auth::UserStore.new(@data_dir, @replicator)
-      authenticator = Auth::Chain.create(@users)
+      @jwks_fetcher = Auth::JWT::JWKSFetcher.new(@config.oauth_issuer_url, @config.oauth_jwks_cache_ttl)
```
carlhoerberg: we always instantiate it? even if `oauth_issuer_url` is nil?

src/lavinmq/config/options.cr
```diff
@@ -342,6 +342,23 @@ module LavinMQ
       def amqp_default_consumer_prefetch=(value)
         @default_consumer_prefetch = value
       end
+
+      @[IniOpt(section: "oauth", ini_name: issuer)]
+      property oauth_issuer_url : URI = URI.new
```
carlhoerberg:
```suggestion
      property oauth_issuer_url : URI? = nil
```

src/lavinmq/auth/jwt/jwks_fetcher.cr
carlhoerberg: One single idling fiber is not a big cost. But can they accumulate? Eg. if you change the JWTS URI in the config and reload repeatedly?

src/lavinmq/auth/jwt/jwks_fetcher.cr
carlhoerberg: Nope doesn't look like it. Remove?

src/lavinmq/auth/jwt/jwks_fetcher.cr
carlhoerberg: everything doesn't have to configurable, i'm all for a "good enough" default. the sheer number of config options can be overwhelming for beginners.

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -0,0 +1,229 @@
...
+          public_keys = extract_public_keys_from_jwks(jwks)
+          ttl = extract_jwks_ttl(headers)
+          JWKSResult.new(public_keys, ttl)
+        end
+
+        private def extract_public_keys_from_jwks(jwks : JWKSResponse)
+          jwks_array = jwks.keys
+
+          public_keys = {} of String => String
+          jwks_array.each_with_index do |key, idx|
```
carlhoerberg, line 119:
```suggestion
          public_keys = Hash(String, String).new
          jwks.keys.each_with_index do |key, idx|
```

src/lavinmq/auth/base_user.cr
```diff
@@ -0,0 +1,82 @@
...
+          write:     p[:write],
+        }
+      end
+
+      def can_write?(vhost : String, name : String, cache : PermissionCache) : Bool
+        permission_revision = @permission_revision.lazy_get
+        if permission_revision != cache.revision
+          cache.clear
+          cache.revision = permission_revision
+        end
```
carlhoerberg, line 41:
```suggestion
        if @permission_revision.swap(cache.revision, :relaxed) != cache.revision
          cache.clear
        end
```

src/lavinmq/auth/chain.cr
```diff
@@ -1,23 +1,41 @@
...
+            verifier.fetcher.start_refresh_loop
+            authenticators << OAuthAuthenticator.new(verifier)
+          else
+            raise "Unsupported authentication backend: #{backend}"
+          end
+        end
+        if authenticators.empty?
+          # Default to local auth if no backends configured
+          authenticators << LocalAuthenticator.new(users)
+        end
```
carlhoerberg, line 40:
```suggestion
        [LocalAuthenticator.new(users)] if backends.empty?
        backends.map do |backend|
          case backend
          when "local" then LocalAuthenticator.new(users)
          when "oauth"
            verifier.fetcher.start_refresh_loop
            OAuthAuthenticator.new(verifier)
          else
            raise "Unsupported authentication backend: #{backend}"
          end
        end
```

src/lavinmq/auth/chain.cr
```diff
@@ -1,23 +1,41 @@
...
-        authenticators = [LocalAuthenticator.new(users)] of Authenticator
-        new(authenticators)
+      def self.create(backends : Array(String), users : UserStore, verifier : JWT::TokenVerifier) : Chain
+        authenticators = [] of Authenticator
+        backends.each do |backend|
+          case backend
+          when "local"
+            authenticators << LocalAuthenticator.new(users)
+          when "oauth"
+            verifier.fetcher.start_refresh_loop
```
carlhoerberg: we have a `verfier` even if no oauth backend is configured? Should we create one here instead?

src/lavinmq/server.cr
```diff
@@ -47,10 +48,12 @@ module LavinMQ
       @vhosts = VHostStore.new(@data_dir, @users, @replicator)
       @mqtt_brokers = MQTT::Brokers.new(@vhosts, @replicator)
       @parameters = ParameterStore(Parameter).new(@data_dir, "parameters.json", @replicator)
-      @authenticator = Auth::Chain.create(@users)
+      @jwks_fetcher = Auth::JWT::JWKSFetcher.new(@config.oauth_issuer_url, @config.oauth_jwks_cache_ttl)
```
carlhoerberg: I think we should rethink this more. Do we really need to instantiate a "fetcher" and "verifier" if no oauth endpoint is configured?

src/lavinmq/auth/oauth_user.cr
snichme: Moving this to a separate PR

PR #1636, opened 2026-01-20
Deprecate duplicate INI options, forward to [main]

viktorerlingsson:
## Summary
- Reverts the multi-section INI option approach in favor of deprecation
- Adds deprecated options for settings that were previously configurable in multiple INI sections
- Deprecated options forward their values to the primary option in `[main]` and log a warning

## Deprecated options

| Option | Deprecated Section | Primary Section |
|--------|-------------------|-----------------|
| `tls_cert` | amqp | main |
| `tls_key` | amqp | main |
| `tls_cert` | mgmt | main |
| `tls_key` | mgmt | main |
| `set_timestamp` | amqp | main |
| `consumer_timeout` | amqp | main |
| `default_consumer_prefetch` | amqp | main |

When users configure these options in the deprecated sections, they'll see a warning like:
```
Config tls_cert is deprecated, use tls_cert in [main] instead
```

### DIFF
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -291,6 +291,57 @@ module LavinMQ
 
       @[IniOpt(section: "amqp")]
       property max_consumers_per_channel = 0
+
+      # Deprecated options - these forward to the primary option in [main]
+
+      @[IniOpt(ini_name: tls_cert, section: "amqp", deprecated: "tls_cert in [main]")]
+      @amqp_tls_cert = ""
+
+      def amqp_tls_cert=(value)
+        @tls_cert_path = value
+      end
+
+      @[IniOpt(ini_name: tls_key, section: "amqp", deprecated: "tls_key in [main]")]
+      @amqp_tls_key = ""
+
+      def amqp_tls_key=(value)
+        @tls_key_path = value
+      end
+
+      @[IniOpt(ini_name: tls_cert, section: "mgmt", deprecated: "tls_cert in [main]")]
+      @mgmt_tls_cert = ""
+
+      def mgmt_tls_cert=(value)
+        @tls_cert_path = value
+      end
+
+      @[IniOpt(ini_name: tls_key, section: "mgmt", deprecated: "tls_key in [main]")]
+      @mgmt_tls_key = ""
+
+      def mgmt_tls_key=(value)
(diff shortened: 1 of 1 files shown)


viktorerlingsson:
@spuun  @kickster97 WDYT of making it possible to set multiple sections for an `IniOpt` like this?

spuun:
It's actually possible to handle it already like this:
```crystal
@[IniOpt("section": "main")]
property option = "default_value"

# Should probably make sure ini_name value must be a string:
@[IniOpt("section": "amqp", ini_name: option, deprecated: "option")] 
@dummy_option = "default_value"
def dummy_option=(value)
  option = value
end
```
then tweak the deprecation message to say `<option x> in <section> is deprecated, use <option y> in <other section>`

I don't think we want the possibility to configure it from multiple sections, all but one should be deprecated? 🤔

viktorerlingsson:
> It's actually possible to handle it already like this:

Yes, this PR was an attempt at making something more generic that is simpler to configure so we don't need to write the same code multiple times. 
I'd also like to make sure that we don't overwrite the value if you set it both in main and amqp for example.

> I don't think we want the possibility to configure it from multiple sections, all but one should be deprecated? 🤔

Hmm, yeah maybe that's better. But then I think we need to make a decision that the options mentioned in the description here are actually meant to be deprecated from the extra sections? Before #917 they weren't deprecated, at least not "officially".

spuun:
> Hmm, yeah maybe that's better. But then I think we need to make a decision that the options mentioned in the description here are actually meant to be deprecated from the extra sections? Before #917 they weren't deprecated, at least not "officially".

To me it doesn't make sense to be able to set it from multiple places. 

According to comments in the old code `tls_cert`/`tls_key` under `[amqp]` was kept for backward compatibility, so i guess they can be deprecated? Same for `tls_cert`/`tls_key` under `[mgmt]`.

viktorerlingsson:
> To me it doesn't make sense to be able to set it from multiple places.
> 
> According to comments in the old code `tls_cert`/`tls_key` under `[amqp]` was kept for backward compatibility, so i guess they can be deprecated? Same for `tls_cert`/`tls_key` under `[mgmt]`.

Yeah, agreed. Updated this PR to instead deprecate the duplicate INI options!

PR #1637, opened 2026-01-21
Fixup: Add support for Sec-WebSocket-Protocol to HTTP::WebSocketHandler

spuun:
This fixes some parts that #1621 missed.

The specs added in #1621 did only verify that the websocket handler picked the right client handler based on `Sec-WebSocket-Protocol`. What they didn't verify was if `Sec-WebSocket-Protocol` was sent back to the client.

Because of how Crystal's `HTTP::WebSocketHandler` and `HTTP::Server:Response#upgrade` works I just copied `HTTP::WebSocketHandler` to make my own fixed version. https://github.com/crystal-lang/crystal/pull/16574 would fix this in the stdlib.

Run specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Add Sec-WebSockets-Protocol to websocket responses [#1621](https://github.com/cloudamqp/lavinmq/pull/1621), [#1637](https://github.com/cloudamqp/lavinmq/pull/1637)
 - A delayed exchange didn't deliver messages on time [#1600](https://github.com/cloudamqp/lavinmq/pull/1600)
 
 ## [2.7.0-alpha.1] - 2025-12-07


viktorerlingsson:
I don't know if it matters much in this context, but we parse through the headers twice for the same info now (in `handle_sub_protocol` and `pick_protocol` (and basically duplicate the same string check). Could we get away with doing that just once and passing the information instead? 

Besides that, I think it looks good!

spuun:
> I don't know if it matters much in this context, but we parse through the headers twice for the same info now (in `handle_sub_protocol` and `pick_protocol` (and basically duplicate the same string check). Could we get away with doing that just once and passing the information instead?
> 
> Besides that, I think it looks good!

Yeah, I know. I tried to keep it simple and wait for https://github.com/crystal-lang/crystal/pull/16574 , but we might as well do it better and see what that PR ends up with.

src/lavinmq/http/handler/websocket.cr
```diff
@@ -1,22 +1,86 @@
...
       AMQP
     end
 
-    def self.new(server : Server)
-      ::HTTP::WebSocketHandler.new do |ws, ctx|
-        req = ctx.request
-        protocol, header_value = pick_protocol(req)
+    SUPPORTED_PROTOCOLS = {
+      /^amqp/ => Protocol::AMQP,
+      /^mqtt/ => Protocol::MQTT,
```
viktorerlingsson: This will be case-sensitive as compared to the earlier `/^amqp/i`, right? maybe doesn't matter?

viktorerlingsson (approved):
LGTM! 👍

src/lavinmq/http/handler/websocket.cr
spuun: Right! Maybe it should remain case insensitive? hm...

spuun: Lets stick with case insensitive because why not

PR #1638, opened 2026-01-21
Refactor: break out more stuff from Queue to "Argument"

spuun:
Before breaking out more "extension" to their own classes, I felt it was in need to do some more work with dead lettering to set a standard.

Run specs

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -323,11 +323,14 @@ module LavinMQ::AMQP
           return true
         end
       else
-        # Instead of passing @effective_args here we can remove "x-#{key}" from
-        # effective_args if apply_policy_argument returns true. At least for now
-        # it seems like the policy argument and queue argument is the same except
-        # for the "x-" prefix.
-        return true if @dead_letter.apply_policy_argument(key, value, @effective_args)
+        # If something can be configured by both policy arguments and queue
+        # arguments, we assume that the keys are the same with an "x-" prefix
+        # for queues, e.g. policy "max-length" queue "x-max-length". So if the
+        # policy argument is applied we can remove the key from effective args.
+        if @dead_letter.apply_policy_argument(key, value)
+          @effective_args.delete("x-#{key}")
+          return true
+        end
       end
       false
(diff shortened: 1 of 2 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -334,6 +322,12 @@ module LavinMQ::AMQP
           @effective_args.delete("x-consumer-timeout")
           return true
         end
+      else
+        # Instead of passing @effective_args here we can remove "x-#{key}" from
+        # effective_args if apply_policy_argument returns true. At least for now
+        # it seems like the policy argument and queue argument is the same except
+        # for the "x-" prefix.
+        return true if @dead_letter.apply_policy_argument(key, value, @effective_args)
```
spuun: Thoughts on this?

viktorerlingsson: I think the comment sounds good!

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -334,6 +322,15 @@ module LavinMQ::AMQP
           @effective_args.delete("x-consumer-timeout")
           return true
         end
+      else
+        # If something can be configured by both policy arguments and queue
+        # arguments, we assume that the keys are the same with an "x-" prefix
+        # for queues, e.g. policy "max-length" queue "x-max-length". So if the
+        # policy argument is applied we can remove the key from effective args.
+        if @dead_letter.apply_policy_argument(key, value)
```
snichme, line 330: Would it be a nicer solution to do 
```suggestion
        @dead_letter.apply_policy_argument(key, value)
        if @dead_letter.active?
```

would that be doable?

spuun: Not following how you mean. Feels like that would mean something different? Also, it may be active already before the call tp `#apply_policy_argument`. 
However, maybe `apply_policy_argument` should be renamed to `apply_policy_argument?`.

snichme: Looked at this again, I think it's good as it is :)

PR #1639, opened 2026-01-21
typescript

snichme:

Migrate all JS to TS


snichme:
Will not migrate to typescript now. Maybe in the future when there is more apps or logic in the frontend.

PR #1641, opened 2026-01-22
Server-side auth validation for views

spuun:
Move the very simple js authentication check in views to backend.

This could be improved by doing proper authorization, but that would imply doing it for each view request which would cause some extra load.

Run specs (including playwright)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add Sec-WebSockets-Protocol to websocket responses [#1621](https://github.com/cloudamqp/lavinmq/pull/1621), [#1637](https://github.com/cloudamqp/lavinmq/pull/1637)
 - A delayed exchange didn't deliver messages on time [#1600](https://github.com/cloudamqp/lavinmq/pull/1600)
 
+### Changed
+- Basic authentication check moved to backend to get rid of inline javascript [#1641](https://github.com/cloudamqp/lavinmq/pull/1641)
+
 ## [2.7.0-alpha.1] - 2025-12-07
 
 This release introduces the ability to restart closed queues, enhanced TLS capabilities including SNI and mTLS support, improved stream performance, improved clustering and federation management, and various performance optimizations.


src/lavinmq/http/controller/views.cr
```diff
@@ -62,14 +63,22 @@ module LavinMQ
...
-            context.response.headers.add("Content-Security-Policy", "default-src 'none'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; script-src 'self' 'sha256-9nCxy0qjWUXfAqDk5MjMKgu+tHDTvI8ZUAmbmYoCEF8='")
+            context.response.headers.add("Content-Security-Policy", "default-src 'none'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; script-src 'self'")
             {{ block.body if block }}
             render {{ view }}
           end
           context
         end
       end
 
+      macro redirect_unless_logged_in!
```
carlhoerberg, line 74: Does this have to be a macro?

src/lavinmq/http/controller/views.cr
```diff
@@ -48,9 +48,10 @@ module LavinMQ
       #   context.response.content_type = "application/json"
       # end
       # ```
-      macro static_view(path, view = nil, &block)
+      macro static_view(path, *, auth_required = true, view = nil, &block)
         {% view = path[1..] if view.nil? %}
         get {{ path }} do |context, params|
+          {% if auth_required %} redirect_unless_logged_in! {% end %}
```
carlhoerberg: Cant be a normal if?

src/lavinmq/http/controller/views.cr
spuun: Nope

src/lavinmq/http/controller/views.cr
spuun: I guess. It will be removed by the compiler anyway I guess.

src/lavinmq/http/controller/views.cr
spuun: Actually, yes. At least to not have to add support for `HaltRequest` in view controller etc. That can be an exercise in an upcoming PR.

spuun:
> Looks good to me! But I guess we need to add the cookie to the tests in `static_requests.sh` as well?
> 
> ```
> local body_size=$(curl --silent --output /dev/null --write-out --cookie "m=|:Z3Vlc3Q6Z3Vlc3Q%3D" \
>   "%{size_download}\n" "http://127.0.0.1:${PORT}/${path}")
> ```

Hm, it's the request to `/` that fails, which it should. I guess that request is to be removed.

views/partials/head.ecr
```diff
@@ -1,7 +1,3 @@
-<script>
-  if (!document.cookie.startsWith("m=") && document.cookie.indexOf("; m=") === -1)
-    window.location.assign("login")
```
carlhoerberg, line 3: the client side redirect shouldn't be moved to `getCookie` in `auth.js`, if the cookie is missing?

spuun: I don't think we'll need any client side check? But sure, we can do something in the `getAuth` method in `auth.js`.

PR #1643, opened 2026-01-23
oauth2: Use structs instead of JSON::Any

spuun:

Use structs instead of `JSON::Any`. The codes get cleaner without all string indexing and `as_X?` calls.

Many properties are nilable now, but maybe they don't have to be that?

Specs

### DIFF
--- a/src/lavinmq/auth/jwt/jwt.cr
+++ b/src/lavinmq/auth/jwt/jwt.cr
@@ -85,6 +85,8 @@ module LavinMQ
             [@aud]
           in Array(String)
             @aud
+          in Nil
+            nil
           end
         end
       end


src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -15,32 +15,50 @@ module LavinMQ
...
-              return username
+            # Check standard JWT properties first
+            username = case claim
+                       when "sub"
+                         payload.sub
+                       when "iss"
+                         payload.iss
+                       else
+                         nil
+                       end
```
spuun, line 34: May `claim` ever be `sub` or `iss`? If not, this case statement can be removed.

src/lavinmq/auth/jwt/token_verifier.cr
```diff
@@ -86,23 +86,26 @@ module LavinMQ
...
-                      when .as_a? then aud.as_a.map(&.as_s)
-                      when .as_s? then [aud.as_s]
-                      else             return
+                      in String
+                        [aud]
+                      in Array(String)
+                        aud
+                      in Nil
+                        return
                       end
```
spuun: Maybe this can be moved into `Payload`

src/lavinmq/auth/jwt/token_claim.cr
snichme: No idea. based on the previous code it could be any key so better leave it like that.

src/lavinmq/auth/jwt/token_verifier.cr
```diff
@@ -59,25 +59,25 @@ module LavinMQ
...
+          alg = header.alg
           raise JWT::DecodeError.new("Missing algorithm in header") unless alg
           raise JWT::DecodeError.new("Expected RS256, got #{alg}") unless alg == "RS256"
 
           payload_str = JWT::RS256Parser.base64url_decode(parts[1])
-          payload = JSON.parse(payload_str)
-          exp = payload["exp"]?.try(&.as_i64?)
+          payload = JWT::Payload.from_json(payload_str)
+          exp = payload.exp
           raise JWT::DecodeError.new("Missing exp claim in token") unless exp
```
snichme, line 69: @kickster97 according to the specs https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4 the `exp` field is optional but here we raise if we cannot find it.
It could make sense to not allow a claim without expiration time but we must be clear that either we follow the specs or we have a good reason not to. 
Do you remember why we require it here?

kickster97: Good question! I think its good to keep the validation since we're implementing OAuth 2.0/OIDC authentication, not just a generic JWT validator. 
With that said maybe this check is in the wrong place, and should not be linked to the JWT code? rather when we validate with the public keys, it seems like we have had a check there previously if we look at the code in public_keys.cr:48

src/lavinmq/auth/jwt/token_claim.cr
kickster97: you can set the lavinmq configs oauth_preferred_username_claims to `sub` or `iss`, and then we will match against that in the payload. so its not that the claim actually is "sub" or "iss", its that we look for those fields in the payload and use the value we have there. so for me this feels redundant?

src/lavinmq/auth/jwt/token_verifier.cr
kickster97: Its also motivated by the way RabbitMQ handles the JWT validation 
https://www.rabbitmq.com/docs/3.13/oauth2#:~:text=Must%20not%20be%20expired%E2%80%8B,NOT%20be%20accepted%20for%20processing. 

(I also feel like this can motivate that it is in the right place as well, in `prevalidate_token()`, so i dont think we need changes here)

src/lavinmq/auth/jwt/token_claim.cr
spuun: Yeah, but that's what this code does. `sub` and `iss` must be handled this way since they are properties in the struct now and not part of a "generic hash" which we can look into.

I guess my questions is if one ever would put `sub` or `iss` in `oauth_preferred_username_claims`, or if these fields maybe always is used for something else?

spuun: I guess `sub` but not `iss` may be used. Well, keeping the code like this makes anything possible.

PR #1645, opened 2026-01-23
Fix OAuth2 scope format to match RabbitMQ spec

kickster97:
 This PR fixes our OAuth2 scope parsing to match the RabbitMQ OAuth2 specification. Previously we were using a different format that wasn't compatible with how RabbitMQ handles JWT token scopes.

RabbitMQ uses the format `permission:vhost/pattern` (with a slash separator).

I added proper wildcard-to-regex conversion. Before, we were just passing wildcards through directly, but now we properly escape special regex characters and anchor the patterns. This means `queue*` becomes `^queue.*$` so it only matches resources starting with "queue" rather than matching anywhere in the string.

The RabbitMQ spec requires URL-encoding for special characters in patterns, so we now decode patterns like `queue%2A` back to `queue*` before processing them. This lets you distinguish between a wildcard `*` and a literal asterisk.

For OAuth users specifically, we added support for wildcard vhost patterns. If your token has permissions for `dev-*`, that will now match vhosts like `dev-test` or `dev-prod`. This makes it easier to grant permissions across multiple environments without listing each one explicitly.

  - [RabbitMQ OAuth2 Documentation](https://www.rabbitmq.com/docs/oauth2)

All the tests have been updated to use the correct RabbitMQ scope format.

### DIFF
--- a/src/lavinmq/auth/jwt/token_claim.cr
+++ b/src/lavinmq/auth/jwt/token_claim.cr
@@ -127,6 +127,9 @@ module LavinMQ
           return if !perm_type.in?("read", "write", "configure") || remainder.empty?
 
           parts = remainder.split("/")
+
+          # Allow 2 or 3 parts: vhost/pattern or vhost/pattern/routing_key
+          # RabbitMQ allows a third part (routing_key), but we don't use it
           return unless parts.size == 2 || parts.size == 3
 
           vhost = URI.decode_www_form(parts[0])


src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -104,37 +103,39 @@ module LavinMQ
...
-          perm_type, vhost, pattern = parts[0], parts[1], parts[2]
-          return if !perm_type.in?("configure", "read", "write")
+          perm_type, _, remainder = role.partition(":")
+          return if !perm_type.in?("read", "write", "configure") || remainder.empty?
+
+          parts = remainder.split("/")
+          return unless parts.size == 2 || parts.size == 3
 
-          pattern = ".*" if pattern == "*"
+          vhost, pattern = parts[0], parts[1]
```
viktorerlingsson: should `vhost` also be `URI.decode_www_form`-ed?

src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -104,37 +103,39 @@ module LavinMQ
...
-            Log.warn { "Skipping scope '#{role}': Expected format 'permission:vhost:pattern'" }
-            return
-          end
-          perm_type, vhost, pattern = parts[0], parts[1], parts[2]
-          return if !perm_type.in?("configure", "read", "write")
+          perm_type, _, remainder = role.partition(":")
+          return if !perm_type.in?("read", "write", "configure") || remainder.empty?
+
+          parts = remainder.split("/")
+          return unless parts.size == 2 || parts.size == 3
```
viktorerlingsson, line 133: We don't seem to handle `parts[2]` anywhere. Should we? If not, maybe that should be documented here

kickster97: yeah.. that is for topic exchanges.. Rabbitmq allows for parsing that as well as part of the "remainder"
.. but i'm leaning towards not having it

src/lavinmq/auth/jwt/token_claim.cr
kickster97: ah yeah i think your right!

src/lavinmq/auth/jwt/token_claim.cr
viktorerlingsson: Okay. Then I think just adding a little comment about that would be good!

src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -104,37 +103,39 @@ module LavinMQ
...
+          vhost = URI.decode_www_form(parts[0])
+          pattern = URI.decode_www_form(parts[1])
+          regex_pattern = wildcard_to_regex(pattern)
 
-          permissions[vhost] ||= {
-            config: Regex.new("^$"),
-            read:   Regex.new("^$"),
-            write:  Regex.new("^$"),
-          }
+          permissions[vhost] ||= {config: Regex.new("^$"), read: Regex.new("^$"), write: Regex.new("^$")}
```
snichme: Instead of this we could do 
```crystal 
          permissions = Hash(String, User::Permissions).new do |h, k|
            h[k] = {config: Regex.new("^$"), read: Regex.new("^$"), write: Regex.new("^$")}
          end
```
when constructing the hash.

src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -104,37 +103,39 @@ module LavinMQ
...
-                               when "config"    then permissions[vhost].merge({config: regex})
                                when "read"      then permissions[vhost].merge({read: regex})
                                when "write"     then permissions[vhost].merge({write: regex})
                                else                  permissions[vhost]
                                end
         end
+
+        private def wildcard_to_regex(pattern : String) : String
+          return ".*" if pattern == "*"
+          escaped = pattern.gsub(/[.+?^${}()|[\]\\]/) { |m| "\\#{m}" }
```
snichme: Could we have comment about this `gsub`? Hard to grasp what it means.

src/lavinmq/auth/oauth_user.cr
```diff
@@ -52,6 +52,16 @@ module LavinMQ
...
         @token_updated.send nil
       end
+
+      def find_permission(vhost : String) : Permissions?
+        permissions[vhost]? || permissions.find { |pattern, _| wildcard_match?(pattern, vhost) }.try(&.[1])
+      end
+
+      private def wildcard_match?(pattern : String, value : String) : Bool
+        return true if pattern == "*"
+        regex = pattern.gsub(/[.+?^${}()|[\]\\]/) { |m| "\\#{m}" }.gsub("*", ".*")
```
snichme: same here, a comment on what this means.

src/lavinmq/auth/jwt/token_claim.cr
kickster97: I'm open, but why? if we do this we will overwrite the permissions with a new permissions hash for each role that we parse

snichme (approved):

src/lavinmq/auth/jwt/token_claim.cr
kickster97: Ah when we construct! read too quickly, yes that works.

PR #1650, opened 2026-01-28
Fix: Register log system on config parse (startup)

### LINKED ISSUE #1648: Live logs broken since config rewrite
The `/api/livelog` endpoint returns no data because `Log::InMemoryBackend` is never registered with the logging system on startup.

**Cause:**

https://github.com/cloudamqp/lavinmq/pull/917 removed the `reload_logger` call from `Config#parse`. It's now only called on SIGHUP, not during initial startup.

**Fix:**

Add `reload_logger` at the end of `Config#parse`:

```crystal
def parse(argv = ARGV)
  # ...
  parse_cli(argv)
  reload_logger
end
```

jage:
In #917 reload_logger was removed from Config#parse, thus logging wasn't setup properly.

When fixing this the specs broke due to `Log.capture` being unable to verify warnings since Log.setup replaced all backends now that parse actually restarted the logging, thus fixing the Log.setup(...) override in spec_helper.

Also noticed that we now try to log `/var/log/lavinmq.log` for real which in most systems isn't allowed, use `/tmp` instead.

Close #1648

### DIFF
--- a/src/lavinmq/config.cr
+++ b/src/lavinmq/config.cr
@@ -32,6 +32,7 @@ module LavinMQ
       parse_ini(@config_file)
       parse_env()
       parse_cli(argv)
+      reload_logger
     end
 
     private def parse_config_from_cli(argv)
--- a/spec/config_spec.cr
+++ b/spec/config_spec.cr
@@ -89,7 +89,7 @@ describe LavinMQ::Config do
           [main]
           data_dir = /tmp/lavinmq-test
           log_level = debug
-          log_file = /var/log/lavinmq.log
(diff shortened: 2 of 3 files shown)


src/lavinmq/config.cr
```diff
@@ -32,6 +32,7 @@ module LavinMQ
       parse_ini(@config_file)
       parse_env()
       parse_cli(argv)
+      reload_logger
```
jage, line 35: This method should be called `initialize_logger` IMO, separate PR coming.

PR #1656, opened 2026-01-30
 Pass version env var to deb/RPM builds for commit tracking

### LINKED ISSUE #1640: Add commit hash to Prometheus version label in main/PR builds (deb packages)
When testing out `main` or a `PR` I want to note _exactly_ which version I'm testing in our metrics, right now we need to manually bump the version which reduces the resolution (since we don't do it for every commit). 

Example on an annotation that's very useful.

Image

I propose we have something like this:

```
# Release build
lavinmq_identity_info{lavinmq_version="2.8.0",lavinmq_cluster="hostname"} 1

# For a dev build it would show:
lavinmq_identity_info{lavinmq_version="2.7.0-alpha.1-gab12cd3",lavinmq_cluster="hostname"} 1
```

I'm running QA systems that auto-update on latest `main` build, so I want to track any improvements/regressions with these labels. I'm installing via the deb packages we build in `main`.

kickster97:
The Crystal macro in version.cr checks for `$version` env var. Since it's not set, it falls back to` git describe --tags`.. but i dont think `.git` works with the Docker image.. So the version is probably falling back to `shards version`

This PR makes the version available as an env variable during the build process so the crystal macro can access it. 
Fixes #1640 

Just want to try if this works when we build the deb package on PR

### DIFF
--- a/packaging/debian/Dockerfile
+++ b/packaging/debian/Dockerfile
@@ -13,7 +13,7 @@ RUN tar -czf ../lavinmq_${version}.orig.tar.gz -C /usr/src lavinmq_${version}
 COPY packaging/debian/ debian/
 RUN sed -i -E "s/^(lavinmq) \(.*\)/\1 \(${version}-1\)/" debian/changelog
 ARG DEB_BUILD_OPTIONS="parallel=2"
-RUN debuild -us -uc
+RUN debuild -us -uc -e=version
 
 FROM ubuntu:24.04 AS test
 COPY --from=builder /usr/src/*deb .


carlhoerberg:
`ARG` already adds an environment variable (just not stored in the container), the only thing you need is `--preserve-envvar version`

packaging/rpm/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN dnf install -y --nodocs --setopt=install_weak_deps=False \
 RUN rpmdev-setuptree
 COPY packaging/rpm/lavinmq.spec /root/rpmbuild/SPECS/
 ARG version
+ENV version=$version
```
carlhoerberg:
```suggestion
```

packaging/debian/Dockerfile
```diff
@@ -3,6 +3,7 @@ ARG build_image=84codes/crystal:latest-ubuntu-24.04
 FROM $build_image AS builder
 RUN apt-get update && apt-get install -y devscripts help2man lintian debhelper liblz4-dev
 ARG version
+ENV version=$version
```
carlhoerberg:
```suggestion
```

PR #1659, opened 2026-02-02
Use sysusers.d for user creation in install target

### LINKED ISSUE #1655: Broken install on Arch Linux
**Describe the bug**

The `install` target in Makefile requires superuser permissions since it's using `getent passwd lavinmq`. This breaks installation on the AUR build for Arch linux and possible other systems as well. 

What done in the PKGBUILD for Arch is that it's runned like this `make install DESTDIR=$DIR` where this is a tmp directory. This shouldn't require super user but since it's running `getent passwd lavinmq` this fails. 

The makefile was changed here: https://github.com/cloudamqp/lavinmq/commit/0dca86a7b825dac208e0218576ea721ece1b6198 so any build from that commit includes this issue. 

Running the `make install` with `sudo` fixes this but shouldn't be needed.

carlhoerberg:
## Summary
- Replace inline `useradd` command with a `sysusers.d` file for user creation
- Allows `make install DESTDIR=$DIR` to work without superuser privileges
- The lavinmq user is created by systemd-sysusers on boot or during package post-install

Fixes #1655

## Test plan
- [x] Verify `make install DESTDIR=/tmp/test` succeeds without sudo
- [x] Verify sysusers file is installed to `/usr/lib/sysusers.d/lavinmq.conf`
- [x] Verify `systemd-sysusers` creates the lavinmq user correctly


carlhoerberg:
`sysusers.d` is not available in ubuntu 20.04, but maybe we should drop support for it anyway now, EOL was in May 2025.

carlhoerberg:
Tested with fedora, ubuntu 22.04, and arch. creates users on all

carlhoerberg:
@snichme Arch PKGBBUILD file added. I've tested and it works.

PR #1662, opened 2026-02-03
chore(client.cr): add duration of connection in logs

lukas8219:
Add `duration` in connection disconnected log line

- Manual
  - Use `amqpcat -z` to check connection and if log works.

### DIFF
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -177,8 +177,13 @@ module LavinMQ
       ensure
         cleanup
         close_socket
-        duration = (RoughTime.unix_ms - @connected_at) / 1000
-        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
+      end
+
+      private def duration
+        ms = RoughTime.unix_ms - @connected_at
+        seconds = (ms / 1000).round.to_i
+        Time::Span.new(seconds: seconds)
       end
 
       private def frame_size_ok?(frame) : Bool
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -84,8 +84,13 @@ module LavinMQ
         @broker.remove_client(self)
         @waitgroup.done
         close_socket
-        duration = (RoughTime.unix_ms - @connected_at) / 1000
-        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
+      end
+
+      private def duration
+        ms = RoughTime.unix_ms - @connected_at
+        seconds = (ms / 1000).round.to_i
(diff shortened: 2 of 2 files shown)


src/lavinmq/amqp/client.cr
```diff
@@ -177,7 +177,8 @@ module LavinMQ
       ensure
         cleanup
         close_socket
-        @log.info { "Connection disconnected for user=#{@user.name}" }
+        duration = (RoughTime.unix_ms - @connected_at) / 1000
```
lukas8219: I am a bit concerned with `divide by zero errors`. But I did `amqpcat -z` which instantly creates connection and closes. And no issues!

carlhoerberg: but you divide by 1000? `0 / 1000` is not an issue, only `x / 0` is

lukas8219: Lol. Yeah. Multi tasking problems

src/lavinmq/amqp/client.cr
```diff
@@ -177,7 +177,8 @@ module LavinMQ
...
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
```
carlhoerberg: You have a nicer output from `Time::Span#to_s` if the duration is long. And doing the calculation inside the log block saves a couple of cpu cycles if the log level is above info. Using `//` does a truncated division and returns an integer, will result in an rounding error, but maybe that's ok, or else go with `.round.to_i`.

```suggestion
        @log.info { "Connection disconnected for user=#{@user.name} duration=#{Time::Span.new(seconds: (RoughTime.unix_ms - @connected_at) // 1000)}" }
```


```suggestion
        def duration
          ms = RoughTime.unix_ms - @connected_at
          seconds = (ms / 1000).round.to_i
          Time::Span.new(seconds: seconds)
        end
        
        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
```

lukas8219: Is defining 'duration' function like that efficient in Crystal? Or do a private fn for the class itself?

carlhoerberg: Do a private method outside this method

lukas8219:
```bash
2026-02-04T22:04:03.303198Z   INFO - lmq.amqp.client: Connection disconnected for user=guest duration=00:00:00 -- vhost: "/", address: "127.0.0.1:54584", name: "AMQPCat 1.0.2-1-g27ac244"
```

jage:
> ```shell
> duration=00:00:00
> ```

Nice addition! Did you consider something like `connected=15s`? I'm more used to seeing that format from logfmt (Go, Loki etc), not sure if there's a good "standard" definition here.

lukas8219:
@jage That was the pattern before @carlhoerberg suggested using TimeSpan. Checking now, i see there's a option `.seconds` which may be a better fit while still using `Time::Span`. I'll open a PR for that

lukas8219:
@jage https://github.com/cloudamqp/lavinmq/pull/1674

Issue #1612, opened 2026-01-15
HTTP shovel/webhook acks and removes message on HTTP 404; message does not return to source queue; pausing shovel does not stop in-flight retries


tenfortyeight:
Summary

When using an HTTP destination shovel (webhook) sourced from a queue, a message delivered to an endpoint that responds 404 is removed from the source queue and does not return/requeue. This prevents any queue-based retry/DLX strategy (since the message is no longer in the queue). Additionally, pausing the shovel in the UI does not stop ongoing retry attempts for an in-flight message.

Related Slack thread: https://lavinmq.slack.com/archives/C01K2BK1PT6/p1768484120860459

Setup (created via UI)
	•	Exchange: topic exchange (durable)
	•	Queue: durable queue
	•	Binding: queue bound to exchange with routing key # (all messages)
	•	Shovel: source = queue, destination = HTTP endpoint, ack-mode = on-confirm
	•	Shovel created via the web UI

Repro steps
	1.	Create a fanout exchange (durable).
	2.	Create a durable queue.
	3.	Bind queue to exchange using routing key #.
	4.	Create a shovel (via UI):
	•	src-type: queue
	•	src-queue: the queue above
	•	dest-uri: HTTP endpoint URL that responds 404
	•	ack-mode: on-confirm
	5.	Publish a message to the exchange (any routing key, since # binding).
	6.	Observe queue depth and shovel behavior.
	7.	Pause the shovel in the UI and observe whether delivery attempts stop.

Expected
	•	If the HTTP endpoint responds with an error (e.g. 404), the message should remain in (or return to) the source queue until successfully delivered, so queue-based retry/DLX policies can apply.
	•	Pausing the shovel should stop delivery attempts (at least for anything not successfully delivered).

Observed
	•	On HTTP 404 response, the message is no longer present in the source queue and does not reappear.
	•	Because the message leaves the queue, it cannot be handled by retry/DLX mechanisms (no opportunity for dead-lettering/TTL-based retry).
	•	After pausing the shovel in the UI, delivery attempts for the in-flight message continue (retries still occur).

Notes / Evidence
	•	LavinMQ logs show repeated 404-related activity during the test window, and the shovel appears to continue attempting delivery while the message is not visible in the queue.
	•	The shovel appears to treat “HTTP connection + response received” as sufficient progress to remove the message from the queue, regardless of HTTP status. But the http/webhook "re-trier" keeps retrying.

lukas8219:
[...]

PR #1667, opened 2026-02-04
feat(shovels/http_destination): implement retry + backoff in Shovel H…

### LINKED ISSUE #1137: Improvments to HTTP shovel with exponential backoff and retry limits

lukas8219:
1. Implements Retrier in HTTPDestination Shovel - configured by parameters
  - `dest-backoff`
  - `dest-jitter`
  - `dest-timeout`
  - `dest-max-retries`
2. Fixes HTTPDestination not taking `no_ack` mode into consideration - as all HTTPDestinations were defaulted as OnConfirm/OnPublish.

Specs:
- [x] Expose mock server to receive HTTP requests. Fail at least 2 times and accept third time.
- [x] Tests focused on the Retrier logic 

### RFC
- Should we actually implement a proper Circuit Breaker instead of just Retrier?
- Currently using `?.try` is [misleading](https://github.com/cloudamqp/lavinmq/pull/1215#discussion_r2595426705). Is there a better way to handle it? 

Closes https://github.com/cloudamqp/lavinmq/issues/1137


lukas8219:
Aims to solve https://github.com/cloudamqp/lavinmq/issues/1612 also

src/lavinmq/shovel/http_destination.cr
```diff
@@ -44,14 +64,26 @@ module LavinMQ
                else
                  "/"
                end
-        response = c.post(path, headers: headers, body: msg.body_io)
+        success = push_and_maybe_retry(@ack_mode) do
+          c.post(path, headers: headers, body: msg.body_io).success?
```
spuun, line 82: The io must be rewinded before each try.

src/lavinmq/shovel/retrier.cr
```diff
@@ -0,0 +1,17 @@
...
+  module Shovel
+    module Retrier
+      def self.push_with_retry(max_retries : Int32, jitter : Float64, backoff : Float64, & : -> Bool) : Bool
+        retries = 0
+        while retries <= max_retries
+          return true if yield
+          retries += 1
+          base_delay = (backoff ** retries).seconds
+          jitter_delay = jitter.seconds * Random.rand(0.0..1.0)
+          sleep base_delay + jitter_delay
```
spuun, line 12: This sleep will happen even after the last retry. Maybe not a big deal, but I guess it could be nice to skip that.

Maybe change the `while` to `loop` and add a `break if retries > max_retries`?

src/lavinmq/shovel/http_destination.cr
lukas8219: Hmm.  That is curious @spuun . Why would the Specs and Manual Testing work without rewind? What exactly does rewind do?

moves some buffer offset back to 0?

src/lavinmq/shovel/retrier.cr
lukas8219: Yes. Good catch. I noticed that on the Specs but didn't manage to add a `TODO` for myself.

src/lavinmq/shovel/http_destination.cr
lukas8219: Ok. The Specs was not reading the body and I don't trust my manual tests. Thanks @spuun

snichme:
What is the status of this PR?  
If we add more options to the destination, should that also be available in the UI when configuring? 

I'm also leaning to having a bigger overhaul of the shovels feature since the code is quite difficult to follow now and there are some pending issues, like this one and #1357. Just rewriting it for the sake of refactor I don't think is the right approach but if we could find a solution without adding more complexity would be nice. 
Also shovels are a good feature to have but they are hard to debug since you have little insight into what is happening when it fails or stops. 
Bigger discussion but worth taken into consideration here i think.

lukas8219:
There were only old comments from Jon here @snichme that I've been adjusting since this PR is really really old. **I'll add the UI buttons for the new options**. I am +1 to giving more visibility to what's happening inside the shovel - perhaps metrics and shovel-specific logs showing on UI?

As to the overhaul I am not sure if it's necessary? The code looks much better after the modularization. What aspects are hard to deal now?

lukas8219:
@snichme Do you think it's worth to join Monday weekly to discuss this? I wanted to either close it or move it forward. This has been here for too long.

snichme:
PR #2128 is relevant as well, might need to merge the ideas from both

PR #1674, opened 2026-02-06
chore(client.cr): use `total_seconds` instead of normal `to_s` to format into seconds

lukas8219:
https://github.com/cloudamqp/lavinmq/pull/1662 introduced the duration on the disconnected log but on a weird format. 

- Manual (`amqpcat -z`)

```bash
2026-02-06T16:47:42.138068Z  INFO lmq.amqp.client[vhost: "/", address: "127.0.0.1:54655", name: "AMQPCat 1.0.2-3-g968b65e"] Connection disconnected for user=guest duration=0.0s
```

### DIFF
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -186,13 +186,13 @@ module LavinMQ
       ensure
         cleanup
         close_socket
-        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
       end
 
       private def duration
         ms = RoughTime.unix_ms - @connected_at
         seconds = (ms / 1000).round.to_i
-        Time::Span.new(seconds: seconds)
+        Time::Span.new(seconds: seconds).total_seconds
       end
 
       private def frame_size_ok?(frame) : Bool
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -84,13 +84,13 @@ module LavinMQ
         @broker.remove_client(self)
         @waitgroup.done
         close_socket
-        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
       end
 
       private def duration
         ms = RoughTime.unix_ms - @connected_at
         seconds = (ms / 1000).round.to_i
-        Time::Span.new(seconds: seconds)
+        Time::Span.new(seconds: seconds).total_seconds
       end
 
       def read_and_handle_packet


src/lavinmq/amqp/client.cr
```diff
@@ -186,13 +186,13 @@ module LavinMQ
...
         close_socket
-        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}" }
+        @log.info { "Connection disconnected for user=#{@user.name} duration=#{duration}s" }
       end
 
       private def duration
         ms = RoughTime.unix_ms - @connected_at
         seconds = (ms / 1000).round.to_i
-        Time::Span.new(seconds: seconds)
+        Time::Span.new(seconds: seconds).total_seconds
```
carlhoerberg, line 195: If we want to do this we can skip time::span completely. But the human readability of `duration=185832s` is lost. Compare it to `duration=2.03:37:12`. And to me logs are primarily for humans. 

```suggestion
        (ms / 1000).round.to_i

carlhoerberg (commented):
The human readability is lost with `duration=185832s`. Compare it to `duration=2.03:37:12`. Aren't logs primarily for humans? @jage

lukas8219:
@carlhoerberg HHHmmmmss might be much harder to transform logs into metrics compared to seconds-only

Like promql

carlhoerberg:
Prometheus histogram feels like a better fit if you want to track connection times, no?

lukas8219:
Sorta. But way less flexible and costs more memory.

Imagine I have 3 distinct client types, maybe IP or name or protocol

Only 1 of them is having issues.

Prom would have to label them (too many)

Logs could easily be parsed and transformed into metrics


Besides If we already have the log, might as well prefer it than creating prom only for that?

@carlhoerberg

jage:
> The human readability is lost with `duration=185832s`. Compare it to `duration=2.03:37:12`. Aren't logs primarily for humans? @jage

Yes, agree. I think both are not super nice to read but ok.

I'm used to how Loki/LogQL does it and they can parse `duration=51h37m12s`. I think a lot of tooling are built/compatible with https://pkg.go.dev/time#ParseDuration see https://grafana.com/docs/loki/latest/query/log_queries/#label-filter-expression

Personally I think that's a good format, but no idea how "standard" it is.

Just FYI, data could also include prefix (`83KiB`) instead of just using bytes. But as previously mentioned, I'm not sure how "standard" this is, but I like it. :)

carlhoerberg:
i like the `51h37m12s` format! maybe monkey patch `Time::Span` with `.humanize_time` that outputs that format.  

> Just FYI, data could also include prefix (83KiB) instead of just using bytes. But as previously mentioned, I'm not sure how "standard" this is, but I like it. :)

yes, size prefix is easy with `.humanize_bytes`

lukas8219:
@carlhoerberg Should we move to that format then? https://github.com/cloudamqp/lavinmq/pull/1674#issuecomment-3908689969

PR #1676, opened 2026-02-06
add http module for lavinmqperf

lukas8219:
Please fill me in.

### DIFF
--- a/src/lavinmqperf.cr
+++ b/src/lavinmqperf.cr
@@ -2,6 +2,7 @@ require "log"
 require "./lavinmq/version"
 require "./lavinmqperf/amqp/*"
 require "./lavinmqperf/mqtt/*"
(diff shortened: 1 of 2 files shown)


lukas8219:
I'll invest some time on this for real. This was only a POC I needed to convince myself Lavin was able to support good HTTP workloads

lukas8219:
I believe this is really not need. There's already many HTTP benchmark tools. Much easier to use Gatling or K6 than dealing with scaling HTTP clients etc

carlhoerberg:
I'd missed this! Not against it, but as the reviews says, reuse the http clients!

lukas8219:
I feel like that's solved somewhere else. If we need to benchmark for real, better use gold(or industry)-standard tools!

PR #1678, opened 2026-02-08
Track open mmap and fiber counts

carlhoerberg:
## Summary
- Add an atomic counter on `MFile` to track the number of open file-backed mmaps
- Expose `MFile.mmap_count` and `Fiber.count` as Prometheus gauge metrics (`mfile_mmap_count`, `fiber_count`)
- Display both counts in the nodes detail view in the management UI

## Test plan
- [ ] Verify `mmap_count` increments/decrements correctly by checking the Prometheus `/metrics` endpoint while creating and closing queues
- [ ] Verify `fiber_count` reflects active fibers on the `/metrics` endpoint
- [ ] Check the nodes UI page shows both "Open mmaps" and "Fibers" rows with correct values

### DIFF
--- a/src/lavinmq/http/controller/nodes.cr
+++ b/src/lavinmq/http/controller/nodes.cr
@@ -54,6 +54,7 @@ module LavinMQ
       end
 
       private def node_stats
+        fiber_count = Fiber.count
         {
           os_pid:             Process.pid.to_s,
           fd_total:           System.file_descriptor_limit[0],
@@ -76,11 +77,11 @@ module LavinMQ
           disk_free:          @amqp_server.disk_free,
           disk_free_details:  {log: @amqp_server.disk_free_log},
           partitions:         Tuple.new,
-          proc_used:          Fiber.count,
+          proc_used:          fiber_count,
           run_queue:          0,
           sockets_used:       @amqp_server.vhosts.sum { |_, v| v.connections.size },
-          mmap_count:         MFile.mmap_count,
-          fiber_count:        Fiber.count,
+          mmap_file_count:    MFile.count,
+          fiber_count:        fiber_count,
           followers:          @amqp_server.followers,
         }
       end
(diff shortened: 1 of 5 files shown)


carlhoerberg:
Call it `file_mmap_count` or `mfile_count` instead? `mmap_count` is a bit misleading as each fiber creates an anonmoyus mmap for each fiber stack. So `fiber_count` is also the number of anonomous mmap's, while `mfile_count` is the same as number of file backed mmaps.

spuun:
> Call it `file_mmap_count` or `mfile_count` instead? `mmap_count` is a bit misleading as each fiber creates an anonmoyus mmap for each fiber stack. So `fiber_count` is also the number of anonomous mmap's, while `mfile_count` is the same as number of file backed mmaps.

There's a stack pool for fiber stacks, so I guess there may be more mmaps than fibers. What about monkey patching `Crystal::System::Fiber.allocate_stack` and `Crystal::System::Fiber.self.free_stack`?

jage:
> Call it `file_mmap_count` or `mfile_count` instead? `mmap_count` is a bit misleading as each fiber creates an anonmoyus mmap for each fiber stack. So `fiber_count` is also the number of anonomous mmap's, while `mfile_count` is the same as number of file backed mmaps.

IMO `mfile_count`, that's what it is, and that's what it's called in https://github.com/cloudamqp/lavinmq/pull/1627. 

We must have both mmap count from the kernel (https://github.com/cloudamqp/lavinmq/pull/1627) and these two metrics to get an understanding what's going on in the broker.

carlhoerberg:
> There's a stack pool for fiber stacks, so I guess there may be more mmaps than fibers. What about monkey patching `Crystal::System::Fiber.allocate_stack` and `Crystal::System::Fiber.self.free_stack`?

Yes, look at the first, but didn't know they could diverge! It's probably faster too, as `Fiber.count` traverses a linked list.

carlhoerberg:
#1627

PR #1681, opened 2026-02-09
refactor: Break out queue stats to `QueueStats` module

spuun:
One step closer of decoupling `MQTT::Session` from `AMQP::Queue`

This will break out queue stats to a `QueueStats` module.

Specs still passes

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -3,7 +3,6 @@ require "../../logger"
 require "../../segment_position"
 require "../../policy"
 require "../../observable"
-require "../../stats"
 require "../../sortable_json"
 require "../../client/channel/consumer"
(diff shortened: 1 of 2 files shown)


spuun:
This should go even further in the future, removing the access to instance variables added by the module in favor of methods to increase/decrease counters.

snichme:
Is this so the module QueueStats can be used for Session or something else? I'm not following the reason for the new module. 
Not saying it's a bad thing, just curious

spuun:
> Is this so the module QueueStats can be used for Session or something else? I'm not following the reason for the new module. Not saying it's a bad thing, just curious

Yes, exactly.

snichme:
Would be nice to find a common name for `Queue`, `Stream` and `Session`

snichme (approved):

PR #1682, opened 2026-02-09
Track publish events for HTTP endpoints

### LINKED ISSUE #1677: `Publish` metric is not increment on `Overview` page when doing HTTP POST to publish messages
**Describe the bug**
When testing https://github.com/cloudamqp/lavinmq/pull/1676 with `-q perf` I didn't see the metrics on the `Overview` page

**Describe your setup**
Spin Dedicated instance on CloudAMQP

**How to reproduce**

1. Run on shell
```bash
./bin/lavinmqperf http throughput --host test-windy-dog.lmq.cloudamqp.com --port 443 --vhost havpiyfn --password XXXX --username havpiyfn -x 100 -y 100
```
5. Check Overview
6. Confirm theres no movement/data on the UI

**Expected behavior**
Overview should track the published messages

kickster97:
Fixes vhost-level message rate stats not updating when publishing or consuming messages via the HTTP API, by adding event_tick calls for these operations. 

Fixes #1677

see #1677

### DIFF
--- a/src/lavinmq/http/controller/exchanges.cr
+++ b/src/lavinmq/http/controller/exchanges.cr
@@ -147,7 +147,6 @@ module LavinMQ
             Log.debug { "Post to exchange=#{e.name} on vhost=#{e.vhost.name} with routing_key=#{routing_key} payload_encoding=#{payload_encoding} properties=#{properties} size=#{size}" }
             ok = e.vhost.publish(msg)
             e.vhost.event_tick(EventType::ClientPublish)
-            e.vhost.add_recv_bytes(size)
             {routed: ok}.to_json(context.response)
           end
         end
--- a/src/lavinmq/http/controller/main.cr
+++ b/src/lavinmq/http/controller/main.cr
@@ -7,7 +7,7 @@ module LavinMQ
     class MainController < Controller
       include StatsHelpers
 
-      OVERVIEW_STATS = {"ack", "deliver", "get", "deliver_get", "publish", "confirm", "redeliver", "reject", "recv_oct", "send_oct"}
+      OVERVIEW_STATS = {"ack", "deliver", "get", "deliver_get", "publish", "confirm", "redeliver", "reject"}
       EXCHANGE_TYPES = {
         {name: "direct", human: "Direct"},
         {name: "fanout", human: "Fanout"},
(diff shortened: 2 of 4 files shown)


lukas8219 (approved):

snichme (commented):
Adding `vhost.event_tick` seems to be enough to resolve the issue. 
With this change we are also adding bytes in and out counters, but only for HTTP not for AMQP.  If we add these metrics I think we should have it for all incoming and outgoing data unless there is a reason not to. And if we want it introduce it in a separate PR, and have this one focus the related issue.

kickster97:
> Adding `vhost.event_tick` seems to be enough to resolve the issue.
> With this change we are also adding bytes in and out counters, but only for HTTP not for AMQP. If we add these metrics I think we should have it for all incoming and outgoing data unless there is a reason not to. And if we want it introduce it in a separate PR, and have this one focus the related issue.

AMQP/MQTT track bytes in/out at connection level, which does not fit short-lived HTTP. without the bytes tracking for HTTP, we see the publish graph move for AMQP/MQTT and HTTP, but in the data rates graph we only see values for AMQP/MQTT. 
I could split the PR up, but i still think the bytes metrics for http are nessecary. Maybe there is a better way to keep it more consistent with AMQP/MQTT?

carlhoerberg:
i agree with @snichme . a global counter is probably a good idea, but should also include all amqp/mqtt connections, they can also be shortlived, in another PR.

snichme:
> AMQP/MQTT track bytes in/out at connection level, which does not fit short-lived HTTP. without the bytes tracking for HTTP, we see the publish graph move for AMQP/MQTT and HTTP, but in the data rates graph we only see values for AMQP/MQTT.

Ahh ok. Yes makes sense to also show the data rates for HTTP I agree on that :) 
But I think we should in that case find a way that works for all types of clients instead of adding bytes_in and bytes_out on a vhost level but it only counts for some clients. 

We need to keep track of data rates per connection since we show that in the UI so we cannot just move it, but maybe a add those rates to the vhost counter as well?

snichme (approved):

kickster97:
I think we merged in a weird way, so when i rebased stuff on main this became obsolete

PR #1685, opened 2026-02-10
Add option to disable message timestamp validation

### LINKED ISSUE #1684: Add option to disable message timestamp validation

RabbitMQ doesn't enforce the AMQP 0-9-1 requirement that message timestamps be POSIX `time_t`. In contrast, LavinMQ enforces this (to a degree), which prevents applications that rely on RabbitMQ's behavior (e.g., applications using Gregorian timestamps) from working with LavinMQ.

Add a configuration option to disable message timestamp validation.

**Describe alternatives you've considered**

While it is possible to fix code that violates AMQP 0-9-1, adding a configuration option that enables mirroring RabbitMQ's behavior allows for quicker and easier adoption of LavinMQ.

rdnn:

This pull request adds a new `validate_timestamp` configuration option that allows message timestamp validation to be explicitly disabled.

Fixes #1684

Specs have been added to `server_spec.cr‎` and updated in `config_spec.cr‎`.

### DIFF
--- a/spec/config_spec.cr
+++ b/spec/config_spec.cr
@@ -92,6 +92,7 @@ describe LavinMQ::Config do
           log_file = /tmp/lavinmq-test.log
           stats_interval = 10000
           stats_log_size = 240
+          validate_timestamp = false
           set_timestamp = true
           socket_buffer_size = 32768
           tcp_nodelay = true
@@ -175,6 +176,7 @@ describe LavinMQ::Config do
       config.log_file.should eq "/tmp/lavinmq-test.log"
       config.stats_interval.should eq 10000
       config.stats_log_size.should eq 240
+      config.validate_timestamp?.should be_false
       config.set_timestamp?.should be_true
       config.socket_buffer_size.should eq 32768
       config.tcp_nodelay?.should be_true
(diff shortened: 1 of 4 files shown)


rdnn:

carlhoerberg:
The exception is raised here https://github.com/cloudamqp/amq-protocol.cr/blob/1f176f1b7f9f9b70e8044d9389b5ff910f253e88/src/amq/protocol/properties.cr#L408

And maybe we should return nil instead

carlhoerberg:
Or maybe better: 

```
        props.timestamp = RoughTime.utc if props.timestamp_raw.nil? && Config.instance.set_timestamp?
```

rdnn:
> The exception is raised here https://github.com/cloudamqp/amq-protocol.cr/blob/1f176f1b7f9f9b70e8044d9389b5ff910f253e88/src/amq/protocol/properties.cr#L408
> 
> And maybe we should return nil instead

I favored changing the behavior of LavinMQ over modifying the library in the event that someone relied on timestamp validation.

> Or maybe better:
> 
> ```
>         props.timestamp = RoughTime.utc if props.timestamp_raw.nil? && Config.instance.set_timestamp?
> ```

This is certainly a simpler option if everyone is okay with dropping the timestamp validation.

carlhoerberg:
Yes, it has never been meant as a validation, only implicitly has it become one.

PR #1689, opened 2026-02-10
PoC: Add user tags as classes on html element

spuun:
There is no easy way of styling the view based on the user's permissions. As it is, a request must be done to get the user's tags, then hide or show elements based on this. This is very inconvenient for ux development and it's easy to break because of the javascript dependency. It may also cause flicker for the user.

This PR introduces `user-tag-X` classes added to the `<html>` element, where `X` is a tag e.g. `user-tag-administrator`. By doing this it's possible to show/hide elements with css depending on the user's tags. One use case is hiding menu items, another is hiding definitions forms on the overview page if the user is lacking administrator permissions.

The ETag will now contain an integer representing the tags the user has. This is done by summarizing the int value for each tag, which is possible by the `Tag` enum now being annotated as `Flags`.

This PR will add some extra CPU cycles on view loading, but IMO it's totally worth it to be able to make a better UX.

Run specs

### DIFF
--- a/src/lavinmq/auth/base_user.cr
+++ b/src/lavinmq/auth/base_user.cr
@@ -19,7 +19,7 @@ module LavinMQ
       def user_details
         {
           name: name,
-          tags: tags.map(&.to_s.downcase).join(","),
+          tags: tags.map(&.to_downcase_s).join(","),
         }
       end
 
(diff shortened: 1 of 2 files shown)


spuun:
A follow up to this is to store theme and menu state in cookies instead of local storage and render classes for this too.

carlhoerberg:
Im against adding any runtime stuff to the html. I think they ideally
should be 100% static. Maybe the html should even be generated in the
Makefile, and not even be ecr files. Im convinced flickering can be avoided
in other ways.

spuun:
Superseded by #1698

PR #1692, opened 2026-02-11
refactor: Read inlined js from file to automate csp hash calculation

spuun:
It's cumbersome to edit the javascript inlined in `head.ecr`. Besides recompiling one must calculate a new hash of the script and update the controller code with this hash. 

This PR will break out the inlined script to `inline.js` and read and embed this file on compile time. The hash is automatically calculated.

If built without `--release` and without `-Dbake_static` the file is read from disk and hash is calculated per load (on full refresh / cached disabled).

Run front specs / test manually

### DIFF
--- a/src/lavinmq/http/controller/views.cr
+++ b/src/lavinmq/http/controller/views.cr
@@ -87,7 +87,7 @@ module LavinMQ
         end
       {% else %}
         macro inline_js
-          File.read("./static/js/_inline.js")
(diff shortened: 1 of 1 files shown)


snichme (approved):

spuun:
> add all localstorage items with a certain prefix to the document classlist? and then do the rest in external css/js?

I actually had this in mind, too.

carlhoerberg:
closed in preference for #1698

PR #1695, opened 2026-02-12
Use policymaker role consistently for global-parameters endpoints

kickster97:
This adds consistent permission access across the global parameters endpoints. 
This follows rabbitmq's access management [For global parameters. Must be policymaker.](https://github.com/rabbitmq/rabbitmq-server/blob/008cf25f58dba5162f237c2c282d8e0c85bce495/deps/rabbitmq_management/src/rabbit_mgmt_util.erl#L185)

**Be clear in releasenotes that this may break someones workflow!**

run added specs

### DIFF
--- a/spec/api/parameters_spec.cr
+++ b/spec/api/parameters_spec.cr
@@ -516,6 +516,17 @@ describe LavinMQ::HTTP::ParametersController do
     end
   end
   describe "DELETE /api/global-parameters/name" do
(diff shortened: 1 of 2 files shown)


snichme (approved):

snichme:
We should be aware that this might break someones workflow so we announce it clearly in the release notes.

PR #1699, opened 2026-02-12
Aggregate data rates at vhost level for all protocols

kickster97:

Track data rates at vhost level across all protocols

Previously, byte rates in the overview were aggregated from per-connection stats, which meant HTTP API traffic was invisible in the data rates graph since HTTP has no persistent connections. This adds `recv_oct` and `send_oct` counters to the `vhost`, fed by all three protocols: AMQP and MQTT clients bubble up their per-connection byte counts to the vhost alongside existing connection-level tracking, and HTTP reports bytes directly to the vhost. The overview endpoint now reads byte rates from the vhost instead of summing connections, giving a complete picture of all traffic.
  
Based off discussion in #1682 

specs to come

### DIFF
--- a/spec/vhost_stats_spec.cr
+++ b/spec/vhost_stats_spec.cr
@@ -1,4 +1,5 @@
 require "./spec_helper"
+require "./mqtt/spec_helper"
 require "../src/lavinmq/config"
 
 describe LavinMQ::VHost do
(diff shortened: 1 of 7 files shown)


carlhoerberg (approved):
Did it have any performance impact?

carlhoerberg:
HTTP revc/send doesn't include the whole request/response size, but i guess that's ok.

snichme (approved):

jage:
Can we add specs? Ideally specs that break if we add additional protocols and miss this.

Issue #1680, opened 2026-02-09
Broken tooltip for debugging shovel statuses


oskgu360:
**Summary**  

When debugging shovels, the the messages is not shown in its full shape.

**Current Behavior**  

<img width="2984" height="578" alt="Image" src="https://github.com/user-attachments/assets/687dbf55-24a7-458d-a996-5634065f540d" />

On Google Chrome Version 144.0.7559.133, 3024x1964

**Steps to Reproduce (if reporting an issue)**  

Create a non-working shovel and hover over the status

PR #1700, opened 2026-02-12
Bugfix -  UI tooltip sizing

oskgu360:

Fixes https://github.com/cloudamqp/lavinmq/issues/1680

Prior to change:
Screenshot 2026-02-12 at 15 36 57
Screenshot 2026-02-12 at 15 37 20

After change
Screenshot 2026-02-12 at 15 30 47
Screenshot 2026-02-12 at 15 34 17

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@ VIEW_SOURCES := $(wildcard views/*.ecr)
 VIEW_TARGETS := $(patsubst views/%.ecr,static/%.html,$(VIEW_SOURCES))
 VIEW_PARTIALS := $(wildcard views/partials/*.ecr)
 JS := static/js/lib/chunks/helpers.segment.js static/js/lib/chart.js static/js/lib/luxon.js static/js/lib/chartjs-adapter-luxon.esm.js static/js/lib/elements-8.2.0.js static/js/lib/elements-8.2.0.css $(wildcard static/js/*.js)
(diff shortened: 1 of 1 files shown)


kickster97 (commented):
sorry for nitpick and no actual value in review :P but the changes should be in two separate PRs, otherwise nice changes! :)

carlhoerberg:
> but the changes should be in two separate PRs, otherwise nice changes! :)

Agree. 

You can also `make bin/lavinmq CRYSTAL_FLAGS=` to make a non-release build, in which static assets are not baked in, which make iterating on static resources way faster (no recompile, just editing the JS/CSS and ctrl-r in the browser). But how to make that clearer for developers? At the very least we should drop `--release` from `make bin/lavinmq-debug`.

oskgu360:
> You can also `make bin/lavinmq CRYSTAL_FLAGS=` to make a non-release build, in which static assets are not baked in, which make iterating on static resources way faster (no recompile, just editing the JS/CSS and ctrl-r in the browser). But how to make that clearer for developers? At the very least we should drop `--release` from `make bin/lavinmq-debug`.

Agree :) I figured out eventually to, but as you say, would be nice if the "default option" was the best approach. A section in the readme stating that would be enough imo. (as it does in CLAUDE.md)

However, dropping the make stuff from this PR then

PR #1702, opened 2026-02-12
Add support for mqtts to `lavinmqperf`

walro:

Adding support for mqtts to `lavinmqperf`

`lavinmqperf mqtt throughput --uri=mqtts://<user>:<pass>@<host>`

### DIFF
--- a/src/lavinmqperf/mqtt/throughput.cr
+++ b/src/lavinmqperf/mqtt/throughput.cr
@@ -132,8 +132,10 @@ module LavinMQPerf
 
         if tls
           ssl_context = OpenSSL::SSL::Context::Client.new
-          if @uri.query_params["verify"]? =~ /^none$/i
-            ssl_context.verify_mode = OpenSSL::SSL::VerifyMode::NONE
+          if verify_param = @uri.query_params["verify"]?
+            if verify_mode = OpenSSL::SSL::VerifyMode.parse?(verify_param)
+              ssl_context.verify_mode = verify_mode
+            end
           end
           begin
             ssl_socket = OpenSSL::SSL::Socket::Client.new(tcp_socket, context: ssl_context, sync_close: true, hostname: host)


walro:

walro:
> URI fallback always hardcodes `mqtt://` scheme, breaking `mqtts` for bare `host:port` input

Not sure how to address this, I find "This means TLS can never be activated via bare host:port input" acceptable for now.

> Please verify this compiles, or use a type restriction / `responds_to?` check.

It compiles (and runs).

carlhoerberg (approved):
for the amqp client we support verify mode via the uri, like:
```
bin/lavinmqperf amqp throughput --uri amqps://localhost\?verify=none
```
just an idea

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -125,6 +124,30 @@ module LavinMQPerf
...
+        tcp_socket = TCPSocket.new(host, port)
+        tcp_socket.keepalive = true
+        tcp_socket.tcp_nodelay = false
+        tcp_socket.sync = false
+
+        if tls
+          ssl_context = OpenSSL::SSL::Context::Client.new
+          if @uri.query_params["verify"]? =~ /^none$/i
+            ssl_context.verify_mode = OpenSSL::SSL::VerifyMode::NONE
+          end
```
carlhoerberg: suggestion: 

```suggestion
          if verify_param = @uri.query_params["verify"]?
            if verify_mode = OpenSSL::SSL::VerifyMode.parse?(verify_param)
              ssl_context.verify_mode = verify_mode
            end
          end
```

src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -125,6 +124,30 @@ module LavinMQPerf
...
+          begin
+            ssl_socket = OpenSSL::SSL::Socket::Client.new(tcp_socket, context: ssl_context, sync_close: true, hostname: host)
+          rescue ex
+            tcp_socket.close rescue nil
+            raise ex
+          end
+          ssl_socket.sync = false
```
carlhoerberg, line 146: to avoid double buffering:
```suggestion
          ssl_socket.sync = false
          tcp_socket.sync = true
```
maybe ssl socket is already doing it "behind the scenes", i'm not sure.

carlhoerberg (approved):

src/lavinmqperf/mqtt/throughput.cr
walro: Yes, it seems to do it behind the scenes, I see a familiar face in https://github.com/crystal-lang/crystal/commit/4a5c263ba3ce1452f6ced6e0f9216fe7c0757908 😁 Code was recently moved in https://github.com/crystal-lang/crystal/commit/37d25b037331f1053317ef4d9e4ae98491140170, but seems to be doing the same thing still.

src/lavinmqperf/mqtt/throughput.cr
walro: Yes, that is better. FTR I was inspired by https://github.com/cloudamqp/amqp-client.cr/blob/v1.3.2/src/amqp-client.cr#L89 so we may wanna change there too to allow for all possible values.

walro: Pushed in cf95fd0

snichme (approved):

PR #1703, opened 2026-02-12
Replace ECR templates with SHTML + shell build

carlhoerberg:
## Summary
- Replace ECR templates with SHTML (SSI syntax) processed by a POSIX shell script (`views/render.sh`), removing Crystal as a dependency for view rendering
- Convert `views/*.ecr` to `views/*.shtml` and `views/partials/*.ecr` to `views/partials/*.shtml`
- Delegate view file serving from ViewsController to StaticController by rewriting the request path to `.html` and calling next — eliminates duplicate file embedding in the binary
- ViewsController now only handles auth checks and security headers (CSP, X-Frame-Options, Referrer-Policy)
- Move active-menu highlighting to CSS via the existing per-page `<html class="…">` (drops the JS fallback)

## Test plan
- [x] `make clean-views && make views` generates HTML files with substituted TITLE/VERSION
- [x] `make bin/lavinmq CRYSTAL_FLAGS=` builds successfully
- [x] `make lint` clean
- [x] `make test SPEC=spec/http_views_spec.cr` — CSP regression + partial-include specs pass
- [x] Pages render with security headers (CSP, X-Frame-Options, Referrer-Policy)
- [x] Auth redirect works — `/queues` without cookie redirects to `/login`
- [x] `/login` works without auth
- [x] Static assets (CSS, JS, images) still served normally
- [x] `/docs/` API docs still work


carlhoerberg:
Much faster view generation with `make views`, good for @bengtmagnus 
The HTML is stored compressed in the binary unlike today, saving 120KiB of `bin/lavinmq` 
Server-Side-Include/SHTML because it's an old standard.

static/js/layout.js
```diff
@@ -1,6 +1,17 @@
...
+// Highlight active menu item based on current path
+;(function () {
+  const path = window.location.pathname.replace(/^\//, '') || '.'
+  document.querySelectorAll('#menu-content > li > a').forEach(link => {
+    const href = link.getAttribute('href')
+    if (href === path || href === path + 's' || (href === '.' && (path === '' || path === '/'))) {
+      link.parentElement.classList.add('active')
+    }
+  })
+})()
```
spuun: I'm not a fan of this.

What if we put a class in the html element on each page with the "path", then we can probably use css all the way. E.g. in overview.html we have `<html class="overview">` and in the menu we have `<li class="overview">`, then we can do like `html.overview li.overview { active css }`.

I guess that also would make it possible to move all icons to the css instead of using img tags? But maybe that doesn't make sense.

carlhoerberg: fixed!

carlhoerberg:
rebase on your #1709 @spuun

snichme:
With this change I feel that we loose some functionality in our views and not sure if it's worth the benefits. 
I think it's nice to have the ability to render elements from a loop which you can do in ecr. 
Another feature I think it's good is to be able to pass params to partials, so you can reuse one partial for different views but customize them with some values.

carlhoerberg:
For rapid iteration of views we need fast compilation, which ECR is unable to provide. Iterations should imho be done client side. Otherwise we have a mix of both server side and client side rendering. And I guess my dream of putting this UI in front of Rabbit still lives on.

carlhoerberg:
> Another feature I think it's good is to be able to pass params to partials, so you can reuse one partial for different views but customize them with some values.

This is exactly what i wanted to achieve with this PR, to prevent sneaking in server side rendering. If we were to go with server side rendering it's easy to forget that we need to invalidate all caches etc. I think it will be easier for everybody if we are strict about static html + JS for client side data fetching and rendering.

spuun:
We could use ECR and `_render.cr` instead of shtml and `render.sh`. Slightly slower rendering, but a very rich templating. Still renderd to html on compile time and included as static files, so the `controllers/views.cr` and `controllers/static.cr` would be what they look like in this PR.

carlhoerberg:
We can use erb/haml too, would be rich and very fast, but another
dependency.

carlhoerberg:
SSI is just easy to do in a Shelly script with minimal dependencies. And
partial injections is the only thing we do currently. Can always reach for
something else when we feel we need it.

Or some python templating, maybe more common to be available in CI envs
etc.

spuun:
> We can use erb/haml too, would be rich and very fast, but another dependency.

Yes, but ECR isn't a new dependency. We'd get the same result as with this PR, but with the possibility to do more advanced things. 

I'm however fine with this PR. I guess it will force us to think twice and make clean html.

Makefile
```diff
@@ -18,13 +19,13 @@ all: $(BINS)
 bin/%: src/%.cr $(SOURCES) lib $(JS) $(DOCS) | bin
 	crystal build $< -o $@ $(CRYSTAL_FLAGS)
 
-bin/lavinmq: src/lavinmq.cr $(SOURCES) $(VIEW_SOURCES) $(VIEW_PARTIALS) lib $(JS) $(DOCS) | bin
+bin/lavinmq: src/lavinmq.cr $(SOURCES) $(VIEW_TARGETS) lib $(JS) $(DOCS) | bin
```
ThomasSarlin, line 22: We lost `$(VIEW_PARTIALS)` here and in `bin/lavinmq-debug` section, is it intentional?

carlhoerberg: Yes, it's a transient dependency, the view targets depends on the partials

ThomasSarlin: Thanks for clarifying 👍

src/lavinmq/http/controller/views.cr
```diff
@@ -37,36 +32,16 @@ module LavinMQ
...
-            context.response.headers.add("ETag", ETag)
-            context.response.headers.add("X-Frame-Options", "SAMEORIGIN")
-            context.response.headers.add("Referrer-Policy", "same-origin")
-            context.response.headers.add("Content-Security-Policy", "default-src 'none'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; script-src 'self' 'sha256-7WqJMeRnYJHbEWj9TgKVXAh9Vz/3wErO1WSGhQ2LTf4='")
-            {{ block.body if block }}
-            render {{ view }}
-          end
+          context.response.headers.add("X-Frame-Options", "SAMEORIGIN")
+          context.response.headers.add("Referrer-Policy", "same-origin")
+          context.response.headers.add("Content-Security-Policy", "default-src 'none'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; script-src 'self' 'sha256-7WqJMeRnYJHbEWj9TgKVXAh9Vz/3wErO1WSGhQ2LTf4='")
```
carlhoerberg: Move this to static controller, and add when content-type/file-ending is html?

PR #1704, opened 2026-02-13
Replace Observable/Observer with typed callback blocks

carlhoerberg:
## Summary
- Remove the macro-based `Observable`/`Observer` pattern and replace with explicit, per-event typed callbacks (`on_bind`, `on_unbind`, `on_deleted`) on `Exchange` and (`on_added`, `on_deleted`, `on_closed`) on `VHostStore`
- Remove dead `QueueLink` observer code that was never registered
- Delete unused `ExchangeEvent`, `QueueEvent` enums and `observable.cr`
- Net reduction of 80 lines across 15 files

## Test plan
- [x] `make test SPEC=spec/observable_spec.cr` — passes
- [x] `make test` — all 1545 tests pass, 0 failures
- [x] `make lint` — 0 issues


spec/observable_spec.cr
spuun, line 1: Looks like this file can be removed.

PR #1709, opened 2026-02-13
Refactor menu to not need the active class (no javascript)

spuun:
This refactors the menu to not need the active class in the menu item. Instead it relies on a class on the HTML tag together with a unique class per menu item. The menu CSS is also grouped better using nesting selectors.

This also fixes so the icons in the menu respect light and dark mode (before this only light icon is used).

This will support #1703 and get rids of the JavaScript to set active class.

Some repetitive code in the CSS to handle light/dark mode and active color. Part of this can probable removed if #1698 is merged and modified to handle system color mode in another way.

Manually

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -67,7 +67,7 @@
   --color-theme-switcher-active: light-dark(#ffffff, rgba(255, 255, 255, 0.15));
 
   scrollbar-width: thin;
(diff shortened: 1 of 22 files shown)


carlhoerberg (approved):
👍

PR #1710, opened 2026-02-16
fix: gracefully close message store on corrupt segments

### LINKED ISSUE #1701: Corrupt message segment crashes server on startup with clustering
When a message segment file contains corrupt data, MessageStore correctly detects the error during initialization and calls close to mark the store as closed. However, two problems prevent graceful handling:

1. Execution continues after close — in load_segments_from_disk, the corrupt file is still added to `@segments` and the constructor keeps running. In load_stats_from_segments, the loop continues iterating over segments that may have been closed.
2. With a replicator (clustering), close spawns a fiber to close MFiles asynchronously (message_store.cr:263-270). This fiber captures `@segments` by reference. At the next Fiber.yield, the fiber runs and closes MFiles that were added after the close call — causing an unhandled IO::Error: Closed mfile that crashes the entire server.

This affects any type of segment corruption detected during startup — invalid schema version, OverflowError, FrameDecode, etc. Without clustering (no replicator), the close path is synchronous and happens to work because `@segments` is empty at the time.

To reproduce:
1. Create a queue and publish a message on a clustered LavinMQ instance
2. Stop LavinMQ
3. Write junk data into a segment file (e.g. echo -n "abcd" | dd of=msgs.0000000001 conv=notrunc)
4. Start LavinMQ — server crashes instead of closing just the affected queue

Expected behavior: The queue should be marked as closed, the server should continue running, and the queue can be restarted via the API.

Root cause: MessageStore#close (message_store.cr:263) checks `if replicator = @replicator` and takes an async path that spawns a fiber, even when no followers are connected (startup). The fix should: (a) make close synchronous when there are no followers, (b) return from load_segments_from_disk after close, and (c) skip remaining initialization when `@closed` is true. The same pattern applies to produce_metadata which also calls close mid-iteration in load_stats_from_segments.

This affects v2.7.0-rc.1

viktorerlingsson:
## Summary
- Skip async replicator close path when no followers are connected (startup), eliminating the race condition where a spawned fiber closes MFiles out from under the constructor
- Skip post-load processing (`load_stats_from_segments`, `delete_unused_segments`, etc.) when the store was closed during `load_segments_from_disk`
- Break out of `load_stats_from_segments` loop after `produce_metadata` calls `close`, preventing access to already-closed MFiles

Fixes #1701

## Testing
Run new specs

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -427,7 +427,9 @@ module LavinMQ
             end
           rescue ex
             @log.error { "Could not initialize segment #{seg}, closing message store: #{ex.message}" }
+            @segments[seg] = file
             close
+            break
           end
         end
         file.pos = 4


src/lavinmq/message_store.cr
```diff
@@ -260,7 +263,7 @@ module LavinMQ
       @empty.close
       # To make sure that all replication actions for the segments
       # have finished wait for a delete action of a nonexistent file
-      if replicator = @replicator
+      if (replicator = @replicator) && !replicator.all_followers.empty?
```
carlhoerberg, line 266: there's no issue if there are followers?

viktorerlingsson: No, I think this was only an issue during startup when there can be no followers. Other places where `close` is called is protected by mutexes and `raise ClosedError.new if @closed` as far as I can tell

src/lavinmq/message_store.cr
```diff
@@ -33,15 +33,18 @@ module LavinMQ
...
+        load_stats_from_segments
+        delete_unused_segments
+      end
+
       @wfile_id = @segments.last_key
       @wfile = @segments.last_value
       @rfile_id = @segments.first_key
       @rfile = @segments.first_value
-      @empty.set empty?
+      @empty.set empty? unless @closed
```
kickster97, line 47:
```suggestion
      unless @closed
        delete_orphan_ack_files
        load_deleted_from_disk
        load_stats_from_segments
        delete_unused_segments
        @wfile_id = @segments.last_key
        @wfile = @segments.last_value
        @rfile_id = @segments.first_key
        @rfile = @segments.first_value
        @empty.set empty?
      end
```

Should we move the `@wfile`/`@rfile`/`@empty` assignments into the `unless @closed` guard to avoid referencing closed MFiles when `load_segments_from_disk` triggers a close?

viktorerlingsson: The instance variables must be assigned in the constructor or they will be nilable, and then we'd have to handle that in all the places we use them. I think this is pretty safe, because the message_store will be `@closed`, and all paths that try to use the message_store should have guards for that.

kickster97 (approved):

PR #1712, opened 2026-02-16
fix: prevent consumer starvation in lavinmqperf throughput

### LINKED ISSUE #1711: lavinmqperf: consumer stops receiving messages while publisher continues
When running lavinmqperf throughput against remote instances, the consume rate can drop to 0 permanently while publishing
  continues at full speed. No errors are logged.

###  Symptoms:
  - Happens with both single and multiple consumers
  - Consumer connection remains open, consumer is still registered on the queue
  - Queue has messages ready for delivery
  - The consumer fiber appears to still be alive (no crash, no exception)
  - More likely to reproduce against cloud/remote instances, hard to reproduce locally
  - Lowering the server-side yield_each_received_bytes / yield_each_delivered_bytes settings seems to have some impact

####  Example:
```
  lavinmqperf throughput -z 60 -x 1 -y 1 -s 16 --uri=amqps://...
  Publish rate: 456806 msgs/s Consume rate: 34519 msgs/s
  Publish rate: 461952 msgs/s Consume rate: 0 msgs/s
  Publish rate: 463446 msgs/s Consume rate: 0 msgs/s
```

### What I've investigated/tried so far:
  - Fixed RateLimiter to use per-consumer counter for Fiber.yield frequency (was using global counter, reducing yield frequency per consumer as more consumers were added)
  - Fixed ack decision to use per-consumer counter (global counter could cause deadlock with multiple consumers and prefetch)
  - Added reconnect-on-normal-exit and catch-all exception handling to rerun_on_exception — no new output observed, confirming  the consumer fiber is alive but idle

  **Unknown**: Why the server stops delivering messages to a connected, registered consumer with messages available in the queue.

viktorerlingsson:
## Summary
- Add `Fiber.yield` every 128k publishes when no rate limit is set, preventing the publisher from monopolizing a shared thread and starving consumer fibers
- Fix ack/tx_commit to use per-consumer local counters instead of the global atomic counter, preventing deadlocks with multiple consumers and prefetch
- Fix `RateLimiter` to use its local counter for yield frequency (was using global counter or stuck at 0)

## Context
When running `lavinmqperf throughput` on machines with few CPU cores, `Fiber::ExecutionContext::Parallel` has limited threads. The publisher and consumer fibers (plus AMQP client internal fibers) can share a thread. The publisher's tight loop with zero yielding monopolized the thread, causing the consumer to stop reading from its socket. The server's TCP send buffer filled up and delivery stalled permanently.

Closes #1711

## Test plan
- [x] Tested on 2-core ARM cloud instance with `lavinmqperf throughput -z 60 -x 1 -y 2 -s 16` — consume rate no longer drops to 0

### DIFF
--- a/src/lavinmqperf/amqp/throughput.cr
+++ b/src/lavinmqperf/amqp/throughput.cr
@@ -384,9 +384,9 @@ module LavinMQPerf
             end
             queue_idx += 1 if @queue_pattern
             pubs = @pubs.add(1, :relaxed) + 1
-            ch.tx_commit if @pub_in_transaction > 0 && (pubs % @pub_in_transaction) == 0
-            break if pubs >= @pmessages > 0
             local_pubs &+= 1
+            ch.tx_commit if @pub_in_transaction > 0 && (local_pubs % @pub_in_transaction) == 0
+            break if pubs >= @pmessages > 0
             Fiber.yield if @rate.zero? && local_pubs % (128*1024) == 0
             unless @rate.zero?
               pubs_this_second += 1
@@ -458,11 +458,13 @@ module LavinMQPerf
           q.bind(@exchange, @routing_key) unless @exchange.empty?
           wait_until_all_are_connected(connected)
           rate_limiter = RateLimiter.new(@consume_rate)
+          local_consumes = 0_u64
           loop do
             if msg = q.get(no_ack: @ack.zero?)
               consumes = @consumes.add(1, :relaxed) + 1
+              local_consumes &+= 1
               record_latency(msg.body_io.to_slice) if @measure_latency
-              msg.ack(multiple: true) if @ack > 0 && consumes % @ack == 0
+              msg.ack(multiple: true) if @ack > 0 && local_consumes % @ack == 0
               break if @stopped || consumes >= @cmessages > 0
               rate_limiter.limit
             end


src/lavinmqperf/amqp/throughput.cr
```diff
@@ -385,6 +385,7 @@ module LavinMQPerf
             pubs = @pubs.add(1, :relaxed) + 1
             ch.tx_commit if @pub_in_transaction > 0 && (pubs % @pub_in_transaction) == 0
             break if pubs >= @pmessages > 0
+            Fiber.yield if @rate.zero? && pubs % (128*1024) == 0
```
carlhoerberg: should not have `local_pubs` like in the consumer case?

src/lavinmqperf/amqp/throughput.cr
```diff
@@ -418,23 +419,25 @@ module LavinMQPerf
           q.bind(@exchange, @routing_key) unless @exchange.empty?
           wait_until_all_are_connected(connected)
           rate_limiter = RateLimiter.new(@consume_rate)
+          local_consumes = 0_u64
           q.subscribe(tag: "c", no_ack: @ack.zero?, block: true, args: @consumer_args) do |m|
-            handle_consumed_message(ch, m, data, rate_limiter)
+            local_consumes += 1
```
carlhoerberg: tiny bit faster and will wrap around on overflow
```suggestion
            local_consumes &+= 1
```

src/lavinmqperf/amqp/throughput.cr
```diff
@@ -492,16 +495,15 @@ module LavinMQPerf
           @total_consumes = 0_u64
         end
 
-        def limit(consume_count : UInt64? = nil)
+        def limit
+          @total_consumes += 1
```
carlhoerberg: not introduced here, but why doesn't it have to be an `Atomic`?

viktorerlingsson: Each `RateLimiter` instance is created inside consume() and only ever accessed from that single consumer's fiber

src/lavinmqperf/amqp/throughput.cr
viktorerlingsson: Yes, that will be more correct. Fixed 👍

PR #1716, opened 2026-02-17
fix: prevent stack overflow from cascading dead-letter overflow cycles

viktorerlingsson:
## Summary
- When queues with `x-max-length` dead-letter to each other, each overflow shifts out a different message with no x-death history, so x-death based cycle detection can't catch the cross-queue chain — causing infinite recursion and SEGV
- Adds a `dlx_depth` parameter threaded through `route → publish → drop_overflow → expire_msg → route`, limiting recursive dead-letter routing to 128 levels
- No shared mutable state — the depth flows explicitly through the call chain

Run `make test SPEC=spec/queue_dead_lettering_spec.cr` to verify.

### DIFF
--- a/src/lavinmq/amqp/argument/dead_lettering.cr
+++ b/src/lavinmq/amqp/argument/dead_lettering.cr
@@ -15,12 +15,9 @@ module LavinMQ::AMQP
       class DeadLetterer
         property dlx : String? = nil
         property dlrk : String? = nil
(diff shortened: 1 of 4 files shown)


viktorerlingsson:
Just testing some different ways to handle infinite recursion in dlx

viktorerlingsson:
Superseded by #1723 and #1721

PR #1717, opened 2026-02-17
fix: prevent stack overflow from cascading DLX overflow cycles

jage:
When queues with max-length dead-letter to each other (A→B→A), each overflow shifts a different message with no x-death history, so x-death based cycle detection can't catch the cross-queue chain, causing infinite synchronous recursion and SEGV.

Track visited queue names per fiber in DeadLetterer#route, using the same Set#add? guard pattern as Exchange#find_queues. This allows arbitrary-depth chains (A→B→C) but stops immediately when a queue is revisited (A→B→A).

### DIFF
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -530,6 +530,34 @@ module DeadLetteringSpec
         end
       end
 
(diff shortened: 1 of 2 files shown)


jage:
This will not be merged, we're exploring some different ways to handle DLX recursive loops

PR #1719, opened 2026-02-18
Fix RPM test stage to run on target OS

magnushoerberg:
## Summary
- The RPM Dockerfile test stage was hardcoded to `fedora:42`, so packages built for other targets (fc43, el-9) were never tested on the actual target OS.
- The final `scratch` stage only referenced the `builder` stage via `COPY --from=builder`, so BuildKit skipped the `test` stage entirely — it was never executed in CI.
- This allowed a broken fc43 RPM (missing `Provides: user(lavinmq)` / `group(lavinmq)`) to be published, causing `dnf install lavinmq` to fail with unresolvable dependencies.

The fix parameterizes the test image via a `test_image` build arg (set per matrix entry in CI) and routes the RPMS through the test stage so BuildKit must execute it.

## Test plan
- [x] Verified `podman build` with `test_image=fedora:42` passes
- [x] Verified `podman build` with `test_image=fedora:43` passes
- [x] Verified output image structure matches CI expectations (`x86_64/*.rpm` + `*.src.rpm`)
- [ ] CI RPM workflow should pass for all matrix entries (fedora-41, fedora-42, fedora-43, el-9)

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -39,6 +39,14 @@ jobs:
       - name: Generate version string
         run: echo "version=$(./packaging/rpm/rpm-version)" >> $GITHUB_ENV
 
+      - name: Determine test image
+        run: |
+          OS="${{ matrix.os }}"
+          case "$OS" in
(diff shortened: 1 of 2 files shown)


magnushoerberg:

PR #1720, opened 2026-02-18
Only allow one follower to do bulk sync at a time

### LINKED ISSUE #1708: LavinMQ crashes when multiple followers try to full_sync simultaneously
**Describe the bug**
Leader crashes with `SIGSEGV` when two followers connect simultaneously during a full sync. The clustering follower handling uses `Fiber::ExecutionContext::Parallel `(server.cr:157) which runs fibers on real OS threads. `full_sync` (server.cr:187) is called outside `@lock`, so `files_with_hash` iterates `@files` and writes to `@checksums` concurrently from multiple threads. Crystal's Hash isn't thread-safe, so this corrupts internal state and segfaults.

**Describe your setup**
3-node cluster, leader with ~80 GiB stream data (19208 files). LavinMQ v2.7.0-rc.1.

**How to reproduce**
Have two followers connect to the leader at roughly the same time with a large dataset. The crash happens during hash calculation at varying points.

**Expected behavior**
Multiple followers should be able to connect and sync concurrently without crashing.

kickster97:

Serializes follower `full_sync` so only one follower syncs at a time. Two followers doing `full_sync` simultaneously crashes the leader with `SIGSEGV` because both iterate `@files` and write to `@checksums` from separate threads without synchronization.

Fixes #1708 

TBA

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -185,6 +185,10 @@ module LavinMQ
           end
           @followers << follower # Starts in Syncing state
         end
+        # Only allow one follower to do bulk sync at a time
+        # The bandwidth between nodes should be very high, so
+        # better with one fully synced than 2 partially synced followers
+        # Also @files and @checksums are not protected by locks
         @sync_lock.synchronize do
           follower.full_sync # sync the bulk
         end


src/lavinmq/clustering/server.cr
```diff
@@ -184,7 +185,9 @@ module LavinMQ
           end
           @followers << follower # Starts in Syncing state
         end
-        follower.full_sync # sync the bulk
+        @sync_lock.synchronize do
```
carlhoerberg, line 192:
```suggestion
        # Only allow one follower to do bulk sync at a time
        # The bandwidth between nodes should be very high, so
        # better with one fully synced than 2 partially synced followers
        # Also @files and @checksums are not protected by locks
        @sync_lock.synchronize do
```

carlhoerberg:
What will happen to the other followers that are waiting, will they timeout?

jage:
> What will happen to the other followers that are waiting, will they timeout?

So far only done a single test run, but the other follower waited patiently, no timeout/errors.

carlhoerberg (approved):

src/lavinmq/clustering/server.cr
```diff
@@ -184,7 +185,13 @@ module LavinMQ
...
-        follower.full_sync # sync the bulk
+        # Only allow one follower to do bulk sync at a time
+        # The bandwidth between nodes should be very high, so
+        # better with one fully synced than 2 partially synced followers
+        # Also @files and @checksums are not protected by locks
+        @sync_lock.synchronize do
+          follower.full_sync # sync the bulk
+        end
         @lock.synchronize do
           follower.full_sync    # sync the last
```
spuun, line 196: This also need the lock, no? (The "sync the last" sync)

PR #1721, opened 2026-02-18
Reuse the `message_expire_loop` for drop_overflow

snichme:
#### reason for the change

When a queue overflows, `drop_overflow` is called synchronously in the `publish` method.
`drop_overflow` dead-letters messages via `expire_msg`, which calls `vhost.publish` to
route the dead-lettered message to another queue. If that target queue also overflows, it
dead-letters back to the original queue, creating a recursive publish → drop_overflow →
publish → drop_overflow chain. The call stack grows with each cycle until the process
crashes.

#### solution 

Reuse the existing `message_expire_loop` fiber to perform `drop_overflow` asynchronously.
A new zero-buffered `@drop_overflow_channel` (`Channel(Nil)`) signals the fiber that
overflow needs to be checked. The `publish` and `requeue` methods send a non-blocking
signal (`try_send?`) instead of calling `drop_overflow` synchronously. This breaks the
recursive call stack because the dead-lettering happens in a separate fiber, not in the
caller's stack frame.

#### changed behaviour

**Overflow is no longer immediate**

Previously, `drop_overflow` ran synchronously inside `publish`, so the queue was trimmed
before `publish` returned. Now the overflow signal is processed by the
`message_expire_loop` fiber, which means there is a brief window where the queue
exceeds `max-length` or `max-length-bytes` until the fiber wakes up and trims it.
In practice this window is very short (a single fiber yield).

**Extra fiber wakeup per empty→non-empty transition**

The original loop was fully parked when consumers existed — it blocked on
`@consumers_empty.when_true.receive` at the top and never woke until all consumers
disconnected.

Now the loop blocks on `@msg_store.empty.when_false.receive`, which fires on the
empty→non-empty *state transition* (not per message). When consumers are present, the
fiber wakes, checks `@consumers.empty?`, enters the `select` in the `else` branch, and
parks again. This is one extra wakeup per transition, not per publish, so the overhead
is negligible.

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -671,20 +671,17 @@ module LavinMQ::AMQP
 
     private def expire_messages : Nil
       i = 0
-      @msg_store_lock.synchronize do
-        loop do
-          env = @msg_store.first? || break
-          msg = env.message
+      loop do
+        env = @msg_store_lock.synchronize do
+          first_env = @msg_store.first? || break
+          msg = first_env.message
           @log.debug { "Checking if next message #{msg} has expired" }
-          if has_expired?(msg)
-            # shift it out from the msgs store, first time was just a peek
-            env = @msg_store.shift? || break
-            expire_msg(env, :expired)
-            i += 1
-          else
-            break
-          end
-        end
+          break unless has_expired?(msg)
+          # shift it out from the msgs store, first time was just a peek
+          @msg_store.shift?
+        end || break
+        expire_msg(env, :expired)
+        i += 1
       end
       @log.info { "Expired #{i} messages" } if i > 0
     end
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -229,6 +229,7 @@ module LavinMQ::AMQP
       # Recreate channels that were closed
       @queue_expiration_ttl_change = ::Channel(Nil).new
       @message_ttl_change = ::Channel(Nil).new
+      @drop_overflow_channel = ::Channel(Nil).new
       @paused = BoolChannel.new(false)
       @consumers_empty = BoolChannel.new(true)
       @single_active_consumer_change = ::Channel(Client::Channel::Consumer).new
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -284,20 +284,20 @@ module LavinMQ::AMQP
         unless @max_length.try &.< value.as_i64
           @max_length = value.as_i64
           @effective_args.delete("x-max-length")
-          cleanup_messages(CleanupReason::Overflow)
+          @cleanup_message_channel.try_send?(CleanupReason::Overflow)
           return true
         end
       when "max-length-bytes"
         unless @max_length_bytes.try &.< value.as_i64
           @max_length_bytes = value.as_i64
           @effective_args.delete("x-max-length-bytes")
-          cleanup_messages(CleanupReason::Overflow)
+          @cleanup_message_channel.try_send?(CleanupReason::Overflow)
           return true
         end
       when "message-ttl"
         unless @message_ttl.try &.< value.as_i64
           @message_ttl = value.as_i64
-          cleanup_messages(CleanupReason::TTLChange)
+          @cleanup_message_channel.try_send?(CleanupReason::TTLChange)
           @effective_args.delete("x-message-ttl")
           return true
         end
@@ -371,7 +371,7 @@ module LavinMQ::AMQP
       @effective_args << "x-max-length-bytes" if @max_length_bytes
       @message_ttl = parse_header("x-message-ttl", Int).try &.to_i64
       @effective_args << "x-message-ttl" if @message_ttl
-      cleanup_messages(CleanupReason::TTLChange)
+      @cleanup_message_channel.try_send?(CleanupReason::TTLChange)
       @delivery_limit = parse_header("x-delivery-limit", Int).try &.to_i64
       @effective_args << "x-delivery-limit" if @delivery_limit
       overflow = parse_header("x-overflow", String)
@@ -534,7 +534,7 @@ module LavinMQ::AMQP
         @msg_store.push(msg)
       end
       @publish_count.add(1, :relaxed)
-      cleanup_messages(CleanupReason::Overflow)
+      @cleanup_message_channel.try_send?(CleanupReason::Overflow)
       true
     rescue ex : MessageStore::Error
       @log.error(ex) { "Queue closed due to error" }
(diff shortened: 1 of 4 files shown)


spuun:
`@drop_overflow_channel` must be recreated in `#reset_queue_state`, then specs passes.

snichme:
@claude you are wrong about this 1. Bug: Missing @consumers_empty.when_false in no-consumer / no-TTL select (lines 135-144)

spuun:
FYI I've found some kind of dead-lock. Trying to narrow it down.

kickster97 (approved):
To me this looks ok, I ran a few tests and i worked well

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -550,67 +563,72 @@ module LavinMQ::AMQP
       end
     end
 
-    private def drop_overflow_if_no_immediate_delivery : Nil
-      drop_overflow if (@max_length || @max_length_bytes) && !immediate_delivery?
+    private def signal_drop_overflow : Nil
```
snichme: @spuun do we really need this method? 
All the three conditions are checked in the `drop_overflow` method so it only leave the `try_send` which IMO we dont need a method for.

spuun: Yeah... I guess it may prevent the expire loop from waking unnecessarily. With many queues it may make a difference because we prevent unnecessary fiber scheduling.

I'd say we can let it be, and investigate it in upcoming PRs.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -550,67 +563,72 @@ module LavinMQ::AMQP
...
+      @drop_overflow_channel.try_send?(nil) if (@max_length || @max_length_bytes) && !immediate_delivery?
     end
 
     private def drop_overflow : Nil
+      return if immediate_delivery?
```
spuun, line 564: This was added as a result of a failing spec.

spuun:
Claude is right about `cleanup_messages`. I guess this is why some specs fails.

snichme:
Some code to tests some cases: https://gist.github.com/snichme/cb23deb8e06363deb6d7b759c6eaff09

snichme:
# Queue Overflow Handling

How the `Queue` class enforces `x-max-length`, `x-max-length-bytes`, and related limits.

## Configuration

- `@max_length` — from `x-max-length` argument or policy
- `@max_length_bytes` — from `x-max-length-bytes` argument or policy
- `@reject_on_overflow` — `true` when `x-overflow` is `"reject-publish"`, otherwise `false` (default is `"drop-head"`)
- `@dead_letter.dlx` — dead-letter exchange, if configured

## Overflow modes

There are two overflow strategies, selected by `x-overflow`:

1. **`reject-publish`** — refuse the new message before it enters the store
2. **`drop-head`** (default) — accept the new message, then drop the oldest message(s) from the head of the queue

## Methods

### `reject_on_overflow(msg)` — pre-publish guard

Called at the start of `publish`, **before** the message is pushed to the store.

- Only active when `@reject_on_overflow` is `true`
- Checks `@msg_store.size >= @max_length` or `@msg_store.bytesize + msg.bytesize >= @max_length_bytes`
- Raises `RejectOverFlow` if the queue is already at capacity — the message is never stored

### `drop_head_on_overflow` — synchronous drop (under lock)

Called **inside `@msg_store_lock.synchronize`**, immediately after pushing a message to the store (in `publish`) or requeuing a message (in `reject`).

- Returns `true` if overflow is fully resolved (or doesn't apply)
- Returns `false` if it couldn't complete — specifically when messages need dead-letter routing (`@dead_letter.dlx` is set or `segment_position.has_dlx?`)
- Skipped when `@reject_on_overflow` is true, no limits are set, or consumers can deliver immediately

Why it can't always finish: dead-letter routing involves I/O (publishing to another exchange) which shouldn't happen while holding the msg_store lock.

### `drop_head_on_overflow_async(wait)` — signal the expire loop

Called after `drop_head_on_overflow` returns `false`, or when a policy change sets a new limit.

- Sends `CleanupReason::Overflow` to `@cleanup_message_channel`
- `wait=true` (default, used on publish/requeue): blocks with `select` — waits until either the store becomes empty or the signal is sent
- `wait=false` (used for policy changes): non-blocking `try_send?`, fire-and-forget

### `drop_overflow` — async drop (in expire loop)

Called at the top of each `message_expire_loop` iteration when the store is non-empty.

- Acquires `@msg_store_lock` only for the peek/shift, releases it before calling `expire_msg`
- This is the method that handles the DLX case — it can do the dead-letter routing outside the lock
- Yields the fiber every 16K messages to avoid starving other work
- Skipped when no limits are set or consumers can deliver immediately

### `expire_msg(env, reason)` — the actual drop

Both sync and async paths converge here. Routes the message to the dead-letter exchange (if configured) and deletes it from the store.

## Publish flow

```
publish(msg)
  │
  ├─ reject_on_overflow(msg)          # if reject-publish mode, raise before storing
  │
  ├─ @msg_store_lock.synchronize:
  │    ├─ @msg_store.push(msg)
  │    └─ drop_head_on_overflow        # try sync drop under lock
  │         returns true?  ──→ done
  │         returns false? ──→ drop_head_on_overflow_async  ──→ signals expire loop
  │
  └─ publish_count += 1
```

## Expire loop (`message_expire_loop`)

```
loop:
  wait for msg_store non-empty
  │
  ├─ drop_overflow()                   # handle any pending overflow (async path)
  │
  ├─ if no consumers:
  │    ├─ if TTL set: wait for TTL timeout, cleanup signal, store empty, or consumers arrive
  │    └─ if no TTL: wait for cleanup signal or store empty
  │
  └─ if consumers exist:
       wait for cleanup signal or consumers to leave
```

The `@cleanup_message_channel` (buffered capacity 1) carries `CleanupReason::Overflow` or `CleanupReason::TTLChange` to wake the loop.

## Requeue flow (nack with requeue)

Same two-phase pattern as publish:

```
reject(sp, requeue: true)
  │
  ├─ @msg_store_lock.synchronize:
  │    ├─ @msg_store.requeue(sp)
  │    └─ drop_head_on_overflow
  │
  └─ drop_head_on_overflow_async unless drop_done
```

## Policy changes

When `max-length` or `max-length-bytes` is applied via policy, `drop_head_on_overflow_async(false)` is called (non-blocking) to signal the expire loop.

## Summary table

| Method                        | Called from             | Holds lock? | Handles DLX? | Blocking? |
|-------------------------------|------------------------|-------------|--------------|-----------|
| `reject_on_overflow`          | `publish` (before push)| No          | N/A          | No        |
| `drop_head_on_overflow`       | `publish`, `reject`    | Yes (caller)| No — returns false | No  |
| `drop_head_on_overflow_async` | `publish`, `reject`, policy | No     | N/A (signals)| Optionally|
| `drop_overflow`               | `message_expire_loop`  | Briefly     | Yes          | No        |
| `expire_msg`                  | All of the above       | No          | Yes          | Possibly  |

---

In an attempt to get a better understand of how it's fits togheter, a summary. It's a lot of cases. 
Note that this summary includes these changes (which is not yet merged): https://github.com/cloudamqp/lavinmq/commit/db569ee410eb5a4c82bc88dfdf1fc60507e68ddd

PR #1722, opened 2026-02-19
On-demand deliver_loop fibers for AMQP consumers

### LINKED ISSUE #1595: Possible segfault after `Cannot allocate memory` error

viktorerlingsson:
## Summary

Consumer `deliver_loop` fibers are now spawned on-demand instead of unconditionally at consumer creation. Each fiber exits after 30 seconds of idle waiting on an empty queue and restarts when new messages are published or requeued. This frees up a few kb's of memory per consumer, plus the virtual memory of fiber stacks.

Follows the same pattern as #1614 (idle fiber management for `message_expire_loop`): an atomic boolean tracks whether the fiber is running, and `ensure_deliver_loop` uses an atomic swap to guarantee at most one fiber per consumer. On publish, fibers are only woken when the queue transitions from empty to non-empty, so we don't pay N atomic swaps per message under sustained load.

## Performance

Per-consumer memory drops by ~9 KB RSS and ~8 MB virtual (one Crystal fiber stack) once a consumer has been idle past the 30s timeout. A consumer churn workload (16 workers, each cycling 128 subscribe/unsubscribe pairs on empty queues) runs ~15% faster on the branch (2262 vs 1972 cycles/sec) with ~23% lower peak RSS, since `add_consumer` no longer spawns a fiber.

Memory of N idle consumers, one consumer per durable queue with no message flow, sampled after the 30s deliver_loop idle timeout:

| Consumers | RSS (main) | RSS (branch) | Δ RSS         | VSZ (main) | VSZ (branch) |
|----------:|-----------:|-------------:|--------------:|-----------:|-------------:|
|     1,000 |      52 MB |        44 MB |  -8 MB (-15%) |      24 GB |        16 GB |
|     5,000 |     202 MB |       157 MB | -44 MB (-22%) |     118 GB |        79 GB |
|    10,000 |     388 MB |       300 MB | -88 MB (-23%) |     235 GB |       157 GB |

## Testing

Three new specs cover: consumer on empty queue receiving published messages, consumer on non-empty queue, requeued message delivery. Full suite passes (1553 examples).

### DIFF
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -21,6 +21,7 @@ module LavinMQ
       @metadata : ::Log::Metadata
       @unacked = Atomic(UInt32).new(0_u32)
       getter has_capacity : BoolChannel
+      @deliver_loop_running = Atomic(Bool).new(false)
 
       def initialize(@channel : AMQP::Channel, @queue : Queue, frame : AMQP::Frame::Basic::Consume)
         @tag = frame.consumer_tag
@@ -31,10 +32,9 @@ module LavinMQ
         @flow = @channel.flow?
         @metadata = @channel.@metadata.extend({consumer: @tag})
         @log = Logger.new(Log, @metadata)
-        fiber_name = "Consumer vhost=#{@queue.vhost.name} queue=#{@queue.name}"
-        @queue.deliver_loop_wg.spawn(name: fiber_name) { deliver_loop }
         @flow_change = BoolChannel.new(@flow)
         @has_capacity = BoolChannel.new(true)
+        ensure_deliver_loop unless @queue.empty?
       end
 
       def close
@@ -43,6 +43,13 @@ module LavinMQ
         @has_capacity.close
         @flow_change.close
         @queue.rm_consumer(self)
+      end
+
+      def ensure_deliver_loop
+        return if @closed
+        return if @deliver_loop_running.swap(true, :acquire_release)
+        fiber_name = "Consumer vhost=#{@queue.vhost.name} queue=#{@queue.name}"
+        @queue.deliver_loop_wg.spawn(name: fiber_name) { deliver_loop }
       end
 
       @notify_closed = ::Channel(Nil).new
@@ -89,8 +96,13 @@ module LavinMQ
             Fiber.yield
           end
         end
+      rescue IdleError
+        @log.debug { "deliver loop idle, exiting fiber" }
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
         @log.debug { "deliver loop exiting: #{ex.inspect}" }
+      ensure
+        @deliver_loop_running.set(false, :release)
+        ensure_deliver_loop unless @closed || @queue.empty?
       end
 
       private def wait_for_global_capacity
@@ -157,6 +169,9 @@ module LavinMQ
           select
           when @queue.empty.when_false.receive
           when @notify_closed.receive
+          when timeout 30.seconds
+            @log.debug { "Deliver loop idle timeout" }
+            raise IdleError.new
           end
           return true
         end
@@ -276,6 +291,8 @@ module LavinMQ
       end
 
       class ClosedError < Error; end
+
+      private class IdleError < Exception; end
     end
   end
 end
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -34,7 +34,6 @@ module LavinMQ
         @log = Logger.new(Log, @metadata)
         @flow_change = BoolChannel.new(@flow)
         @has_capacity = BoolChannel.new(true)
-        ensure_deliver_loop unless @queue.empty?
       end
 
       def close
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -101,7 +101,6 @@ module LavinMQ
         ensure_deliver_loop unless @closed || @queue.empty?
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
         @log.debug { "deliver loop exiting: #{ex.inspect}" }
-      ensure
         @deliver_loop_running.set(false, :release)
       end
 
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -97,11 +97,12 @@ module LavinMQ
         end
       rescue IdleError
         @log.debug { "deliver loop idle, exiting fiber" }
+        @deliver_loop_running.set(false, :release)
+        ensure_deliver_loop unless @closed || @queue.empty?
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
         @log.debug { "deliver loop exiting: #{ex.inspect}" }
       ensure
         @deliver_loop_running.set(false, :release)
-        ensure_deliver_loop unless @closed || @queue.empty?
       end
 
       private def wait_for_global_capacity
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -106,6 +106,10 @@ module LavinMQ
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
         @log.debug { "deliver loop exiting: #{ex.inspect}" }
         @deliver_loop_running.set(false, :release)
+      rescue ex
+        @log.debug { "deliver loop exiting unexpectedly: #{ex.inspect}" }
+        @deliver_loop_running.set(false, :release)
+        raise ex
       end
 
       private def wait_for_global_capacity
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -174,7 +174,7 @@ module LavinMQ
           :ready
         when @notify_closed.receive
           raise ClosedError.new
-        when timeout 30.seconds
+        when timeout Config.instance.deliver_loop_idle_timeout
           @log.debug { "Deliver loop idle timeout" }
           :timeout
         end
--- a/src/lavinmq/amqp/consumer.cr
+++ b/src/lavinmq/amqp/consumer.cr
@@ -77,7 +77,15 @@ module LavinMQ
             raise ClosedError.new if @closed
             next if wait_for_global_capacity
             next if wait_for_priority_consumers
-            next if wait_for_queue_ready
+            case wait_for_queue_ready
+            when :timeout
+              @log.debug { "deliver loop idle, exiting fiber" }
+              @deliver_loop_running.set(false, :release)
+              ensure_deliver_loop unless @queue.empty?
+              return
+            when :ready
+              next
+            end
             next if wait_for_paused_queue
             next if wait_for_flow
             break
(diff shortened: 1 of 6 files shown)


viktorerlingsson:
Would be good to test if this has any performance impact, and if there's any impact on latency.

src/lavinmq/amqp/queue/queue.cr
kickster97, line 896: If a publish arrives between `Consumer.new` and `Queue#add_consumer`, neither consumer or publisher spawns the `deliver_loop`: the publish iterates an empty consumer list, and the consumer's own check already saw the queue as empty, so the message will sit in the queue and the consumer stays silent. 

 should we re-check `@msg_store.empty?` once the consumer is visible under `@consumers_lock`?

```suggestion
        @consumers << consumer
        consumer.ensure_deliver_loop unless @msg_store.empty?
```

viktorerlingsson: I don't think that race is possible when running single threaded, but it doesn't hurt to be defensive! Did that in [dbfab63](https://github.com/cloudamqp/lavinmq/pull/1722/commits/dbfab63c35ccb47bcb1405a086f550f1f0e94c02)

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -526,11 +526,13 @@ module LavinMQ::AMQP
         d.add(msg)
       end
       reject_on_overflow(msg)
+      was_empty = @msg_store.empty?
       @msg_store_lock.synchronize do
```
carlhoerberg: safer:
```suggestion
      was_empty = false
      @msg_store_lock.synchronize do
        was_empty = @msg_store.empty?
```

src/lavinmq/amqp/consumer.cr
```diff
@@ -157,6 +168,9 @@ module LavinMQ
           select
           when @queue.empty.when_false.receive
           when @notify_closed.receive
+          when timeout 30.seconds
```
kickster97: should this be configurable?

src/lavinmq/amqp/consumer.cr
```diff
@@ -89,8 +95,13 @@ module LavinMQ
             Fiber.yield
           end
         end
+      rescue IdleError
+        @log.debug { "deliver loop idle, exiting fiber" }
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
         @log.debug { "deliver loop exiting: #{ex.inspect}" }
+      ensure
+        @deliver_loop_running.set(false, :release)
+        ensure_deliver_loop unless @closed || @queue.empty?
```
kickster97: Should we move this into the `IdleError` rescue instead? As is, every exit path respawns, so on `AMQP::Channel::ClosedError` / `IO::Error` / `::Channel::ClosedError` before `@closed` has propagated, the new fiber re-raises and we tight-loop until close lands. Only the idle race needs the respawn. `stream_consumer.cr` has no `IdleError` path, so the respawn can be dropped, `notify_new_message` already covers it.

src/lavinmq/amqp/consumer.cr
viktorerlingsson: I don't think there's much to gain from changing the timeout, so I don't think it's worth the "overhead" (having another config value for users to understand, documentation, etc) of having this configurable. I think we can revisit that in the future if we ever feel the need for it!

src/lavinmq/amqp/consumer.cr
viktorerlingsson: Yeah, that makes sense!

src/lavinmq/amqp/consumer.cr
```diff
@@ -89,8 +95,13 @@ module LavinMQ
             Fiber.yield
           end
         end
+      rescue IdleError
```
spuun: Isn't it possible to rework something to get rid of this error? It's not really an error but control flow, so I think it would be nice to handle in another way. What if `wait_for_queue_ready` returns symbols and change the `next if …` to something like

```crystal
case wait_for_queue_ready
when :timeout
  @deliver_loop_running.set(false, :release)
  @log.debug { "deliver loop idle, exiting fiber" }
  return
when :ready
  next
end
```

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -556,6 +559,14 @@ module LavinMQ::AMQP
       end
     end
 
+    private def ensure_consumers_deliver_loops : Nil
```
spuun, line 595: I think there's a risk that one forgets to call this method when refactoring or adding stuff to Queue. I wonder if we maybe could add a callback on message store that triggers on empty state change, and then handle this in that callback? Then we don't have to remember this in the rest of the code.

src/lavinmq/amqp/consumer.cr
```diff
@@ -89,8 +95,13 @@ module LavinMQ
...
+        @log.debug { "deliver loop idle, exiting fiber" }
+        @deliver_loop_running.set(false, :release)
+        ensure_deliver_loop unless @closed || @queue.empty?
```
spuun: We still have a race condition here, no?

No need to check closed, it's being checked in `ensure_deliver_loop`
```suggestion
        @deliver_loop_running.set(false, :release)
        ensure_deliver_loop unless @queue.empty?
```

src/lavinmq/amqp/consumer.cr
viktorerlingsson: Yeah, agree that is cleaner 👍

src/lavinmq/amqp/queue/queue.cr
viktorerlingsson: Yeah, makes sense. I explored that a bit, but didn't find a good solution without impacting bigger parts of the codebase. Instead, I made sure to have specs that cover the current ways the deliver loop is spawned. If it isn't spawned, those specs will fail. That should ensure that we don't forget to call this method.

src/lavinmq/amqp/consumer.cr
viktorerlingsson: yes, I think you're right. Removed the `@closed` check

carlhoerberg:
is there a throughput performance impact?

viktorerlingsson:
> is there a throughput performance impact?

Not as far as I can tell. I ran `benchmarks.sh`, as well as some more targeted benchmarks, and everything is within a few % vs main. There _might_ be a slight latency impact on a consumer that has been idle long enough for the fiber to shutdown, since that fiber has to be respawned. But that should be very small. 


### Throughput: `on-demand-deliver-loop-fibers` vs `main`

  `extras/benchmark.sh`, averaged over 2 runs each (10s/scenario).  Δ = (branch − main) / main.

  | Scenario | Branch (msgs/s) | main (msgs/s) | Δ |
  |---|--:|--:|--:|
  | Baseline 1→1 (16 B) | 1,040,664 | 1,040,073 | +0.1% |
  | 1→1 with properties | 897,127 | 770,814 | +16.4% |
  | Publish only* | 2,041,582 | 1,982,339 | +3.0% |
  | Consume only (drain) | 1,814,862 | 1,786,710 | +1.6% |
  | Fanout 1→4 | 1,457,607 | 1,427,592 | +2.1% |
  | Large messages (100 KB) | 8,661 | 8,505 | +1.8% |
  | Prefetch 1000, ack 100 | 113,586 | 101,632 | +11.8% † |
  | Publish confirm | 97,156 | 33,875 | +186.8% ‡ |
  | Publish in tx | 25,929 | 9,834 | +163.7% ‡ |
  | Ack in tx | 830 | 865 | −4.0% |
  | Stream 1→1 | 91,323 | 92,404 | −1.2% |
  | Stream + properties | 114,058 | 122,661 | −7.0% |
  | MT fan-in 64→1 | 216,026 | 200,011 | +8.0% |
  | MT competing 1→64 | 684,723 | 698,453 | −2.0% |
  | MT symmetric 64→64 | 1,015,906 | 903,472 | +12.4% |
  | MT sharded 64→64 | 899,309 | 803,785 | +11.9% |
  | MT stream multi 1→64 | 1,041,836 | 941,862 | +10.6% |

  \* publish rate (no consumer).
  † high run-to-run variance, treat as noise.
  ‡ **unreliable** in `benchmark.sh` - single 10s run swings ~30k-160k on *both* builds; ignore (controlled harness shows
  confirm ~equal at ~160k).

  **No regression in any scenario.** Core paths (baseline, drain, fanout, large, competing) are flat. Several multi-consumer
  scenarios trend faster on the branch, but on single-run samples I'd treat that as unconfirmed rather than a claimed speedup.
  Note: the +16% on *1→1 properties* also appears on an unrelated feature branch vs the same `main` binary, so it's likely a
  shared-base/build artifact, not an effect of this change.

src/lavinmq/amqp/consumer.cr
```diff
@@ -91,6 +105,7 @@ module LavinMQ
         end
       rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
```
kickster97, line 106: should we clear `@deliver_loop_running` on unexpected `deliver_loop` exits too? The handled errors cover the expected close/IO paths, but if an internal bug raises anything else, the fiber exits while the flag stays `true`, and future `ensure_deliver_loop` calls will not respawn it.

  ```suggestion
        rescue ex : ClosedError | Queue::ClosedError | AMQP::Channel::ClosedError | ::Channel::ClosedError | IO::Error
          @log.debug { "deliver loop exiting: #{ex.inspect}" }
          @deliver_loop_running.set(false, :release)
        rescue ex
          @deliver_loop_running.set(false, :release)
          raise ex
        end

kickster97 (approved):

src/lavinmq/amqp/consumer.cr
viktorerlingsson: Yeah, that's probably safest. I don't think it matters a whole lot in reality though. If there's an unexpected exception it's probably pretty likely that the consumer/channel/connection closes anyway (and in main the deliver loop exits on all errors and there's no respawn mechanism at all). But better safe than sorry!

PR #1723, opened 2026-02-19
Bugfix: dead letter cycle detection

spuun:
Fixes a bug in dead letter cycle detection where cycles were only checked against the source queue rather than against each destination queue individually. This could cause infinite loops when a dead letter exchange routes to multiple queues.

The previous implementation checked for cycles once before routing to any destination queues. It compared the source queue name against the x-death header to detect cycles. However, when a dead letter exchange routes to multiple queues, this approach was insufficient — each destination queue needs to be checked.

Specs

### DIFF
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -577,7 +577,7 @@ module DeadLetteringSpec
         end
       end
 
-      pending "should detect cycle with three queues in chain" do
+      it "should detect cycle with three queues in chain" do
         with_amqp_server do |s|
           with_channel(s) do |ch|
             qargs1 = {


spec/queue_dead_lettering_spec.cr
```diff
@@ -545,6 +545,74 @@ module DeadLetteringSpec
           msg.body_io.to_s.should eq "msg1"
         end
       end
+
+      pending "should detect cycle with two queues in chain" do
```
snichme:
```suggestion
      it "should detect cycle with two queues in chain" do
```

spec/queue_dead_lettering_spec.cr
```diff
@@ -545,6 +545,74 @@ module DeadLetteringSpec
...
+            # to q2. When q2 now overflows m1 shouldn't be dead lettered to q1,
+            # but instead be dropped.
+
+            get1(q1).body_io.to_s.should eq "m2"
+            get1(q2).body_io.to_s.should eq "m3"
+          end
+        end
+      end
+
+      pending "should detect cycle with three queues in chain" do
```
snichme:
```suggestion
      it "should detect cycle with three queues in chain" do
```

Issue #1696, opened 2026-02-12
Connection details/properties design inconsistencies


jage:
This is nitpick and not an issue that breaks the usability of LavinMQ, but still once seen can't be unseen. :) 
 
3. Capabilities under "Client Properties" has a weird indentation.
4. The table borders under "details" looks a bit inconsistent to me<img width="1363" height="471" alt="Image" src="https://github.com/user-attachments/assets/3fc0ffd0-a050-4795-ac65-3f8cc1f2f8db" />

This is a local shovel (e.g. shovels messages between queues on the same broker).

PR #1732, opened 2026-02-20
Add empty cells in connection details for consistent rendering

snichme:
### WHAT is this pull request doing?

Fixes #1696 by adding empty cells in details.

image

### HOW can this pull request be tested?

Visually

PR #1733, opened 2026-02-20
fix: collapse whitespace in client capabilities display

hananasen:
### Summary
As described in https://github.com/cloudamqp/lavinmq/issues/1696, the capabilities under "Client Properties" has an odd indentation. its because the `jsonToText` separates each field in the json object by newlines and adds a 2 space indentation before them, however trims the whitespaces on the first row.
Then we preserve the newlines by adding the pre-wrap to the pre-object. One solution could be to modify the jsonToText or separate each field in a div. But I figured this would be the smallest change.

Visually 👁️ 

https://github.com/user-attachments/assets/9af0071f-5c22-4a10-a2db-df72086a4f9f

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -1382,6 +1382,7 @@ footer .logo {
 
 #cp-capabilities {
   white-space: pre-line;
+  margin: 0
 }
 
 #q-features span,


hananasen:

kickster97 (approved):
lgtm, but mange.li can have last words on adding a new css class :)

snichme:
Two PR for the same issue: https://github.com/cloudamqp/lavinmq/pull/1732 not sure which is best :)

hananasen:
> Two PR for the same issue: #1732 not sure which is best :)

No strong opinions from me. I see in the css library that you don't seem to work so much with ids there. And mine applies to the cp-capabilities. Could get quite messy? So that is one argument to go with https://github.com/cloudamqp/lavinmq/pull/1732 instead :)

But I think the best would be to add `pre-line` on the global pre-object, however that could mess with message bodies with intentional indentation.. Leaving it to the css gurus and your team to decide 😄

snichme:
I like the css `pre-line` solution better than my regex to remove white spaces. Didn't know that existed. So a combination might be the best solution here

snichme:
One benefit of the regex that I saw now is that it also strips a newline in the beginning and end to it gives less padding above and below the text

regex: 
image

---

without regex:
image

hananasen:
Nice! Have you tried applying pre-line on the details table? 🤔 could be a compromise :D but then again, if the regex solves more problems. I think we could go ahead with it!

snichme:
Yes, still the same result. 
Probably because the `jsonToText` strips the first `{` and just leaves an empty line and the second `.replace` in my PR removes that line.. Will try some more

snichme:
hmm, looks like we only need a `margin:0` to the `<pre>` as well and it will solve it.

static/main.css
```diff
@@ -1380,6 +1380,10 @@ footer .logo {
   color: var(--color-text-muted);
 }
 
+#cp-capabilities {
+  white-space: pre-line;
```
snichme, line 1384:
```suggestion
  white-space: pre-line;
  margin: 0
```

I think this will do it. Please verify

hananasen: yes that worked :tada: but still, wondering if we want this just on this element or if we'd like all tables using pre- to use the same format :) in that case it is probably best to add both these properties to the global pre.. no strong opinions, just raising the question

snichme: I think that we might want `<pre>` that can be indented in some places. I think we should start like this, we can move it to a class later on if we needed in more places.

hananasen:
Not able to merge since I am not part of the team, but merge if you feel like it is ok @snichme :)

PR #1734, opened 2026-02-20
fix: return 400 for invalid binding routing key

### LINKED ISSUE #1724: Binding queue to exchange without routing key fails but LavinMQ says it worked.
**Describe the bug**

When I bind a queue to a Consistent Hash exchange, if I only enter the name of the exchange but not enter any value for the binding key (should be a number for this exchange) I get a warning of an internal server error but also a notification about the queue has been bound to the exchange but it was actually never bound. 

Reproduce
Setup a Consistent Hash Exchange
Create a queue
From the queue page bind the queue to the exchange without any value in the "Routing key" field. 

LavinMQ Version: 2.7.0-rc.1

viktorerlingsson:
## Summary
 - Catch `LavinMQ::Error::PreconditionFailed` centrally in `ApiErrorHandler` and return `400 Bad Request`, instead of handling it in individual controller actions
  - Remove redundant begin/rescue blocks from bindings and queues controllers
  - Add is_error guard in the binding form handlers (queue, exchange, stream pages) to prevent showing false success notifications when the request fails

Fixes #1724

### DIFF
--- a/src/lavinmq/http/controller/bindings.cr
+++ b/src/lavinmq/http/controller/bindings.cr
@@ -62,11 +62,7 @@ module LavinMQ
             unless routing_key
               bad_request(context, "Field 'routing_key' is required")
             end
-            begin
-              ok = e.vhost.bind_queue(q.name, e.name, routing_key, arguments)
-            rescue ex : Error::PreconditionFailed
-              bad_request(context, ex.message)
-            end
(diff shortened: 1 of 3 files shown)


snichme:
> A `PreconditionFailed` should probably always be a 400, so what about handling this in the error handler?

I think this makes sense

PR #1737, opened 2026-02-23
Reapply policies after queue restart

### LINKED ISSUE #1736: Policies not reapplied after queue restart
**Describe the bug**
Queue policies (e.g. `max-length`) are not enforced after restarting a closed queue. The policy shows as applied in the API response, but the limits are not actually enforced. Queue arguments (e.g.`x-max-length`) work correctly after restart.

  **Describe your setup**
  LavinMQ v2.7.0-rc.1-4-gab0aeafe7, macOS

  **How to reproduce**
  1. Create a policy with `max-length: 2` matching a durable queue
  2. Close the queue (e.g. via data corruption)
  3. Restart the queue via API (`PUT /api/queues/%2f/<name>/restart`)
  4. Publish more than 2 messages to the queue
  5. All messages are accepted — the policy limit is ignored

  **Expected behavior**
  After restart, the policy `max-length` should be enforced and excess messages should be dropped.

kickster97:

Reapplies policies after restarting a closed queue. Previously, only queue arguments (e.g. `x-max-length`) were restored via `handle_arguments`, but policies set via the management API were not reapplied. 

Fixes #1736 

Run the new spec: `make test SPEC=spec/queue_spec.cr`

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -204,7 +204,7 @@ module LavinMQ::AMQP
       @msg_store = init_msg_store(@data_dir)
       @empty = @msg_store.empty
       start.tap do |started|
-        apply_policy(@policy, @operator_policy) if started
+        reapply_policy if started
       end
     end
 
--- a/src/lavinmq/policy.cr
+++ b/src/lavinmq/policy.cr
@@ -7,6 +7,10 @@ module LavinMQ
     getter operator_policy : OperatorPolicy?
     getter effective_policy_args = Array(String).new
 
+    def reapply_policy
+      apply_policy(@policy, @operator_policy)
+    end
+
     def apply_policy(policy : Policy?, operator_policy : OperatorPolicy?)
       clear_policy
       effective_policy_args = Array(String).new


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -203,7 +203,9 @@ module LavinMQ::AMQP
       reset_queue_state
       @msg_store = init_msg_store(@data_dir)
       @empty = @msg_store.empty
-      start
+      start.tap do |started|
+        apply_policy(@policy, @operator_policy) if started
```
spuun: Queue should probably avoid accessing `@policy` and  `@operator_policy` and use a method from `PolicyTarget` instead. We want less coupling between everything in Queue.

I suggest adding a `reapply_policy`or something to `PolicyTarget` which is after called `start`.

kickster97: yeah good thinking!

kickster97: We could also remove the redundant `getter policy` / `getter operator_policy` from `Queue` and `Exchange` (they shadow the ones from `PolicyTarget`) and stop accessing `@policy`/`@operator_polic`y directly outside the module. That could squeeze into this PR, or I can open another one?

snichme (approved):

PR #1746, opened 2026-02-23
Disconnect MQTT clients on OAuth token expiration

### LINKED ISSUE #1744: MQTT OAuth: clients not disconnected on token expiration
**Describe the bug**

MQTT clients are not disconnected when their OAuth token expires. The AMQP client sets up an `on_expiration` callback during initialization that closes the connection when the token expires. The MQTT client has no equivalent, so MQTT clients with expired tokens remain connected indefinitely.

  **Describe your setup**

LavinMQ 2.7.0-rc.1, MQTT client connecting with a short-lived JWT token obtained via OAuth client credentials flow.

 **How to reproduce**

  1. Connect an MQTT client using a short-lived JWT token (e.g. 60s expiry)
  2. Wait for the token to expire
  3. The MQTT client remains connected and can continue publishing/subscribing

 **Expected behavior**

  MQTT clients should be disconnected when their OAuth token expires, consistent with AMQP.

kickster97:

Fixes #1744

The AMQP client sets up an `on_expiration` callback that closes the connection when an OAuth token expires. The MQTT client had no equivalent, so MQTT clients with expired tokens remained connected indefinitely.

Adds the same `on_expiration` callback to the MQTT client, closing the connection with reason "token expired" when the token lifetime runs out.

  Run with: `make test SPEC=spec/mqtt/oauth_expiration_spec.cr`

### DIFF
--- a/spec/mqtt/oauth_expiration_spec.cr
+++ b/spec/mqtt/oauth_expiration_spec.cr
@@ -15,7 +15,7 @@ module MqttSpecs
 
         # Create OAuthUser with a very short-lived token
         permissions = {"/" => {config: /.*/, read: /.*/, write: /.*/}}
-        user = OAuthUserHelper.create_user(RoughTime.utc + 1.second, permissions)
+        user = OAuthUserHelper.create_user(RoughTime.utc + 50.milliseconds, permissions)
 
         broker = server.@mqtt_brokers.@brokers["/"]
         packet = MQTT::Protocol::Connect.new(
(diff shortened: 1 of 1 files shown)


kickster97:
@claude this is fixed in #1745

spec/mqtt/oauth_expiration_spec.cr
```diff
@@ -0,0 +1,77 @@
...
+          will: nil,
+        )
+
+        broker.add_client(mqtt_io, conn_info, user, packet)
+
+        # Client should be connected
+        reader.closed?.should be_false
+
+        # Wait for token to expire and the on_expiration callback to fire
+        sleep 1.5.seconds
```
viktorerlingsson: Can we use shorter sleeps and expirations in the specs? Would be nice to not add too much runtime to the specs! 
Looks like `auth_spec` only uses 50ms to test expiration

kickster97: I'll do you one better and replaced that sleep with `wait_for` / `select` where possible

PR #1747, opened 2026-02-23
Encapsulate collections and extract stores (prep for MT safety)

carlhoerberg:
## Motivation
First step toward multi-threading safety. LavinMQ's hot collections (`@exchanges`, `@queues`, `@connections`, `@channels`, `@consumers`, `@vhosts`, `@users`, `@parameters`, `@shovels`) and stat counters were exposed as raw `getter`/`forward_missing_to`, leaving every caller free to mutate or iterate them without coordination. That's fine under single-threaded `preview_mt`, but blocks any move to true concurrent execution.

Encapsulating the collections behind explicit accessors is the prerequisite. With every read/write going through a known method, the next step can swap each backing store for a thread-safe primitive without touching call sites:
- collections protected by `Sync::Shared` (or per-store mutex),
- counters replaced with `Atomic(UInt64)`,
- booleans (`@flow`, `@closed`, …) replaced with `Atomic(Bool)` / `BoolChannel`.

This PR does no locking changes itself — it just narrows the surface so a future PR can.

## Summary
- Encapsulate all mutable collections behind explicit accessor methods; remove `getter`/`forward_missing_to` exposure
- Extract three stores from `VHost`: `ConnectionStore`, `DirectReplyConsumerStore`, `DefinitionsStore` (queues + exchanges + definitions-file persistence)
- Replace `Iterator(T)` returns with `Array(T)` snapshots across HTTP controllers and collection accessors — iterators hold references to underlying collections and are not safe under concurrent mutation
- Replace `Server` and store `getter?` booleans with explicit methods for future atomic-safe access

## Key changes
- **VHost**: snapshot accessors are plain plurals (`queues`, `exchanges`, `connections`); block-form `each_*` retained for internal iteration; `definitions` is now private
- **Safe registration**: `register_exchange`/`register_queue` replace `*_unsafe_put` — locked, drop the explicit name parameter (read from the object), idempotent on duplicate to support `init_delayed_queue` re-registration
- **Stores** (VHostStore, UserStore, ParameterStore, Shovel::Store): single `values` accessor (no more parallel `values`/`values_dup`)
- **AMQP/MQTT Client**: `channels` snapshot accessor
- **Exchange types**: `bindings_details` returns `Array(BindingDetails)` instead of `Iterator`
- **HTTP `page()`**: accepts `Array(SortableJSON)` — since it always fully traverses for sorting/pagination, lazy evaluation provided no benefit
- **Queue**: `unacked_messages` and `bindings` return arrays; `consumers` encapsulated with accessor methods

## Next steps (separate PRs)
- Wrap collections with `Sync::Shared` / per-store mutex
- Convert stat counters to `Atomic`
- Convert `@flow`, `@closed`, etc. to atomic primitives

## Test plan
- [x] `make bin/lavinmq CRYSTAL_FLAGS=` compiles cleanly
- [x] `make lint` passes (0 failures)
- [x] `make test` passes (1653 examples, 0 failures, 9 pending)

### DIFF
--- a/spec/api/bindings_spec.cr
+++ b/spec/api/bindings_spec.cr
@@ -65,7 +65,7 @@ describe LavinMQ::HTTP::BindingsController do
         response = http.post("/api/bindings/%2f/e/be1/q/bindings_q1", body: body)
         response.status_code.should eq 201
         response.headers["Location"].should eq "bindings_q1/rk"
-        s.vhosts["/"].exchanges["be1"].bindings_details.first.routing_key.should eq "rk"
+        s.vhosts["/"].exchange("be1").bindings_details.first.routing_key.should eq "rk"
       end
     end
 
@@ -142,7 +142,7 @@ describe LavinMQ::HTTP::BindingsController do
         props = binding[0]["properties_key"].as_s
         response = http.delete("/api/bindings/%2f/e/be1/q/bindings_q1/#{props}")
         response.status_code.should eq 204
-        s.vhosts["/"].exchanges["be1"].bindings_details.empty?.should be_true
+        s.vhosts["/"].exchange("be1").bindings_details.empty?.should be_true
       end
     end
   end
--- a/spec/api/definitions_spec.cr
+++ b/spec/api/definitions_spec.cr
@@ -84,7 +84,7 @@ describe LavinMQ::GlobalDefinitions do
       begin
         with_amqp_server do |s|
           LavinMQ::GlobalDefinitions.import_from_file(tmpfile, s)
-          s.vhosts["/"].queues.has_key?("load_def_q1").should be_true
+          s.vhosts["/"].queue_exists?("load_def_q1").should be_true
         end
       ensure
         File.delete?(tmpfile)
--- a/spec/api/definitions_spec.cr
+++ b/spec/api/definitions_spec.cr
@@ -264,7 +264,7 @@ describe LavinMQ::HTTP::Server do
         body = %({ "queues": [{ "name": "import_q1", "vhost": "/", "durable": true, "auto_delete": false, "arguments": {} }] })
         response = http.post("/api/definitions", body: body)
         response.status_code.should eq 200
-        s.vhosts["/"].queues.has_key?("import_q1").should be_true
+        s.vhosts["/"].queue_exists?("import_q1").should be_true
       end
     end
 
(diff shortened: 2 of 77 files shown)


carlhoerberg:
"error_max_turns" from claude code review, fyi @viktorerlingsson

carlhoerberg:
extracted from #1683

carlhoerberg:
this is the first step to multi threading

1. change signatures
2. make collections threadsafe using `Sync::Shared` and `Atomic`
3. enabled multi threading

spuun:
In `ParameterStore` the `forward_missing_to` is removed and the necessary methods are added explicitly. I like this approach.

Why not make this for other things too, like `@exchanges` in `VHost`? Add an entitiy `Exchanges` that's thread safe with only necessary methods? Same with `Queues`, `Connections` (or maybe `Clients`) etc? Hiding the non thread-safe Hashes and Arrays in custom entities prevents you from accidentally accessing non-safe structures.

carlhoerberg:
### Gemini review

The encapsulate-collections branch performs a comprehensive refactoring of the VHost class and related components in LavinMQ. The primary goal is to encapsulate internal collections
  (exchanges, queues, connections, etc.) into specialized store classes and "tighten" the public API to prevent direct access to internal data structures.

  Key Changes

   1. New Store Classes:
       * DefinitionsStore: Encapsulates exchanges, queues, and their persistence logic (formerly in VHost).
       * ConnectionStore: Manages the list of active client connections.
       * DirectReplyConsumerStore: Handles AMQP direct reply-to consumers.
   2. API Tightening:
       * VHost no longer exposes internal Hash objects for exchanges and queues via getter. Instead, it provides specific accessor methods (exchange?, queue?, etc.) and returns snapshots
         (Array) for bulk access.
       * ParameterStore, UserStore, and VHostStore now use explicit delegation instead of forward_missing_to, hiding internal Hash methods.
       * Redundant _dup suffix methods (like channels_dup, values_dup) have been renamed to their base names or removed, as Hash#values in Crystal already returns a copy.
   3. Refactored Persistence:
       * The management of definitions.amqp is now entirely within DefinitionsStore, including loading, compacting, and syncing.
   4. Performance Optimizations:
       * Introduced each_* iterator methods (e.g., each_connection, each_channel) to avoid unnecessary array allocations during common iteration tasks, such as stats collection.

  Architectural Impact

   * Improved Maintainability: VHost.cr is significantly reduced in size and complexity. Responsibilities are clearly delineated between stores.
   * Safety: By returning Array snapshots instead of mutable Hash references, the internal state of VHost is better protected from accidental external modification.
   * Separation of Concerns: Entity management (AMQP frames, persistence) is now decoupled from the higher-level vhost logic.

  Recommendation

  The refactoring is high-quality, follows idiomatic Crystal patterns, and improves the overall architecture of the codebase. All mechanical changes in controllers and specs appear correct
  and consistent with the new API.

  Status: Approved

kickster97 (approved):

PR #1751, opened 2026-02-25
Handle closed MFiles during clustering full_sync

viktorerlingsson:
## Summary

MFiles can be closed while a full_sync is in progress (e.g. queue deletion, compaction, vhost removal). This caused two issues:

- `files_with_hash` crashes with `IO::Error("Closed mfile")` when iterating. Now skips closed MFiles.
- `with_file` yields a closed MFile to the caller. Now yields `nil` instead (callers already handle nil).

Same fix for main in #1752

## Test plan

- Added spec: `files_with_hash` skips MFiles closed during iteration
- Added spec: `with_file` yields nil for a closed MFile

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -106,8 +106,8 @@ module LavinMQ
               yield({path, hash})
             end
           rescue ex : IO::Error
-            Log.error(exception: ex) { "MFile closed during hash calculation: #{path}" }
-            raise ex
+            Log.warn(exception: ex) { "MFile closed during hash calculation: #{path}" }
+            next
           end
         end
       end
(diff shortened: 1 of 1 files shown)


src/lavinmq/clustering/server.cr
```diff
@@ -86,9 +86,11 @@ module LavinMQ
         sha1 = Digest::SHA1.new
         @files.each do |path, mfile|
```
spuun, line 87: Maybe we should add an `rescue` in this loop, log the exception together with `path` and re-raise? I guess there is a tiny risk that a file is closed between then closed check and the usage.

viktorerlingsson: Yes, that's probably best. But shouldn't technically be needed in v2.6.x, right? The parallel context is only introduced in 2.7.0, so in v2.6.x there should be no way for the file to be closed between the closed check and usage. But I'll add it to both this and #1752 .

spuun: Right! I missed the double PRs at first.

viktorerlingsson: Hmm, should we really re-raise? or just skip and move on to the next?

viktorerlingsson:
see https://github.com/cloudamqp/lavinmq/issues/1757#issuecomment-3985085374

viktorerlingsson:
Superseded by #1792

PR #1752, opened 2026-02-25
Handle closed MFiles during clustering full_sync

viktorerlingsson:
## Summary

Equivalent of #1751 for main.

MFiles can be closed while a full_sync is in progress (e.g. queue deletion, compaction, vhost removal). This caused two issues:

- `files_with_hash` crashes with `IO::Error("Closed mfile")` when iterating. Now skips closed MFiles.
- `with_file` yields a closed MFile to the caller. Now yields `nil` instead (callers already handle nil).

## Test plan

- Added spec: `files_with_hash` skips MFiles closed during iteration
- Added spec: `with_file` yields nil for a closed MFile

### DIFF
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -107,8 +107,8 @@ module LavinMQ
               yield({path, hash})
             end
           rescue ex : IO::Error
-            Log.error(exception: ex) { "MFile closed during hash calculation: #{path}" }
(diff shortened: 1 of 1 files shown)


viktorerlingsson:
see https://github.com/cloudamqp/lavinmq/issues/1757#issuecomment-3985085374

viktorerlingsson:
Superseded by #1792

PR #1753, opened 2026-02-25
Add Sync::Shared lock for @files and @checksums in Clustering::Server

viktorerlingsson:
## Summary

  - `@files` and `@checksums` in `Clustering::Server` are accessed from multiple threads without synchronization: main server threads mutate them (via `register_file`, `delete_file`, `append`, `replace_file`) while clustering follower threads read them (via `files_with_hash`, `with_file`) on the `@mt` parallel execution context.                                                   
  - Wraps both in a `Sync::Shared` (`@file_index`) which uses a `RWLock` — multiple followers can read concurrently        
  (`.shared`), while mutations require exclusive access (`.lock`). In `files_with_hash`, we hold the shared lock during SHA1      
  computation to prevent files from being deleted/unmapped during the calculation.
  - Related to #1708. Complements #1752 (closed MFile handling) — both address different races in the same code paths.

### DIFF
--- a/spec/clustering_spec.cr
+++ b/spec/clustering_spec.cr
@@ -98,14 +98,14 @@ describe LavinMQ::Clustering::Client, tags: "etcd" do
       msg_store.@segments.size.should be > 1
 
       msg_store.close
-      files_before = replicator.file_keys.sort!
+      files_before = replicator.@file_index.shared { |files, _| files.keys }.sort!
       replicator.close
 
       # Re-open the message store and verify the same files are registered
       replicator = LavinMQ::Clustering::Server.new(LavinMQ::Config.instance, etcd, node_id + 1)
       msg_store = LavinMQ::MessageStore.new(msg_dir, replicator)
       msg_store.close
-      files_after = replicator.file_keys.sort!
+      files_after = replicator.@file_index.shared { |files, _| files.keys }.sort!
 
       files_before.size.should be > 2
       files_before.find(&.ends_with?("meta.0000000001")).should_not be_nil
--- a/src/lavinmq/clustering/server.cr
+++ b/src/lavinmq/clustering/server.cr
@@ -35,18 +35,18 @@ module LavinMQ
       @dirty_isr = true
       @id : Int32
       @config : Config
-      @file_index_lock : Sync::Shared(Tuple(Hash(String, MFile?), Checksums))
+      @file_index : Sync::Shared(Tuple(Hash(String, MFile?), Checksums))
 
       def initialize(config : Config, @etcd : Etcd, @id : Int32)
         Log.info { "ID: #{@id.to_s(36)}" }
         @config = config
         @data_dir = @config.data_dir
         @password = password
-        @file_index_lock = Sync::Shared.new({Hash(String, MFile?).new, Checksums.new(@data_dir)}, :unchecked)
+        @file_index = Sync::Shared.new({Hash(String, MFile?).new, Checksums.new(@data_dir)}, :unchecked)
       end
 
       def clear
-        @file_index_lock.lock do |(files, checksums)|
+        @file_index.lock do |files, checksums|
           files.clear
           checksums.clear
         end
@@ -54,7 +54,7 @@ module LavinMQ
 
       def register_file(path : String)
         path = strip_datadir path
-        @file_index_lock.lock do |(files, checksums)|
+        @file_index.lock do |files, checksums|
           files[path] = nil
           checksums.delete(path)
         end
@@ -62,7 +62,7 @@ module LavinMQ
 
       def register_file(file : File)
         path = strip_datadir file.path
-        @file_index_lock.lock do |(files, checksums)|
+        @file_index.lock do |files, checksums|
           files[path] = nil
           checksums.delete(path)
         end
@@ -70,7 +70,7 @@ module LavinMQ
 
       def register_file(mfile : MFile)
         path = strip_datadir mfile.path
-        @file_index_lock.lock do |(files, checksums)|
+        @file_index.lock do |files, checksums|
           files[path] = mfile
           checksums.delete(path)
         end
@@ -78,7 +78,7 @@ module LavinMQ
 
       def replace_file(path : String) # only non mfiles are ever replaced
         path = strip_datadir path
-        @file_index_lock.lock do |(files, checksums)|
+        @file_index.lock do |files, checksums|
           files[path] = nil
           checksums.delete(path)
         end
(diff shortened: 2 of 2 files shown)


spec/clustering/server_spec.cr
```diff
@@ -40,4 +40,104 @@ describe LavinMQ::Clustering::Server, tags: "etcd" do
...
+
+      mfiles = Array(MFile).new
+      10.times do |i|
+        mfile = MFile.new(File.join(data_dir, "concurrent_test_#{i}"), 1024)
+        mfile.print "content_#{i}"
+        server.register_file(mfile)
+        mfiles << mfile
+      end
+
+      mt = Fiber::ExecutionContext::Parallel.new("test-concurrent-hash", 2)
```
spuun: Looks like only one fiber is ever spawned in this EC, so maybe use an `Isolated` instead? Then we won't have a dangling thread for the rest of the spec run, as the thread in an `Isolated` context exits when the fiber does.

Not for this PR, but we should probably make a helper for EC:s in specs so we reuse EC:s instead of creating new ones, since it's never stopped, so we won't have "tons" of EC's running during specs.

spec/clustering/server_spec.cr
```diff
@@ -40,4 +40,104 @@ describe LavinMQ::Clustering::Server, tags: "etcd" do
...
+      10.times do |i|
+        mfile = MFile.new(File.join(data_dir, "concurrent_wf_#{i}"), 1024)
+        mfile.print "content_#{i}"
+        server.register_file(mfile)
+        mfiles << mfile
+      end
+
+      mt = Fiber::ExecutionContext::Parallel.new("test-concurrent-wf", 2)
+      done = Channel(Nil).new
+      iterations = 200
```
spuun: I think this should be much higher, like 1000000. With 200 it crashes very rarely, with a million pretty often.
```suggestion
      iterations = 1_000_000
```

spec/clustering/server_spec.cr
```diff
@@ -40,4 +40,104 @@ describe LavinMQ::Clustering::Server, tags: "etcd" do
...
+        end
+      end
+
+      select
+      when done.receive
+      when timeout(30.seconds)
+        fail "Timed out waiting for concurrent operations"
+      end
+    ensure
+      mfiles.try &.each { |mf| mf.delete rescue nil }
```
spuun, line 139:
```suggestion
      mfiles.try &.each { |mf| mf.delete rescue nil }
      FileUtils.rm_rf data_dir
```

spec/clustering/server_spec.cr
viktorerlingsson: > Looks like only one fiber is ever spawned in this EC, so maybe use an Isolated instead? Then we won't have a dangling thread for the rest of the spec run, as the thread in an Isolated context exits when the fiber does.

Yes, fixed!

src/lavinmq/clustering/server.cr
```diff
@@ -106,7 +122,7 @@ module LavinMQ
               sha1.file filename
             end
             hash = sha1.final
-            @checksums[path] = hash
+            @files_lock.synchronize { @checksums[path] = hash if @files.has_key?(path) }
```
spuun: I think we should benchmark how bad it is to lock the entire hash calculation section.

What if we can tell the replicator to calculate the hash for a file, like when the message store is rolling over to the next segment? Then there wouldnt't be many calculations here?

viktorerlingsson: Did some benchmarking with main, the code in this PR, and with locking around the hash generation.

| Build | Runs | Avg sync (s) | Avg publish (msg/s) | Avg consume (msg/s) |                                                    
  |-------|------|-------------|--------------------|--------------------|                                                     
  | `main` | 3 | 37.0 | 112,453 | 112,443 |                                                                                      
  | `#1753` (snapshot) | 3 | 34.1 | 115,484 | 115,475 |                                                                          
  | `#1753` + lock around hash | 3 | 29.6 | 84,790 | 84,787 | 

Benchmark: 10GB pre-loaded data, `full_sync` with concurrent throughput load (`-x 2 -y 2 -s 1024`).


So holding the lock for the hash calculation lowers throughput by ~25% in this benchmark, while just doing the snapshoting and briefly aquiring lock to verify the file still exists a couple of times doesn't really add any overhead.

viktorerlingsson: > What if we can tell the replicator to calculate the hash for a file, like when the message store is rolling over to the next segment? Then there wouldnt't be many calculations here?

Yeah, that would make adding a new follower quicker, but it could also "cost" a lot of hash calculations in a scenario where you never intended to add any followers. But then you probably shouldn't have clustering enabled.. Not sure what the best approach is! But regardless, I think we can do that in another PR if we want to?

viktorerlingsson:
Not sure about holding a shared lock in `with_file` like claude suggests, that drops throughput to almost zero during file sync...

Edit: Wait, that shouldn't even be needed? there's a lock in message_store.close that prevents closing the mfiles during sync.

spuun:
> Not sure about holding a shared lock in `with_file` like claude suggests, that drops throughput to almost zero during file sync...
> 
> Edit: Wait, that shouldn't even be needed? there's a lock in message_store.close that prevents closing the mfiles during sync.

Yeah, we should be safe with the current approach.

src/lavinmq/clustering/server.cr
```diff
@@ -31,92 +32,120 @@ module LavinMQ
       @sync_lock = Mutex.new(:unchecked)
       @followers = Array(Follower).new(4)
       @password : String
-      @files = Hash(String, MFile?).new
       @dirty_isr = true
       @id : Int32
       @config : Config
+      @file_index_lock : Sync::Shared(Tuple(Hash(String, MFile?), Checksums))
```
spuun: To me this doesn't need the suffix. A `Shared(T)` is a "special version of T", so IMO it's closer to `T` than a lock.
```suggestion
      @file_index : Sync::Shared(Tuple(Hash(String, MFile?), Checksums))
```

src/lavinmq/clustering/server.cr
```diff
@@ -31,92 +32,120 @@ module LavinMQ
...
         @data_dir = @config.data_dir
         @password = password
-        @checksums = Checksums.new(@data_dir)
+        @file_index_lock = Sync::Shared.new({Hash(String, MFile?).new, Checksums.new(@data_dir)}, :unchecked)
       end
 
       def clear
-        @files.clear
-        @checksums.clear
+        @file_index_lock.lock do |(files, checksums)|
```
spuun: Parentheses not necessary
```suggestion
        @file_index_lock.lock do |files, checksums|
```

spec/clustering_spec.cr
```diff
@@ -98,14 +98,14 @@ describe LavinMQ::Clustering::Client, tags: "etcd" do
       msg_store.@segments.size.should be > 1
 
       msg_store.close
-      files_before = replicator.@files.keys.sort!
+      files_before = replicator.file_keys.sort!
```
spuun: I think it's fine to do
```suggestion
      files_before = replicator.@file_index_lock.shared { |files, _| files.keys }.sort!
```
to not have to add a getter in `Server` just for specs

src/lavinmq/clustering/server.cr
```diff
@@ -173,7 +202,8 @@ module LavinMQ
 
       def listen(server : TCPServer)
         server.listen
-        @checksums.restore
+        # called before accepting followers, no lock needed
```
carlhoerberg: left over comment?

PR #1754, opened 2026-02-25
Prevent publisher starvation in MQTT throughput perf tool

viktorerlingsson:
## Summary
 - Add `Fiber.yield` every 128k publishes when no rate limit is set in `lavinmqperf mqtt throughput`, ensuring fairness between multiple publishers sharing the same thread
 - Mirrors the approach applied to AMQP throughput in #1712

Note: MQTT consumers and publishers run in different threads, so this doesn't address the consume rate = 0 issue reported by @walro — that appears to be server-side.

### DIFF
--- a/src/lavinmqperf/mqtt/throughput.cr
+++ b/src/lavinmqperf/mqtt/throughput.cr
@@ -208,6 +208,7 @@ module LavinMQPerf
 
         start = Time.instant
         pubs_this_second = 0
+        local_pubs = 0_u64
         packet_id_generator = (1_u16..).each
         wait_until_all_are_connected(connected)
         until @stopped
@@ -233,7 +234,9 @@ module LavinMQPerf
           end
 
           pubs = @pubs.add(1, :relaxed) + 1
+          local_pubs &+= 1
           break if @pmessages > 0 && pubs >= @pmessages
+          Fiber.yield if @rate.zero? && local_pubs % (128*1024) == 0
 
           if !@rate.zero?
             pubs_this_second += 1


walro (approved):

spuun:
Is this a problem with mqtt? Consumers and publishers runs in different threads.

viktorerlingsson:
> Is this a problem with mqtt? Consumers and publishers runs in different threads.

Haven't actually tested myself yet 🙈 but @walro reported this same issue with mqtt. Will do some testing tomorrow!

walro:
Yes, I got a run like this:

```
Summary:
Average publish rate: 257415.1 msgs/s
Average consume rate: 0.0 msgs/s
Total published: 15445717
Total consumed: 0
```

Worth nothing that this was against a single-node target (running from a 2 core host). I will try a build from this PR.

spuun:
I guess having yields too isn't a problem. It will just ensure that if we have multiple publishers each of the will be allowed to publish.

walro:
Yeah, this PR does not seem to solve my problem so I think you are right Jon. Maybe the server-side is the culprit this time around?

```
$ lavinmqperf --version
2.7.0-rc.1-21-g5f773d5c

$ lavinmqperf mqtt throughput -z 60 -x 1 -y 1 -s 16 --uri=mqtt://<user>:<pass>@test-chilly-hound.lmq.cloudamqp.com/obpnzjbs

# ....

Summary:
Average publish rate: 501267.7 msgs/s
Average consume rate: 0.0 msgs/s
Total published: 30077593
Total consumed: 0
```

spuun:
> Yeah, this PR does not seem to solve my problem so I think you are right Jon. Maybe the server-side is the culprit this time around?

So it seems 😞

viktorerlingsson:
>I guess having yields too isn't a problem. It will just ensure that if we have multiple publishers each of the will be allowed to publish.

Yeah, this change should still be good for multi-publisher cases I think

>Yeah, this PR does not seem to solve my problem so I think you are right Jon. Maybe the server-side is the culprit this time around?

Okay!

walro:
Asked Claude to investigate how this is, as always, take this with some grain of salt:

```
  The publisher's read_loop (client.cr:56-68) only yields after receiving 128KB of data (yield_each_received_bytes). With ~29 bytes per PUBLISH packet (16-byte payload + MQTT
  overhead), that's ~4,500 messages between yields. At 500k msgs/s, the TCP receive buffer always has data queued (Nagle batching from the perf tool's tcp_nodelay=false), so
  read_and_handle_packet almost never blocks on I/O — the read_loop just keeps running.

  Why consume is exactly 0 (not just low)

  The deliver_loop (session.cr:30-45) needs to:
  1. Acquire @msg_store_lock (same lock the publish path holds during push)
  2. Read the message from disk (MFile)
  3. Build the MQTT packet
  4. Acquire @lock on the consumer's client and write + flush to its socket

  Each of these steps can block or be preempted. On a 1 vCPU machine with cooperative scheduling, the brief yield windows (~every 9ms) aren't enough for the deliver_loop to complete
   a full delivery cycle including the network flush to the consumer. The publisher's fiber is immediately reschedulable (socket has data), so it gets picked up again before the
  deliver_loop finishes.

  Contrast with AMQP

  The AMQP consumer (src/lavinmq/amqp/consumer.cr) has extensive fairness:
  - Prefetch-based flow control (wait_for_capacity)
  - Channel-level flow control (wait_for_global_capacity)
  - Byte-based yield threshold (yield_each_delivered_bytes)

  The MQTT deliver_loop has none of these — no prefetch, no backpressure, and only a coarse Fiber.yield every 32,768 messages.

  What would help (server-side)

  The most impactful fix would be making the deliver_loop more competitive against the read_loop:
  - Lower the effective yield threshold for MQTT publishers (or yield more aggressively on the publish path when there are pending deliveries)
  - Add byte-aware yielding to the MQTT deliver_loop
  - Consider some form of backpressure/flow control for fast MQTT publishers

  There's nothing further to tweak in lavinmqperf that would fix this — the client already separates publishers and consumers onto different threads. The server needs to schedule
  delivery work more fairly alongside ingest.
```

viktorerlingsson:
>Worth nothing that this was against a single-node target (running from a 2 core host). I will try a build from this PR.

@walro does this only happen when the target is single core?

walro:
> @walro does this only happen when the target is single core?

Probably, didn't happen when I tested against 2 core machines the other day. I can change instance type on the current target and retest (after the current LavinMQ workshop).

walro:
```
# single core target
Average publish rate: 482769.5 msgs/s
Average consume rate: 0.0 msgs/s
Total published: 28967641
Total consumed: 0

# dual core target
Average publish rate: 215885.5 msgs/s
Average consume rate: 213259.2 msgs/s
Total published: 12953709
Total consumed: 12796121
```

viktorerlingsson:
@walro thanks for checking, just wanted to make sure we're not chasing the wrong thing 🙂

viktorerlingsson:
Probably not needed for now. Re-open if we reconsider!

PR #1756, opened 2026-02-26
Chart color added

bengtmagnus:
Adds a chart color. 

On a queue - see the color of "Dedup" in the Rates chart

### DIFF
--- a/static/js/chart.js
+++ b/static/js/chart.js
@@ -12,7 +12,7 @@ Chart.register(Title)
 Chart.register(Filler)
 
 const chartColors = ['#54be7e', '#4589ff', '#d12771', '#d2a106', '#08bdba', '#bae6ff', '#ba4e00',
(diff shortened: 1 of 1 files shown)


static/js/chart.js
```diff
@@ -12,7 +12,7 @@ Chart.register(Title)
 Chart.register(Filler)
 
 const chartColors = ['#54be7e', '#4589ff', '#d12771', '#d2a106', '#08bdba', '#bae6ff', '#ba4e00',
-  '#d4bbff', '#8a3ffc', '#33b1ff', '#007d79']
+  '#d4bbff', '#8a3ffc', '#33b1ff', '#007d79', '#770F1C']
```
walro: The others are in lower-case, so to follow suite?

```suggestion
  '#d4bbff', '#8a3ffc', '#33b1ff', '#007d79', '#770f1c']
```

walro (approved):
didn't test, but makes sense to me! I guessed it "wrapped around" prior to this change

PR #1761, opened 2026-02-27
Prevent theme flash by applying the theme before loading the css

hananasen:

I noticed when trying out the light theme but my OS theme was dark that it first renders the page in darkmode, causing a flash.
Read [this article](https://dev.to/gaisdav/how-to-prevent-theme-flash-in-a-react-instant-dark-mode-switching-o20) on best practices in react applications for avoiding problems like this and asked claude for some suggestions. And keep in mind this is just a suggestion on how this can be solved :) 

Following the article, This PR:
1. Extracts all theme-apply logic into a single global appyTheme function defined in a new shared partial (partials/theme.ecr
2. Adds the shared partial to head 
3. Adds the shared partial to login page

This can be tested visually. Apply a theme on your OS and switch between themes. logout etc.

https://github.com/user-attachments/assets/9d3e265a-9786-45ac-a2d6-13207b12660f

### DIFF
--- a/static/js/layout.js
+++ b/static/js/layout.js
@@ -84,21 +84,8 @@ class ThemeSwitcher {
   }
 
   applyTheme (theme) {
-    const html = document.documentElement
-
-    // Remove existing theme classes
-    html.classList.remove('theme-light', 'theme-dark')
-
-    if (theme === 'light') {
-      html.classList.add('theme-light')
-      html.style.colorScheme = 'light'
-    } else if (theme === 'dark') {
-      html.classList.add('theme-dark')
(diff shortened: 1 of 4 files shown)


hananasen:
oh quite a lot of things failing 🙃 For the lint-fails it looks like you can tell the linter the global functions of the file so it doesn't fail on it not being declared in that file.. but I am also not sure what you think of making applyTheme global like this.. 

I'll convert this to a draft for now

spuun:
I think this will be resolved by #1698. That PR will handle classes like theme and menu state better.

hananasen:
ah nice! I'll close this in favor of that one.

PR #1762, opened 2026-02-27
Dont log stack traces when client present a bad certificate

snichme:

Stop logging stack traces when a client connects over TLS using a bad certificate. Instead logs the message on why the connection was closed. 

Stop logging stack traces when loading TLS context and the config points to a file that doesn't exist. Still logs as error with the message. After logging stops the process.  

Configure TLS and point to a cert file that doesn't exists then start lavinmq.


snichme:
The reason is that the error isn't code related so the stack traces isn't helping in recover the error but it will generate addition logs without any value.

spuun (dismissed):

spuun (commented):
Second thought..

Is it possible to "validate" values and just log error on reload, and not exit? 🤔
Claude is right about exit being ok on startup, but not reload.

kickster97:
@snichme Maybe this should be split in two based on Jons comment

PR #1763, opened 2026-02-28
Non-blocking follower action sending

baelter:
## Summary
- Replace blocking `@actions.send` with `@actions.try_send` in `Follower#send_action`
- When a follower's action channel is full, disconnect it immediately instead of stalling the leader's main fiber
- Adds `FollowerTooSlowError` caught in `Server#each_follower` which logs a warning and closes the slow follower

## Motivation
When the leader is CPU-saturated, slow or unresponsive followers cause `send_action` to block on the bounded channel. This stalls the leader's main fiber, blocking ALL message processing for all clients. Now the leader disconnects slow followers and continues serving traffic.

## Test plan
- [ ] Existing clustering spec updated to verify non-blocking disconnect behavior
- [ ] `make test SPEC=spec/clustering_spec.cr`
- [ ] Manual test: saturate a leader, add a slow follower, verify it gets disconnected without stalling

### DIFF
--- a/spec/clustering_spec.cr
+++ b/spec/clustering_spec.cr
@@ -214,20 +214,18 @@ describe LavinMQ::Clustering::Client, tags: "etcd" do
     end
   end
 
-  it "won't deadlock under high load when a follower disconnects [#926]" do
+  it "disconnects slow follower instead of blocking leader [#926]" do
     LavinMQ::Config.instance.clustering_max_unsynced_actions = 1
     replicator = LavinMQ::Clustering::Server.new(LavinMQ::Config.instance, LavinMQ::Etcd.new("localhost:12379"), 0)
     tcp_server = TCPServer.new("localhost", 0)
     spawn(replicator.listen(tcp_server), name: "repli server spec")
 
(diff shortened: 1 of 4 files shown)


carlhoerberg:
or should we have a more aggressive write timeout for follower connections?

carlhoerberg:
no, this is probably better. a very low write_timeout could cause issues with non loaded followers but with suddenly elevated latency.

baelter:
- Test the scenario in 2.7 where we have better MT.
- Can we block less when @actions is full?

Adding a follower should have lower prio then keeping the leader health.

oskgu360:
> Adding a follower should have lower prio then keeping the leader health.

Just a note on the subject (not on these particular changes), when running HA cluster, isn't highest prio to get all nodes in sync? IMO blocking producers/consumers would be the action to allow for followers to catch up? Otherwise whats the point of HA?

Not saying we should let the leader die, but maybe not allow for more actions until follower catch up? I.e the `min_isr` take, either config or up to a mininmum number of nodes for quorom at least.

PR #1764, opened 2026-02-28
Add Log::RateLimitBackend to suppress repeated log entries

baelter:
## Summary
- Add `Log::RateLimitBackend`, a `Log::Backend` wrapper that suppresses repeated log entries within a configurable window
- Wired up globally in `Config#setup_logger` wrapping the existing broadcast backend
- No changes to business logic — connection factories call `log.info { "Authentication failure..." }` as normal

## How it works
- Keys on `{source, message}` (null-byte delimited)
- First occurrence: passes through immediately
- Subsequent occurrences within the window (default 5s): suppressed and counted
- When window expires and the next entry arrives: emits a summary ("… suppressed N times") then the new entry
- Capped at 10,000 tracked keys; entries beyond that fail-open (pass through unconditionally)
- Periodic cleanup fiber removes stale keys after 60s inactivity

## Motivation
Under connection churn, a client with wrong credentials can generate hundreds of auth failure log lines per second. This wastes CPU on log formatting and I/O. Rather than tracking this in each connection factory, the suppression is handled transparently by the logging layer and applies to any log source.

## Test plan
- [ ] `make test SPEC=spec/log/rate_limit_backend_spec.cr` (9 specs)
- [ ] Manual test: connect with invalid credentials repeatedly, verify log suppression

### DIFF
--- a/spec/log/rate_limit_backend_spec.cr
+++ b/spec/log/rate_limit_backend_spec.cr
@@ -1,0 +1,222 @@
+require "log"
+require "spec"
+require "../../src/lavinmq/log/rate_limit_backend"
+
+# A minimal backend that records every entry written to it.
+private class CapturingBackend < Log::Backend
+  getter entries = Array(Log::Entry).new
+
+  def initialize
+    super(:direct)
+  end
+
+  def write(entry : Log::Entry) : Nil
+    @entries << entry
+  end
+end
+
+private def make_entry(source : String, message : String, severity = Log::Severity::Info) : Log::Entry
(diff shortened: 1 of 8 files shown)


carlhoerberg:
We could take it further, make it more generic, by hashing each log message, in `Log`, compare it to a size limited array/hash/set, and not print repeated log messages, much like syslog does.

baelter:
Made it generic but only configured it for auth logs since it adds a performance cost on every log call.

baelter:
Maybe push this into a log backend/dispatcher instead?

The motivation looks faulty, followers dont start amqp listeners.

PR #1765, opened 2026-02-28
Add connection rate limiting

baelter:
## Summary
- Add token bucket rate limiter that rejects new connections at the TCP level before the AMQP/MQTT handshake
- Two config options: `connection_rate_limit` (global max new connections/sec) and `connection_rate_limit_per_ip` (per source IP), both default to 0 (unlimited)
- Rate limiting runs synchronously in the accept loop before spawning a fiber, so rejected connections never allocate a fiber or enter the handshake
- Rate-limited log messages (max 1/sec) to avoid log spam when rejecting

## Motivation
When LavinMQ is CPU-saturated, handshake timeouts cause clients to disconnect and immediately reconnect, creating a death spiral. Each reconnection goes through the full AMQP handshake (confirm_header, start, authenticate, tune, open), consuming CPU that should serve existing connections. Rejecting at the TCP level is orders of magnitude cheaper.

## Test plan
- [ ] `make test SPEC=spec/connection_rate_limit_spec.cr` (11 specs)
- [ ] Manual test: set `connection_rate_limit = 10`, connect many clients, verify excess connections are rejected immediately
- [ ] Verify existing connection tests still pass

### DIFF
--- a/src/lavinmq/connection_rate_limiter.cr
+++ b/src/lavinmq/connection_rate_limiter.cr
@@ -2,7 +2,8 @@ module LavinMQ
   # Token bucket rate limiter for incoming connections.
   # Fiber-safe (single-threaded Crystal runtime).
   class ConnectionRateLimiter
-    Log = LavinMQ::Log.for "rate_limiter"
+    Log             = LavinMQ::Log.for "rate_limiter"
+    MAX_TRACKED_IPS = 100_000
 
     @global_tokens : Float64
     @global_last_refill : Time::Instant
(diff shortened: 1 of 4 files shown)


src/lavinmq/connection_rate_limiter.cr
```diff
@@ -0,0 +1,119 @@
...
+      else
+        @per_ip_tokens[ip] = tokens
+        false
+      end
+    end
+
+    private def consume_token(
+      limit : Int32,
+      tokens_ptr : Pointer(Float64),
+      last_refill_ptr : Pointer(Time::Instant),
```
carlhoerberg: this is odd, why not modify the instance variables directly?

PR #1768, opened 2026-03-01
Add OAuth2/OIDC SSO login to the management UI

### LINKED ISSUE #1690: Add OAuth2 authorization flow for management UI login
The management UI only supports username/password login. The backend already supports OAuth2/JWT for AMQP connections, but there's no interactive OAuth2 flow for the UI.

- OAuth2 Authorization Code Flow with [PKCE](https://oauth.net/2/pkce/) (/oauth/authorize, /oauth/callback endpoints)
- "Sign in with SSO" button on the login page when oauth_issuer_url is configured

kickster97:
Adds OAuth2/OIDC SSO login to the management UI using the authorization code flow with PKCE. When an OAuth issuer, client ID, and management base URL are configured, a "Sign in with SSO" button appears on the login page.

Clicking it discovers the IdP's endpoints via OIDC well-known configuration, redirects the user to authenticate, then exchanges the authorization code for an access token on callback. The token is validated against the existing JWT authenticator and stored as the `m` cookie. A separate `oauth_user` cookie holds the display name for the signed-in user. The `AuthHandler` is updated to detect JWT-shaped cookies and authenticate them directly, and clears stale tokens on rejection. A server-side `/oauth/logout` endpoint clears both cookies and redirects back to the login page.

Adds `oauth_client_id` and `oauth_mgmt_base_url` config options. `oauth_mgmt_base_url` must use https in production; http is only allowed for localhost, 127.0.0.1, and [::1] for development. Refactors `JWKSFetcher` to expose OIDC discovery so the controller can reuse the cached configuration.

Would like input from @bengtmagnus on frontend solution.

Fixes #1690

start LavinMQ with Oauth & oauth client-id configured and authenticate against your identity provider.
also run specs


kickster97:
TODO: add a config variable that allows a user to disable basic auth for the mgmt interface

static/js/layout.js
```diff
@@ -32,7 +32,11 @@ document.getElementById('userMenuVhost').addEventListener('change', (e) => {
 })
 
 document.getElementById('signoutLink').addEventListener('click', () => {
-  Auth.logout()
+  if (document.cookie.split('; ').some(c => c.startsWith('oauth_user='))) {
+    window.location.assign('/oauth/logout')
+  } else {
```
spuun: Maybe this should be part of `Auth.logout`?

static/js/login.js
```diff
@@ -1,4 +1,39 @@
 import * as Auth from './auth.js'
+
+function showLoginError (message) {
+  const el = document.getElementById('login-error')
+  el.textContent = message
+  el.style.display = 'block'
+}
+
+window.fetch('/oauth/enabled')
```
spuun: Use relative url

static/js/login.js
```diff
@@ -1,4 +1,39 @@
...
+  .then(resp => resp.json())
+  .then(data => {
+    if (data.enabled) {
+      const ssoBtn = document.getElementById('sso-btn')
+      ssoBtn.style.setProperty('display', 'flex', 'important')
+      document.getElementById('sso-separator').style.display = ''
+      ssoBtn.addEventListener('click', () => {
+        window.fetch('/oauth/authorize')
```
spuun: Use relative url

static/js/layout.js
```diff
@@ -32,7 +32,11 @@ document.getElementById('userMenuVhost').addEventListener('change', (e) => {
 })
 
 document.getElementById('signoutLink').addEventListener('click', () => {
-  Auth.logout()
+  if (document.cookie.split('; ').some(c => c.startsWith('oauth_user='))) {
+    window.location.assign('/oauth/logout')
```
spuun: Use relative url

src/lavinmq/auth/chain.cr
```diff
@@ -9,22 +9,23 @@ module LavinMQ
   module Auth
     class Chain < Authenticator
       Log = LavinMQ::Log.for "auth.chain"
-      @backends : Array(Authenticator)
+      property backends : Array(Authenticator)
+      property oauth_fetcher : Auth::JWT::JWKSFetcher?
```
spuun: How come this is part of `Chain`? Should only be needed by the OAuthAuthenticator, no?

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,167 @@
...
+require "../../auth/jwt/jwks_fetcher"
+
+module LavinMQ
+  module HTTP
+    class OAuthController
+      include Router
+
+      Log = LavinMQ::Log.for "http.oauth"
+
+      def initialize(@authenticator : Auth::Authenticator)
```
spuun: I'd say this should be typed as `OAuthAuthenticator` since it's what it needs.

src/lavinmq/http/http_server.cr
```diff
@@ -22,6 +22,7 @@ module LavinMQ
           StrictTransportSecurity.new,
           StaticController.new,
           AuthHandler.new(@amqp_server.authenticator, @amqp_server.users.direct_user),
+          OAuthController.new(@amqp_server.authenticator),
```
spuun: Have you concider putting `OAuthController` before `AuthHandler` and let it do the oauth authentication? Then we wouldn't need to have oauth stuff in two controllers.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,167 @@
...
+        token_response = OAuth2::TokenExchange.new(token_ep, client_id)
+          .exchange(code, build_redirect_uri, code_verifier)
+
+        auth_context = Auth::Context.new("", token_response.access_token.to_slice, context.request.remote_address)
+        user = @authenticator.authenticate(auth_context)
+        oauth_redirect_error(context, "Token validation failed") if user.nil? || user.tags.empty?
+
+        max_age = token_response.expires_in.try { |e| Math.min(e, 8.hours.total_seconds.to_i64).seconds } || 8.hours
+
+        set_cookie(context, "m", token_response.access_token,
```
spuun: Are we using the `m` cookie because this is something rabbitmq does?

static/js/auth.js
```diff
@@ -1,8 +1,14 @@
...
+  const m = getCookie('m')
+  if (!m || m.startsWith('eyJ')) return
   return window.atob(getAuth()).split(':')[0]
 }
 
 function getPassword () {
+  const m = getCookie('m')
+  if (!m || m.startsWith('eyJ')) return null
   return window.atob(getAuth()).split(':')[1]
 }
```
spuun, line 17: `getPassword` is only used for "move message" functionality, but dunno if that's even needed today.

Right now that functionality won't work with oauth.

src/lavinmq/http/handler/auth_handler.cr
spuun: Maybe this should be changed to create a `Auth::Context` which is then used with `@authenticator.authenticate(auth_context)`. Now we have two paths that both are calling `@authenticator.authenticate`.

src/lavinmq/http/controller/oauth.cr
kickster97: not sure, i just continued with what we already had. 
should we consider something else?

spuun: The `m` cookie is just there to be compatible with rabbitmq. Using a different name for the oauth stuff would make you get rid of the `eyJ` checks, for example.

src/lavinmq/http/http_server.cr
```diff
@@ -18,12 +18,18 @@ module LavinMQ
       Log = LavinMQ::Log.for "http.server"
 
       def initialize(@amqp_server : LavinMQ::Server)
+        oauth_authenticator = @amqp_server.authenticator.as?(Auth::Chain)
+          .try(&.backends.find { |b| b.is_a?(Auth::OAuthAuthenticator) })
+          .as?(Auth::OAuthAuthenticator)
```
spuun: To get rid of some casting you can do
```suggestion
          .try(&.backends.select(Auth::OAuthAuthenticator).first)
```

src/lavinmq/http/controller/oauth.cr
kickster97: Yes, smart! thats better

static/js/queue.js
```diff
@@ -278,7 +278,11 @@ document.querySelector('#getMessages').addEventListener('submit', function (evt)
     })
 })
 
-document.querySelector('#moveMessages').addEventListener('submit', function (evt) {
+const moveMessagesForm = document.querySelector('#moveMessages')
+if (Auth.getPassword() === null) {
+  moveMessagesForm.classList.add('hide')
+}
```
spuun, line 284: Instead of hiding it we could probably refactor the move message functionality, but maybe that's for another PR. 

We can probably just remove credentials from the uri's and it will use the direct user if the current user has permissions to src and dst.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,190 @@
...
+
+module LavinMQ
+  module HTTP
+    class OAuthController
+      include Router
+
+      Log = LavinMQ::Log.for "http.oauth"
+
+      def initialize(@oauth_authenticator : Auth::OAuthAuthenticator?)
+        register_routes
```
spuun, line 16: What about not registering the routes if `@oauth_authenticator` is nil? (as in oauth isn't enabled)

The `/oauth/enabled` can just be a `204` as in "yes, enabled" and when disabled it will be a 404 since the route doesn't exist.

Or maybe not even add the OAuthController if oauth is disabled? Then `@oauth_authenticator` won't have to be nilable and you get rid of some nil checks.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,190 @@
...
+      private def oauth_error(context, status_code, message) : NoReturn
+        context.response.status_code = status_code
+        context.response.content_type = "application/json"
+        {reason: message}.to_json(context.response)
+        raise OAuthError.new
+      end
+
+      private def oauth_redirect_error(context, message) : NoReturn
+        context.response.status = ::HTTP::Status::FOUND
+        context.response.headers["Location"] = "/login?error=#{URI.encode_path_segment(message)}"
```
spuun: Maybe `Config.instance.oauth_mgmt_base_url` should be used here too?

spuun (commented):
What about using a cookie `oauth_token` instead of `m`, then use `m` instead of `oauth_user`? Then you'd put the username together with a dummy password in `m` and "everything" in the frontend will work the same as with other authentication methods (except for move messages, but that we can fix in another PR).

src/lavinmq/http/controller/oauth.cr
kickster97: When this error runs, the browser is already on the LavinMQ site, so sending it to `/login` should be enough to reach the login page. The callback URL `redirect_uri` is different because the identity provider needs the full address to send the user back, so that one has to be absolute. 
 
also, `oauth_mgmt_base_url` raises if it isn't configured, and since this code is the error handler itself, depending on it could make the error handler crash while trying to report an error.

kickster97:
> What about using a cookie `oauth_token` instead of `m`, then use `m` instead of `oauth_user`? Then you'd put the username together with a dummy password in `m` and "everything" in the frontend will work the same as with other authentication methods (except for move messages, but that we can fix in another PR).

the `eyJ` checks are already gone via the `oauth_token` rename, i did that in a local commit before you commented this :) .. so dropping `oauth_user` now would mean a placeholder password in m and temporarily breaking move-messages. Do you think the simplification is worth it, or are you ok keeping `oauth_user` as a small username and session marker cookie?

src/lavinmq/http/controller/oauth.cr
spuun: Ah, then I guess it should be relative!

src/lavinmq/http/http_server.cr
```diff
@@ -45,6 +47,10 @@ module LavinMQ
           NodesController.new(@amqp_server),
           LogsController.new(@amqp_server),
         ] of ::HTTP::Handler
+        if oauth_authenticator
+          auth_index = handlers.index! { |h| h.is_a?(AuthHandler) }
+          handlers.insert(auth_index, OAuthController.new(oauth_authenticator))
+        end
         handlers.unshift(::HTTP::LogHandler.new(log: Log)) if Log.level == ::Log::Severity::Debug
```
spuun: Instead of this we can (probably) do
```crystal
handlers = [
  ::HTTP::LogHandler.new(log: Log)) if Log.level == ::Log::Severity::Debug,
  …
  StaticController.new,
  oauth_authenticator && OAuthController.new(oauth_authenticator),
  AuthHandler.new(@amqp_server.authenticator, @amqp_server.users.direct_user),
  …
].select(::HTTP::Handler) # remove nil values and types the array to Array(::HTTP::Handler)
```

Then it's easy to follow see the actual order of the handlers.

static/js/layout.js
```diff
@@ -32,7 +32,11 @@ document.getElementById('userMenuVhost').addEventListener('change', (e) => {
 })
 
 document.getElementById('signoutLink').addEventListener('click', () => {
-  Auth.logout()
+  if (document.cookie.split('; ').some(c => c.startsWith('oauth_user='))) {
+    window.location.assign('oauth/logout')
+  } else {
+    Auth.logout()
+  }
```
spuun: I think
```suggestion
  Auth.logout()
```
is enough?

`logout` will clear all cookies and redirect to login, which is also what `oauth/logout` does.

However, you had the `m` cookie (now `oauth_token`) as a http only but changed that in 0e7863fc6b09fb68838164522580941e4838b149. Maybe it makes sense to let `oauth_token` be http only, and then a redirect to `oauth/logout` would be needed

Maybe logout in in `auth.js` could be
```javascript
function logout () {
  window.localStorage.removeItem(whoAmICacheKey)
  stateClasses.remove(/^user-is-/)
  document.cookie = 'm=; max-age=0'
  if (getCookie('oauth_user')) {
    window.location.assign('oauth/logout')
  } else { 
    window.location.assign('login')
  }
}
```

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,187 @@
...
+
+        user = authenticate_token(token_response.access_token, context.request.remote_address)
+        oauth_redirect_error(context, "Token validation failed") if user.nil?
+
+        max_age = token_response.expires_in.try { |e| Math.min(e, 8.hours.total_seconds.to_i64).seconds } || 8.hours
+
+        set_cookie(context, "oauth_token", token_response.access_token,
+          path: "/", http_only: true, samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
+        set_cookie(context, "oauth_user", URI.encode_path_segment(user.name),
+          path: "/", samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
```
spuun: What if we do something like this instead
```suggestion
        value = Base64.strict_encode("#{user.name}:DUMMY") # DUMMY password
        set_cookie(context, "m", "|:#{URI.encode_path_segment(value)}",
          path: "/", samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
```
Then we can remove the changes to `getUsername` in `auth.js`. Dunno if it will screw with something, but things worked just fine when the content of `oauth_token` was stored in `m`, so should be fine.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,187 @@
...
+
+        context.response.status = ::HTTP::Status::FOUND
+        context.response.headers["Location"] = "/"
+        context
+      rescue OAuthError
+        context
+      rescue ex
+        Log.error(exception: ex) { "OAuth callback failed (#{ex.class})" }
+        context.response.status = ::HTTP::Status::FOUND
+        context.response.headers["Location"] = "/login?error=#{URI.encode_path_segment("OAuth authentication failed")}"
```
spuun: Should be relative.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,187 @@
...
+        max_age = token_response.expires_in.try { |e| Math.min(e, 8.hours.total_seconds.to_i64).seconds } || 8.hours
+
+        set_cookie(context, "oauth_token", token_response.access_token,
+          path: "/", http_only: true, samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
+        set_cookie(context, "oauth_user", URI.encode_path_segment(user.name),
+          path: "/", samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
+        expire_cookie(context, "oauth_state", "/oauth")
+
+        context.response.status = ::HTTP::Status::FOUND
+        context.response.headers["Location"] = "/"
```
spuun: Must be relative

spuun (commented):
I realized the cookies paths must be handled. You'll have to use the request path and strip away the "current mapped path" to get the "base path" of the mgmt ui. E.g. when setting `oauth_token` you should use the request path minus `/oauth/callback`.

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,203 @@
...
+
+        user = authenticate_token(token_response.access_token, context.request.remote_address)
+        oauth_redirect_error(context, "Token validation failed") if user.nil?
+
+        max_age = token_response.expires_in.try { |e| Math.min(e, 8.hours.total_seconds.to_i64).seconds } || 8.hours
+
+        set_cookie(context, "oauth_token", token_response.access_token,
+          path: "/", http_only: true, samesite: ::HTTP::Cookie::SameSite::Strict, max_age: max_age)
+        value = Base64.strict_encode("#{user.name}:")
+        set_cookie(context, "m", "|oauth:#{URI.encode_path_segment(value)}",
```
spuun, line 115: Nice touch with the `oauth` prefix 👌

src/lavinmq/http/http_server.cr
```diff
@@ -18,11 +18,15 @@ module LavinMQ
       Log = LavinMQ::Log.for "http.server"
 
       def initialize(@amqp_server : LavinMQ::Server)
+        oauth_authenticator = @amqp_server.authenticator.as?(Auth::Chain)
+          .try(&.backends.select(Auth::OAuthAuthenticator).first?)
```
spuun: I guess the `@amqp_server.authenticator` could be a `OAuthAuthenticator`, so maybe
```suggestion
        oauth_authenticator = case authenticator = @amqp_server.authenticator
                                when Auth::Chain
                                  authenticator.backends.select(Auth::OAuthAuthenticator).first?
                                when Auth::OAuthAuthenticator
                                  authenticator
                              end
```
... but we do know that it will be a `Chain` for now, so maybe no action is needed.

src/lavinmq/http/handler/auth_handler.cr
```diff
@@ -10,11 +10,11 @@ module LavinMQ
       end
 
       def call(context)
+        return call_next(context) if context.user
```
spuun: I think it may be good to add a comment that a previous handler already may have authenticated the request.

src/lavinmq/http/handler/auth_handler.cr
```diff
@@ -10,11 +10,11 @@ module LavinMQ
...
+
         if internal_unix_socket?(context)
           context.user = @direct_user
-        end
-
-        if auth = cookie_auth(context) || basic_auth(context)
+        elsif auth = cookie_auth(context) || basic_auth(context)
```
spuun: Maybe we should consider what claude points out about this. I think it's reasonable that given credentials will override the direct user.

src/lavinmq/config/options.cr
```diff
@@ -382,6 +382,10 @@ module LavinMQ
       property oauth_audience : String? = nil
       @[IniOpt(section: "oauth", ini_name: jwks_cache_ttl)]
       property oauth_jwks_cache_ttl : Time::Span = 1.hours
+      @[IniOpt(section: "oauth", ini_name: client_id)]
+      property oauth_client_id : String? = nil
+      @[IniOpt(section: "oauth", ini_name: mgmt_base_url)]
+      property oauth_mgmt_base_url : String? = nil
```
spuun:
```suggestion
      property oauth_mgmt_base_url : URI? = nil
```

spuun (commented):
We need to make sure that the cookie path is correct. I think you just have to use the path of `oauth_mgmt_base_url` for that.

Also, to me the redirects doesn't really work when i login. I end up on the login page. I think this is because of SameSite being set to strict for the auth cookies.

src/lavinmq/http/http_server.cr
kickster97: No we should be explicit i think... good suggestion

spuun (approved):
:+1:

src/lavinmq/http/controller/oauth.cr
```diff
@@ -0,0 +1,210 @@
...
+        expected_state = cookie_value[0...sep]
+        provided_state = context.request.query_params["state"]? || oauth_redirect_error(context, "State mismatch")
+        oauth_redirect_error(context, "State mismatch") unless Crypto::Subtle.constant_time_compare(expected_state, provided_state)
+        code = context.request.query_params["code"]? || oauth_redirect_error(context, "Missing authorization code")
+
+        {code, cookie_value[sep + 1..], client_id}
+      end
+
+      private def set_cookie(context, name, value, path = "/", http_only = false,
+                             samesite = ::HTTP::Cookie::SameSite::Strict,
```
viktorerlingsson: Nit: we only call `set_cookie` with `samesite: ::HTTP::Cookie::SameSite::Lax`, should that be the default instead of `strict`?

viktorerlingsson (approved):

src/lavinmq/http/controller/oauth.cr
kickster97: yes, youre right!

spuun:
@kickster97 I actually found an issue now: try add a vhost as an oauth user with administrator rights.

kickster97:
> @kickster97 I actually found an issue now: try add a vhost as an oauth user with administrator rights.

ah, we use userstore for that and oauth users are not in that store... looking at a fix

PR #1772, opened 2026-03-02
Fix: reset pos when crossing segment boundary in find_offset_in_segments

### LINKED ISSUE #1771: `find_offset_in_segments` missing pos reset when crossing segment boundaries
**Describe the bug**

PR #1479 removed the `pos = 4_u32` reset when `find_offset_in_segments` crosses a segment boundary. When the method walks past the end of one segment and advances to the next, pos retains the old segment's file size instead of being reset to 4 (past the 4-byte schema version header). The subsequent `BytesMessage.from_bytes(rfile.to_slice + pos)` then reads from a wrong byte position in the new segment.

When the new segment is smaller than the old pos value, this causes an `IndexError` from slice bounds checking, which closes the client connection with INTERNAL_ERROR. When the new segment is larger, the read lands on garbage bytes inside the segment which are passed to `ShortString.from_bytes` → `String.new(pointer, garbage_length)` with no bounds check, which can `SEGFAULT`.

**How to reproduce**
1. Create a stream queue
2. Publish 7 x 1MB messages (fills the 8MB segment without spilling)
3. Wait 3 seconds to create a timestamp gap
4. Publish 3 more x 1MB messages (the 8th message triggers `open_new_segment`, giving segment 2 a `first_ts` after the gap)
5. Consume with x-stream-offset set to a Time that falls in the 3-second gap

The consumer subscription fails with INTERNAL_ERROR (541).

 ```
 Server log:
  ERROR lmq.amqp.client Unexpected error, while reading:
    path=.../msgs.0000000002 pos=4 size=3145846
    (LavinMQ::MessageStore::Error)
      from stream_message_store.cr:122 in 'find_offset_in_segments'
      from stream_message_store.cr:58 in 'find_offset'
      from stream_consumer.cr:24 in 'initialize'
  Caused by: Index out of bounds (IndexError)
      from slice.cr:1251 in 'check_size'
      from slice.cr:169 in '+'
      from stream_message_store.cr:113 in 'find_offset_in_segments'
```

**Expected behavior**

When crossing a segment boundary, pos should be reset to `4_u32` so the method reads from the start of the new segment's message data (after the schema version header). The consumer should receive the first message in segment 2 whose timestamp matches.

### PR DESCRIPTION
**Pull request by kickster97**

Fixes #1771

Restores the `pos = 4u32` reset in `find_offset_in_segments` when advancing to the next segment.

In #1479, the original `pos = 4_u32` was changed to `rfile.pos = 4u32`. During [review](https://github.com/cloudamqp/lavinmq/pull/1479#discussion_r2530350896) it was mentioned that the method uses a local `pos` variable, not `rfile.pos`. When resolving, the line was removed entirely instead of being changed back to `pos = 4u32`.

Without the reset, when a timestamp-based `x-stream-offset` falls between two segments, the method walks all messages in the first segment, crosses to the next segment, and reads from the old segment's end position, causing an `IndexError` or potentially a SEGFAULT.

  Run added spec

### DIFF
--- a/spec/stream_queue_spec.cr
+++ b/spec/stream_queue_spec.cr
@@ -230,9 +230,8 @@ describe LavinMQ::AMQP::Stream do
         # Fill segment 1 — two half-segment messages won't fit, so second triggers new segment
         q.publish_confirm data
         # Sleep to create a timestamp gap
-        sleep 0.5.seconds
+        sleep 1.seconds
         target_time = Time.utc
-        sleep 0.5.seconds
         # This publish can't fit in current segment, creates a new one with first_ts > target
         q.publish_confirm data
         # Consume from timestamp in the gap — find_offset_in_segments must cross segment boundary

### CODE REVIEW

File: spec/stream_queue_spec.cr [hunk 1]
```diff
@@ -221,6 +221,33 @@ describe LavinMQ::AMQP::Stream do
...
+      with_channel(s) do |ch|
+        q = ch.queue("stream-ts-across-segments", args: stream_queue_args)
+        # Use half-segment messages so exactly 1 fits per segment
+        data = Bytes.new(LavinMQ::Config.instance.segment_size // 2)
+        # Fill segment 1 — two half-segment messages won't fit, so second triggers new segment
+        q.publish_confirm data
+        # Sleep to create a timestamp gap
+        sleep 0.5.seconds
+        target_time = Time.utc
+        sleep 0.5.seconds
```
viktorerlingsson:
```suggestion
        sleep 1.seconds
        target_time = Time.utc
```

Hmm, AMQP field table timestamps (the `x-stream-offset` header) are only precise to the second, while messages are stored with ms-precision. So with 0.5 seconds of sleep between the first message and `target_time = Time.utc` there's a chance that target_time is truncated to the same second as message 1, making it <= message 1's ms-precision timestamp, causing message 1 to be returned. 
1s sleep guarantees that they fall in different seconds. And that means that one sleep should be enough, message 2 published  in the same second as target_time still works because the truncated offset is always <= any ms-precision timestamp within that second.

PR #1774, opened 2026-03-02
add x-stream-offset header to StreamReader responses

### LINKED ISSUE #1773: HTTP API missing x-stream-offset header
### Describe the bug

The HTTP API stream endpoint (`POST /api/queues/:vhost/:name/stream`) no longer returns `x-stream-offset` in message headers.

  ### To Reproduce

  1. Create a stream queue and publish messages
  2. Fetch messages via `POST /api/queues/%2f/:name/stream` with `{"count": 1, "offset": "first"}`
  3. Check `properties.headers` in the response

  ### Expected behavior

  Messages should include `x-stream-offset` in their headers, as they did before #1479.

  ### Root cause

#1479 removed offsets from disk storage. `shift?()` was updated to add the header dynamically via `add_offset_header`, but `read()` (used by `StreamReader`) was not.

kickster97:
After stream offsets were removed from disk storage(#1479), the `x-stream-offset` header is added dynamically at read time. The `shift?()` path handles this correctly, but the `read()` path (used by `StreamReader` for the `/api/queues/:vhost/:name/stream` endpoint) was missed.

Fixes #1773 

This PR adds an offset parameter to `StreamMessageStore#read()` and updates `StreamReader#each` to track and pass the offset through, so the header is added consistently across both code paths.

Run the added apec

### DIFF
--- a/src/lavinmq/amqp/stream/stream_message_store.cr
+++ b/src/lavinmq/amqp/stream/stream_message_store.cr
@@ -212,14 +212,13 @@ module LavinMQ::AMQP
       end
     end
 
-    def read(segment : UInt32, position : UInt32, offset : Int64) : Envelope?
+    def read(segment : UInt32, position : UInt32) : Envelope?
       return if @closed
       rfile = @segments[segment]
       return if position == rfile.size
       begin
         msg = BytesMessage.from_bytes(rfile.to_slice + position)
         sp = SegmentPosition.new(segment, position, msg.bytesize.to_u32)
-        msg.properties.headers = add_offset_header(msg.properties.headers, offset)
         Envelope.new(sp, msg, redelivered: false)
       rescue ex
         puts "read segment=#{segment} position=#{position}"
--- a/src/lavinmq/amqp/stream/stream_reader.cr
+++ b/src/lavinmq/amqp/stream/stream_reader.cr
@@ -14,8 +14,13 @@ module LavinMQ::AMQP
       offset, segment, position = store.find_offset(@start_offset)
       loop do
         break if store.closed
-        env = store.read(segment, position, offset)
+        env = store.read(segment, position)
         if env
+          if headers = env.message.properties.headers
+            headers["x-stream-offset"] = offset
+          else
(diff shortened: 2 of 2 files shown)


src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -212,13 +212,16 @@ module LavinMQ::AMQP
       end
     end
 
-    def read(segment : UInt32, position : UInt32) : Envelope?
+    def read(segment : UInt32, position : UInt32, offset : Int64? = nil) : Envelope?
```
viktorerlingsson: read is only called from `each(&)`, right? so then we could just make offset non-nilable and not need the `if offset` check

src/lavinmq/amqp/stream/stream_message_store.cr
```diff
@@ -212,13 +212,14 @@ module LavinMQ::AMQP
       end
     end
 
-    def read(segment : UInt32, position : UInt32) : Envelope?
+    def read(segment : UInt32, position : UInt32, offset : Int64) : Envelope?
```
snichme: I dont think `offset` is a good argument to this method. 

The idea for this method "Give me the message/envelope in segment X and position Y" `offset` is added just to set an additional flag on the message which can be done outside of the method.  So either make the `add_offset_header` public or just copy to same code to the StreamReader

src/lavinmq/amqp/stream/stream_reader.cr
```diff
@@ -11,12 +11,13 @@ module LavinMQ::AMQP
     def each(&)
       stream = @stream
       store = stream.stream_msg_store
-      _, segment, position = store.find_offset(@start_offset)
+      offset, segment, position = store.find_offset(@start_offset)
       loop do
         break if store.closed
-        env = store.read(segment, position)
+        env = store.read(segment, position, offset)
         if env
```
snichme:
```suggestion
        env = store.read(segment, position)
        if env
          if headers = env.message.properties.headers
            headers["x-stream-offset"] = offset
          else
            env.message.properties.headers = AMQP::Table.new({"x-stream-offset": offset})
          end
```

something like this

PR #1779, opened 2026-03-02
[oauth] Default to "sub" and "client_id" when preferred_username_claims is not configured

hananasen:

Unlike RabbitMQs oauth2.0 implementation, LavinMQ does not have a built-in default for `preferred_username_claims`. Without it set, LavinMQ won't know which JWT claim to use as the username. 
`DEBUG lmq.oauth2 authentication failed for user "": Could not decode token - No username found in JWT claims (tried: )`

Hence, it needs to be set in the config. I think it would be a good idea to fall back to 'sub', 'client_id' in those cases, as that is the behavior for RabbitMQ .

Specs.

### DIFF
--- a/src/lavinmq/auth/jwt/token_claim.cr
+++ b/src/lavinmq/auth/jwt/token_claim.cr
@@ -22,7 +22,6 @@ module LavinMQ
 
         private def extract_username(payload : JWT::Payload) : String
           claims = @config.oauth_preferred_username_claims
-          claims = {"sub", "client_id"} if claims.empty?
           claims.each do |claim|
             # Check standard JWT properties first
             username = case claim
--- a/spec/oauth_config_spec.cr
+++ b/spec/oauth_config_spec.cr
@@ -12,9 +12,9 @@ describe LavinMQ::Config do
       config.oauth_jwks_cache_ttl.should eq(1.hours)
     end
 
-    it "sets default oauth_preferred_username_claims to empty array" do
+    it "sets default oauth_preferred_username_claims to sub and client_id" do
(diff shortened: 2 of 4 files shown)


src/lavinmq/auth/jwt/token_claim.cr
```diff
@@ -22,6 +22,7 @@ module LavinMQ
 
         private def extract_username(payload : JWT::Payload) : String
           claims = @config.oauth_preferred_username_claims
+          claims = ["sub"] if claims.empty?
```
spuun: You can probably make it a `Tuple` instead of an `Array` to get rid of an allocation. 
```suggestion
          claims = {"sub"} if claims.empty?
```

hananasen: ah yes, good find, cause now it would allocate a new object on the heap on every request?

hananasen: with tuple then we can easily add `client_id` as well, thats nice. but could that result for some undefined methods if we try to use some array method on that later? or is the crystal compiler smart enough for that 😄

hananasen:
read a bit more and since sub and client_id are required fields in the jwt data structure in oauth https://datatracker.ietf.org/doc/rfc9068/ it feels like it makes sense to fallback on them. I have no strong opinions here, but just good to know whether we should make it a required field on cloudamqp or not to help customers that configure it.

src/lavinmq/auth/jwt/token_claim.cr
spuun: Compiler will handle it!

PR #1783, opened 2026-03-03
Honor the `yield_each_delivered_bytes` setting for MQTT

walro:

This is the "same" as we do it for AMQP.

The default `yield_each_delivered_bytes`of 1 MB roughly means that we will yield twice as often with a 16 bytes payload:

```
For a 16-byte payload MQTT message, the per-message sp.bytesize is roughly:

32 bytes (fixed overhead) + topic_length + 1 + 16 bytes (payload) ≈ 49 + topic_length

So with the old count-based yield every 32,768 messages:

┌────────────────────┬──────────────────────┬───────────────────────────────────────┐
│    Topic length    │ Per-message bytesize │ Equivalent yield_each_delivered_bytes │
├────────────────────┼──────────────────────┼───────────────────────────────────────┤
│ short (~8 bytes)   │ ~58 bytes            │ ~1.9 MB                               │
├────────────────────┼──────────────────────┼───────────────────────────────────────┤
│ medium (~16 bytes) │ ~66 bytes            │ ~2.2 MB                               │
└────────────────────┴──────────────────────┴───────────────────────────────────────┘
```

Benchmarks (`bin/lavinmqperf mqtt throughput -z 30 -s <size>`): 

```
## this commit

# 16 bytes
Average publish rate: 709338.3 msgs/s
Average consume rate: 708276.4 msgs/s

# 256 bytes
Average publish rate: 595813.9 msgs/s
Average consume rate: 595835.3 msgs/s

# 65536 bytes
Average publish rate: 22164.2 msgs/s
Average consume rate: 22228.1 msgs/s

## main

# 16 bytes
Average publish rate: 717025.2 msgs/s
Average consume rate: 715138.9 msgs/s

# 256 bytes
Average publish rate: 587857.8 msgs/s
Average consume rate: 587646.0 msgs/s

# 65536 bytes
Average publish rate: 22045.0 msgs/s
Average consume rate: 22040.7 msgs/s
```

Looks like the difference is basically just noise.

Tests + for instance benchmarking: `bin/lavinmqperf mqtt throughput -z 30 -s 16`

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -28,16 +28,20 @@ module LavinMQ
       end
 
       private def deliver_loop
-        i = 0
+        delivered_bytes = 0_i32
         loop do
           break if @closed
           next @msg_store.empty.when_false.receive? if @msg_store.empty?
           next @consumers_empty.when_false.receive? if @consumers.empty?
           consumer = @consumers.first.as(MQTT::Consumer)
-          get_packet do |pub_packet|
+          get_packet do |pub_packet, bytesize|
             consumer.deliver(pub_packet)
+            delivered_bytes &+= bytesize
           end
-          Fiber.yield if (i &+= 1) % 32768 == 0
+          if delivered_bytes > Config.instance.yield_each_delivered_bytes
+            delivered_bytes = 0
+            Fiber.yield
+          end
         rescue ex
           @log.error(exception: ex) { "Failed to deliver message in deliver_loop" }
           @consumers.each &.close
@@ -107,7 +111,7 @@ module LavinMQ
         @vhost.unbind_queue(@name, EXCHANGE, rk, arguments || AMQP::Table.new)
       end
 
-      private def get_packet(& : MQTT::Publish -> Nil) : Bool
+      private def get_packet(& : MQTT::Publish, UInt32 -> Nil) : Bool
         raise ClosedError.new if @closed
         loop do
           env = @msg_store_lock.synchronize { @msg_store.shift? } || break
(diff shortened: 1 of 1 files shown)


kickster97 (approved):
nice find!

src/lavinmq/mqtt/session.cr
```diff
@@ -28,16 +28,20 @@ module LavinMQ
...
+        delivered_bytes = 0_i32
         loop do
           break if @closed
           next @msg_store.empty.when_false.receive? if @msg_store.empty?
           next @consumers_empty.when_false.receive? if @consumers.empty?
           consumer = @consumers.first.as(MQTT::Consumer)
-          get_packet do |pub_packet|
+          get_packet do |pub_packet, bytesize|
             consumer.deliver(pub_packet)
+            delivered_bytes &+= bytesize
```
spuun, line 39: I think this should work
```suggestion
          get_packet do |pub_packet|
            consumer.deliver(pub_packet)
            delivered_bytes &+= pub_packet.bytesize
```

spuun (commented):
`MQTT::Protocol::Packet` has a `#bytesize` which I hope is correct :)

src/lavinmq/mqtt/session.cr
walro: Yes, that works but then it would be a bit different to how we do it for AMQP, essentially it would be "on wire size" vs "message store size" if I get it right. I guess for AMQP they are closer to each other(?). No strong opinion here though and it's all the same to you I think I prefer to keep what we have 😇 

It's "just" yielding anyway.

PR #1790, opened 2026-03-04
Fix 1787/overlapping sources in logs

### LINKED ISSUE #1787: Log text overlapping
**Describe the bug**
Log text can overlap if it's too long, like "lmq.amqp.connection_factory"

Image

ThomasSarlin:
Close #1787 

The source text overlaps the message if the text is too long, to prevent this we'll give the source a bit more space and if it were to still extend its bounds we limit it with overlap and ellipsis. The full text will still be searchable with the browser find method and it will look more neat. To make things more clear we'll add a title to the source to make the full text appear on hover.

:eyes:

image

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -1867,12 +1867,16 @@ pre.arguments .inactive-argument:before {
   font-weight: 600;
 }
 
+#livelog table td {
+  text-overflow: ellipsis;
+  max-width: 200px;
+  overflow:hidden;
+}
+
 #livelog table td,
 #livelog table th {
   padding: 4px 8px;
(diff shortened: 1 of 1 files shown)


ThomasSarlin:
> Without these changes, the ellipsis will never appear and the title tooltip (which is correct) is the only functional part of this PR.

I changed so this change only affects the `td`, not the `th`. However the ellipsis effect will take place as there are no whitespace in the source field. If there whitespace we would not have this issue as a line break would be applied.

snichme (approved):

PR #1791, opened 2026-03-04
[oauth] enable parsing permissions from multiple additional claims

hananasen:
### Context

Supported in rmq since 4.1. Figured it wasn't a huge change to add it here. But crystal team can decide if it makes sense to add it. 
I would however argue that it is a bit confusing that the config setting is called additional_scopes_key (singular) when we support a list. But no strong opinions.

In rmq this config setting it is a string separated by blank spaces. Since other config settings are separated by commas in lavinmq I chose to do the same here to keep things consequent in the ini-config.

Changes additional_scopes_key from string to array and traverses the token with the claims found in the config.

Specs.
Tested manually by adding hardcoded claims on my keycloak client that issues the jwt.
lmq config
```
[main]
auth_backends = oauth
log_level = debug

[oauth]
issuer = http://localhost:8080/realms/master
additional_scopes_key = additional_scope1,additional_scope2
preferred_username_claims = sub
verify_aud = false
```
`bin/lavinmq --data-dir ./tmp/data -c tmp/lavinmq.ini`

### DIFF
--- a/spec/oauth_config_spec.cr
+++ b/spec/oauth_config_spec.cr
@@ -17,9 +17,9 @@ describe LavinMQ::Config do
       config.oauth_preferred_username_claims.should eq(["sub", "client_id"])
     end
 
-    it "sets default oauth_additional_scopes_key to empty array" do
+    it "sets default oauth_additional_scopes_keys to empty array" do
       config = LavinMQ::Config.new
-      config.oauth_additional_scopes_key.should be_empty
+      config.oauth_additional_scopes_keys.should be_empty
     end
 
     it "sets default oauth_scope_prefix to nil" do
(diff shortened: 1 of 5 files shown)


kickster97:
This is pre-existing behavior, but if we support multiple claim sources it's more likely to be hit: if two claims both define e.g. `read:myvhost/queue1` and `read:myvhost/queue2`, the second one overwrites the first. Should we consider adding so that we combine the regex patterns instead of replacing?

hananasen:
> This is pre-existing behavior, but if we support multiple claim sources it's more likely to be hit: if two claims both define e.g. `read:myvhost/queue1` and `read:myvhost/queue2`, the second one overwrites the first. Should we consider adding so that we combine the regex patterns instead of replacing?

that's a really good question. I researched a bit in the rabbitmq repo. and they seem to keep both as it looks like in later versions (haven't tested myself yet). No strong opinions here, but if it isn't a large effort to fix or have other side effects it would be nice to be able to combine the patterns instead of replacing them(I could look into that as well). Otherwise it would be great to clarify it in the documentation in some way.

kickster97:
> that's a really good question. I researched a bit in the rabbitmq repo. and they seem to keep both as it looks like in later versions (haven't tested myself yet). No strong opinions here, but if it isn't a large effort to fix or have other side effects it would be nice to be able to combine the patterns instead of replacing them(I could look into that as well). Otherwise it would be great to clarify it in the documentation in some way.

I think we should try and go for combining the patterns, and if we get it ready before 2.7.0 thats great, but i don't think its a blocker. We'll make sure the docs map out how this works 👍 

I created #1798, feel free to look into it :)

snichme:
Which is more correct then? 
`additional_scopes_key` `additional_scope_keys` or `additional_scopes_keys`?

hananasen:
> Which is more correct then? `additional_scopes_key` `additional_scope_keys` or `additional_scopes_keys`?

I think `additional_scopes_keys` since they can include more than one scope and there can be more than one key in the jwt

dentarg:
I guess Claude does not know we haven't made a real release with `additional_scopes_key` (so no need to deprecate it)

PR #1792, opened 2026-03-04
Bugfix: replication shouldn't break if mfiles are closed

spuun:
If a queue is closed because of a bad segment, the segment files are still registered in the replicator and will be replicated during a full sync. Because they are closed an error will be raised which aborts the replication. There's also a possibility that the close is happening while a checksum is being calculated which then will result in a segfault.

This PR will do two things:
1. Wait for any full sync to be done before closing files
2. When mfiles are closed, they are "re-registered" in the replicator with path only to make sure the files are still being replicated.

Specs should cover it

### DIFF
--- a/spec/clustering_spec.cr
+++ b/spec/clustering_spec.cr
@@ -21,15 +21,19 @@ private def do_full_sync(tcp_server, replicator, wg : WaitGroup? = nil) : Fiber:
   Fiber::ExecutionContext::Isolated.new("test-follower") do
     client_io = TCPSocket.new("localhost", tcp_server.local_address.port)
     begin
+      # Handshake and authentication
       client_io.write LavinMQ::Clustering::Start
       client_io.write_bytes replicator.password.bytesize.to_u8, IO::ByteFormat::LittleEndian
       client_io.write replicator.password.to_slice
       client_io.read_byte
       client_io.write_bytes 2i32, IO::ByteFormat::LittleEndian
       client_io.flush
-      wg.try &.done # negotiation done — server is now entering files_with_hash on @mt
+      # Signal that full sync is about to start
+      wg.try &.done
       sha1_size = Digest::SHA1.new.digest_size
       client_lz4 = Compress::LZ4::Reader.new(client_io)
+      # Do the full sync two times without requesting files (everything is up
+      # to date)
       2.times do
         loop do
           filename_len = client_lz4.read_bytes Int32, IO::ByteFormat::LittleEndian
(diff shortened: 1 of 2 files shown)


viktorerlingsson:
Should we `wait_for_sync` in `delete_file` as well?

spuun:
> Should we `wait_for_sync` in `delete_file` as well?

Yeah, maybe. I think I managed to reason about it and that it wasn't necessary, but maybe better be safe than sorry... hm.

viktorerlingsson:
> Yeah, maybe. I think I managed to reason about it and that it wasn't necessary, but maybe better be safe than sorry... hm.

Yeah, I think it's more unlikely to cause problems, but I think it can cause problems at least on queue delete or purge_all 🤔

spec/message_store_spec.cr
```diff
@@ -3,6 +3,64 @@ require "file_utils"
 require "time"
 require "../src/lavinmq/message_store"
 
+class SpyReplicator
```
viktorerlingsson, line 7: I think this needs a comment - what is this and why is it here?

spec/clustering_spec.cr
```diff
@@ -5,6 +5,47 @@ require "../src/lavinmq/clustering/controller"
...
+  Fiber::ExecutionContext::Isolated.new("test-follower") do
+    client_io = TCPSocket.new("localhost", tcp_server.local_address.port)
+    begin
+      client_io.write LavinMQ::Clustering::Start
+      client_io.write_bytes replicator.password.bytesize.to_u8, IO::ByteFormat::LittleEndian
+      client_io.write replicator.password.to_slice
+      client_io.read_byte
+      client_io.write_bytes 2i32, IO::ByteFormat::LittleEndian
+      client_io.flush
+      wg.try &.done # negotiation done — server is now entering files_with_hash on @mt
```
viktorerlingsson: i think this should have some comments, like "authenticating follower" or something

spec/clustering_spec.cr
```diff
@@ -5,6 +5,47 @@ require "../src/lavinmq/clustering/controller"
...
+      sha1_size = Digest::SHA1.new.digest_size
+      client_lz4 = Compress::LZ4::Reader.new(client_io)
+      2.times do
```
viktorerlingsson, line 37: I'd like a comment here as well, explaining why this runs 2 times and maybe what happens in the loop

spuun:
> ### Bug: `wg.wait` inside `@sync_lock` in `delete_file` risks holding the lock indefinitely

The `WaitGroup` will always be resolved. When a follower is disconnected `#close` will be called which will resolve all pending actions.

viktorerlingsson (approved):
Not sure how to properly test this besides the spec, but the code looks good to me 👍

PR #1796, opened 2026-03-05
fix: print usage on unexpected arguments in lavinmqperf

oskgu360:
Abort run if unexpected arguments are provided. 

Stupid users (me) need help. I got confused why stuff worked when it defaulted to localhost behind the scenes, but real cause of error was that I forgot to pass option flags :) 

Prior (connecting to default localhost)
```
lavinmqperf throughput amqp://nouser:nopw@nodomain.com:5672/test
Publish rate: 1519814 msgs/s Consume rate: 1519778 msgs/s
...
```
After
```
bin/lavinmqperf throughput amqp://nouser:nopw@nodomain.com:5672/test
Unexpected argument: amqp://nouser:nopw@nodomain.com:5672/test

Usage: bin/lavinmqperf [protocol] [throughput | bind-churn | queue-churn | connection-churn | connection-count | queue-count] [arguments]
    -h, --help                       Show this help
....
```


spuun:
What about using `OptionsParser#unknown_args` and place it in `Perf`? I think that's doable, that "argumetns" are never used so any argument is an unknown argument?

PR #1799, opened 2026-03-06
Use default `wait_for` timeout in spec

dentarg:
This spec sometimes times out in CI on macOS. CI uses shared infrastructure so there are no guarantees how performant the runners are at a given time.

Most other wait_for calls does not specify a timeout, so I don't think we have any special requirements here about how quick this needs to be.

The timeout was tweaked before in e45a1f7dd94e77dfd0a755f1ff3c1ea030dc40bf

Recent CI timeouts that I noticed:

* https://github.com/cloudamqp/lavinmq/actions/runs/22345888663/job/64660272287#step:5:3716
* https://github.com/cloudamqp/lavinmq/actions/runs/22345870018/job/64660207705#step:5:3711
* https://github.com/cloudamqp/lavinmq/actions/runs/21860578235/job/63088351486#step:5:3679
* https://github.com/cloudamqp/lavinmq/actions/runs/21747588807/job/62737224067#step:5:3349

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,6 +311,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: compile
     continue-on-error: true
+    if: github.event.pull_request.head.repo.full_name == github.repository
 
     steps:
       - name: Install dependencies


dentarg:
Somehow I travled back to 2023 ... let me try again.

spuun (approved):
I wonder why we ever want to set the timeout…

Issue #1727, opened 2026-02-20
Form resets on warnings from backend


snichme:
**Describe the bug**

When creating an Exchange and you happen to select a vhost for which your user doesn't have permissions to create exchanges on you get back an message from backend saying that you dont have permissions and then the whole form is reset. 

Create a new vhost called test
Create a new user that has full access to test but not `/`
Create an Topic exchange on `/` with name "My Exchange" and enter `{ "x-alternate-exchange": "X" }` 
When you submit you get a permissions warning but now the whole form is reset so you have to fill everything in again. 

Ideas for solving this
* Dont reset the form 
* Only show the vhosts that you have permissions to create exchanges on in the "Add exchange form"

snichme:
Not really a bug, more of a feature improvement.

spuun:
It's not only this form that's affected. It would be nice with a generic solution for this.

PR #1802, opened 2026-03-09
Centralize form submit handling

ThomasSarlin:
## WHY
Adds to #1727 — form inputs were cleared even when the server returned an error, forcing users to re-enter all values

## HOW
  - Introduces a shared HTTP.submitForm() helper that centralizes the submit-then-reset pattern and only resets the form on success, it is also made async for the possibility of add chaining for when the submitForm method is completed.
  - Refactors all form submit handlers across the management UI to use the new helper
  - Adds Playwright specs covering both the success (form resets) and error (form preserves values) behaviors for exchanges and queues forms

## Test plan

- [ ] Run Playwright specs: make test-frontend (or npx playwright test spec/frontend/submit_form.spec.js)
- [ ] Verify exchanges form: create an exchange successfully — form should reset; attempt to create one that fails (e.g. no permission) — form should keep your input
- [ ] Verify queues form: same success/error check
- [ ] Spot-check other forms (policies, shovels, federation upstreams, users, bindings, vhosts) to confirm they still submit and reset correctly on success


static/js/http.js
```diff
@@ -66,8 +68,21 @@ function noencode (v) {
...
 
+async function submitForm (form, method, url, options = {}) {
+  const { body, table, toast, callback } = options
+  return request(method, url, { body })
+    .then(res => {
+      if (res?.is_error) return
+      if (table) table.reload()
+      if (toast) DOM.toast(toast)
+      form.reset()
+      if (callback) callback()
```
carlhoerberg: do we need to keep this, promisies and callbacks? isn't it better that callers of this method chain instead?

static/js/http.js
```diff
@@ -1,3 +1,5 @@
+import * as DOM from './dom.js'
```
carlhoerberg: do every user of the http module now also have to load the dom module?

Isn't this easy enough?

```js
HTTP.submitForm(evt.target, 'PUT', url, {
    body,
    table: upstreamsTable,
  }).then(() => DOM.toast(`Upstream ${name} saved`)
```

static/js/http.js
```diff
@@ -66,8 +68,21 @@ function noencode (v) {
   return new NoUrlEscapeString(v)
 }
 
+async function submitForm (form, method, url, options = {}) {
+  const { body, table, toast, callback } = options
+  return request(method, url, { body })
+    .then(res => {
+      if (res?.is_error) return
```
carlhoerberg: now we just ignore errors? should we raise instead? so that callers can chain `catch`?

carlhoerberg:
```suggestion
      if (res?.is_error) throw new Error(`HTTP error: ${res.status}`);
```

ThomasSarlin: `if (res?.is_error) return` is how it was handled today in queue, queues, exchange and streams so that followed along. We could absolutely opt for throwing an error if that is present.

ThomasSarlin: This does however bypass the check in
https://github.com/cloudamqp/lavinmq/blob/0123288b2ee23141a1f1e3b2780e5752dd772aac/static/js/exchange.js#L111-L116

Need to work on how to resolve so that message is presented anyway.

static/js/http.js
```diff
@@ -66,8 +66,19 @@ function noencode (v) {
   return new NoUrlEscapeString(v)
 }
 
+async function submitForm (form, method, url, options = {}) {
+  const { body, table } = options
+  return request(method, url, { body })
+    .then(res => {
+      if (res?.is_error) throw res
+      if (table) table.reload()
```
snichme: Do we need to give `table` as options, we return a promise so `table.reload()` can be done in a `.then()` from the caller, think that is easier to follow.

ThomasSarlin: That is true, would make for a more clear read. 👍

kickster97:
Superseded by #2214

PR #1807, opened 2026-03-09
Add RateLimiter and reduce clustering sync log verbosity

### LINKED ISSUE #1805: Logging is too verbose when syncing followers
If a lavinmq cluster has a lot of files, we can overwhelm journald with a huge amount of logs during the initial sync of a new follower. 

We should change it so many of the info logs are debug logs instead, and add some progress logging every x file instead. 

`systemd-journald[3473972]: [🡕] Suppressed 4448 messages from lavinmq.service`

viktorerlingsson:
## Summary
- Add time-based `RateLimiter` utility for throttled progress logging
- Apply it (2s interval) to follower hash calculation/upload, client compare/receive loops, and `MessageStore` segment loading (replacing the hardcoded every-128 counter)
- Downgrade noisy per-file sync logs (checksum calc, match/mismatch, delete, request, receive, upload) from info to debug
- Add summary lines at start/end of each sync phase

Closes #1805

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -1,6 +1,6 @@
 require "../data_dir_lock"
 require "../clustering"
-require "../log_rate_limiter"
+require "../rate_limiter"
 require "./checksums"
 require "./proxy"
 require "lz4"
@@ -154,7 +154,7 @@ module LavinMQ
         requested_files = Array(String).new
         file_count = 0
         Log.info { "Calculating checksums and comparing files" }
-        log_limiter = LogRateLimiter.new(2.seconds)
+        log_limiter = RateLimiter.new(2.seconds)
         loop do
           filename_len = lz4.read_bytes Int32, IO::ByteFormat::LittleEndian
           break if filename_len.zero?
@@ -215,7 +215,7 @@ module LavinMQ
           Log.warn(exception: ex) { "Failed to delete #{path}" }
         end
         received_count = 0
-        log_limiter = LogRateLimiter.new(2.seconds)
+        log_limiter = RateLimiter.new(2.seconds)
         requested_files.each do |filename|
           file_from_socket(filename, lz4)
           received_count &+= 1
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -151,6 +151,7 @@ module LavinMQ
         remote_hash = Bytes.new(sha1.digest_size)
         files_to_delete, dirs_to_delete = ls_r(@data_dir)
         requested_files = Array(String).new
+        file_count = 0
         loop do
           filename_len = lz4.read_bytes Int32, IO::ByteFormat::LittleEndian
           break if filename_len.zero?
@@ -166,7 +167,7 @@ module LavinMQ
           end
           if File.exists? path
             unless local_hash = @checksums[filename]?
-              Log.info { "Calculating checksum for #{filename}" }
+              Log.debug { "Calculating checksum for #{filename}" }
               sha1.file(path)
               local_hash = sha1.final
               @checksums[filename] = local_hash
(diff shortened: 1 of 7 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -161,35 +162,42 @@ module LavinMQ
...
               request_file(filename, socket)
             else
-              Log.info { "Matching hash: #{path}" }
+              Log.debug { "Matching hash: #{path}" }
             end
           else
             requested_files << filename
             request_file(filename, socket)
           end
+          Log.info { "Compared #{file_count} files" } if (file_count &+= 1) % 128 == 0
```
spuun: `128` can probably be even higher?

I wonder if it would be possible to use `Time.instant` and log only every second or what ever we want instead of per nth file?

src/lavinmq/clustering/client.cr
```diff
@@ -161,35 +162,42 @@ module LavinMQ
...
+          Log.info { "Deleting #{files_to_delete.size} files not on leader" }
+          files_to_delete.each do |path|
+            Log.debug { "File not on leader: #{path}" }
+            File.delete path
+          end
         end
+        received_count = 0
         requested_files.each do |filename|
           file_from_socket(filename, lz4)
+          Log.info { "Received #{received_count}/#{requested_files.size} files" } if (received_count &+= 1) % 128 == 0
```
spuun: Ditto...

src/lavinmq/clustering/follower.cr
```diff
@@ -146,10 +147,13 @@ module LavinMQ
           total_time_taken = (Time.instant - start).total_seconds
           bps = (sent_bytes / total_time_taken).round.to_u64
           time_left = bps > 0 ? (total_requested_bytes / bps).round(1) : 0
-          Log.info { "Uploaded #{filename} in #{bps.humanize_bytes}/s" }
-          Log.info { "#{total_requested_bytes.humanize_bytes} left, expected #{time_left}s left" }
+          Log.debug { "Uploaded #{filename} at #{bps.humanize_bytes}/s" }
+          if (uploaded_count &+= 1) % 128 == 0
+            Log.info { "Uploaded #{uploaded_count}/#{requested_files.size} files at #{bps.humanize_bytes}/s, #{total_requested_bytes.humanize_bytes} left (~#{time_left}s)" }
+          end
```
spuun: Ditto again.

spuun (approved):
Maybe we can make a little resuable "rate limiter" like

```crystal
log_limiter = TimedRateLimiter.new(3.seconds)
log_limiter.do { Log.info { "this is line is only logged once every 3rd second" } }
```

I think we have more places than this were it could be useful?

viktorerlingsson:
> Maybe we can make a little resuable "rate limiter" like
> 
> ```crystal
> log_limiter = TimedRateLimiter.new(3.seconds)
> log_limiter.do { Log.info { "this is line is only logged once every 3rd second" } }
> ```
> 
> I think we have more places than this were it could be useful?

Sounds neat! I gave it a shot with `LogRateLimiter`. WDYT?

spuun:
> > Maybe we can make a little resuable "rate limiter" like
> > ```crystal
> > log_limiter = TimedRateLimiter.new(3.seconds)
> > log_limiter.do { Log.info { "this is line is only logged once every 3rd second" } }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think we have more places than this were it could be useful?
> 
> Sounds neat! I gave it a shot with `LogRateLimiter`. WDYT?

Yes, exactly. We can probably use this in multiple places!

spuun (approved):

spuun:
> Sounds neat! I gave it a shot with `LogRateLimiter`. WDYT?

But maybe give it a more generic name so we can use it to more things than logging?

viktorerlingsson:
> But maybe give it a more generic name so we can use it to more things than logging?

Yeah, guess it can be used to rate limit anything!

viktorerlingsson:
@spuun should we merge this now or wait for 2.7.1 you think? 🤔

PR #1816, opened 2026-03-12
Automatic retries with backoff using internal queue

snichme:

This branch adds a retry feature for queues that allows messages to be automatically retried after rejection with exponential backoff — it introduces a RetryQueue class and retry argument parsing, modifies the AMQP queue and client to support retry behavior on nack/reject.

In the [research.md](https://github.com/cloudamqp/lavinmq/blob/78ec99c1afb695be27322ff95273c3f778270b13/research.md) file there are some motivations to the current implementation

Fixes #1815 

There are specs for the changes  and also a ruby script that uses different combinations of the queue arguments to they can be tested easily.


snichme:
Currently this solution doesn't offer a solution for poison messages, so there is no way to send a message directly to DLX/discard if it's contain bad data or something else. What will happen now is that the message will be retried as with any message. 
A solution to handle poison message would be good.

snichme:
Other things to consider, in this PR or new ones based on this: 
* Apply this features using policies as well as queue arguments
* How can this be observed? Are internal queues exposed in prometheus metrics? Are there any other metrics that could be relevant? Currently internal queues aren't listed in the UI, should we show something on Queue view related to this?
* How will this work with Streams? Do we need to consider MQTT Sessions?
*

spec/queue_retry_spec.cr
```diff
@@ -0,0 +1,242 @@
...
+          args = AMQP::Client::Arguments.new({
+            "x-retry-max-count" => 3,
+            "x-retry-delay"     => 100,
+          })
+          q = ch.queue(QUEUE_NAME, args: args)
+          ch.default_exchange.publish_confirm("msg1", q.name)
+
+          msg = wait_for { q.get(no_ack: false) }
+          msg.not_nil!.body_io.to_s.should eq "msg1"
+          msg.not_nil!.reject(requeue: false)
```
spuun:
```suggestion
          msg = wait_for { q.get(no_ack: false) }
          msg = msg.should_not be_nil
          msg.body_io.to_s.should eq "msg1"
          msg.reject(requeue: false)
```
Using the `should_not be_nil` (or `should be_a(type)`) to avoid `not_nil!` is nice in specs IMO.

PR #1824, opened 2026-03-23
fix: deprecated --guest-only-loopback flag silently ignored

viktorerlingsson:
## Summary
- `--guest-only-loopback` no longer used `property?`; replaced with a bare ivar and custom setter that forwards to `@default_user_only_loopback`
- Added a generic test harness (`deprecated_config_spec.cr`) that validates all deprecated options: forwarding (INI + CLI), deprecation warnings, and completeness (fails if a new deprecated annotation is added without a corresponding test entry)

Closes #1822

### DIFF
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -251,7 +251,7 @@ module LavinMQ
       @[IniOpt(section: "main")]
       property? default_user_only_loopback : Bool = true
 
-      @[CliOpt("", "--guest-only-loopback=BOOL", "Limit guest user to only connect from loopback address", deprecated: "Use --default-user-only-loopback instead.", section: "options")]
+      @[CliOpt("", "--guest-only-loopback=BOOL", "Limit guest user to only connect from loopback address", deprecated: "Deprecated: Use --default-user-only-loopback instead.", section: "options")]
       @[IniOpt(section: "main", deprecated: "default_user_only_loopback")]
       @guest_only_loopback : Bool = true
 
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -253,7 +253,11 @@ module LavinMQ
 
       @[CliOpt("", "--guest-only-loopback=BOOL", "Limit guest user to only connect from loopback address", deprecated: "Use --default-user-only-loopback instead.", section: "options")]
       @[IniOpt(section: "main", deprecated: "default_user_only_loopback")]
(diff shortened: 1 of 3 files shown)


src/lavinmq/config/options.cr
```diff
@@ -250,7 +250,11 @@ module LavinMQ
 
       @[CliOpt("", "--guest-only-loopback=BOOL", "Limit guest user to only connect from loopback address", deprecated: "Use --default-user-only-loopback instead.", section: "options")]
       @[IniOpt(section: "main", deprecated: "default_user_only_loopback")]
-      property? guest_only_loopback : Bool = true
+      @guest_only_loopback : Bool = true
```
kickster97, line 256: Nothing prevents someone from using `property?` here again and reintroducing the bug for another deprecated config. We could consider adding a `forward_to` parameter to the annotation and letting the macros handle the forwarding instead:

```suggestion
      @[IniOpt(section: "main", deprecated: "default_user_only_loopback", forward_to: "default_user_only_loopback")]
      property? guest_only_loopback : Bool = true
```


This would require two changes in `config.cr`

In `parse_section`, add `forward_to: anno[:forward_to]` to the map at line 190, and replace line 205 with something like:

```
{% if (forward = var[:forward_to]) %}
self.{{forward.id}} = parse_value(v, {{var[:transform]}})
{% else %}
self.{{var[:var_name]}} = parse_value(v, {{var[:transform]}})
{% end %}
````
In `parse_cli`, replace line 86 with something like:
```
{% ini_opt = ivar.annotation(IniOpt) %}
{% if ini_opt && ini_opt[:forward_to] %}
self.{{ini_opt[:forward_to].id}} = parse_value(value, {{value_parser}})
{% else %}
self.{{ivar.name.id}} = parse_value(value, {{value_parser}})
{% end %}
```

viktorerlingsson: Okay, after discussions with the team, this will be problematic. Mainly because of transformers and different data types. Lets just keep the setters for now and instead add some "smart" specs to monitor all deprecated values

viktorerlingsson: Added some specs to make sure we won't miss this in the future: https://github.com/cloudamqp/lavinmq/pull/1824/changes/4a4a76ba391e151ec5ccf464f3b6dec3fbe63e11#diff-21b4aac926bdb6c1030222ddae6484b4c21f58f26b27465275bb4d094b46e84a

spuun (approved):

PR #1828, opened 2026-03-24
Add load_definitions config option

baelter:
## Motivation

When running LavinMQ in Docker Compose, there's no clean way to pre-configure vhosts, users, exchanges, and queues. The workaround is to start the container, then exec `lavinmqctl import_definitions` inside it, which means the broker briefly runs without the expected topology. RabbitMQ solves this with `load_definitions`, and this PR brings the same capability to LavinMQ.

With this, you mount a definitions JSON file into the container and point the config at it. The broker imports it on startup before accepting connections.

## Changes

- New `load_definitions` config option under `[main]` that takes a path to a JSON definitions file
- `GlobalDefinitions.import_from_file` class method that reads and imports the file, reusing the same logic as `POST /api/definitions` and `lavinmqctl import_definitions`
- Extracts `Definitions`, `GlobalDefinitions`, and `VHostDefinitions` from the HTTP controller into `src/lavinmq/definitions.cr` so the launcher doesn't pull in the HTTP layer
- Boot-time import is non-destructive: existing users, permissions, policies, and parameters are not overwritten (matches RabbitMQ behavior). HTTP API imports keep their upsert behavior.
- A fresh node with `load_definitions` configured skips creating the default "/" vhost and "guest" user, letting the definitions file be the sole source of truth

## Usage

```ini
[main]
load_definitions = /path/to/definitions.json
```

```yaml
# docker-compose.yml
services:
  lavinmq:
    image: cloudamqp/lavinmq
    volumes:
      - ./definitions.json:/etc/lavinmq/definitions.json:ro
      - ./lavinmq.conf:/etc/lavinmq/lavinmq.conf:ro
```

## Test plan

- [x] Spec: imports definitions from a JSON file on startup
- [x] Spec: raises `File::NotFoundError` for missing file
- [x] Spec: raises `JSON::ParseException` for invalid JSON
- [x] Spec: boot-time import does not overwrite existing users
- [x] Spec: boot-time import does not overwrite existing permissions
- [x] Spec: fresh node skips default vhost/user when load_definitions is configured
- [x] Existing definitions specs pass
- [x] Linter clean


src/lavinmq/launcher.cr
```diff
@@ -62,6 +63,7 @@ module LavinMQ
       @amqp_server = amqp_server = LavinMQ::Server.new(@config, @replicator)
       @http_server = http_server = LavinMQ::HTTP::Server.new(amqp_server)
       setup_log_exchange(amqp_server)
+      load_definitions(amqp_server)
```
kickster97: We should load the definitions file before setting up the log exchange so the "/" vhost exists when starting from a clean data directory with `load_definitions` configured. If we don't, `setup_log_exchange` will crash since the default vhost is not created when `load_definitions` is set. (I guess this is by design?) 

```suggestion
      load_definitions(amqp_server)
      setup_log_exchange(amqp_server)
```

src/lavinmq/definitions.cr
```diff
@@ -0,0 +1,411 @@
...
+    private def import_permissions(body, skip_existing = false)
+      if permissions = body["permissions"]?
+        permissions.as_a.each do |p|
+          vhost = p["vhost"].as_s
+          user = p["user"].as_s
+          next if skip_existing && @amqp_server.users[user]?.try(&.permissions[vhost]?)
+          configure = p["configure"].as_s
+          read = p["read"].as_s
+          write = p["write"].as_s
+          @amqp_server.users[user].permissions[vhost] = {
```
kickster97: This is not really introduced by this PR, but if we import definitions for a user that doesn't exist we will crash. Not as big of a deal when the code was only in the HTTP controller, but if we run this at boot its more problematic. 
We should nil-guard and log a warning, something like: 

```suggestion
          unless u = @amqp_server.users[user]?
             Log.warn { "No user named #{user}, can't import permissions" }
             next
           end
           u.permissions[vhost] = {
```

src/lavinmq/launcher.cr
```diff
@@ -135,10 +137,19 @@ module LavinMQ
       {% end %}
     end
 
+    private def load_definitions(amqp_server)
+      path = @config.load_definitions
+      return if path.empty?
+      GlobalDefinitions.import_from_file(path, amqp_server)
```
snichme, line 143: We should have some handling here is the file pointed to by the config doesn't exists or is invalid. 

I think the broker should just not start but it should be handled in a graceful way not just `FileNotFound`

baelter: Now rescued in the launcher's `load_definitions` and aborts with a clear message for missing file, permission denied, invalid JSON, and a generic fallback. Importer still raises raw exceptions (existing specs cover that), the launcher just turns them into a clean shutdown.

src/lavinmq/definitions.cr
```diff
@@ -0,0 +1,417 @@
...
+      end
+    end
+
+    def fetch_vhost?(json) # ignore vhost property, always use the specified one
+      @vhost
+    end
+  end
+
+  class GlobalDefinitions < DefinitionsImporter
+    Log = LavinMQ::Log.for "definitions"
```
kickster97:
```suggestion

```

src/lavinmq/definitions.cr
```diff
@@ -0,0 +1,417 @@
...
+
+    def fetch_vhost?(json)
+      if name = json["vhost"]?.try(&.as_s)
+        if vhost = vhosts[name]?
+          vhost
+        else
+          Log.warn { "No vhost named #{name}, can't import #{name}" }
+        end
+      else # if vhost property is missing, use first/default vhost
+        vhosts.first_value
```
kickster97:
```suggestion
     elsif vhost = vhosts.first_value?
       vhost # if vhost property is missing, use first/default vhost
     else
      Log.warn { "No vhost defined, can't import entry without vhost" }
```

kickster97 (approved):

src/lavinmq/definitions.cr
baelter: Removed the duplicate `Log` constant in commit 6c6c545e.

src/lavinmq/definitions.cr
baelter: Applied the suggestion -- `fetch_vhost?` now uses `first_value?` and warns when no vhost is defined. Same commit (6c6c545e).

kickster97:
Don't forget to add documentation on this :)

PR #1829, opened 2026-03-25
Handle 404 errors gracefully when viewing deleted entities in UI

### LINKED ISSUE #889: Handle 404 responses from api nicer
**Describe the bug**
If you have an exchange view open and the exchange is deleted you'll get a "Something went wrong: Error fetching data: [object Object]" error above the bindings table. We're probably having similar messages in other views.

**How to reproduce**
Open the management UI and have the exchange view open, while an exchange gets deleted via amqp.

**Expected behavior**
We should handle 404s better and display that the entity is (probably) deleted.

queelius:
## Summary

Fixes #889

When an entity (exchange, queue, connection, etc.) is deleted while its management UI detail page is open, the UI previously showed the unhelpful error "Something went wrong: Error fetching data: [object Object]". This PR makes the UI handle 404 responses gracefully:

- **Entity detail pages** now catch 404 errors, stop the periodic refresh interval, and replace the page content with a clear "not found" message and a link back to the entity list page
- **Table data sources** now emit a distinct `not_found` event on 404, which the table renders as a user-friendly message instead of the generic error
- **Error serialization** is fixed so error objects always have a `message` property (preventing `[object Object]` in string concatenation)

## Root cause

The error object created in `http.js` on non-OK responses had no `message` property. When the datasource caught this error and tried to emit it, the object was passed through to `toggleDisplayError()` which concatenated it into a string, producing `[object Object]`. Additionally, none of the entity detail pages (exchange.js, queue.js, etc.) had `.catch()` handlers for their periodic API fetch calls.

## Files changed

- `static/js/http.js` - Add `message` property to error objects
- `static/js/datasource.js` - Emit `not_found` event on 404; ensure error messages are always strings
- `static/js/table.js` - Handle `not_found` event; fix error display fallback
- `static/js/dom.js` - Add `showEntityNotFound()` helper function
- `static/js/exchange.js` - Handle 404 on entity fetch
- `static/js/queue.js` - Handle 404 on entity fetch
- `static/js/connection.js` - Handle 404 on entity fetch
- `static/js/channel.js` - Handle 404 on entity fetch
- `static/js/vhost.js` - Handle 404 on entity fetch
- `static/js/user.js` - Handle 404 on entity fetch
- `static/js/stream.js` - Handle 404 on entity fetch

## Test plan

- [ ] Open an exchange detail page, delete the exchange via AMQP, verify the UI shows "Exchange not found" with a link to the exchanges list
- [ ] Open a queue detail page, delete the queue, verify the UI shows "Queue not found" with a link to the queues list
- [ ] Verify the same behavior for connections, channels, vhosts, users, and streams
- [ ] Verify the bindings table on exchange/queue pages shows "This resource no longer exists" instead of `[object Object]`
- [ ] Verify normal (non-404) errors still display correctly
- [ ] Run `npx standard static/js` to verify linting passes

### DIFF
--- a/static/js/channel.js
+++ b/static/js/channel.js
@@ -94,6 +94,7 @@ const prefetchHandler = () => {
 
 const prefetch = prefetchHandler()
 document.getElementById('ch-prefetch').appendChild(prefetch.el)
+let channelUpdateInterval = null
 function updateChannel () {
   HTTP.request('GET', channelUrl).then(item => {
     Chart.update(chart, item.message_stats)
@@ -118,7 +119,12 @@ function updateChannel () {
       chMode.replaceChildren(confirmSpan)
     }
     document.getElementById('ch-global-prefetch').textContent = Helpers.formatNumber(item.global_prefetch_count)
+  }).catch(e => {
+    if (e.status === 404) {
+      clearInterval(channelUpdateInterval)
+      DOM.showEntityNotFound('Channel', channel, 'channels')
+    }
   })
 }
 updateChannel()
-setInterval(updateChannel, 5000)
+channelUpdateInterval = setInterval(updateChannel, 5000)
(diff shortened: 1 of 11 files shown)


queelius:
Closing this to keep your queue clean. Happy to reopen if useful, or feel free to pick up any of it. Thanks!

PR #1832, opened 2026-03-26
Fix flaky config specs caused by STDERR.reopen

viktorerlingsson:
## Summary
- Add `@io` constructor param to `Config` (defaults to `STDERR`) and route all warning/error output through it
- Specs that verify warning text now pass `IO::Memory` instead of using `STDERR.reopen`
- Remove duplicate `Config.initialize` reopening in `spec_helper.cr`
- Simplify `Option` struct by removing `@deprecation_warn_msg` and `@io` fields

## Root cause
`STDERR.reopen` in the "log warning" specs corrupted Crystal's event loop fd registration for fd 2, causing intermittent `File not open for writing` errors in other config specs under `--order random`.


kickster97 (approved):
nice

PR #1836, opened 2026-03-30
Fix MQTT subscribe permission check logic

kickster97:
The subscribe permission check used `&&` instead of `||`, allowing users with only read or only write permissions to subscribe. Both permissions are required: read on the exchange (to consume messages) and write on the session queue (to route messages into it). This aligns with how RabbitMQ handles MQTT subscribe permissions.

Run added spec

### DIFF
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -164,7 +164,7 @@ module LavinMQ
 
       def recieve_subscribe(packet : MQTT::Subscribe)
         if Config.instance.mqtt_permission_check_enabled?
-          if !user.can_read?(@broker.vhost.name, EXCHANGE) && !user.can_write?(@broker.vhost.name, "mqtt.#{client_id}")
+          if !user.can_read?(@broker.vhost.name, EXCHANGE) || !user.can_write?(@broker.vhost.name, "mqtt.#{client_id}")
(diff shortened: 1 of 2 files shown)


src/lavinmq/mqtt/client.cr
```diff
@@ -164,7 +164,7 @@ module LavinMQ
 
       def recieve_subscribe(packet : MQTT::Subscribe)
         if Config.instance.mqtt_permission_check_enabled?
-          if !user.can_read?(@broker.vhost.name, EXCHANGE) && !user.can_write?(@broker.vhost.name, "mqtt.#{client_id}")
+          if !user.can_read?(@broker.vhost.name, EXCHANGE) || !user.can_write?(@broker.vhost.name, "mqtt.#{client_id}")
```
hananasen:
```suggestion
          unless user.can_read?(@broker.vhost.name, EXCHANGE) && user.can_write?(@broker.vhost.name, "mqtt.#{client_id}")
```
I find that easier to understand, but no strong opinions :)

Also, nice to update the docs around this, I can look into it.

hananasen (approved):
LGTM atleast 😄 just a minor comment, feel free to ignore if you don't agree on it

src/lavinmq/mqtt/client.cr
kickster97: I agree :)

PR #1838, opened 2026-03-31
Recognize IPv4-mapped IPv6 loopback addresses

walro:

When the server binds to `::` on a dual-stack machine, IPv4 clients get peer addresses like `::ffff:127.0.0.1`. This was not recognized as loopback, causing `default_user_only_loopback` to reject the default user connecting from localhost over IPv4.

Specs

### DIFF
--- a/spec/connection_info_spec.cr
+++ b/spec/connection_info_spec.cr
@@ -1,0 +1,40 @@
+require "./spec_helper"
+
+describe LavinMQ::ConnectionInfo::IPAddress do
+  describe "#loopback?" do
+    it "returns true for IPv4 loopback" do
+      addr = LavinMQ::ConnectionInfo::IPAddress.new(Socket::IPAddress.new("127.0.0.1", 0))
+      addr.loopback?.should be_true
+    end
+
+    it "returns true for IPv4 loopback subnet" do
+      addr = LavinMQ::ConnectionInfo::IPAddress.new(Socket::IPAddress.new("127.0.0.2", 0))
+      addr.loopback?.should be_true
+    end
+
+    it "returns true for IPv6 loopback" do
+      addr = LavinMQ::ConnectionInfo::IPAddress.new(Socket::IPAddress.new("::1", 0))
+      addr.loopback?.should be_true
+    end
+
+    it "returns true for IPv4-mapped IPv6 loopback" do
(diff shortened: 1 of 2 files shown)


spuun:
I wonder why we ain't using `Socket::IPAddress#loopback?`? 🤔

walro:
> I wonder why we ain't using `Socket::IPAddress#loopback?`? 🤔

A few lines up we have this comment: https://github.com/cloudamqp/lavinmq/blob/bde4c17ad4d1b32eadadf986ba56a2d399ba68ee/src/lavinmq/connection_info.cr#L27. Possibly not a problem anymore? We could create a throwaway `Socket::IPAddress` instance and delegate to`#loopback?` but seems a bit wasteful.

(or better, we capture it in the initializer)

walro:
@spuun cb12fc2813a69f1d69c5f0dac98a380eddb4df70 WDYT? It will do the loopback check for every IP now, as compared with before it did it lazily. But it seems to be evaluated on every AMQP connection anyway and the bitwise checking in stdlib ought to be quicker than the previous implementation. It will take a toll on the MQTT path where we don't seem to be concerned with `#loopback?` tho.

Personally I think the simpler code is worth it.

spuun:
> > I wonder why we ain't using `Socket::IPAddress#loopback?`? 🤔
> 
> A few lines up we have this comment:
> 
> https://github.com/cloudamqp/lavinmq/blob/bde4c17ad4d1b32eadadf986ba56a2d399ba68ee/src/lavinmq/connection_info.cr#L27
> . Possibly not a problem anymore? We could create a throwaway `Socket::IPAddress` instance and delegate to`#loopback?` but seems a bit wasteful.
> 
> (or better, we capture it in the initializer)

I wonder what kind of memory problem. Leak? Segfault? hm... oh well

spuun:
> @spuun [cb12fc2](https://github.com/cloudamqp/lavinmq/commit/cb12fc2813a69f1d69c5f0dac98a380eddb4df70) WDYT? It will do the loopback check for every IP now, as compared with before it did it lazily. But it seems to be evaluated on every AMQP connection anyway and the bitwise checking in stdlib ought to be quicker than the previous implementation. It will take a toll on the MQTT path where we don't seem to be concerned with `#loopback?` tho.
> 
> Personally I think the simpler code is worth it.

I agree!

spuun (approved):

PR #1842, opened 2026-04-07
Fix: handle valueless boolean CLI flags at compile time

kickster97:
Valueless boolean CLI flags like `--clustering` and `--raise-gc-warn` are broken. `OptionParser` passes an empty string for flags without `=VALUE`, and `parse_value("", Bool)` evaluates to `false` instead of `true`.

Fixed by detecting valueless boolean flags at compile time in the `parse_cli` macro and assigning the correct literal (`true`, or `false` for --no- prefixed flags) instead of always going through `parse_value`.

maybe this code is a bit clunky? I'd love to discuss if there are more elegant solutions

run updated specs

### DIFF
--- a/spec/config_spec.cr
+++ b/spec/config_spec.cr
@@ -281,6 +281,8 @@ describe LavinMQ::Config do
       "--default-user=cliuser",
       "--default-user-only-loopback=false",
       "--no-data-dir-lock",
+      "--clustering",
+      "--raise-gc-warn",
       "--clustering-advertised-uri=lavinmq://test:5679",
       "--clustering-bind=0.0.0.0",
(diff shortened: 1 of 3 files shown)


spuun:
Try
```crystal
      @[CliOpt("", "--clustering", "Enable clustering", ->(String) { true }, section: "clustering")]
 ```

kickster97:
> Try
> 
> ```crystal
>       @[CliOpt("", "--clustering", "Enable clustering", ->(String) { true }, section: "clustering")]
> ```

Think that makes for a better solution!

viktorerlingsson (approved):

snichme (approved):

PR #1844, opened 2026-04-07
Fix data_dir and config_file not reading systemd env vars

kickster97:
The config rewrite (#917) replaced the systemd-standard `STATE_DIRECTORY` and `CONFIGURATION_DIRECTORY` env vars with `LAVINMQ_`-prefixed variants. Since the systemd unit file (`extras/lavinmq.service`) declares `StateDirectory=lavinmq` and `ConfigurationDirectory=lavinmq`, systemd sets these vars automatically, and they are no longer read. 

This means anyone using a non-default `StateDirectory= path` in a systemd override will have it silently ignored, falling back to the hardcoded `/var/lib/lavinmq`. Similarly, a custom `ConfigurationDirectory= `in a systemd override would be ignored when resolving the default config file path.

This PR restores both systemd env vars as fallbacks. `LAVINMQ_DATADIR` and `LAVINMQ_CONFIGURATION_DIRECTORY` still take precedence when set.

run added spec

### DIFF
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -13,13 +13,13 @@ module LavinMQ
       DEFAULT_PASSWORD_HASH = Auth::Password::SHA256Password.new("+pHuxkR9fCyrrwXjOD4BP4XbzO3l8LJr8YkThMgJ0yVHFRE+") # Hash of 'guest'
 
       @[CliOpt("-c CONFIG", "--config=CONFIG", "Path to config file", section: "options")]
-      @[EnvOpt("LAVINMQ_CONFIGURATION_DIRECTORY")]
       property config_file = ""
 
       @[CliOpt("-D DIRECTORY", "--data-dir=DIRECTORY", "Data directory", section: "options")]
       @[IniOpt(section: "main")]
+      @[EnvOpt("STATE_DIRECTORY")]
       @[EnvOpt("LAVINMQ_DATADIR")]
-      property data_dir : String = ENV.fetch("STATE_DIRECTORY", "/var/lib/lavinmq")
+      property data_dir : String = "/var/lib/lavinmq"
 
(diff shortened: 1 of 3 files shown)


src/lavinmq/config/options.cr
```diff
@@ -19,7 +19,7 @@ module LavinMQ
       @[CliOpt("-D DIRECTORY", "--data-dir=DIRECTORY", "Data directory", section: "options")]
       @[IniOpt(section: "main")]
       @[EnvOpt("LAVINMQ_DATADIR")]
-      property data_dir : String = "/var/lib/lavinmq"
+      property data_dir : String = ENV.fetch("STATE_DIRECTORY", "/var/lib/lavinmq")
```
snichme: The code looks to support multiple `EnvOpt`  which means you could add 
`@[EnvOpt("STATE_DIRECTORY")] ` after the existing `EnvOpt` and it should pick one of them, the last should have precedence. Have you tested that?

kickster97: hm, yeah.. but the thing is that Crystals `.annotation()` (single) only returns the last annotation, so stacking `@[EnvOpt]` would silently drop all but the last env var. I'd need to change `parse_env` to use `.annotations()` (plural). 

That works for `data_dir`, but `config_file` has to be resolved before `parse_env` runs because `parse_ini` needs the path first, so it still needs special handling. So I end up changing the macro, adding a transform lambda for the directory-to-filepath conversion, and still keeping the early lookup.

What do you think is the downside of the current solution?

spuun: I think it's a tiny change to the code to support multiple `EnvOpt`. 

With the proposed solution in this PR, won't ini config values take precedence over these envs?

kickster97: > With the proposed solution in this PR, won't ini config values take precedence over these envs?

ah yes thats true, need to fix that!

snichme (approved):

PR #1845, opened 2026-04-08
Fix stack overflow with circular or long dead-letter chains

spuun:
 Dead-lettering was implemented via direct recursive calls, causing a stack overflow when two queues mutually dead-lettered to each other or when a sufficiently long dead-letter chain was traversed.

Changes:
  - Replaced recursive dead-letter routing with a trampoline pattern: the DeadLetterer now collects pending (queue, message, callback) tuples into a Context deque and drains them iteratively, keeping the call stack flat regardless of chain length
  - Added a routed callback to DeadLetterer#route so the source message is deleted only after it has been successfully enqueued at the destination (preserving message safety)
  - Introduced publish_internal on queues to accept and propagate the dlx_context through publish → drop_overflow → expire_msg → route, avoiding allocating a new context per hop
  - Moved drop_overflow calls triggered by policy updates into a spawned fiber to avoid re-entrancy

Supersedes #1721

Specs

### DIFF
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -993,5 +993,67 @@ module DeadLetteringSpec
         #  end
       end
     end
+
+    describe "Routed Callback" do
+      # The routed callback deletes the source message from storage.
+      # With a fanout DLX routing to two queues, the old code stored
+      # the same callback once per destination and called it N times.
+      # Correct behavior: source empties to 0 (called at least once)
+      # and both destinations receive all messages (routing completed).
+      # A double-call would attempt to delete the same segment twice,
+      # risking store corruption or stat errors.
+      it "is called once when dead lettering to multiple queues" do
+        with_amqp_server do |s|
+          with_channel(s) do |ch|
+            ch.exchange_declare("dlx_fanout", "fanout")
+            src = ch.queue("src_fanout", args: AMQP::Client::Arguments.new({
+              "x-dead-letter-exchange" => "dlx_fanout",
+            }))
+            dst1 = ch.queue("dst_fanout_1")
+            dst2 = ch.queue("dst_fanout_2")
+            dst1.bind("dlx_fanout", "")
+            dst2.bind("dlx_fanout", "")
+
+            n = 3
+            n.times { |i| ch.default_exchange.publish_confirm("msg#{i + 1}", src.name) }
+            get_n(n, src, &.reject(requeue: false))
+
+            wait_for { src.message_count == 0 }
+            wait_for { dst1.message_count == n }
+            wait_for { dst2.message_count == n }
+
+            src.message_count.should eq 0
+            dst1.message_count.should eq n
+            dst2.message_count.should eq n
+          end
+        end
+      end
+
+      # When cycle detection drops all destinations the routed callback
+      # must still fire so the source message is removed from storage.
+      # The queue self-binds to the fanout DLX: on first TTL the message
+      # is re-queued with an x-death entry; on the second TTL the cycle
+      # is detected and every destination is skipped. Without the fix
+      # the callback is never enqueued and the message leaks, leaving
+      # message_count > 0 after the sleep.
+      it "is called once when all dead letter destinations are cycles" do
+        with_amqp_server do |s|
+          with_channel(s) do |ch|
+            ch.exchange_declare("dlx_fanout_cycle", "fanout")
+            q = ch.queue("q_fanout_cycle", args: AMQP::Client::Arguments.new({
+              "x-message-ttl"          => 1,
+              "x-dead-letter-exchange" => "dlx_fanout_cycle",
+            }))
+            q.bind("dlx_fanout_cycle", "")
+
+            2.times { |i| ch.default_exchange.publish_confirm("msg#{i + 1}", q.name) }
+
+            sleep 0.1.seconds
+
+            q.message_count.should eq 0
+          end
+        end
+      end
+    end
   end
 end
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -577,6 +577,59 @@ module DeadLetteringSpec
         end
       end
 
+      it "should not stack overflow with two mutually dead-lettering queues" do
+        with_amqp_server do |s|
+          with_channel(s) do |ch|
+            n = 10_000
+            q1 = ch.queue("dlx_mutual_q1", args: AMQP::Client::Arguments.new({
+              "x-max-length"              => n,
+              "x-dead-letter-exchange"    => "",
+              "x-dead-letter-routing-key" => "dlx_mutual_q2",
+            }))
+            q2 = ch.queue("dlx_mutual_q2", args: AMQP::Client::Arguments.new({
+              "x-max-length"              => n,
+              "x-dead-letter-exchange"    => "",
+              "x-dead-letter-routing-key" => "dlx_mutual_q1",
+            }))
+            n.times { |i| ch.default_exchange.publish("q1-#{i}", q1.name) }
+            n.times { |i| ch.default_exchange.publish("q2-#{i}", q2.name) }
+            wait_for { q1.message_count == n && q2.message_count == n }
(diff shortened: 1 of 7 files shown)


src/lavinmq/amqp/argument/dead_lettering.cr
```diff
@@ -12,9 +12,25 @@ module LavinMQ::AMQP
         add_argument_validator "x-dead-letter-routing-key", ArgumentValidator::DeadLetteringValidator.new
       end
 
+      alias MessageRoutedCallback = Proc(Nil)
+      alias Task = {AMQP::Queue, Message} | MessageRoutedCallback
+
+      struct Context
```
viktorerlingsson: I think `Context` is a bit vague of a name. How about renaming this to something more descriptive, like `TaskQueue` or something like that?

spec/queue_dead_lettering_spec.cr
```diff
@@ -940,5 +993,67 @@ module DeadLetteringSpec
...
+            ch.exchange_declare("dlx_fanout_cycle", "fanout")
+            q = ch.queue("q_fanout_cycle", args: AMQP::Client::Arguments.new({
+              "x-message-ttl"          => 1,
+              "x-dead-letter-exchange" => "dlx_fanout_cycle",
+            }))
+            q.bind("dlx_fanout_cycle", "")
+
+            2.times { |i| ch.default_exchange.publish_confirm("msg#{i + 1}", q.name) }
+
+            sleep 0.1.seconds
```
viktorerlingsson, line 1051: can we use `wait_for` instead of `sleep` here?

viktorerlingsson:
I've ran a bunch of tests against both this branch and main, and everything seems to be working as expected 👍 

## Test report

  ### The bug this PR fixes

  | Scenario | `main` | This PR |
  |---|---|---|
  | Two queues dead-lettering into each other, triggered to cascade (5000 msgs/queue) | **SEGFAULT** (stack overflow) | PASS,  final counts correct |
  | 100-queue linear DLX chain, trigger cascades end-to-end | Messages don't reach the tail (chain_q99=2) | PASS, full chain  completes in 26 ms |

  ### Existing DLX behaviour is preserved

  Identical results on both branches:

  - `max-length` overflow dead-letters the head (200 msgs, limit 100 → 100 in src, 100 in DLQ)
  - `max-length-bytes` overflow dead-letters by bytes (200×256 B, 10 kB cap → 35 in src, 165 in DLQ)
  - TTL expiry dead-letters (500 msgs, ttl=100 ms → 0 in src, 500 in DLQ)
  - Immediate expiry, TTL=0 (5000 msgs → 0 in src, 5000 in DLQ)
  - Fanout DLX routes to every bound queue (50 msgs, limit 10 → 10 in src, 40 in each of 2 DLQs)
  - Cycle detection still drops self-DLX loops to zero
  - Persistent dead-lettered messages survive a server restart with correct counts (500 msgs → 100 in src, 400 in DLQ after
  restart)

  ### Throughput

  No measurable regression in release mode. 100k 64 B messages, publisher confirms:

  | Scenario | `main` | This PR |
  |---|---|---|
  | Publish with `max-length=1000` + DLX (steady overflow) | 99,206 msg/s | 99,304 msg/s |
  | Publish with `max-length=1000`, no DLX (drop-head overflow) | 49,751 msg/s | 49,825 msg/s |

### Stress / concurrency (PR branch)

  Each scenario is a stress variant of the bug-reproduction cases above, designed to surface any race or memory issue that a single-pair/single-chain test wouldn't:

  - **Parallel mutual DLX** — 5 independent mutual-DLX queue pairs, all triggered at the same instant. Exercises the trampoline under concurrent cascades on different queues. All pairs settle to correct counts.
  - **Parallel DLX chains** — 5 independent 10-queue chains, all triggered at the same instant. Like the 100-queue chain test, but exercising multiple cascades running through the trampoline in parallel. All chains complete.
  - **Parallel overflow** — 5 queues with `max-length` + DLX, all receiving high-rate publishes in parallel. Exercises `drop_overflow` + dead-letter routing under simultaneous load across queues. No crash, all src/DLQ counts match expectations.
  - **Sustained load with a consumer** — single queue, 10 s of concurrent publishers and a consumer draining the DLQ. Exercises the interaction of publishing, overflow dead-lettering, and consumer delivery under steady pressure. Server stable, no message loss or corruption.

All tests pass.

src/lavinmq/amqp/argument/dead_lettering.cr
snichme: Out of curiosity, can `Context` or what we decide to name it instead be a `alias Context = Deque(Task)`?

src/lavinmq/amqp/argument/dead_lettering.cr
```diff
@@ -69,18 +87,47 @@ module LavinMQ::AMQP
...
+        end
+
+        private def drain_context
+          while task = @context.shift?
+            case task
+            when MessageRoutedCallback
+              begin
+                task.call
+              rescue ex
+                @log.warn(exception: ex) { "Unexpected error in dead letter routed callback" }
```
snichme: These should probably be `@log.error` since they are unexpected.

src/lavinmq/amqp/argument/dead_lettering.cr
```diff
@@ -69,18 +87,47 @@ module LavinMQ::AMQP
...
+              end
+            else
+              dst_q, msg = task
+              begin
+                dst_q.publish_internal(msg, dlx_context: @context)
+              rescue Queue::RejectOverFlow
+                # noop
+              rescue ex
+                @log.warn(exception: ex) { "Unexpected error when dead lettering to #{dst_q.name}, messages dropped" }
```
snichme: These should probably be `@log.error` since they are unexpected.

snichme: And is it possible to type the `ex` or is this just a catch all just to be safe?

snichme (commented):
This solution looks much cleaner than the other. Nicely done.

src/lavinmq/amqp/argument/dead_lettering.cr
spuun: Yeah, or change the interface of it. When it was introduced I thought that it maybe would include more stuff.

Rename the methods to `push_task` and `shift_task?`. Or maybe just go with @snichme's approach and use an alias for `Deque(Task)`.

spec/queue_dead_lettering_spec.cr
spuun: Probably. I guess claude used `wait_for` since it's used in other places in that file.

src/lavinmq/amqp/argument/dead_lettering.cr
spuun: True!

src/lavinmq/amqp/argument/dead_lettering.cr
spuun: I want to catch all. I reasoned that we want the dead letter to continue even if some queues we're dead lettering too are in a bad state.

src/lavinmq/amqp/argument/dead_lettering.cr
viktorerlingsson: Yeah maybe just go with the alias!

src/lavinmq/amqp/argument/dead_lettering.cr
snichme: Yes, makes sense. can we type it to `RuntimeError` or whatever the "highest" type is, just to show that we haven't forgotten about setting a type but we still catch what we want :)

spec/queue_dead_lettering_spec.cr
spuun: I'll go with sleep for now, since it's used a lot in that spec file. We can reconsider this and refactor in another PR.

spuun:
> Out of curiosity, can Context or what we decide to name it instead be a alias Context = Deque(Task)?

Yes, IIRC I made the Context a Deque at first, but wanted something more generic with a more strict interface.

snichme (approved):

PR #1846, opened 2026-04-09
Fix flaky etcd spec caused by duplicate ports

viktorerlingsson:
Use `Array#sample` instead of repeated `rand` calls to guarantee unique port numbers in `EtcdCluster#initialize`, preventing random collisions that caused bind failures.

### DIFF
--- a/spec/etcd_spec.cr
+++ b/spec/etcd_spec.cr
@@ -184,7 +184,7 @@ class EtcdCluster
   @ports : Array(Int32)
 
   def initialize(nodes = 3)
-    @ports = (100..998).to_a.sample(nodes)
+    @ports = (100..999).sample(nodes)
   end
 
   def endpoints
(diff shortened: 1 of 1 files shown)


spec/etcd_spec.cr
```diff
@@ -184,7 +184,7 @@ class EtcdCluster
   @ports : Array(Int32)
 
   def initialize(nodes = 3)
-    @ports = nodes.times.map { rand(899) + 100 }.to_a
+    @ports = (100..998).to_a.sample(nodes)
```
spuun: I think you can use sample on a range
```suggestion
    @ports = (100..998).sample(nodes)
```

Any reason you didn't include 999? :)

spuun (approved):

snichme (approved):

spec/etcd_spec.cr
viktorerlingsson: ah, changed to sample on range instead! 

> Any reason you didn't include 999? :)

No, I think Claude just made that up 😬

PR #1847, opened 2026-04-13
Fix SNI config parsing wildcard host corruption

kickster97:
Fix SNI config parsing overwriting wildcard host when exact-match host is parsed after it. 

`parse_sni` used `get_host` which includes wildcard resolution, so a section like `test.example.com` parsed after `*.example.com` would match the wildcard instead of creating a new host. The wildcard's cert got overwritten and the exact-match host was never registered. Adds `get_exact_host` to `SNIManager` that only does direct key lookup, and uses it in `parse_sni`.

Added spec

### DIFF
--- a/spec/sni_spec.cr
+++ b/spec/sni_spec.cr
@@ -233,7 +233,7 @@ describe LavinMQ::Config do
       wildcard.tls_cert.should eq("spec/resources/wildcard_example_certificate.pem")
 
       # Exact-match host is registered separately
-      exact = config.sni_manager.get_exact_host("test.example.com").not_nil!
+      exact = config.sni_manager.get_exact_host("test.example.com").should_not be_nil
       exact.tls_cert.should eq("spec/resources/foobar_localhost_certificate.pem")
 
       # Runtime lookup: exact match takes precedence over wildcard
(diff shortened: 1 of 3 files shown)


spec/sni_spec.cr
```diff
@@ -188,6 +204,45 @@ describe LavinMQ::Config do
...
+    INI
+
+    config_file = File.tempname("lavinmq", ".ini")
+    File.write(config_file, config_content)
+
+    begin
+      config.parse(["-c", config_file])
+
+      # Wildcard host keeps its own cert
+      wildcard = config.sni_manager.get_exact_host("*.example.com").not_nil!
```
spuun:
```suggestion
      wildcard = config.sni_manager.get_exact_host("*.example.com").should_not be_nil
```

spec/sni_spec.cr
```diff
@@ -188,6 +204,45 @@ describe LavinMQ::Config do
...
+      wildcard.tls_cert.should eq("spec/resources/wildcard_example_certificate.pem")
+
+      # Exact-match host is registered separately
+      exact = config.sni_manager.get_exact_host("test.example.com").not_nil!
```
spuun:
```suggestion
      exact = config.sni_manager.get_exact_host("test.example.com").should_not be_nil
```

spuun (approved):

snichme (approved):

PR #1848, opened 2026-04-14
Styling updates for tables

bengtmagnus:
Adds styling to queue table rows that has class "internal".

Removes the even/odd table styling and adds a hover for table rows.

New styling for table head.

Add an exchange with the delayed option checked and look at your queues.

Look at tables in general.

### DIFF
--- a/spec/frontend/queues.spec.js
+++ b/spec/frontend/queues.spec.js
@@ -75,12 +75,12 @@ test.describe("queues", _ => {
 
   test.describe('pagination', _ => {
     test('is visible for page_count=10', async ({ page }) => {
-      await expect(page.locator('#table .pagination .page-item')).toContainText(['Previous', 'Next'], { timeout: 10 })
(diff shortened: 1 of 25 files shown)


spuun:
What about adding the same hover coloring (at least background color change) to the pagination links?

bengtmagnus:
> What about adding the same hover coloring (at least background color change) to the pagination links?

Yes absolutely! I fix

bengtmagnus:
@spuun I think we should add some like class ".current" on the current pagination link and style it?

spuun:
> @spuun I think we should add some like class ".current" on the current pagination link and style it?

Actually, it has `active`. I found a bug however so it's not always applied.

spuun:
I like the internal queue styling, but I'd like the text to be a little bit "grayed out" or something. I tried putting opacity (dunno what, maybe 0.7) and I think that looked good.

We should also display just an `I` in the feature column.

spuun (approved):

PR #1849, opened 2026-04-14
lavinmqctl reset — offline command to wipe node data and cluster identity

### LINKED ISSUE #1797: lavinmqctl reset state
Would be nice to have a command to wipe the data directory and reset LavinMQ to a "virgin" state. rabbitmqctl has this.

annieblomgren:
Closes #1797.

Adds a `lavinmqctl reset` command that places a node back in a virgin state.

## What it does

- Removes all configured users and vhosts (`users.json`, `vhosts.json`, `parameters.json`)
- Deletes all persistent messages by removing the vhost data directories
- Removes the node's cluster identity (`.clustering_id`) so it rejoins any cluster as a fresh node on next startup

## How it works

The command operates directly on the data directory — no running server required. The data directory is resolved in priority order:

1. `LAVINMQ_DATADIR` environment variable
2. `data_dir` from the `[main]` section of the config file (`/etc/lavinmq/lavinmq.ini` by default, or the directory set by `LAVINMQ_CONFIGURATION_DIRECTORY`)
3. Default: `/var/lib/lavinmq`

Before wiping anything, the command:

1. Asks for confirmation — prompts `Reset node at <data-dir>? All messages and metadata will be permanently deleted. [y/N]`. Use `--force` to skip.
2. Checks that LavinMQ is not running by attempting a non-blocking exclusive flock on the `.lock` file. If the lock is held, the command raises with a clear message:

```
LavinMQ is running. Please stop it first with 'lavinmqctl stop_app'.
```

The flock is held for the entire duration of the reset so LavinMQ cannot start mid-wipe.

## Usage

```
lavinmqctl reset
lavinmqctl reset --force
```

## Cluster behaviour

Deleting `.clustering_id` is sufficient to remove a node from a cluster. On restart the node is assigned a new cluster ID; the leader automatically updates the ISR set when the node goes offline (etcd lease expires on stop), so no manual cluster reconfiguration is needed.

## Testing

Three new specs cover:
- data directory is fully wiped (users, vhosts, clustering ID, message directories), resolving data dir via `LAVINMQ_DATADIR`
- data dir is correctly read from a config file via `LAVINMQ_CONFIGURATION_DIRECTORY`
- command raises if LavinMQ is already running (lock held)

### DIFF
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -864,14 +864,12 @@ class LavinMQCtl
     end
 
     lock_path = File.join(data_dir, ".lock")
-    lock = File.exists?(lock_path) ? File.open(lock_path, "r") : nil
-    if lock
-      begin
-        lock.flock_exclusive(blocking: false)
-      rescue IO::Error
-        lock.close
-        raise Exception.new("LavinMQ is running. Please stop it first with 'lavinmqctl stop_app'.")
-      end
+    lock = File.open(lock_path, "a+")
+    begin
+      lock.flock_exclusive(blocking: false)
+    rescue IO::Error
+      lock.close
+      raise Exception.new("LavinMQ is running. Please stop it first with 'lavinmqctl stop_app'.")
     end
 
     @io.puts "Resetting node ..." unless quiet?
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -7,6 +7,8 @@ require "../lavinmq/shovel/constants"
 require "../lavinmq/federation/constants"
 require "../lavinmq/definitions_generator"
 require "../lavinmq/auth/user"
+require "file_utils"
+require "ini"
 
 class LavinMQCtl
   @options = {} of String => String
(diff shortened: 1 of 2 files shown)


spuun:
I think it would be nice to ask the user for confirmation about `resetting <data dir>`. Maybe add an option to skip that.

annieblomgren:
> I think it would be nice to ask the user for confirmation about `resetting <data dir>`. Maybe add an option to skip that.

Sounds good @spuun . I added the requirement that lavinmq must be stopped first, as a safe guard for this (it's what a similar broker is doing), but we could be even more cautious here and do a confirm as well.

src/lavinmqctl/cli.cr
```diff
@@ -843,6 +853,57 @@ class LavinMQCtl
...
+    end
+    File.delete?(File.join(data_dir, "users.json"))
+    File.delete?(File.join(data_dir, "vhosts.json"))
+    File.delete?(File.join(data_dir, "parameters.json"))
+    File.delete?(File.join(data_dir, ".clustering_id"))
+  ensure
+    lock.try(&.close)
+  end
+
+  private def resolve_data_dir : String
```
kickster97, line 888: This only checks `LAVINMQ_DATADIR` and the ini under `LAVINMQ_CONFIGURATION_DIRECTORY` / `/etc/lavinmq`. It ignores `-c/--config=PATH`, so `lavinmq -c /opt/my.ini` + `lavinmqctl reset` falls through to the default and can wipe the wrong directory. 

We could add a `-c`to the `reset`cammond and check it here to avoid that :)

PR #1851, opened 2026-04-15
refactor: Replace per-feature localStorage keys with unified stateClasses manager

spuun:

- Introduces a stateClasses singleton in helpers.js that manages "stateful" HTML classes (theme and menu state), in a single lmq.stateclasses localStorage key. These classes are applied with a simple IIFE to prevent flickering.
- Rewrites ThemeSwitcher to use stateClasses instead of a separate theme key, and to react to OS color scheme changes via matchMedia
- Clean up some theme css
- Make login page follow system color scheme

Supersedes parts of #1698, the other part will come in another PR.

Some specs, but manually testing involves:
- Menu collapsed state is kept on page change
- Theme is kept on page change
- If system theme is selected, the right theme is rendered and kept on page change


kickster97 (approved):
clicked around a bit and it worked well

spec/frontend/menu.spec.js
```diff
@@ -2,118 +2,6 @@ import * as helpers from './helpers.js'
...
-  })
-
-  test('users menu item has active class on users page', async ({ page }) => {
-    await page.goto('/users')
-    const activeItems = await page.locator('#menu-content .active')
-    const activeItem = await activeItems.first()
-    await expect(activeItems).toHaveCount(1)
-    await expect(activeItem).toHaveText('Users')
-  })
-
```
viktorerlingsson, line 116: what's the reasoning for removing all of these specs? maybe add something to the PR description/commit message about it

static/main.css
```diff
@@ -72,6 +72,26 @@
 
 :root.menu-collapsed { --sidebar-width: 68px; }
 
+/* Theme */
+html.theme-light {
+  color-scheme: light !important;
+  .dark {
```
viktorerlingsson, line 78: maybe we should scope the .dark/.light to something like `.login-logo.light .theme-icon.light` so we don't accidentally hide something with .dark/.light later?

spec/frontend/menu.spec.js
spuun: Hm, I think this is some sloppy work from me separating PRs. I think I should keep them in this, but they'll be removed in another.

static/main.css
spuun: `dark` and `light` are generic classes that should be used for hiding/showing elements depending theme. 

Before this PR it they were applied to some elements but not added to the css. Instead there was more specific styles for elements to show and hide them depending on theme (logo and theme buttons). This is a clean up so the element specific code is gone and the this more generic is used.

viktorerlingsson: okay!

viktorerlingsson (approved):

PR #1830, opened 2026-03-26
Group lavinmqctl -h commands by category

annieblomgren:
### WHAT is this pull request doing?
the list of lavinmqctl commands are growing and listing them with -h starts to look messy.
This will group lavinmqctl commands according to their usage:

```
  Global options
      -p vhost, --vhost=vhost          Specify vhost
      -H URI, --host=URI               Specify URI (Deprecated, use --uri or --hostname)
      -U URI, --uri=URI                Specify URI
      --hostname=hostname              Specify hostname
      --user=user                      Specify user
      --password=password              Specify password
      -P port, --port=port             Specify port (15672)
      --scheme=scheme                  Specify scheme (http)
      -n node, --node=node             Specify node
      -q, --quiet                      suppress informational messages
      -s, --silent                     suppress informational messages and table formatting
      -f format, --format=format       Format output (text, json)
      -v, --version                    Show version
      --build-info                     Show build information
      -h, --help                       Show this help

  User Management
      add_user                         Creates a new user
      delete_user                      Delete a user
      list_users                       List user names and tags
      change_password                  Change the user password
      set_user_tags                    Sets user tags
      set_permissions                  Set permissions for a user
      hash_password                    Hash a password

  Virtual Hosts
      add_vhost                        Creates a virtual host
      delete_vhost                     Deletes a virtual host
      list_vhosts                      Lists virtual hosts
      set_vhost_limits                 Set VHost limits (max-connections, max-queues)

  Queues
      list_queues                      Lists queues and their properties
      purge_queue                      Purges a queue (removes all messages in it)
      pause_queue                      Pause all consumers on a queue
      resume_queue                     Resume all consumers on a queue
      restart_queue                    Restarts a closed queue
      delete_queue                     Delete queue
      create_queue                     Create queue

  Exchanges
      list_exchanges                   Lists exchanges
      delete_exchange                  Delete exchange
      create_exchange                  Create exchange

  Policies
      list_policies                    Lists all policies in a virtual host
      clear_policy                     Clears (removes) a policy
      set_policy                       Sets or updates a policy

  Connections
      list_connections                 Lists AMQP 0.9.1 connections for the node
      close_connection                 Instructs the broker to close a connection by pid
      close_all_connections            Instructs the broker to close all connections for the specified vhost or entire node

  Definitions
      export_definitions               Exports definitions in JSON
      import_definitions               Import definitions in JSON
      definitions                      Generate definitions json from a data directory

  Shovels
      list_shovels                     Lists shovels
[...]

PR #1852, opened 2026-04-15
Use annotations for lavinmqctl command registration

kickster97:
Replace command-group constants with `@[Cmd]` annotations on handler methods. A macro generates the `OptionParser` registrations grouped by section and the case dispatch in `run_cmd`. Follows the same pattern used in `config/options.cr` with `@[CliOpt]`/`@[IniOpt]`/`@[EnvOpt]`.

TBA if we like this approach

### DIFF
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -36,35 +36,38 @@ class LavinMQCtl
   def parse_cmd
     {% begin %}
       {% for section in SECTIONS %}
+        {%
+          methods = @type.methods.select(&.annotation(Cmd))
+            .select { |m| m.annotation(Cmd)[:section] == section }
+            .sort_by(&.name)
+        %}
         @parser.separator({{"\n" + section}})
-        {% for method in @type.methods.select(&.annotation(Cmd)).sort_by(&.name) %}
+        {% for method in methods %}
           {% cmd = method.annotation(Cmd) %}
-          {% if cmd[:section] == section %}
-            {% name = method.name.stringify %}
-            {% description = cmd[0] %}
-            {% usage = cmd[1] %}
-            @parser.on({{name}}, {{description}}) do
-              @cmd = ->{ {{method.name.id}}; nil }
-              self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{usage.id}}"
-              {% for opt in method.annotations(Opt) %}
-                {%
-                  flag = opt[0]
-                  desc = opt[1]
-                  options_key = opt[:options]
-                  args_key = opt[:args]
-                  literal = opt[:value]
-                  value = literal || (opt[:coerce] == :i64 ? "v.to_i64".id : "v".id)
-                %}
-                @parser.on({{flag}}, {{desc}}) do {% unless literal %}|v|{% end %}
-                  {% if options_key %}
-                    @options[{{options_key}}] = {{value}}
-                  {% else %}
-                    @args[{{args_key}}] = JSON::Any.new({{value}})
-                  {% end %}
-                end
-              {% end %}
-            end
-          {% end %}
+          {% name = method.name.stringify %}
+          {% description = cmd[0] %}
+          {% usage = cmd[1] %}
+          @parser.on({{name}}, {{description}}) do
+            @cmd = ->{ {{method.name.id}}; nil }
+            self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{usage.id}}"
+            {% for opt in method.annotations(Opt) %}
+              {%
+                flag = opt[0]
+                desc = opt[1]
+                options_key = opt[:options]
+                args_key = opt[:args]
+                literal = opt[:value]
+                value = literal || (opt[:coerce] == :i64 ? "v.to_i64".id : "v".id)
+              %}
+              @parser.on({{flag}}, {{desc}}) do {% unless literal %}|v|{% end %}
+                {% if options_key %}
+                  @options[{{options_key}}] = {{value}}
+                {% else %}
+                  @args[{{args_key}}] = JSON::Any.new({{value}})
+                {% end %}
+              end
+            {% end %}
+          end
         {% end %}
       {% end %}
     {% end %}
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -36,7 +36,7 @@ class LavinMQCtl
     {% begin %}
       {% for section in SECTIONS %}
         @parser.separator({{"\n" + section}})
-        {% for method in @type.methods.select(&.annotation(Cmd)) %}
+        {% for method in @type.methods.select(&.annotation(Cmd)).sort_by(&.name) %}
           {% ann = method.annotation(Cmd) %}
           {% if ann[:section] == section %}
             @parser.on({{method.name.stringify}}, {{ann[0]}}) do
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -18,6 +18,7 @@ class LavinMQCtl
   @io : IO
 
   annotation Cmd; end
+  annotation Opt; end
 
   SECTIONS = {"User Management", "Virtual Hosts", "Queues", "Exchanges",
               "Policies", "Connections", "Definitions", "Shovels",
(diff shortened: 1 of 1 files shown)


src/lavinmqctl/cli.cr
```diff
@@ -88,250 +32,181 @@ class LavinMQCtl
...
+        @parser.separator({{"\n" + section}})
+        {% for method in @type.methods.select(&.annotation(Cmd)) %}
+          {% ann = method.annotation(Cmd) %}
+          {% if ann[:section] == section %}
+            @parser.on({{method.name.stringify}}, {{ann[0]}}) do
+              @cmd = {{method.name.stringify}}
+              self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{ann[1].id}}"
+              {% if ann[:has_options] %}
+                parse_{{method.name.id}}_options
+              {% end %}
```
viktorerlingsson: we could even drop `has_options` and just use 

```suggestion
              {% if @type.methods.any? { |m| m.name == "parse_#{method.name}_options" } %}
                parse_{{method.name.id}}_options
              {% end %}
```

viktorerlingsson:
I like this approach! Maybe we can even add the options from each parse_x_options methods as `@[Opt]` annotations for those that need it, parse that in `parse_cmd`, and completely skip the parse_x_options methods 🤔

src/lavinmqctl/cli.cr
spuun: Or rename `has_options` to just `options` and let the value be the method, which I think makes it explicit.

Then the code can probably be just `{{ ann[:options] }}` if the annotation looks like `@[Cmd(…, options: parse_add_user_options)]` (i.e. no condition needed because missing `options` will result in nothing).

src/lavinmqctl/cli.cr
```diff
@@ -88,250 +32,181 @@ class LavinMQCtl
...
-        self.banner = "Usage: #{PROGRAM_NAME} #{cmd[0]} #{cmd[2]}"
-      end
-    end
+  def parse_cmd
+    {% begin %}
+      {% for section in SECTIONS %}
+        @parser.separator({{"\n" + section}})
+        {% for method in @type.methods.select(&.annotation(Cmd)) %}
+          {% ann = method.annotation(Cmd) %}
+          {% if ann[:section] == section %}
```
spuun: One risk is that the methods will change order on a new build, because of what ever reason. Maybe we should sort them?

src/lavinmqctl/cli.cr
```diff
@@ -88,250 +32,181 @@ class LavinMQCtl
...
+            @parser.on({{method.name.stringify}}, {{ann[0]}}) do
+              @cmd = {{method.name.stringify}}
```
spuun: You can make `@cmd` a `Proc(Nil)?` and do `@cmd = ->{{method.name}}` here. Then you don't need the macro in `#run_cmd`, instead you can just do
```crystal
  @parser.parse
  if c = @cmd
    c.call
  else
    @io.puts @parser
    abort
  end
```

kickster97:
> I like this approach! Maybe we can even add the options from each parse_x_options methods as `@[Opt]` annotations for those that need it, parse that in `parse_cmd`, and completely skip the parse_x_options methods 🤔

Good thinking, I implemented this. All options are now annotations above each command method. Now, adding a flag of an existing type is very simple.. you just add an annotation line.. but a thing worth noting is that if we want something that requires a new type we need to edit the macro to handle it.

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,50 @@ class LavinMQCtl
...
+      {% for section in SECTIONS %}
+        @parser.separator({{"\n" + section}})
+        {% for method in @type.methods.select(&.annotation(Cmd)).sort_by(&.name) %}
+          {% cmd = method.annotation(Cmd) %}
+          {% if cmd[:section] == section %}
+            {% name = method.name.stringify %}
+            {% description = cmd[0] %}
+            {% usage = cmd[1] %}
+            @parser.on({{name}}, {{description}}) do
+              @cmd = ->{ {{method.name.id}}; nil }
```
spuun: Is that `nil` really needed?

Note: you can create a "pointer" directly to the method if you make the methods protected (or public):
```suggestion
              @cmd = ->self.{{method.name.id}}
```

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,50 @@ class LavinMQCtl
...
-    @parser.on("cluster_status", "Display cluster status") do
-      @cmd = "cluster_status"
-    end
-
+    {% begin %}
+      {% for section in SECTIONS %}
+        @parser.separator({{"\n" + section}})
+        {% for method in @type.methods.select(&.annotation(Cmd)).sort_by(&.name) %}
+          {% cmd = method.annotation(Cmd) %}
+          {% if cmd[:section] == section %}
```
spuun: To get rid of that if-statement and one level of nesting you could filter before the loop:
```suggestion
        {%
          methods = @type.methods.select(&.annotation(Cmd))
            .select(&.annotation(Cmd)[:section].==(section))
            .sort_by(&.name)
        %}
        @parser.separator({{"\n" + section}})
        {% for method in methods %}
          {% cmd = method.annotation(Cmd) %}
```

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,50 @@ class LavinMQCtl
...
+            @parser.on({{name}}, {{description}}) do
+              @cmd = ->{ {{method.name.id}}; nil }
+              self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{usage.id}}"
+              {% for opt in method.annotations(Opt) %}
+                {% flag = opt[0] %}
+                {% desc = opt[1] %}
+                {% options_key = opt[:options] %}
+                {% args_key = opt[:args] %}
+                {% literal = opt[:value] %}
+                {% value = literal || (opt[:coerce] == :i64 ? "v.to_i64".id : "v".id) %}
```
spuun: Are both `:value` and `:coerce` needed? Wouldn't it be possible to go with `value` only and do
```crystal
@[Opt(value: "literal value") # literal value
@[Opt(value: v.to_i64)) #convert to i64
```
and then
```suggestion
                {% value = opt[:value] || "v".id %}
```

(and remove the `unless literal` condition below)

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,50 @@ class LavinMQCtl
...
+            {% usage = cmd[1] %}
+            @parser.on({{name}}, {{description}}) do
+              @cmd = ->{ {{method.name.id}}; nil }
+              self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{usage.id}}"
+              {% for opt in method.annotations(Opt) %}
+                {% flag = opt[0] %}
+                {% desc = opt[1] %}
+                {% options_key = opt[:options] %}
+                {% args_key = opt[:args] %}
+                {% literal = opt[:value] %}
```
spuun: I guess It's a matter of style, but you can get rid of "macro tags" 
```suggestion
                {% 
                 flag = opt[0]
                 desc = opt[1]
                 options_key = opt[:options]
                 args_key = opt[:args]
                 literal = opt[:value] 
                %}
```

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,50 @@ class LavinMQCtl
...
+        {% for method in @type.methods.select(&.annotation(Cmd)).sort_by(&.name) %}
+          {% cmd = method.annotation(Cmd) %}
+          {% if cmd[:section] == section %}
+            {% name = method.name.stringify %}
+            {% description = cmd[0] %}
+            {% usage = cmd[1] %}
+            @parser.on({{name}}, {{description}}) do
+              @cmd = ->{ {{method.name.id}}; nil }
+              self.banner = "Usage: #{PROGRAM_NAME} {{method.name.id}} {{usage.id}}"
+              {% for opt in method.annotations(Opt) %}
```
spuun: Sort these too?
```suggestion
              {% for opt in method.annotations(Opt).sort_by(&.[0]) %}
```

spuun (commented):
A few thoughts.

src/lavinmqctl/cli.cr
kickster97: ah yeah much nicer!

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,54 @@ class LavinMQCtl
...
+          methods = @type.methods.select(&.annotation(Cmd))
+            .select { |m| m.annotation(Cmd)[:section] == section }
+            .sort_by(&.name)
+        %}
+        @parser.separator({{"\n" + section}})
+        {% for method in methods %}
+          {% cmd = method.annotation(Cmd) %}
+          {% name = method.name.stringify %}
+          {% description = cmd[0] %}
+          {% usage = cmd[1] %}
```
spuun:
```suggestion
          {% 
            cmd = method.annotation(Cmd)
            name = method.name.stringify
            description = cmd[0]
            usage = cmd[1] 
          %}
```

src/lavinmqctl/cli.cr
```diff
@@ -88,246 +33,54 @@ class LavinMQCtl
...
-      @cmd = "cluster_status"
-    end
-
+    {% begin %}
+      {% for section in SECTIONS %}
+        {%
+          methods = @type.methods.select(&.annotation(Cmd))
+            .select { |m| m.annotation(Cmd)[:section] == section }
+            .sort_by(&.name)
+        %}
```
spuun: Maybe not for this PR, and nothing important, but it would be nice to validate that all cmds has valid sections. Now it's possible to use an invalid section and the cmd won't be available. Maybe we also should validate that a section actually has methods in it.

spuun: Something like this will generate nice compile time errors
```crystal
    {% begin %}
      {%
        methods_by_section = {} of String => Object
        SECTIONS.each { |s| methods_by_section[s] = [] of Object }
        methods = @type.methods.select(&.annotation(Cmd)).each do |m|
          cmd = m.annotation(Cmd)
          if !methods_by_section.has_key?(cmd[:section])
            cmd[:section].raise "Invalid section #{cmd[:section]}"
          end
          methods_by_section[cmd[:section]] << m
        end
      %}
      {% for section, methods in methods_by_section %}
        {% section.raise "Section #{section} has no methods" if methods.empty? %}
        @parser.separator({{"\n" + section}})
        {% for method in methods %}
          … # 
        {% end %}
      {% end %}
    {% end %}
```

spuun (commented):
WDYT about adding compile-time checks that sections has cmds and cmds belongs to a section?

spuun:
#1830 got merged, but IMO we should finish this and go with this approach

kickster97:
> #1830 got merged, but IMO we should finish this and go with this approach

I saw, and i agree, i will get to it once i have mgmt-oauth branch at a good state

snichme (approved):

PR #1856, opened 2026-04-17
Fix shovel vhost permission check for named vhosts

ThomasSarlin:
### Description
While looking at some JavaScript in the moveMessage context I was unable to move messages between queues, 
getting `guest can't access exchange '' in /test` as an alert. The reason behind this behavior is that the validator is using uri.path (e.g. "/test") as the vhost key when calling user.can_read?/can_write?/can_config?, but the user’s permission map is keyed by the bare vhost name (e.g. "test"). As a result, a user with full permissions on vhost test would incorrectly be rejected with `can't access queue/exchange ... in /test` etc.

Fixes a permission-check bug in Shovel::Store.validate_config! for shovels targeting a named vhost.

### Changes:
- Add a private vhost_from_uri helper in src/lavinmq/shovel/store.cr that strips the leading / from the URI path, returning "/" for the default vhost.
- Use the helper in both the dest_uris and src_uris loops in place of the inline uri.path / "/" if empty logic.

Specs and manually tried moving messages between queues.

### DIFF
--- a/spec/shovel_spec.cr
+++ b/spec/shovel_spec.cr
@@ -800,4 +800,36 @@ describe LavinMQ::Shovel do
       end
     end
   end
+
+  describe "Store.validate_config!" do
+    it "looks up vhost permissions by bare name (strips leading slash from URI path)" do
+      with_amqp_server do |s|
+        user = s.users.create("shovel_user", "pass")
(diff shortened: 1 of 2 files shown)


spuun (approved):
Nice find!

PR #1857, opened 2026-04-17
ui: Show only `I` as feature for internal queue

spuun:
In #1848 an internal queue will rendered differently in the table to highlight that it's internal. 

I think we also should display only `I` in the feature column for an internal queue, no `D` or `Args`. Since the user haven't configured this queue, the arguments and properties of the queue isn't really of interest.

Create a delayed exchange and check the queue listing

### DIFF
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -87,32 +87,39 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
     }
     const features = document.createElement('span')
     features.className = 'features'
-    if (item.durable) {
+    if (item.internal) {
       const durable = document.createElement('span')
-      durable.textContent = 'D '
-      durable.title = 'Durable'
+      durable.textContent = 'I'
+      durable.title = 'Internal'
       features.appendChild(durable)
-    }
-    if (item.auto_delete) {
-      const autoDelete = document.createElement('span')
-      autoDelete.textContent = 'AD '
-      autoDelete.title = 'Auto Delete'
-      features.appendChild(autoDelete)
-    }
-    if (item.exclusive) {
-      const exclusive = document.createElement('span')
-      exclusive.textContent = 'E '
-      exclusive.title = 'Exclusive'
-      features.appendChild(exclusive)
-    }
-    if (Object.keys(item.arguments).length > 0) {
-      const argsTooltip = Object.entries(item.arguments)
-        .map(([key, value]) => `${key}: ${JSON.stringify(value)}`)
-        .join('\n')
-      const argsSpan = document.createElement('span')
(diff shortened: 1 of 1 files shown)


kickster97 (approved):
Agree!

static/js/queues.js
```diff
@@ -87,32 +87,39 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
     }
     const features = document.createElement('span')
     features.className = 'features'
-    if (item.durable) {
+    if (item.internal) {
       const durable = document.createElement('span')
-      durable.textContent = 'D '
-      durable.title = 'Durable'
+      durable.textContent = 'I'
```
viktorerlingsson: maybe name this var something else than `durable` 🙂

viktorerlingsson:
Not introduced here, but maybe can be cleaned up a bit by adding a 
```
 function addFeature(container, text, title) {
    const span = document.createElement('span')
    span.textContent = text
    span.title = title
    container.appendChild(span)
  }
```

that can be called like 
```
if (item.internal) { addFeature(features, 'I', 'Internal') }
...
if (item.durable)     addFeature(features, 'D ',  'Durable')
```

instead of having to have code for that for every feature. WDYT? (Could be used for `exchanges.js` as well)

static/js/queues.js
spuun: hahaha 🤦

spuun:
> Not introduced here, but maybe can be cleaned up a bit by adding a
> 
> ```
>  function addFeature(container, text, title) {
>     const span = document.createElement('span')
>     span.textContent = text
>     span.title = title
>     container.appendChild(span)
>   }
> ```
> 
> that can be called like
> 
> ```
> if (item.internal) { addFeature(features, 'I', 'Internal') }
> ...
> if (item.durable)     addFeature(features, 'D ',  'Durable')
> ```
> 
> instead of having to have code for that for every feature. WDYT? (Could be used for `exchanges.js` as well)

Yup, we have the same thing in `stream.js`. I think it should be nice to start using web components.

static/js/queues.js
```diff
@@ -87,32 +87,39 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
...
-        .join('\n')
-      const argsSpan = document.createElement('span')
-      argsSpan.textContent = 'Args '
-      argsSpan.title = argsTooltip
-      features.appendChild(argsSpan)
+    if (item.internal) {
+      const internal = document.createElement('span')
+      internal.textContent = 'I'
+      internal.title = 'Internal'
+      internal.appendChild(internal)
```
viktorerlingsson:
```suggestion
      features.appendChild(internal)
```

PR #1858, opened 2026-04-17
bugfix: Replicate .queue file

### LINKED ISSUE #1855: `.queue` file not replicated
The `.queue` file isn't replicated. Even though it would be recreated when a follower is promoted to leader, it should be replicated for easier debugging of a follower data dir.

spuun:
Fixes #1855

Specs

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -177,7 +177,9 @@ module LavinMQ::AMQP
       @log = Logger.new(Log, @metadata)
       dotqueue_file = File.join(@data_dir, ".queue")
       File.open(dotqueue_file, "w") { |f| f.sync = true; f.print @name }
-      @vhost.@replicator.try &.replace_file(dotqueue_file)
+      if durable?
+        @vhost.@replicator.try &.replace_file(dotqueue_file)
+      end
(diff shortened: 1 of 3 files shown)


kickster97 (approved):

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -175,7 +175,9 @@ module LavinMQ::AMQP
       @data_dir = make_data_dir
       @metadata = ::Log::Metadata.new(nil, {queue: @name, vhost: @vhost.name})
       @log = Logger.new(Log, @metadata)
-      File.open(File.join(@data_dir, ".queue"), "w") { |f| f.sync = true; f.print @name }
+      dotqueue_file = File.join(@data_dir, ".queue")
+      File.open(dotqueue_file, "w") { |f| f.sync = true; f.print @name }
+      @vhost.@replicator.try &.replace_file(dotqueue_file)
```
viktorerlingsson: Should only be replicated if queue is durable, I think?

spuun: Yeah, true!
I think I reasoned that we always want it in the directory, but the directory won't be "replicated" as long as we don't replicate any file in it.

viktorerlingsson (approved):

PR #1859, opened 2026-04-17
Fix 404 on /metrics/detailed for follower nodes

walro:

Follower nodes returned 404 for `/metrics/detailed`. Respond 200 with only scrape telemetry instead.

Specs

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -217,19 +217,6 @@ describe LavinMQ::HTTP::PrometheusController do
   end
 end
 
-def with_follower_metrics_server(&)
-  # Passing nil for amqp_server wires up FollowerPrometheusController — the same path a real follower takes.
-  h = LavinMQ::HTTP::MetricsServer.new(nil)
-  begin
-    addr = h.bind_tcp("::1", 0)
-    spawn(name: "follower metrics listen") { h.listen }
-    Fiber.yield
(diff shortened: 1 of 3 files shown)


viktorerlingsson (approved):
Looks safe to me 👍

spec/api/prometheus_spec.cr
```diff
@@ -217,6 +217,40 @@ describe LavinMQ::HTTP::PrometheusController do
   end
 end
 
+def with_follower_metrics_server(&)
```
walro: @viktorerlingsson since you looked already :-)

We could move this into the `spec_helper` file instead (where we have existing `with_metrics_server`)? Or do it when needed at more at once file? I am ok with both ways.

viktorerlingsson: hmm, yeah lets move it to `spec_helper`! Would be kinda nice to also refactor a little so we don't need to duplicate "all that code" in  `with_follower_metrics_server / with_metrics_server` 🤔

walro: [c6cc7bc](https://github.com/cloudamqp/lavinmq/pull/1859/commits/c6cc7bc2a699bce886fb7e9b7356cf81920c7e84) wdyt?

viktorerlingsson: Looks good to me! 👍

Issue #1854, opened 2026-04-16
Orphaned queue directories left on follower nodes after queue deletion


viktorerlingsson:
When a queue is deleted on the leader, individual files are replicated as deletions to followers via `DeleteAction`, but the queue directory itself is never removed. The leader does `FileUtils.rm_rf` on the directory locally, but there's no corresponding replication action for directory removal.

Additionally, `sync_files` cleanup only calls `File.delete` on files not present on the leader, and `ls_r` only yields files, so empty directories are never cleaned up during full sync either.

This means empty queue directories accumulate on follower nodes over time, and persist even if the node is promoted to leader (since queue loading is driven by the definitions file, not filesystem scanning).

PR #1865, opened 2026-04-20
Clean up empty directories on follower after full sync

spuun:

During a full sync, the follower now removes local directories that are absent from the leader, in addition to the existing removal of stale files.
                                                                                                                                                                                                                                                                                        
Previously, when a queue or other directory was deleted on the leader, a full sync would delete the files inside it but leave behind an empty directory on the follower. Over time this could accumulate orphaned empty directories.                                                  
                                                                                                                                                                                                                                                                                        
Changes:                                                                                                                                                                                                                                                                              
  - `#ls_r` now returns both files and directories separately
  - After syncing, directories whose ancestors are all absent from the leader's file list are deleted (deepest paths first to avoid non-empty errors)
  - File and directory deletion errors are caught and logged as warnings rather than aborting the sync                                               
  - Adds specs covering: single dir removal, nested dir tree removal, preserving dirs with leader files, and partial removal when only some dirs are absent 

Ref #1854

Specs

### DIFF
--- a/spec/clustering/client_sync_spec.cr
+++ b/spec/clustering/client_sync_spec.cr
@@ -1,0 +1,164 @@
+require "../spec_helper"
+require "lz4"
+
+module ClientSyncSpec
+  class TestClient < LavinMQ::Clustering::Client
+    def sync_files_public(socket, lz4)
+      sync_files(socket, lz4)
+    end
+  end
+
+  def self.make_client(data_dir : String) : TestClient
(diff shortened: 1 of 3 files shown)


spuun:
Lint and format errors because CI running 1.20.

viktorerlingsson (approved):
Haven't tested it, but I think the code looks good and the specs should cover it

PR #1872, opened 2026-04-22
Append short commit hash to shard version in build info

spuun:
In CI the checkout action is doing a shallow clone so no tags are available and the version string will be made from the fallback `shards version`. Since no hash is available one can't be sure what version that's actually has been built e.g. when the Compile step prints the build info.

This PR appends the short git commit hash to the version string when no tag is available (e.g. 0.x.y+abc1234), making it easier to identify the exact build from shards version fallback.

Manually (if testing is needed)


carlhoerberg:
Or do a full checkout in CI?

carlhoerberg:
```
- uses: actions/checkout@v4
  with:
    fetch-depth: 0
    fetch-tags: true
```

spuun:
That was my first thought, dunno why I didn't go with it. Let's do that instead.

PR #1873, opened 2026-04-22
Remove no longer needed media queries

spuun:
The refactor done by #1851 removed the need of media queries for theming. #1709 was merged after #1851 and included some media query rules for the menu.

This PR removes the unnecessary media query rules for the menu.

Manually verify the menu

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -88,10 +88,6 @@ html.theme-dark {
   }
 }
 
-
-
-
-
 /* === RESET / BASE === */
 @font-face {
   font-family: Inter;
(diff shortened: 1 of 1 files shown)


viktorerlingsson (approved):
LGTM 👍 

Claude said there might be a brief flash where the sidebar icons might not be properly loaded for a first-time visitor where `localStorage` was empty, but I wasn't able to reproduce that.

spuun:
> Claude said there might be a brief flash where the sidebar icons might not be properly loaded for a first-time visitor where `localStorage` was empty, but I wasn't able to reproduce that.

Yes, I think more stuff may flash on the first load but that's a trade off I think we need to live with (or do a bunch of more inline js which I think we want to avoid).

viktorerlingsson:
> > Claude said there might be a brief flash where the sidebar icons might not be properly loaded for a first-time visitor where `localStorage` was empty, but I wasn't able to reproduce that.
> 
> Yes, I think more stuff may flash on the first load but that's a trade off I think we need to live with (or do a bunch of more inline js which I think we want to avoid).

Yeah 🤔. I didn't really notice anything when testing this so I think it's OK.

PR #1876, opened 2026-04-23
fix: update CSP hash for login page inline theme script

viktorerlingsson:
The `login_inline_js_sha` in `views.cr` drifted from the actual inline theme-init script in `views/login.ecr`, so the browser's CSP blocked it and both light/dark logos rendered at once.
image

Updated the hash and added regression coverage at both layers so this class of bug fails CI instead of reaching users:

- Backend: `spec/http_views_spec.cr` extracts every inline `<script>` body from the rendered HTML and asserts its sha256 is listed in the `Content-Security-Policy` header.
- Frontend: `trackCspViolations(page)` helper in `spec/frontend/helpers.js` listens for `securitypolicyviolation` events via `addInitScript`; new tests assert `/login` (unauthed) and `/` (authed) load with zero violations.

### DIFF
--- a/spec/frontend/auth.spec.js
+++ b/spec/frontend/auth.spec.js
@@ -1,11 +1,17 @@
 // @ts-check
-import { test, expect } from '@playwright/test';
+import { test, expect } from '@playwright/test'
+import { trackCspViolations } from './helpers.js'
 
-test.describe('when unauthenticated', _ =>  {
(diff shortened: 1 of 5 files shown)


viktorerlingsson:
Frontend specs will fail until #1875 is merged

spuun (approved):

PR #1880, opened 2026-04-24
feat: simple HTTP publish/consume API (PoC)

viktorerlingsson:
**PoC — opening early to get eyes on the shape of the API before investing more.**

## Motivation

Integrating a non-AMQP system with LavinMQ today requires either an AMQP client or the existing management endpoints (`POST /api/exchanges/.../publish`, `POST /api/queues/.../get`) — which need a JSON envelope with `properties`, `routing_key`, `payload_encoding`, require the `management` tag, and are flagged as testing-only in RabbitMQ's docs.

This adds a dedicated data-plane surface for the simplest case: **raw bytes in, raw bytes out, queue auto-declared, no JSON wrapper**. Target users are shell scripts, webhook receivers, IoT devices — anything where pulling in an AMQP library is disproportionate to the job.

## Endpoints

- `POST /api/simple/:vhost/:name` — publish. Request body is the message payload. Queue is auto-declared (classic, durable) if absent. `201 Created`.
- `POST /api/simple/:vhost/:name/get` — consume one message with auto-ack. `200 OK` with raw bytes on hit, `204 No Content` on empty queue, `404 Not Found` when queue is missing.

POST for the read path because auto-ack is destructive, matching the existing `/api/queues/.../get` convention.

Reuses AMQP regex permissions (`write`/`read`/`configure`); `management` tag not required.

**Publish a message:** 
`curl -u user:pass -X POST -d "a message" http://localhost:15672/api/simple/myvhost/myqueue`

**Get a message:**
`curl -u user:pass -X POST http://localhost:15672/api/simple/myvhost/myqueue/get`

## Open questions

- Auto-created queues currently never expire. Needs an answer before real use — explicit `DELETE`, admin marker, or TTL.
- Streaming request body into publish would avoid a buffer copy but `Exchange#publish` seeks the body IO between deliveries, which HTTP bodies don't support.

### DIFF
--- a/spec/api/simple_http_spec.cr
+++ b/spec/api/simple_http_spec.cr
@@ -1,0 +1,196 @@
+require "../spec_helper"
+
+describe LavinMQ::HTTP::SimpleHttpController do
+  describe "POST /api/simple/:vhost/:name" do
+    it "auto-declares the queue and publishes a raw body" do
+      with_http_server do |http, s|
+        response = http.post("/api/simple/%2f/simple1", body: "hello world")
+        response.status_code.should eq 201
+        q = s.vhosts["/"].queues["simple1"].as(LavinMQ::AMQP::Queue)
+        q.durable?.should be_true
+        q.message_count.should eq 1
+      end
+    end
+
+    it "reuses an existing queue instead of re-declaring" do
+      with_http_server do |http, s|
+        s.vhosts["/"].declare_queue("simple2", true, false)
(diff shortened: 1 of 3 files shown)


lukas8219:
👍🏻  on this. Think we should move forward. Opt for 2 interfaces (SQS and a very simple one)

viktorerlingsson:
Added a discussion https://github.com/cloudamqp/lavinmq/discussions/1889 to get some more input on this before more work is done

PR #1883, opened 2026-04-27
refactor: centralize mgmt authentication logic and add whoami caching

spuun:
Consolidates login, logout, and session validation into auth.js. Introduces a whoAmI() function that caches the API response in localStorage (under lmq.whoami) with a 1-hour TTL, stripping sensitive fields (password_hash, hashing_algorithm) before storage. On API failure, whoAmI automatically calls logout() and redirects to /login.

Adds Playwright specs covering login, logout, and whoAmI behavior (caching, cache miss re-fetch, sensitive field exclusion, and failure redirect).

This PR is parts of #1698 and prepares for the "user tags classes".

Specs, but probably good to do some manual too.

### DIFF
--- a/static/js/auth.js
+++ b/static/js/auth.js
@@ -34,19 +34,8 @@ function logout () {
   window.location.assign('login')
 }
 
-async function whoAmI (forceReload = false) {
-  if (!forceReload) {
-    const data = window.localStorage.getItem('lmq.whoami')
-    let whoAmI = null
-    if (data) {
-      whoAmI = JSON.parse(data)
-    }
-    // Do we have cached and valid info?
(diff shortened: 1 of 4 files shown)


static/js/auth.js
```diff
@@ -20,7 +20,70 @@ function getCookie (key) {
...
+  })
+}
+
+function logout () {
+  window.localStorage.removeItem('lmq.whoami')
+  document.cookie = 'm=; max-age=0'
+  window.location.assign('login')
+}
+
+async function fetchAndCacheWhoAmI () {
```
snichme: Nitpick: I dont like functions with `And` in them since that means it does two things and they should do one thing and they you compose or call them after each other.  
It makes sense here that the `fetch` also caches so that is why I consider this nitpick.

snichme (approved):

PR #1887, opened 2026-04-28
Fix: exclusive queue references leak on auto-delete

kickster97:
`Client#@exclusive_queues` kept strong references to exclusive queues, intended to be cleaned up on connection close. Entries were only removed on the explicit `Queue.Delete` frame path, so when a queue with `auto_delete: true` was deleted server-side after the last consumer cancelled, its reference stayed in the array. 

Long-lived connections that churn exclusive `auto-delete` queues (RPC pattern) accumulated dead Queue references and kept the heap growing until the connection finally dropped. 

Fix is to make `Client` an `Observer` of `QueueEvent` and drop the entry on Deleted. The Queue already fires that event, we just pass `self` so the observer can identify it. The pattern matches `federation/link.cr`
  
  
Ran a Node `amqplib` loop that declares an exclusive `auto-delete` queue, subscribes a `no-ack` consumer, cancels and repeats on one connection. 

Without the fix: 122k cycles, RSS grew past 290 MB and only dropped once the connection closed. 
With the fix: 150k+ cycles with RSS holding flat around 13 MB throughout. Reproduces identically on 2.6.0 and 2.6.9.

### DIFF
--- a/spec/queue_spec.cr
+++ b/spec/queue_spec.cr
@@ -897,7 +897,7 @@ describe LavinMQ::AMQP::Queue do
     end
   end
 
-  describe "Exclusive auto-delete queue cleanup" do
+  describe "Exclusive queue cleanup" do
     it "drops the queue from Client#exclusive_queues when auto-deleted server-side" do
       with_amqp_server do |s|
         with_channel(s) do |ch|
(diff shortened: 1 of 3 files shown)


viktorerlingsson (approved):
Just noting for completeness that the same leak probably happens when an exclusive queue is expired via `x-expires`. But in the end that runs the same code path for deletion so this fix should handle that as well. 

Could possibly add a spec that verifies that?

PR #1888, opened 2026-04-28
test: trigger docs-check workflow

kickster97:
Throwaway PR to verify the docs-check workflow.

Touches a single line in src/ without touching docs/, so the workflow should add the missing-docs label. A follow-up commit that touches docs/ should remove it again.

Do not merge.


kickster97:
Test passed, closing throwaway PR.

PR #1891, opened 2026-04-28
Persist publish confirms to disk before acking

carlhoerberg:
Ensures AMQP publisher confirms (`Basic.Ack`) are only sent to clients **after** the published messages have been flushed to disk with `syncfs(2)`.

Implementation:
- New `Persister`, created once per `Server` and shared by all vhosts. Since every vhost lives on the same filesystem, a single `syncfs` flushes data for all of them in one syscall. `syncfs` ownership moved here from `DefinitionsStore`; `VHost#sync` now delegates to the persister.
- On a confirmable publish, the channel records the latest `msgid` in the persister's `@pending_acks` (a `Hash` keyed by channel, coalescing to the newest `msgid`) and rings a 1-slot "doorbell" channel.
- The publish-confirm loop wakes on the doorbell, snapshots and clears `@pending_acks`, runs **one** `syncfs`, then hands each channel its latest cumulative `msgid`. Batching is adaptive and free: while `syncfs` runs, new requests accumulate and are flushed on the next iteration — no artificial delay, and a lone publish is confirmed as soon as the sync completes.
- The loop runs on a dedicated `Fiber::ExecutionContext::Isolated` thread, so the blocking `syncfs` only stalls that thread instead of stealing a worker thread from client connection handling.
- Each channel sends its `Basic.Ack` from a per-channel writer fiber (started lazily on `confirm.select`) fed by a 1-slot mailbox, so a slow/unresponsive client socket can't block the persister or other channels' confirms. Acks are sent with `multiple: true`.
- A `syncfs` failure logs fatal and `exit 1`: in a clustered deployment, fast leader death lets a follower take over sooner rather than risking acking unpersisted data.

- New specs in `spec/publish_confirm_persistence_spec.cr` cover: confirms still reaching the client, a single publish confirming quickly (no batching delay), mixed ack/nack under `reject-publish` overflow, and batching of many concurrent in-flight publishes.
- Manual verification with `strace -e syncfs,write` to confirm `syncfs` is called before the `Basic.Ack` frames are written.

### DIFF
--- a/src/lavinmq/vhost.cr
+++ b/src/lavinmq/vhost.cr
@@ -18,6 +18,8 @@ require "./mqtt/session"
 require "./connection_store"
 require "./direct_reply_consumer_store"
 require "./definitions_store"
+require "./amqp/channel"
+require "./persister"
 
 module LavinMQ
   class VHost
@@ -165,7 +167,7 @@ module LavinMQ
 
     Log = LavinMQ::Log.for "vhost"
 
-    def initialize(@name : String, @server_data_dir : String, @users : Auth::UserStore, @replicator : Clustering::Replicator?, @description = "", @tags = Array(String).new(0))
+    def initialize(@name : String, @server_data_dir : String, @users : Auth::UserStore, @replicator : Clustering::Replicator?, @persister : Persister, @description = "", @tags = Array(String).new(0))
       @log = Logger.new(Log, vhost: @name)
       @dir = Digest::SHA1.hexdigest(@name)
       @data_dir = File.join(@server_data_dir, @dir)
@@ -196,6 +198,10 @@ module LavinMQ
           end
         end
       end
+    end
+
+    def enqueue_ack(channel : AMQP::Channel, msgid : UInt64)
+      @persister.enqueue_ack(channel, msgid)
     end
 
     def max_connections=(value : Int32) : Nil
@@ -549,7 +555,7 @@ module LavinMQ
     end
 
     def sync : Nil
-      definitions.sync
+      @persister.sync
     end
 
     private def definitions : DefinitionsStore
--- a/spec/publish_confirm_persistence_spec.cr
+++ b/spec/publish_confirm_persistence_spec.cr
@@ -1,0 +1,86 @@
+require "./spec_helper"
+
+describe "Publish Confirm Persistence" do
+  it "confirms are still received by the client" do
+    with_amqp_server do |s|
+      with_channel(s) do |ch|
+        q = ch.queue("confirm_test")
+        10.times do |i|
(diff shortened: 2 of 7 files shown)


src/lavinmq/vhost.cr
```diff
@@ -198,6 +203,69 @@ module LavinMQ
...
+        end
+      end
+      return unless acks
+
+      begin
+        sync
+      rescue ex
+        @log.error(exception: ex) { "Failed to sync: #{ex.message}" }
+        exit 1
+      end
```
snichme: Not follow the flow here. `sync` runs the `definition_store.sync` which indeed does a `syncfs` but we aren't syncing definitions here which makes is wierd to look in the DefinitionStore class for what will happen. For this we can just inline the content of DefinitionStore.sync` here.

`exit 1` feels to much. We dont even know what the problem is here. 
If we cannot guarantee that it's written to disk we should not confirm the publish but killing the whole broker?

carlhoerberg:
True, it's because we have a FD there, but we can open a fd of the vhost's
data dir instead and issue a syncfs on that one instead (same result, but
clearer).

Exit 1 can absoluty be debated, my thinking was that all bets are off if we
cant syncfs. It means that it's some issue with the file system, everything
from full disk to broken hardware. With exit 1 we will failover to a
follower (if such exits)

src/lavinmq/vhost.cr
carlhoerberg: True, it's because we have a FD there, but we can open a fd of the vhost's data dir instead and issue a syncfs on that one instead (same result, but clearer). Exit 1 can absoluty be debated, my thinking was that all bets are off if we cant syncfs. It means that it's some issue with the file system, everything from full disk to broken hardware. With exit 1 we will failover to a follower (if such exits)

src/lavinmq/vhost.cr
```diff
@@ -198,6 +203,69 @@ module LavinMQ
...
+
+      acks.each do |channel, msgid|
+        channel.do_confirm_ack(msgid, multiple: true)
```
spuun: Any risk that one channel may block here?

spuun:
Would it make sense to have a global sync instead of on vhost level?

carlhoerberg:
> Would it make sense to have a global sync instead of on vhost level?

yes

src/lavinmq/vhost.cr
carlhoerberg: yes, for up to 15s (write timeout). can either have a more aggressive write timeout, or an extra "write fiber" per client.

carlhoerberg: fixed, one extra fiber per channel that has enabled publish confirm, i think that's doable

Issue #1896, opened 2026-04-30
LavinMQ fails to start after removing password from user


lukas8219:
**Describe the bug**
After creating a user via Management UI and removing it's password (create -> update -> remove password), LavinMQ fails to start with 

```
2026-04-30T19:13:43.192460Z DEBUG lmq.data_dir_lock Data directory lock aquired
2026-04-30T19:13:43.192797Z DEBUG lmq.user_store Loading users from file
2026-04-30T19:13:43.193319Z ERROR lmq.user_store Failed to load users
Expected String but was Null at line 17, column 26 (JSON::ParseException)
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/json/pull_parser.cr:725:5 in 'raise'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/json/pull_parser.cr:716:5 in 'expect_kind'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/json/pull_parser.cr:310:5 in 'read_string'
  from src/lavinmq/auth/user.cr:29:13 in 'initialize'
  from src/lavinmq/auth/user.cr:21:7 in 'new'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/json/from_json.cr:216:11 in 'load!'
  from src/lavinmq/auth/user_store.cr:17:9 in 'initialize'
  from src/lavinmq/auth/user_store.cr:15:7 in 'new'
  from src/lavinmq/server.cr:47:16 in 'initialize'
  from src/lavinmq/server.cr:43:5 in 'new'
  from src/lavinmq/launcher.cr:62:36 in 'start'
  from src/lavinmq/standalone_runner.cr:6:7 in 'run'
  from src/lavinmq.cr:18:1 in '__crystal_main'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/crystal/main.cr:129:5 in 'main_user_code'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/crystal/main.cr:115:7 in 'main'
  from /opt/homebrew/Cellar/crystal/1.20.0/share/crystal/src/crystal/system/unix/main.cr:9:3 in 'main'

Unhandled exception: Expected String but was Null at line 17, column 26 (JSON::ParseException)
[...]

PR #1897, opened 2026-04-30
feature(auth/EXTERNAL): Add support for EXTERNAL Authentication mechanism - X509 Certificates

### LINKED ISSUE #1782: x.509 certificate authentication
## Summary
Add support for the `EXTERNAL` SASL authentication mechanism, allowing clients to authenticate using their TLS client certificate instead of username/password

LavinMQ already supports mTLS with client certificate verification and extracts certificate info (CN, signature algorithm, etc.) into `ConnectionInfo`. However, this information is not used for authentication decisions — clients must still provide credentials via PLAIN/AMQPLAIN.

 With the EXTERNAL mechanism, the server extracts the username from the client's x.509 certificate (e.g. Common Name or Subject Alternative Name) and authenticates against that identity. Any client-provided password is ignored.

### AMQP 0-9-1
  - Advertise `EXTERNAL` in the `mechanisms` field of `Connection.Start` (only on TLS connections with verified peer cert)
  - When a client selects `EXTERNAL`, extract username from the peer certificate CN (or configurable field)
  - Look up the extracted username against configured users/backends

  ### MQTT
  - Support certificate-based authentication as an alternative to username/password in CONNECT
  - When mTLS is enabled and client cert is verified, allow connections without credentials

  ### HTTP Management API
  - Support client certificate authentication for the management API

  ### Configuration
  - Option to select which certificate field to use as username (CN, DN, SAN, etc.)
  - Option to strip/transform the extracted value (e.g. extract CN from full DN)

  ## Existing infrastructure
  - mTLS with `verify_peer` per protocol via SNI config
  - `ConnectionInfo` already stores `ssl_cn`, `ssl_verify`, etc.
  - `peer_certificate` method available on SSL sockets
  - Pluggable `Auth::Chain` with `Auth::Context`

lukas8219:
Adding support for `EXTERNAL` mechanism allowing for X509 Certificate CN or SAN authentication.

WIP. Manually for now

### Work In Progress Tracking
- [ ] Rename configuration as expected
- [ ] Write Specs to ensure the behavior is in-place
- [ ] Move errors to their own classe
- [ ] Move EXTERNAL to separate module/class
- [ ] Evaluate wether we really need to store SSL CN/SAN in-memory for each connection. Too heavy for tons of connections
- [ ] Review log strategy. What needs to be logged?
- [ ] The SAN `type` is not handled properly by X509::Extension. It's a pure string. Should we have parsing in NativeSocket?

Closes https://github.com/cloudamqp/lavinmq/issues/1782


src/lavinmq/amqp/connection_factory.cr
```diff
@@ -90,12 +90,13 @@ module LavinMQ
...
 
-      def credentials(start_ok)
+      def credentials(start_ok, connection_info)
+        Log.debug { "credentials start_ok.mechanism=#{start_ok.mechanism}"}
         case start_ok.mechanism
         when "PLAIN"
           resp = start_ok.response
           if i = resp.index('\u0000', 1)
-            {resp[1...i], resp[(i + 1)..-1]}
+            {resp[1...i], resp[(i + 1)..-1], false}
```
lukas8219: Just for local development for now...

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -104,16 +105,23 @@ module LavinMQ
...
-      def authenticate(socket, remote_address, start_ok, log)
-        username, password = credentials(start_ok)
+      def authenticate(socket, connection_info, start_ok, log)
+        Log.debug { "authenticate#connection_info=#{connection_info}"}
+        remote_address = connection_info.remote_address
+        username, password, is_external_mechanism = credentials(start_ok, connection_info)
         context = Auth::Context.new(
           username,
           password.to_slice,
+          is_ssl_login: is_external_mechanism,
```
lukas8219: Based on RabbitMQ implementation, it looks like it uses the UsersStore (defined users) for ACL.

Meaning, you need to define users so X509 works. All examples in the documentation I've seen was using this manner.

So for LocalAuthenticator logic to work, reused in HTTP etc, there might be the need for some things to move around or simply add a flag like this.

lukas8219: On a second thought: maybe renaming to `allow_passwordless` could be more than enough for this...

src/lavinmq/server.cr
```diff
@@ -237,9 +237,30 @@ module LavinMQ
       conn_info.ssl = true
       conn_info.ssl_version = ssl_client.tls_version
       conn_info.ssl_cipher = ssl_client.cipher
+      conn_info.ssl_san = extract_subject_alternative_name(ssl_client.peer_certificate) 
+      conn_info.ssl_cn = extract_common_name(ssl_client.peer_certificate) 
       handle_connection(ssl_client, conn_info, protocol)
     end
 
+    def extract_subject_alternative_name(peer_certificate) : String?
```
lukas8219: I don't really hate this logic here. But maybe moving the TLS stuff to another place as a starter... `server.cr` seems to be getting very cluttered with TLS stuff...

lukas8219 (commented):
There's also a change in `amqp-protocol.cr` needed it seems...

The `START` packet is not offering EXTERNAL.

Matter of fact  - in RabbitMQ that's configurable - [the ordering](https://github.com/rabbitmq/rabbitmq-server/tree/main/deps/rabbitmq_auth_mechanism_ssl#usage).

Does it make sense for us to also do it?

lukas8219:
Why is Claude being activated on Draft's also? I am being roasted for nothing lol

lukas8219:
Blocked by https://github.com/cloudamqp/lavinmq/issues/1896

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -104,16 +105,23 @@ module LavinMQ
           tbl = AMQP::Table.from_io(io, ::IO::ByteFormat::NetworkEndian, io.bytesize.to_u32)
           user = tbl["LOGIN"]?.as(String?) || ""
           password = tbl["PASSWORD"]?.as(String?) || ""
-          {user, password}
+          {user, password, false}
+        # TODO handle "EXTERNAL"
+        when "EXTERNAL"
+          Log.debug { "EXTERNAL method #{connection_info.ssl_cn}"}
+          {connection_info.ssl_cn.not_nil!, "", true}
```
carlhoerberg:
```suggestion
          if cn = connection_info.ssl_cn
            {cn, "", true}
          else
            raise "EXTERNAL authentication method but no SSL Common Name present"
          end
```

src/lavinmq/server.cr
```diff
@@ -237,9 +237,30 @@ module LavinMQ
...
+      return nil unless peer_certificate
+      san_ext = peer_certificate.extensions.find do |ext|
+        ext.oid == "subjectAltName"
+      end
+      san_ext.try(&.value)
```
carlhoerberg:
```suggestion
      peer_certificate.try &.extensions.each do |ext|
        if ext.oid == "subjectAltName"
          return ext.value
         end
      end
```

src/lavinmq/server.cr
```diff
@@ -237,9 +237,30 @@ module LavinMQ
...
+    end
+
+    def extract_common_name(peer_certificate) : String?
+      Log.debug { "extract_common_name=#{peer_certificate}"}
+      return nil unless peer_certificate
+      maybe_cn = peer_certificate.subject.to_a.find { |name, _value| name == "CN" }
```
carlhoerberg: is `to_a` required here?

src/lavinmq/server.cr
```diff
@@ -237,9 +237,30 @@ module LavinMQ
...
+      Log.debug { "extract_common_name#maybe_cn=#{maybe_cn}"}
+      maybe_cn.try do |entry|
+        Log.debug { "extract_commmon_name#entry=#{entry[0]}"}
+        entry[1]
```
carlhoerberg: what's entry 0 and 1 ehere? key/value? there's no entry.value?

src/lavinmq/server.cr
```diff
@@ -237,9 +237,30 @@ module LavinMQ
       conn_info.ssl = true
       conn_info.ssl_version = ssl_client.tls_version
       conn_info.ssl_cipher = ssl_client.cipher
+      conn_info.ssl_san = extract_subject_alternative_name(ssl_client.peer_certificate)
```
carlhoerberg: SAN is never used? It should right, it should have preference over common name?

src/lavinmq/amqp/connection_factory.cr
carlhoerberg: what is it used for? Just displaying in the UI?

carlhoerberg (changes requested):

src/lavinmq/amqp/connection_factory.cr
lukas8219: @carlhoerberg is used for [this](https://github.com/cloudamqp/lavinmq/pull/1897/changes#diff-ef90027377132e92acd5c1ec39f8512b579f1a1083dc1637a401e735a5496fa0R12) LocalAuthenticator check.

X509 auth is "passwordless" but still uses the Local users

src/lavinmq/server.cr
lukas8219: Yes. It will be based on Config! This was just to make sure the solution was on the right path (testing happy path) etc

src/lavinmq/server.cr
lukas8219: Will look up in docs. That was a quick happy path test!

src/lavinmq/amqp/connection_factory.cr
carlhoerberg: ok, but it should be called something else, right? 
```suggestion
          external_authentication: is_external_mechanism,
```

src/lavinmq/server.cr
carlhoerberg: why not in `ConnectionInfo`, pass the `ssl_client` to the constructor, instead of doing a lot of property assignments in this class.

lukas8219: I did with `with_ssl` builder function on ConnectionInfo

src/lavinmq/server.cr
lukas8219: @carlhoerberg `it should have preference over common name` made me question. Are we going to follow the auth_ssl rabbitmq specification or do we go SAN -> CN in order?

lukas8219:
## Config Keys

RabbitMQ uses root-level config keys such as ssl_cert_login_from.

Should we have an INI section only for that?

I initially started with:

```ini
[external_auth] 
```

## X509 Extension Parsing

The X509::Extension class does not handle the type properly.

X509::Extension.value is a <key:value> string that needs to be parsed, likely by splitting it, for the compliant keys used here:

https://github.com/rabbitmq/rabbitmq-server/blob/a3b748234e3d69a921c5f4e01fc761157e7c8c9b/deps/rabbit/src/rabbit_ssl.erl#L212

## Design Question

What is the usual mindset here?

Should we parse it early, for example at NativeSocket::Server or a similar boundary, or should we leave it as manual handling where it is actually used?

## Current Preference

I am leaning toward not storing the SAN or CN in ConnectionInfo.

Instead, I would load or parse them on demand for callers, because storing them directly could clutter ConnectionInfo and make each connection more memory-expensive.

src/lavinmq/amqp/connection_factory.cr
```diff
@@ -104,16 +105,43 @@ module LavinMQ
...
+            end
+          end
+
+          if external_auth_login_from == "common_name"
+            ssl_cn = connection_info.ssl_cn
+            return {ssl_cn, "", true} unless ssl_cn.nil?
+            raise "EXTERNAL authentication method but no SSL Common Name present"
+          end
+          raise "EXTERNAL is not configured on the server"
         else raise "Unsupported authentication mechanism: #{start_ok.mechanism}"
```
lukas8219: I think it makes sense to extract this `credentials` to a folder + file for each.

There's already a `src/lavinmq/auth/` with a subfolder `authenticator`. Might as well have a `src/lavinmq/auth/mechanisms`

src/lavinmq/connection_info.cr
```diff
@@ -11,19 +11,57 @@ module LavinMQ
...
+    def extract_common_name(peer_certificate) : String?
+      Log.debug { "extract_common_name=#{peer_certificate}"}
+      return nil unless peer_certificate
+      maybe_cn = peer_certificate.subject.to_a.find { |name, _value| name == "CN" }
+      Log.debug { "extract_common_name#maybe_cn=#{maybe_cn}"}
+      maybe_cn.try do |entry|
+        Log.debug { "extract_commmon_name#entry=#{entry[0]}"}
+        entry[1]
+      end
+    end
```
lukas8219: I prefer the idea of not storing it at all. Just calling on-demand

src/lavinmq/config/options.cr
```diff
@@ -382,6 +375,14 @@ module LavinMQ
       property oauth_audience : String? = nil
       @[IniOpt(section: "oauth", ini_name: jwks_cache_ttl)]
       property oauth_jwks_cache_ttl : Time::Span = 1.hours
+
+      @[IniOpt(section: "external_auth", ini_name: login_from)]
+      property external_auth_login_from : Bool = true
```
lukas8219: Temporary. What config-key could it be?

lukas8219:
For proper specs, it needs https://github.com/cloudamqp/amqp-client.cr/pull/74

src/lavinmq/server.cr
lukas8219: to_a.to_h is the best approach to access by key name!

PR #1898, opened 2026-04-30
fix(auth/passwordless): deserializes to empty string and serializes `null` hashing_algorithm to prevent broker start failure

### LINKED ISSUE #1896: LavinMQ fails to start after removing password from user

lukas8219:
Fixes a issue where `passwordless` users would corrupt broker data_dir preventing it from starting again

- [x] Manual - `tls-gen` script to create CA and Client Certs -> `localhost` domain
- [X] Import `passwordless` user from RabbitMQ via Management UI
- [ ] `lavinmqctl import_definitions` from passwordless user
- [X] Spec. (Cherry-picked locally to `main` and ran it to verify the restart actually catches issues)

Closes https://github.com/cloudamqp/lavinmq/issues/1896


lukas8219:
Wonder if ClaudeBot could just wait CI before saying anything. He missed a clear issue from the Specs

src/lavinmq/auth/user.cr
```diff
@@ -28,7 +28,7 @@ module LavinMQ
           when "password_hash"
             hash = pull.read_string
```
kickster97: Pre-fix installs may have written `password_hash: null` to disk. With this still being `read_string`, those installs still cannot start so we need to also loosen this case
  ```suggestion
              hash = pull.read_string_or_null || ""
  ```

src/lavinmq/auth/user.cr
kickster97, line 41: If we allow passwordless users, we should not raise if we are missing `password_hash`?

kickster97: Maybe we need something like this, but i don't like all this complexity in the initializer... 

  ```suggestion
          @password = if hash.empty?
                        nil
                      elsif hash_algo.nil?
                        raise JSON::ParseException.new("password_hash present without hashing_algorithm", *loc)
                      else
                        parse_password(hash, hash_algo, loc)
                      end
  ```

kickster97: The `,` nil arm is unreachable now that line 61 does return `nil` unless `hash_algorithm`.

  ```suggestion
          when /md5$/i    then Password::MD5Password.new(hash)
  ```

kickster97 (dismissed):

src/lavinmq/auth/user.cr
kickster97: Or, i guess its up for discussion if we want to structure the code around that

src/lavinmq/auth/user.cr
lukas8219: That check if mostly for the compiler

src/lavinmq/auth/user.cr
lukas8219: I dont think we need to handle that. The server would not even start if someone had a broken def

kickster97: The point is, if a pre-fix broker ever wrote this to disk, the upgrade will crash. 
  
Old `user_details` serialized `password_hash: @password`, which becomes `null` for passwordless users, and those are reachable via `PUT /api/users/<name>` that accepts `{"password_hash": ""}`. So a healthy pre-fix broker could snapshot `"password_hash": null` to disk, keep running and not validating that value until the broker restarts.

lukas8219: Yes, that's the exact scenario we are solving for.

But currently there's no reason a operator would define a password-less user. The UX is counterintuitive: you need to create the user WITH password, then go to EDIT and then remove it. 

Soon as you do it, the user is useless because we don't support any Authentication method for Password-less users.

I hardly believe we have clusters with those in the wild.

If so, we would have caught it more earlier?

lukas8219:
Discussed with @kickster97 
1. Use `read_string_or_null`. Log if null about malformed JSON (unexpected). Default to `""`

lukas8219:
@kickster97 Do you wanna follow the bot suggestion?

lukas8219:
@kickster97 Any further action here?

kickster97:
@lukas8219 yes, there are still two paths that can crash.

1, when you have non.null algorithm and a null/empty hash, you set it to `hash=""`, then e.g.`parse_password("", "sha256")` falls through to `SHA256Password.new("")` and that needs to be handled.   
2, when password_hash is null on `definitions.cr:203`, A user importing broken definitions exported from an older version will get a `500`

but the root smell of this is that `password_hash` and `hashing_algorithm` are two loosely coupled nullable fields.. Would be nicer to collapse them into one optional password object (or drop the algorithm field entirely and lean on the hash prefix the code already strips on read) so the guarantee "algo and hash are either both present or both absent" becomes structural instead of implicit.

lukas8219:
@kickster97  I am down for your suggestion but it might incur a bit more of refactoring to implement it. (ie, moving the fields to separate class itself).

I am not sure we can reproduce the 1) ?

kickster97:
> I am not sure we can reproduce the 1) ?

why not? Put a `users.json` on disk with `password_hash` null but `hashing_algorithm` set to `"sha256"`, restart the server.

carlhoerberg:
I think this is important: 

> This meant a users.json entry like {"password_hash": "abc123...", "hashing_algorithm": null} would be loaded as an MD5 password — the user could still log in.
> After this PR, the same entry hits return nil unless hash_algorithm and the user silently becomes passwordless (can never authenticate). This is a security-relevant regression: upgrading LavinMQ could lock out existing users whose stored hashing_algorithm is null.

`hash_algorithm` was null in old rabbitmq definitions exports, we still want to support those.

lukas8219:
k. I'll work on this PR tomorrow! Thanks!

lukas8219:
@kickster97 Anything missing here? Do you think we can push this next week or should we wait?

kickster97 (approved):
Maybe the docs could be slightly clearer so clients don't end up using it wrongly? 
  - password_hash omitted leaves password unchanged
  - password_hash: "" = make passwordless
  - password_hash: null = 400 invalid input

lukas8219:
I'll add to my TODO @kickster97

PR #1901, opened 2026-05-04
Fiber-free synchronous replication

carlhoerberg:
Fixes #1806
This PR refactors the clustering replication mechanism to bypass the action channel and instead perform direct socket writes from publisher fibers.

### Key Changes:
- **Direct Socket Writes**: Publisher fibers now write directly to follower sockets in a loop, eliminating the overhead of the action channel and the dedicated action loop fiber.
- **Mutex Protected Writes**: Added a `@write_lock` Mutex in `Follower` to ensure thread-safe access to the socket and LZ4 writer during concurrent appends.
- **Dedicated Ack Loop**: Re-introduced a dedicated `ack_loop` fiber in `Follower` to read asynchronous acknowledgments from followers.
- **Zero-Copy Restoration**: Restored sending `Bytes` slices directly from memory-mapped files in `MessageStore#write_to_disk` to avoid re-serialization overhead.
- **WaitGroup Removal**: Removed redundant `WaitGroup` tracking in the `delete` path across `MessageStore`, `Server`, and `Follower` as writes are now synchronous to the socket buffer.
- **Cleanup**: Removed the unused `Action` structs and simplified the `MessageStore#close` logic.

### Performance & Latency:
By eliminating fiber context switches and channel synchronization on the replication path, this change reduces latency for replicated publishes and decreases CPU usage.


carlhoerberg:
It's for now slower than the buffered Channel approach, about 700k msgs/s compared to 900 msgs/s in main, with a three node cluster.

spuun (approved):
LGTM

PR #1905, opened 2026-05-05
docs: update Homebrew install command for homebrew-core

jage:
LavinMQ is now available in homebrew-core, so the third-party tap prefix is no longer needed.


kickster97 (approved):

jage:
Woops, merged by reflex, dunno how you wanna handle these during release freeze @kickster97 ? (IMHO it makes sense to get it even though freeze since it's not a change to broker) :)

PR #1907, opened 2026-05-06
http: clearer error reason when restart_queue fails

oskgu360:
## Summary
- The `PUT /api/queues/:vhost/:name/restart` endpoint (used by `lavinmqctl restart_queue`) returned `"Queue was not restarted"` on failure, which didn't tell the user what to do about it.
- Replace the message with the queue's current state and the actual requirement, e.g. `"Queue is running; only closed queues can be restarted"`. Restart only succeeds when the queue is in the `closed` state, so surfacing that makes the failure self-explanatory.

## Test plan
- [x] `make test SPEC=spec/api/queues_spec.cr` passes
- [x] `ameba` clean on changed files
- [x] Existing "should not restart if queue is still running" spec extended to assert the new reason string

### DIFF
--- a/spec/api/queues_spec.cr
+++ b/spec/api/queues_spec.cr
@@ -628,6 +628,8 @@ describe LavinMQ::HTTP::QueuesController do
           s.vhosts["/"].queues[queue_name]
           response = http.put("/api/queues/%2f/#{queue_name}/restart")
           response.status_code.should eq 400
+          body = JSON.parse(response.body)
+          body["reason"].as_s.should eq "Queue is running; only closed queues can be restarted"
 
           response = http.get("/api/queues/%2f/#{queue_name}")
           response.status_code.should eq 200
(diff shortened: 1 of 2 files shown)


kickster97 (commented):
maybe too specific for and else case on restart since alot of things can go wrong on a restart.. cloud we have a more detailed rescue here instead where the errors cause is more clear?

me and @snichme had a discussion today about error messaging around closed queues/restarting is pretty lacking today..

oskgu360:
> maybe too specific for and else case on restart since alot of things can go wrong on a restart.. cloud we have a more detailed rescue here instead where the errors cause is more clear?
> 
> me and @snichme had a discussion today about error messaging around closed queues/restarting is pretty lacking today..

Yes, agree :) Lazy attempt on minor improvement. But I agree the long-term fix is probably to return with error reasons and relay them back to user. But not something for now, its a very minor issue 😁  Closing this for now then

PR #1916, opened 2026-05-08
Render lavinmqctl list command output as Unicode box-drawing tables

annieblomgren:
## Summary

- Refactors `output_array` to render results as Unicode box-drawing tables instead of raw tab-separated output
- Adds a `--silent` flag path that falls back to machine-readable TSV for scripting
- Applies to all existing `lavinmqctl list_*` commands

## Test plan

- [ ] Run `lavinmqctl list_queues` and verify table formatting
- [ ] Run `lavinmqctl list_queues --silent` and verify tab-separated output
- [ ] Run `make test` to ensure no regressions

### DIFF
--- a/src/lavinmqctl/cli.cr
+++ b/src/lavinmqctl/cli.cr
@@ -432,29 +432,50 @@ class LavinMQCtl
   end
 
   private def output_array(data : Array, columns : Array(String)?)
-    if columns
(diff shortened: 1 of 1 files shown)


carlhoerberg:
What is "Unicode box-drawing tables"?

PR #1918, opened 2026-05-08
Add lavinmqctl list_etcd_members command

annieblomgren:
## Summary

- Adds `list_etcd_members` command that lists etcd cluster members using the MemberList API, showing name, peer URLs, client URLs, and learner status
- Depends on the `--etcd-endpoints` flag introduced in the `list_in_sync_replicas` PR

**Depends on:** #1917

## Test plan

- [ ] Run `lavinmqctl list_etcd_members` against a running etcd cluster
- [ ] Verify all member fields (name, peer_urls, client_urls, learner) are shown
- [ ] Run `make test` to ensure no regressions

### DIFF
--- a/src/lavinmq/etcd.cr
+++ b/src/lavinmq/etcd.cr
@@ -162,6 +162,18 @@ module LavinMQ
       end
     end
 
+    def member_list : Array(NamedTuple(name: String, peer_urls: String, client_urls: String, learner: Bool))
+      json = post("/v3/cluster/member/list", "{}")
+      members = json["members"]?.try(&.as_a) || return [] of NamedTuple(name: String, peer_urls: String, client_urls: String, learner: Bool)
+      members.map do |m|
+        name = m["name"]?.try(&.as_s) || ""
+        peer_urls = m["peerURLs"]?.try(&.as_a.map(&.as_s).join(",")) || ""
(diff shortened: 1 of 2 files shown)


src/lavinmq/etcd.cr
```diff
@@ -162,6 +162,18 @@ module LavinMQ
       end
     end
 
+    def member_list : Array(NamedTuple(name: String, peer_urls: String, client_urls: String, learner: Bool))
```
spuun, line 165: Maybe introduce a struct instead of this NamedTuple

src/lavinmq/etcd.cr
```diff
@@ -162,6 +162,18 @@ module LavinMQ
...
+      json = post("/v3/cluster/member/list", "{}")
+      members = json["members"]?.try(&.as_a) || return [] of NamedTuple(name: String, peer_urls: String, client_urls: String, learner: Bool)
+      members.map do |m|
+        name = m["name"]?.try(&.as_s) || ""
+        peer_urls = m["peerURLs"]?.try(&.as_a.map(&.as_s).join(",")) || ""
+        client_urls = m["clientURLs"]?.try(&.as_a.map(&.as_s).join(",")) || ""
```
spuun, line 171: I think these two could be `Array(URI)` and formatted when outputted instead

PR #1920, opened 2026-05-08
refactor: Decouple `MQTT::Session` from `AMQP::Queue`

### LINKED ISSUE #1906: Mgmt Console: consumer link returns 404 on mqtt connections.
**Describe the bug**
Management console - when navigating to a consumer channel for a mqtt queue, no information is retrieved about the channel, and the logs report:
```
 lmq.http.apierror method=GET path=/api/channels/xxxx%3A33059%5B0%5D status=404 message="Channel xxxx:33059[0] does not exist"
```

**Describe your setup**
Please describe the environment you are using by including the build info. Please also include any relevant information about the client software or amqp libraries that you may be using. 
2.8.0-rc.2 (beetle shared)

**How to reproduce**
Create a mqtt client with `mosquitto_sub`
navigate to the queue -> click the consumer channel in the management console. 

**Expected behavior**
either see the information about the mqtt connection, or not be able to press the channel.

spuun:
As part of an ongoing effort to better separate AMQP and MQTT concerns, this PR removes the dependency between Session (MQTT) and AMQP::Queue.

  Previously, MQTT sessions were tightly coupled to the AMQP queue abstraction — MQTT::Session inherited from LavinMQ::AMQP::Queue, which in turn inherited from LavinMQ::Queue which was a shared base class for both protocols. This PR removes that base class entirely: LavinMQ::Queue is now a union type alias (AMQP::Queue |
  MQTT::Session) rather than a superclass, keeping the rest of the codebase compiling while the two are fully decoupled. MQTT sessions are stored in their own dedicated structure, consumer code has been moved out of Session, and required modules are now explicit.

  This is a preparatory step toward a cleaner protocol boundary where AMQP and MQTT code can evolve independently. The LavinMQ::Queue alias can hopefully be removed in a future step.

Some future todos are:
- Break out max-length policy code (including drop related code) to a "Argument" (just like DeadLetteringArgument)
- AMQP namespace shouldn't know about MQTT namespace and vice versa
- Separate UI for AMQP and MQTT
- Separate API for AMQP and MQTT?
- Store MQTT definitions separately? 

Addresses #957
Fixes #1920 because MQTT consumers are removed


src/lavinmq/mqtt/session.cr
```diff
@@ -98,9 +159,18 @@ module LavinMQ
       end
 
       def publish(msg : Message) : Bool
-        # Do not enqueue messages with QoS 0 if there are no consumers subscribed to the session
-        return true if msg.properties.delivery_mode == 0 && @consumers.empty?
-        super
+        return true if msg.properties.delivery_mode == 0 && @client.nil?
+        return false if @deleted || @closed
```
spuun:
```suggestion
        return false if @deleted || @closed
        return true if msg.properties.delivery_mode == 0 && @client.nil?
```

spuun:
I rebased this on #1935

src/lavinmq/vhost.cr
```diff
@@ -91,38 +92,80 @@ module LavinMQ
...
+    end
+
+    def sessions_size : Int32
+      definitions.sessions_size
+    end
+
+    def register_session(session : MQTT::Session) : Nil
+      definitions.register_session(session)
+    end
+
```
carlhoerberg: Dead code
```suggestion
```

src/lavinmq/definitions_store.cr
```diff
@@ -136,13 +167,18 @@ module LavinMQ
...
         when AMQP::Frame::Queue::Declare
-          return false if @queues.has_key? f.queue_name
-          q = @queues[f.queue_name] = QueueFactory.make(@vhost, f)
+          return false if @queues.has_key?(f.queue_name) || @sessions.has_key?(f.queue_name)
+          q = QueueFactory.make(@vhost, f)
+          if q.is_a?(MQTT::Session)
+            @sessions[f.queue_name] = q
+          else
+            @queues[f.queue_name] = q.as(AMQP::Queue)
+          end
```
carlhoerberg, line 195:
```suggestion
          case q = QueueFactory.make(@vhost, f)
          in MQTT::Session
            @sessions[f.queue_name] = q
          in AMQP::Queue
            @queues[f.queue_name] = q
          end
```

src/lavinmq/definitions_store.cr
kickster97, line 346: Could durable MQTT sessions be missed here now that they are stored in `@sessions`? `Queue::Declare` puts MQTT sessions in `@sessions`, but compaction only writes durable declarations from `@queues`.

src/lavinmq/mqtt/session.cr
```diff
@@ -20,12 +40,55 @@ module LavinMQ
...
+      end
+
+      def close : Bool
+        return false if @closed
+        @closed = true
+        @msg_store_lock.synchronize do
+          @msg_store.close
+        end
+        true
+      end
```
kickster97, line 89: Should `close` terminate the session delivery loop before closing the message store?

spuun: Yes!
It will eventually be stopped because of @closed, but all channels involved should be closed too (i guess it may be stuck waiting on a channel now).

src/lavinmq/http/controller/prometheus.cr
```diff
@@ -540,6 +552,10 @@ module LavinMQ
             labels = {queue: q.name, vhost: vhost.name}
             writer.write_value("detailed_queue_deduplication", q.dedup_count, labels)
           end
+          vhost.each_session do |s|
+            labels = {queue: s.name, vhost: vhost.name}
+            writer.write_value("detailed_queue_deduplication", s.dedup_count, labels)
```
carlhoerberg: we have dedup on mqtt sessions?

src/lavinmq/mqtt/session.cr
```diff
@@ -238,10 +322,122 @@ module LavinMQ
...
+
+      # def ack(sp : SegmentPosition) : Nil; end
+
+      # def reject(sp : SegmentPosition, requeue : Bool) : Nil; end
+
+      # def basic_get(no_ack : Bool, force : Bool = false, & : Envelope -> Nil) : Bool
+      #  false
+      # end
+
```
carlhoerberg: Leftovers? 
```suggestion
```

src/lavinmq/definitions_store.cr
carlhoerberg: it doesn't work or you don't like it?

src/lavinmq/definitions_store.cr
carlhoerberg: seems to be fixed?

src/lavinmq/vhost.cr
```diff
@@ -291,6 +338,21 @@ module LavinMQ
...
+        ack += s.ack_count
+        confirm += s.confirm_count
+        deliver += s.deliver_count
+        deliver_no_ack += s.deliver_no_ack_count
+        deliver_get += s.deliver_get_count
+        get += s.get_count
+        get_no_ack += s.get_no_ack_count
+        publish += s.publish_count
+        redeliver += s.redeliver_count
+        return_unroutable += s.return_unroutable_count
```
carlhoerberg: sessions can't return messages, so always 0? I guess it's the same for some of the above too?

src/lavinmq/mqtt/session.cr
```diff
@@ -238,10 +322,122 @@ module LavinMQ
...
+      def state : QueueState
+        @closed ? QueueState::Closed : QueueState::Running
```
snichme:
```suggestion
        closed? ? QueueState::Closed : QueueState::Running
```
Now we aren't checking the value if the atomic just that it's there

src/lavinmq/mqtt/session.cr
```diff
@@ -1,45 +1,115 @@
...
+
+      def exclusive? : Bool
+        false
+      end
+
+      def arguments : AMQP::Table
+        ARGUMENTS
+      end
+
+      protected def close : Bool
```
snichme:
```suggestion
      def close : Bool
```
should be public, used from `VHost`

src/lavinmq/mqtt/session.cr
```diff
@@ -238,10 +322,122 @@ module LavinMQ
...
+
+      private def clear_policy_arguments
+        handle_arguments
+      end
+
       private def handle_arguments
-        super
+        @effective_args = Array(String).new
         @effective_args << "x-queue-type"
       end
```
snichme, line 355: Probably not a bug but hard to understand, `clear_policy_arguments` calls 'handle_arguments`  that just appends "x-queue-type". Why? What does this mean :)

snichme (commented):
I get this but not in CI: 
```
In src/lavinmq/mqtt/exchange.cr:60:9

 60 | @bindings.each_value.sum(&.size)
      ^--------
Error: can't infer the type of instance variable '@bindings' of LavinMQ::MQTT::Exchange
```

src/lavinmq/http/controller/prometheus.cr
spuun: This can be removed, I guess. It's just quick copy-pasta to keep the same behavior as before.

src/lavinmq/mqtt/session.cr
spuun: How can this not give a compile-time error? hm...

snichme: was thinking the same, no idea.

kickster97: Hm, in crystal `protected` is based on namespace and not class, so if `close` is defined within the same namespace as the callers in `vhost` its allowed, and both have `LavinMQ::` as a namespece, maybe thats why

so anything wrapped in `LavinMQ` can be accessed by other classes in `LavinMQ`

spuun: > Hm, in crystal `protected` is based on namespace and not class, so if `close` is defined within the same namespace as the callers in `vhost` its allowed, and both have `LavinMQ::` as a namespece, maybe thats why
> 
> so anything wrapped in `LavinMQ` can be accessed by other classes in `LavinMQ`

Right. I've been thinking it was only for classes on the same "namespace level", but it works on the depth.

src/lavinmq/mqtt/session.cr
```diff
@@ -121,7 +194,7 @@ module LavinMQ
...
 
       private def get_packet(& : MQTT::Publish, UInt32 -> Nil) : Bool
-        raise ClosedError.new if @closed
+        raise AMQP::Queue::ClosedError.new if closed?
```
snichme: `AMQP::Queue`  error in mqtt, maybe just duplicate the error instead

src/lavinmq/mqtt/session.cr
```diff
@@ -238,9 +322,110 @@ module LavinMQ
...
+
+      private def clear_policy_arguments
+      end
+
       private def handle_arguments
-        super
-        @effective_args << "x-queue-type"
+      end
+
+      def pause!; end
```
snichme, line 357: Not possible to get rid of these even after the split?

src/lavinmq/mqtt/session.cr
```diff
@@ -238,9 +322,110 @@ module LavinMQ
...
+      end
+
+      def details_tuple
+        stats = queue_stats_details
+        {
+          name:                         @name,
+          durable:                      durable?,
+          exclusive:                    false,
+          auto_delete:                  @auto_delete,
+          arguments:                    AMQP::Table.new,
```
snichme: Session doesn't support arguments, so maybe just remove this and get rid of the AMQP namespace as well?

src/lavinmq/binding_details.cr
```diff
@@ -23,7 +23,7 @@ module LavinMQ
         source:           @source,
         vhost:            @vhost,
         destination:      @destination.name,
-        destination_type: @destination.is_a?(Queue) ? "queue" : "exchange",
+        destination_type: @destination.is_a?(Exchange) ? "exchange" : "queue",
```
snichme, line 26: destination can be exchange, queue and session right? Maybe even stream.

snichme (approved):

src/lavinmq/mqtt/session.cr
spuun: It's still there because I've tried to touch as little as possible, but I've done more than the initially plan, so I could actually do some more. Will fix this.

src/lavinmq/mqtt/session.cr
spuun: Yes. I guess we could do it now by doing some validation and casts in http controllers, but I think that will be a follow up.

src/lavinmq/mqtt/session.cr
spuun: Yes, but I'd like the interface to be the same for Session and Queue as long as they live in the same UI. but this Table can probably be an empty NamedTuple instead.
Also, with MQTT5 we may get "arguments".

src/lavinmq/binding_details.cr
spuun: Yes, but this isn't a change in behavior. Testing vs `Exchange` instead of `Queue` was a trick to get rid of one place where session/queue was used as one type.

carlhoerberg:
@spuun should we merge this one now?

src/lavinmq/mqtt/session.cr
```diff
@@ -238,9 +326,110 @@ module LavinMQ
...
+        end
+      end
+
+      private def clear_policy_arguments
+      end
+
       private def handle_arguments
-        super
-        @effective_args << "x-queue-type"
+      end
```
kickster97, line 355: Since these are no-ops, the policy limits are never reset. `apply_policy` calls `clear_policy` first, but now that we dont have AMQP::Queue that clears here `@max_length`/`@max_length_bytes`/`@reject_on_overflow` stay set, so a removed/relaxed policy keeps dropping messages forever.

kickster97: idk maybe this is considered "future work" with restructuring max-length policy code

src/lavinmq/mqtt/session.cr
```diff
@@ -106,10 +174,19 @@ module LavinMQ
...
+      def publish(msg : Message) : Bool
+        return true if msg.properties.delivery_mode == 0 && @client.nil?
+        return false if @deleted || closed?
+        @msg_store_lock.synchronize do
+          @msg_store.push(msg)
+          drop_overflow
+        end
+        @publish_count.add(1, :relaxed)
+        true
+      end
```
kickster97, line 185: `@reject_on_overflow` is never read here (like it is in queue), so `overflow: reject-publish` silently acts as drop-head.. maybe also considered to be part of future work

src/lavinmq/mqtt/broker.cr
```diff
@@ -39,7 +39,10 @@ module LavinMQ
 
       def add_client(io, connection_info, user, packet)
         if prev_client = @clients[packet.client_id]?
-          prev_client.close("New client #{connection_info.remote_address} (username=#{packet.username}) connected as #{packet.client_id}")
+          prev_client.close(
+            "New client #{connection_info.remote_address} " \
+            "(username=#{packet.username}) connected as #{packet.client_id}")
+          remove_client(prev_client)
```
kickster97, line 45: `prev_client.close` already calls `remove_client` via the read-loop, so takeover double-ticks `ConnectionClosed`?

kickster97 (approved):

src/lavinmq/mqtt/broker.cr
spuun: Yeah, this is a little bit unintuitive. It's not obvious that close will wait for the client to be disconnected and removed, so the remove is here as a "guard" or something.... hm. I'll let this be for now, but things should probably be refactored for clearness.

PR #1921, opened 2026-05-08
avoid double message read in reject(requeue=true)

viktorerlingsson:
## Summary

The expire path in `Queue#reject(sp, requeue: true)` read the message twice - once in `has_expired?(sp)` and again in `expire_msg(sp)` when a DLX was configured. Read once, reuse the envelope across the expiry check and dead-lettering call. Removes the now-unused `has_expired?(sp)` overload.

Behavior is preserved: `expire_msg(env)` calls `@dead_letter.route`, which falls through to `routed.call` (i.e. `delete_message sp`) when no DLX is configured — same end state as the old `expire_msg(sp)` no-DLX branch.

## Test plan

- [x] New regression spec: `reject(requeue=true)` when TTL has elapsed while the message was unacked → asserts DLQ receives the message
- [x] New regression spec: `reject(requeue=true)` when `x-delivery-limit` is exceeded → asserts DLQ receives the message (the existing `policies_spec` only counted the drop, didn't verify DLX routing)
- [x] `make test SPEC=spec/queue_dead_lettering_spec.cr` — 41 examples, 0 failures
- [x] `make lint` — clean

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -629,11 +629,6 @@ module LavinMQ::AMQP
       end
     end
 
-    private def has_expired?(sp : SegmentPosition, requeue = false) : Bool
-      msg = @msg_store_lock.synchronize { @msg_store[sp] }
-      has_expired?(msg, requeue)
-    end
-
     private def has_expired?(msg : BytesMessage, requeue = false) : Bool
       return false if zero_ttl?(msg) && !requeue && !@consumers.empty?
       if expire_at = expire_at(msg)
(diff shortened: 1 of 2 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -816,12 +811,15 @@ module LavinMQ::AMQP
       @unacked_count.sub(1, :relaxed)
       @unacked_bytesize.sub(sp.bytesize, :relaxed)
       if requeue
-        if has_expired?(sp, requeue: true) # guarantee to not deliver expired messages
-          expire_msg(sp, :expired)
+        msg = @msg_store_lock.synchronize { @msg_store[sp] }
```
spuun, line 815: What about creating the envelope before the if-statement? It could be used on line 831 too then I think? Maybe we can get rid of the `expire_msg` taking a `SegmentPosition` then? Or maybe is it used in other places.

viktorerlingsson: yes, but we only need the envelope if the msg has expired, so in other cases we'd be creating an envelope that we never used 🤷 Maybe that's ok though!

viktorerlingsson: ended up with this https://github.com/cloudamqp/lavinmq/pull/1921/commits/3384807c0c279e752374601256ab7cab423683af, wdyt?

spuun: Ahhh, didn't really look into what `expire_msg(SegmentPosition)` did... Maybe it's better to leave it as it was then? I'm for which ever :)

viktorerlingsson: Yeah. I didn't really like the change, I think it made `reject` a little harder to read. I reverted it!

snichme (approved):

PR #1923, opened 2026-05-08
Log security-relevant auth and user management events

johanrhodin:
- Log successful authentication with vhost on connection open
- Log permission grants (already logged removals)
- Log password updates via new UserStore wrappers, including byte length when a plain-text password is provided

Introduce more logging. These can be helpful when a user no longer can login, the operator can find the time of when password was changed.


spuun:
Maybe it's worth logging from `User` instead of introducing another interface for updating `User` objects?

carlhoerberg:
would like to continue the work on this @johanrhodin ?

PR #1924, opened 2026-05-09
MT safe collections

carlhoerberg:
## Summary

Wrap broker-wide mutable state in `Sync::Shared` and adopt `Atomic` for hot-path flags, so the broker is safe to run multi-threaded. MT is **not** enabled by default — set `CRYSTAL_WORKERS=N` to opt in (the launcher resizes `Fiber::ExecutionContext.default` to that count; defaults to `1`).

> **Updated:** base is now `main`. Merged `main` in (commit `d0b1cdf`), integrating #1920 (`MQTT::Session` decoupled from `AMQP::Queue`), #2034/#2035 (split AMQP/MQTT protocol servers + connection-lifecycle rework) and #2090 (zero-alloc routing). The MT-safety below was re-applied onto those new structures — see **Reconciliation with `main`**.

### Sync::Shared coverage
- Store collections: `UserStore`, `VHostStore`, `ConnectionStore`, `DefinitionsStore` (incl. the new decoupled `@sessions` hash), `DirectReplyConsumerStore`, `ParameterStore`, `Shovel::Store`.
- `ProtocolServer#@listeners`, `AMQP::Queue#@consumers`, `AMQP::Channel#@consumers`.
- Exchange bindings: `Direct`/`Fanout`/`Topic`/`Headers`/`ConsistentHash#@bindings` and `MQTT::Exchange#@tree`.
- Reads via `shared { }`, mutations via `lock { }`. `size`/`empty?`/`binding_count` use `unsafe_get` (off-by-one is fine, per project convention).

### Atomic state
- `AMQP::Channel#@running`, `#@flow`.
- `AMQP::Consumer#@closed`, `#@flow`.
- `AMQP::Queue#@closed`, `#@deleted`; `MQTT::Session#@closed`.
- `Server#@flow` (was a plain `Bool`).
- `DefinitionsStore#@definitions_deletes`.

### Concurrency fixes uncovered while MT-testing
- TOCTOU in `UserStore#create` and `VHostStore#create` (existence check + insert under different locks).
- TOCTOU in `AMQP::Queue#rm_consumer` (emptiness check moved inside `@consumers.lock`).
- `User#permissions` now atomically replaced rather than mutated in place.
- `register_exchange` / `register_queue` / `register_session` and `apply` / `compact!` / `store_definition` serialized with `@definitions_lock`.
- `users.json` save moved out of `VHostStore#create`'s vhost lock to avoid I/O under a hot lock.
- Users and vhosts disk-load now happens under the store locks for the full duration.
- `save!` serialized in `UserStore`, `VHostStore`, `ParameterStore`.
- `AMQP::Queue#publish_internal`: `drop_overflow` runs outside the `@msg_store_lock` critical section to break an AB-BA cycle with `DeadLetterer#route`.

### Reconciliation with `main` (merge `d0b1cdf`)
`main` concurrently reworked many of the same hot paths. Resolved by keeping this branch's concurrency model and folding in main's new structures/features:
- **Stores** — kept `save!` public (main added bulk-flush callers: definitions import, `save_policies!`/`save_parameters!`); folded main's OAuth-safe permission grant into `VHostStore#create`; Sync-wrapped the new decoupled `@sessions` hash; kept main's offset-based `store_definition`.
- **Exchanges** — re-applied Sync-wrapping onto main's new structures: the headers `Binding` (pre-parsed match spec) class and the MQTT subscription `@tree`.
- **Channel/Consumer** — kept atomic flags + Sync collections; adopted main's `close_channel` (emits `ChannelClosed`) and lazy, restartable `ensure_deliver_loop` model.
- **Queue** — kept atomic `@closed`/`@deleted` + `@paused` BoolChannel (did not adopt main's `QueueState` enum); combined the drop-overflow-outside-lock deadlock fix with main's `pushed`-based publish-on-close semantics; adopted main's idle-timeout expire-fiber model (fixing a `!@closed` check that was always-false against the now-atomic flag).
- **MQTT::Session** — took main's decoupled design (no longer an `AMQP::Queue`); kept the atomic `@closed`.
- **Server** — took main's split `ProtocolServer` architecture; re-applied MT-safety by making `Server#@flow` atomic and Sync-wrapping `ProtocolServer#@listeners`.

### Launcher
- `Launcher#initialize` resizes the default execution context to `ENV["CRYSTAL_WORKERS"].to_i? || 1`, so MT is opt-in.

### Spec infrastructure
- `MqttHelpers#with_server` and `with_amqp_server`/`amqp_port` set up the split protocol servers and wait for listener registration, removing flakes under MT scheduling.
- Spec accesses to now-`Sync`-wrapped internals (`@bindings`, `@listeners`) go through `unsafe_get`.

## Test plan
- [x] `make lint` (clean; one pre-existing environmental crystal-formatter/ameba discrepancy on an unchanged `main` file)
- [x] `make test` (`-Dpreview_mt -Dexecution_context`) — 2089 examples, 0 failures
- [ ] Smoke-test broker startup + basic publish/consume with `CRYSTAL_WORKERS=4`


src/lavinmq/auth/user_store.cr
```diff
@@ -142,7 +163,7 @@ module LavinMQ
           Log.debug { "Loading users from file" }
           File.open(path) do |f|
             Array(User).from_json(f) do |user|
-              @users[user.name] = user
+              @users.lock { |u| u[user.name] = user }
```
spuun: I guess locking around `#from_json` would make sense here

src/lavinmq/parameter_store.cr
```diff
@@ -8,60 +9,69 @@ module LavinMQ
 
     Log = LavinMQ::Log.for "parameter_store"
 
+    @parameters : Sync::Shared(Hash(ParameterId?, T))
+    # Serializes save! so concurrent admin operations don't both
+    # truncate-and-rewrite the .tmp file from offset zero.
+    @save_lock = Mutex.new
```
spuun, line 15:
```suggestion
    @save_lock = Sync::Mutex.new
```

src/lavinmq/vhost_store.cr
```diff
@@ -15,60 +16,77 @@ module LavinMQ
 
     Log = LavinMQ::Log.for "vhost_store"
 
+    @vhosts : Sync::Shared(Hash(String, VHost))
+    # Serializes save! so concurrent admin operations don't both
+    # truncate-and-rewrite vhosts.json.tmp from offset zero.
+    @save_lock = Mutex.new
```
spuun, line 22:
```suggestion
    @save_lock = Sync::Mutex.new
```

src/lavinmq/auth/user_store.cr
```diff
@@ -12,112 +13,132 @@ module LavinMQ
         DIRECT_USER == name
       end
 
+      @users : Sync::Shared(Hash(String, User))
+      # Serializes save! so concurrent admin operations don't both
+      # truncate-and-rewrite users.json.tmp from offset zero.
+      @save_lock = Mutex.new
```
spuun, line 19: `Mutex` will be replaced by `Sync::Mutex`
```suggestion
      @save_lock = Sync::Mutex.new
```

spuun (dismissed):

viktorerlingsson:
Saw that the last spec run hanged, and the last macOS spec run had an invalid memory access. Tried running the specs locally and got a bunch of errors, most of them complaining about `server.@listeners` being empty 
```
Empty enumerable (Enumerable::EmptyError)
           from /usr/share/crystal/src/enumerable.cr:600:13 in 'first'
           from spec/spec_helper.cr:57:3 in '->'
```


on a second run the specs hanged on/after `shift? handles files with few extra bytes`

src/lavinmq/vhost_store.cr
carlhoerberg: for uncontested locks the original `Mutex` is actually faster

snichme:
With regards to 
> size/empty? use unsafe_get per the convention that off-by-one is acceptable.
don't we have logic that depends on `empty?` being correct, for example immediate delivery or QoS for mqtt sessions? For metrics it's not really a big thing, but if routing logic depends on this than off-by-one might get thing wrong for publishers or consumers.

src/lavinmq/vhost_store.cr
```diff
@@ -15,66 +16,81 @@ module LavinMQ
...
-        return v
+      created = false
+      vhost = @vhosts.lock do |vh|
+        if v = vh[name]?
+          next v
+        end
+        new_vhost = VHost.new(name, @data_dir, @users, @replicator, description, tags)
+        Log.info { "Created vhost #{name}" }
+        @users.add_permission(user.name, name, /.*/, /.*/, /.*/, save: false)
+        @users.add_permission(@users.direct_user, name, /.*/, /.*/, /.*/, save: false)
```
snichme: You can move these outside side of the lock, since they also take a lock it could potentially deadlock if someone were to add a vhost lock in add_permissions. 
Moving the call outside the lock removes that and  I see no problem of doing the `add_permission` outside the vhost lock.

src/lavinmq/auth/user.cr
snichme, line 12: Should we remove this getter? Handing out the Hash here allows someone to get the hash and do `hsh.delete` instead of using the added `delete_permission`  method. Introduce a getter like `permission(vhost : String) : Permission` instead

src/lavinmq/auth/user_store.cr
```diff
@@ -159,23 +182,26 @@ module LavinMQ
...
         save!
       end
 
       private def create_direct_user
-        @users[DIRECT_USER] = User.create_hidden_user(DIRECT_USER)
+        direct = User.create_hidden_user(DIRECT_USER)
+        @users.lock { |u| u[DIRECT_USER] = direct }
         perm = {config: /.*/, read: /.*/, write: /.*/}
-        @users[DIRECT_USER].permissions["/"] = perm
+        direct.set_permission("/", perm)
```
snichme, line 189: move this into `@users.lock` as stated by the doc string for `set_permission`  or is there any reason for this not being in the lock?

src/lavinmq/vhost_store.cr
```diff
@@ -98,44 +114,50 @@ module LavinMQ
...
-            @vhosts[name] = VHost.new(name, @data_dir, @users, @replicator, description, tags)
-            @users.add_permission(@users.direct_user, name, /.*/, /.*/, /.*/)
+          Log.debug { "Loading vhosts.json" }
+          @vhosts.lock do |vhosts|
+            JSON.parse(f).as_a.each do |vhost|
+              name = vhost["name"].as_s
+              tags = vhost["tags"]?.try(&.as_a.map(&.to_s)) || [] of String
+              description = vhost["description"]?.try &.as_s || ""
+              vhosts[name] = VHost.new(name, @data_dir, @users, @replicator, description, tags)
+              @users.add_permission(@users.direct_user, name, /.*/, /.*/, /.*/)
```
snichme, line 159: `add_permission` defaults to `save = true` meaning we write the file on each loop here.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -62,50 +63,55 @@ module LavinMQ::AMQP
     @reject_on_overflow = false
     @exclusive_consumer = false
     @deliveries = Hash(SegmentPosition, Int32).new
```
snichme: `@deliveries` is unsynchronized, it is modified in different places in Queue, for example `delete_message`

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -914,12 +938,12 @@ module LavinMQ::AMQP
...
         end
-      end
-      if @consumers.empty?
-        if @auto_delete
-          delete
-        else
-          notify_consumers_empty(true)
+        if cs.empty?
+          if @auto_delete
+            delete
```
snichme, line 1082: `delete` calls `close` which do `@deliver_loop_wg.wait` that is used in consumer.cr and from what I can see it never goes down to zero so it will wait here while keeping the @consumers.lock.

src/lavinmq/amqp/stream/stream_message_store.cr
snichme, line 230: This method should probably be under some lock as well, it write to the underlying mfile as well as updates consumer_offset_positions.

snichme (commented):
Found some data structures that are not synchronized and a potential issue with `rm_consumer` -> `delete` -> `close` where close waits for a consumer_loop_wg waitgroup.

carlhoerberg:
```
● Benchmark comparison: main vs sync-shared-stores

  Publish / consume rates (msgs/s). Δ is this branch relative to main.

  ┌────────────────────────────┬────────┬───────────┬─────────────┬───────────┐
  │          Scenario          │ Metric │   main    │ this branch │     Δ     │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Baseline 1-to-1            │ pub    │ 1,410,828 │   1,239,968 │    −12.1% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,408,135 │   1,242,662 │    −11.8% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ 1-to-1 + properties        │ pub    │ 1,154,513 │     985,132 │    −14.7% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,151,269 │     983,925 │    −14.5% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Publish only               │ pub    │ 2,631,822 │   2,613,545 │     −0.7% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Consume only               │ con    │ 2,328,066 │   2,203,270 │     −5.4% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Fanout 1→4                 │ pub    │   464,924 │     379,740 │    −18.3% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,848,901 │   1,517,878 │    −17.9% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Large msgs (100 KB)        │ pub    │     8,852 │       9,835 │    +11.1% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │     8,847 │       9,835 │    +11.2% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Prefetch 1000 / ack 100    │ pub    │ 2,119,906 │   2,002,530 │     −5.5% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │   133,202 │     128,064 │     −3.9% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Publish confirm            │ pub    │   117,449 │     100,968 │    −14.0% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Publish in tx              │ pub    │    99,774 │      29,327 │ −70.6% ⚠️ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Ack in tx                  │ pub    │ 2,196,164 │   1,932,627 │    −12.0% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Stream 1-to-1              │ pub    │ 2,191,932 │   1,981,373 │     −9.6% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │    97,651 │     115,343 │    +18.1% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Stream + properties        │ pub    │ 1,753,875 │   1,507,098 │    −14.1% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │   134,056 │     161,026 │    +20.1% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Fan-in 64→1                │ pub    │ 2,390,615 │   2,105,531 │    −11.9% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │   269,596 │     267,473 │     −0.8% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Competing 1→64             │ pub    │ 1,147,343 │     522,929 │ −54.4% ⚠️ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,141,790 │     519,011 │ −54.5% ⚠️ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Symmetric 64→64            │ pub    │ 1,696,989 │   1,436,540 │    −15.3% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,307,900 │   1,167,949 │    −10.7% │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Sharded 64q                │ pub    │ 1,747,544 │   2,486,322 │ +42.3% ✅ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,298,459 │   2,207,760 │ +70.0% ✅ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │ Stream multi-consumer 1→64 │ pub    │   282,557 │     132,679 │ −53.0% ⚠️ │
  ├────────────────────────────┼────────┼───────────┼─────────────┼───────────┤
  │                            │ con    │ 1,586,017 │   1,598,912 │     +0.8% │
  └────────────────────────────┴────────┴───────────┴─────────────┴───────────┘

  Churn benchmarks:

  ┌─────────────────────────────┬─────────────────┬──────────────────┬──────────┐
  │          Scenario           │      main       │   this branch    │    Δ     │
  ├─────────────────────────────┼─────────────────┼──────────────────┼──────────┤
  │ bind durable queue          │ 427/s (2.34 ms) │ 1,830/s (545 µs) │ +328% ✅ │
  ├─────────────────────────────┼─────────────────┼──────────────────┼──────────┤
  │ create/delete durable queue │ 131/s (7.60 ms) │  455/s (2.20 ms) │ +246% ✅ │
  ├─────────────────────────────┼─────────────────┼──────────────────┼──────────┤
  │ connection+channel churn    │         3.92k/s │          4.59k/s │  +17% ✅ │
  ├─────────────────────────────┼─────────────────┼──────────────────┼──────────┤
  │ bind non-durable            │        100.2k/s │          97.2k/s │      −3% │
  ├─────────────────────────────┼─────────────────┼──────────────────┼──────────┤
  │ create/delete transient     │         1.41k/s │          1.24k/s │     −12% │
  └─────────────────────────────┴─────────────────┴──────────────────┴──────────┘

  Read

  The Sync::Shared/Sync::Exclusive store refactor on this branch trades single-queue contention for big wins where many independent stores are touched:

  Clear wins
  - Durable churn is dramatically faster — bind durable +328%, create/delete durable +246%. The store locking change paid off exactly where it was aimed.
  - Sharded 64q throughput +42% pub / +70% con — independent queues no longer serialize on a shared store lock.
  - Stream consume +18–20%, large messages +11%.

  Regressions worth investigating ⚠️
  1. Publish in tx: −70.6% (99.7k → 29.3k). Largest single regression. Transaction commit path likely acquires a now-heavier lock per commit.
  2. Competing consumers 1→64: −54% on both pub and con. Many consumers on a single queue now contend more than before — a Sync::Exclusive (Mutex) where reads
  dominate, or lock granularity, may be the cause. Worth checking against the [[feedback_sync_exclusive_vs_shared]] guidance (RWLock when reads dominate).
  3. Stream multi-consumer 1→64 pub: −53% (282k → 132k). Same single-queue/many-consumer contention shape.
  4. Broad ~12–18% drop across single-queue baselines (baseline, properties, fanout, confirm, symmetric). A consistent per-operation overhead added to the
  common single-store path.

  Caveats on these numbers

  - These look like single runs, not averaged across repetitions — the high variance bands (e.g. ±20% on durable bind, ±24% on connection churn) mean the
  smaller deltas (anything under ~10%) are within noise. The big ones (tx −70%, competing −54%, sharded +70%, durable churn +300%) are real signal.
  - The CHANNEL_ERROR - Channel 1 not open spam appears on stream tests in both branches, so it's not a regression introduced here — but it may be distorting
  the stream pub/con figures.

  The headline: the branch does what it set out to do (independent-store concurrency and durable churn improve massively), but it regressed the single-queue /
  single-store hot path by 12–18% generally, and badly hurt single-queue-with-many-consumers and transaction-commit paths.

PR #1926, opened 2026-05-11
Expose per-vhost message stats and label channel metrics

oskgu360:
## Summary

Two additions to `/metrics/detailed` to unblock per-vhost monitoring use cases that today require polling the management HTTP API.

### New family: `vhost_message_stats`

| Metric | Type | Labels |
|---|---|---|
| `lavinmq_detailed_vhost_messages_published_total` | counter | `vhost` |
| `lavinmq_detailed_vhost_messages_delivered_total` | counter | `vhost` |
| `lavinmq_detailed_vhost_queues` | gauge | `vhost` |

Sourced from the in-memory counters on the `VHost` (`publish_count`, `deliver_get_count`, `queues.size`). Names follow the existing `global_messages_*_total` / `detailed_<subject>_<metric>` conventions.

### Wire-format change: `channel_metrics` family gains labels

`detailed_channel_consumers`, `detailed_messages_unacked`, and `detailed_channel_prefetch` now emit `{channel, vhost, connection}` instead of just `{channel}`. The new labels let consumers attribute channel-level metrics to the owning vhost and connection (the latter is the server-side `remote -> local` name used to address actions on the connection).

This is a backwards-incompatible wire-format change for these three existing metrics. Most Prometheus tooling tolerates added labels (aggregations by the existing `channel` label keep working), but downstream parsers that assume the exact label set will see new keys.

## Notes

- No protocol label on the vhost metrics. LavinMQ's `Server::Protocol` enum is binary (AMQP / MQTT) and the counters are protocol-agnostic at the vhost level. A `{protocol}` label could be added later in a backwards-compatible way.
- Cardinality: bounded by vhost / channel counts on the broker — same scaling as the existing per-X families.
- `MetricLabels` alias extended with new NamedTuple shapes for type-checking.

## Test plan

- [x] `crystal tool format` clean
- [x] `make lint` clean (347 / 0)
- [x] `make test SPEC=spec/api/prometheus_spec.cr` — 16/16 pass (2 new tests covering each addition)
- [x] `make test SPEC=spec/queue_spec.cr` — 60/60 pass (smoke)
- [x] Verify on a live broker via `curl /metrics/detailed?family=vhost_message_stats&family=channel_metrics`

```
curl 'http://guest:guest@localhost:15672/metrics/detailed?family=vhost_message_stats'
# TYPE lavinmq_detailed_vhost_messages_published_total counter
# HELP lavinmq_detailed_vhost_messages_published_total Total messages published to a vhost (all protocols)
lavinmq_detailed_vhost_messages_published_total{vhost="/"} 0
lavinmq_detailed_vhost_messages_published_total{vhost="cloudamqp"} 2
....
```


snichme (approved):

jage:
What is `messages_published_total` here? Messages publish to an exchange? Or number of messages publish to all queues in a vhost? Would be nice with a clear description and/or naming. 

What I'm thinking about are these scenarios:

* 1 publish to fan out exchange to 5 queues, will 1 publish count as 1 or 5?
* 1 publish to an exchange without any bindings, count as 1 or 0?

oskgu360:
> What I'm thinking about are these scenarios:
>
> 1 publish to fan out exchange to 5 queues, will 1 publish count as 1 or 5?
> 1 publish to an exchange without any bindings, count as 1 or 0?

The counter reused (`vhost.publish_count`) is naively counting publishes from clients, no matter what happens in the exchange. That is reflected on messge deliveries though

PR #1928, opened 2026-05-11
Add lavinmqctl stop_app and start_app commands

### LINKED ISSUE #1927: Lavinmqctl stop_app and start_app
Implement stop_app and start_app commands in lavinmqctl, which were already listed in COMPAT_CMDS but had empty handlers.

Relates to [#1797](https://github.com/cloudamqp/lavinmq/issues/1797) stop_app is intended as the required precondition before lavinmqctl reset.

Loofeit:
## Summary

- Adds `PUT /api/broker/stop` and `PUT /api/broker/start` HTTP management endpoints
- Wires `stop_app` and `start_app` commands in `lavinmqctl`
- `stop_app` stops the AMQP/MQTT broker layer (closes connections, rejects new ones) while keeping the HTTP management API and etcd lease alive
- `start_app` restarts the broker layer, reloading vhosts and user store from disk
- Both commands require administrator privileges

## Test plan

- [ ] `lavinmqctl stop_app` exits 0 and sets server to closed state
- [ ] `lavinmqctl start_app` exits 0 and reopens the broker after a stop
- [ ] HTTP management API (`/api/overview`) remains reachable after `stop_app`
- [ ] `make test SPEC=spec/lavinmqctl_spec.cr` — 40 examples, 0 failures

Closes #1927

### DIFF
--- a/src/lavinmq/http/controller/main.cr
+++ b/src/lavinmq/http/controller/main.cr
@@ -162,6 +162,20 @@ module LavinMQ
           Tuple.new.to_json(context.response)
           context
         end
+
+        put "/api/broker/stop" do |context, _params|
+          refuse_unless_administrator(context, user(context))
+          @amqp_server.stop
+          context.response.status_code = 204
+          context
+        end
+
+        put "/api/broker/start" do |context, _params|
+          refuse_unless_administrator(context, user(context))
+          @amqp_server.restart
+          context.response.status_code = 204
+          context
+        end
       end
     end
   end
--- a/spec/lavinmqctl_spec.cr
+++ b/spec/lavinmqctl_spec.cr
@@ -352,6 +352,33 @@ describe "LavinMQCtl" do
       end
     end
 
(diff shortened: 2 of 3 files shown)


src/lavinmq/http/controller/main.cr
```diff
@@ -162,6 +162,20 @@ module LavinMQ
...
+        put "/api/broker/stop" do |context, _params|
+          refuse_unless_administrator(context, user(context))
+          @amqp_server.stop
+          context.response.status_code = 204
+          context
+        end
+
+        put "/api/broker/start" do |context, _params|
+          refuse_unless_administrator(context, user(context))
+          @amqp_server.restart
```
spuun, line 175: This method has only been used in specs... We can probably not rely on it.
We should probably make sure `#stop` and `#restart` are well tested for this kind of stop/start scenario. 

And maybe start should be a noop if "app" is running?

spuun (commented):
I'm not sure it's this easy...

PR #1933, opened 2026-05-12
Updated login design with SSO button

bengtmagnus:
Updates the login page with a two column look and adds a SSO button.


spuun:
I guess this is related to something and shouldn't be merged until the button does something?

bengtmagnus:
Yes, @kickster97 has used some of it already in her oauth PR. I'll tweak the design a bit but now the button exists and is ready to be used.

spuun:
What about closing this and reopening against `mgmt-oauth`? Would that make sense @kickster97?

bengtmagnus:
This is ready from my end

bengtmagnus:
But if it is easier I can redo it in @kickster97 pr...

kickster97:
> What about closing this and reopening against `mgmt-oauth`? Would that make sense @kickster97?


> This is ready from my end


yeah, if its ready then i think we should merge this into the mgmt-oauth pr :)

PR #1935, opened 2026-05-12
bugfix: MQTT max inflight message flow control

spuun:
I discovered some issues when working with #1920.

Two bugs in MQTT::Session around the max_inflight_messages limit:

#### Bug 1 — busy-loop when inflight limit is reached
When `@unacked` hit `max_inflight_messages`, `next_id` returned `nil` and `get_packet` returned false, but `deliver_loop` ignored the return value and looped immediately — spinning at 100% CPU until acks arrived. Fixed by adding a `BoolChannel` `@has_capacity` (initialized true) that the deliver loop waits on when at capacity. It is set to false when the last inflight slot is taken and back to true on each ack.

#### Bug 2 — message loss when next_id returns nil
`get_packet` shifts a message from the store before calling `next_id`. If `next_id` returned `nil`, the method returned early without requeuing — the message was silently dropped. Fixed by requeuing sp before returning false in that path. Since the message won't be acked it's not really lost and will come back on restart.

Specs

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -205,10 +205,11 @@ module LavinMQ
 
       def ack(packet : MQTT::PubAck) : Nil
         id = packet.packet_id
-        sp = @unacked[id]
-        @unacked.delete id
-        @has_capacity.set(true)
-        super sp
+        # TODO: disconncet client if sp is nil?
+        if sp = @unacked.delete(id)
+          super sp
+          @has_capacity.set(true)
+        end
       rescue
         raise ::IO::Error.new("Could not acknowledge package with id: #{id}")
       end
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -70,6 +70,7 @@ module LavinMQ
         end
         @unacked.clear
         @unacked_count.set(0, :release)
+        @unacked_bytesize.set(0, :release)
         @has_capacity.set(true)
 
         @consumers.each do |c|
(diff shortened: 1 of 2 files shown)


src/lavinmq/mqtt/session.cr
```diff
@@ -197,11 +206,16 @@ module LavinMQ
...
         id = packet.packet_id
-        sp = @unacked[id]
-        @unacked.delete id
-        super sp
-      rescue
-        raise ::IO::Error.new("Could not acknowledge package with id: #{id}")
+        if sp = @unacked.delete(id)
+          begin
+            super sp
+          rescue ex
```
viktorerlingsson, line 214: should `@has_capacity.set(true)` be set in this rescue as well?

spuun: Yes! Nice!

src/lavinmq/mqtt/session.cr
```diff
@@ -67,6 +69,9 @@ module LavinMQ
           end
         end
         @unacked.clear
+        @unacked_count.set(0, :release)
+        @unacked_bytesize.set(0, :release)
```
viktorerlingsson, line 78: all other uses of these vars use `:relaxed`. Any reason to use `:release` here?`

viktorerlingsson (approved):

src/lavinmq/mqtt/session.cr
spuun: The idea was to make sure they are set to 0 after the clear, but thinking about it it doesn't matter. There's still a race that another thread would add to `@unacked`, then the first thread clears the hash and set counter to zero before the other threads has increased the counter.

I think doing stuff in another order in `#client=` is the way to go. Since we know there's only one client/consumer we just have to make sure to clear _after_ we've removed the current client.

kickster97 (approved):

PR #1945, opened 2026-05-19
fix: handle past segments and requeued msgs in purge_all

viktorerlingsson:
## Summary

`purge_all` crashed in two ways when segments other than rfile/wfile were present:

- `OverflowError`: `@size -= @segment_msg_count[seg]` underflowed when a segment had been read past but not fully acked. `@segment_msg_count` counts msgs ever written; `@size` only counts ready+requeued.
- `KeyError`: `@requeued.clear` ran at the end of `purge_all`, so the `while shift?` drain looked up an sp's segment after step 1 had already deleted it.

Drain `@requeued` upfront (keeping `@size`/`@bytesize` in sync) so the `shift?` loop never sees a stale sp. Only decrement `@size` by `msg_count` for segments rfile hasn't reached yet. Reset `@size`/`@bytesize` and signal `@empty` at the end (the drain loop won't run when there are no ready msgs left).

Seen in production at LavinMQ 2.7.3 when purging a paused queue with previously-requeued msgs.

## Test plan

Three regression specs in `spec/message_store_spec.cr`:

- [x] `purges multiple segments of unread messages` - catches the EOF regression if the conditional `@size -= msg_count` were dropped.
- [x] `does not raise when an older segment has in-flight messages` - reproduces the original `OverflowError`.
- [x] `does not raise when @requeued points to a segment that gets deleted` - reproduces the original `KeyError`.

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -930,7 +930,7 @@ module LavinMQ::AMQP
 
     def purge(max_count : Int = UInt32::MAX) : UInt32
       delete_count = 0u32
-      @msg_store_lock.synchronize {
+      @msg_store_lock.synchronize do
         if unacked_count == 0 && max_count >= message_count
           # If there's no unacked and we're purging all messages, we can purge faster by deleting files
           delete_count = message_count
@@ -938,7 +938,7 @@ module LavinMQ::AMQP
         else
           delete_count = @msg_store.purge(max_count)
         end
-      }
+      end
       @log.info { "Purged #{delete_count} messages" }
       # Signal expire loop to recalculate wait time for next message
       if delete_count > 0
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -929,13 +929,16 @@ module LavinMQ::AMQP
     end
 
     def purge(max_count : Int = UInt32::MAX) : UInt32
-      if unacked_count == 0 && max_count >= message_count
-        # If there's no unacked and we're purging all messages, we can purge faster by deleting files
-        delete_count = message_count
-        @msg_store_lock.synchronize { @msg_store.purge_all }
-      else
-        delete_count = @msg_store_lock.synchronize { @msg_store.purge(max_count) }
-      end
+      delete_count = 0u32
+      @msg_store_lock.synchronize {
+        if unacked_count == 0 && max_count >= message_count
+          # If there's no unacked and we're purging all messages, we can purge faster by deleting files
+          delete_count = message_count
+          @msg_store.purge_all
+        else
+          delete_count = @msg_store.purge(max_count)
+        end
+      }
       @log.info { "Purged #{delete_count} messages" }
       # Signal expire loop to recalculate wait time for next message
       if delete_count > 0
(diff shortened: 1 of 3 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -929,13 +929,16 @@ module LavinMQ::AMQP
...
     def purge(max_count : Int = UInt32::MAX) : UInt32
-      if unacked_count == 0 && max_count >= message_count
-        # If there's no unacked and we're purging all messages, we can purge faster by deleting files
-        delete_count = message_count
-        @msg_store_lock.synchronize { @msg_store.purge_all }
-      else
-        delete_count = @msg_store_lock.synchronize { @msg_store.purge(max_count) }
-      end
+      delete_count = 0u32
+      @msg_store_lock.synchronize {
```
spuun:
```suggestion
      @msg_store_lock.synchronize do
```

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -929,13 +929,16 @@ module LavinMQ::AMQP
...
+        if unacked_count == 0 && max_count >= message_count
+          # If there's no unacked and we're purging all messages, we can purge faster by deleting files
+          delete_count = message_count
+          @msg_store.purge_all
+        else
+          delete_count = @msg_store.purge(max_count)
+        end
+      }
```
spuun:
```suggestion
      end
```

src/lavinmq/message_store.cr
```diff
@@ -223,8 +229,9 @@ module LavinMQ
         delete(env.segment_position)
       end
 
-      @requeued.clear
+      @size = 0_u32
       @bytesize = 0_u64
+      @empty.set(true)
```
spuun, line 234: If we set the counters to 0 here, why do we even need to bother about them earlier in the method?

Wouldn't it be enough to clear `@requeued` first, then do the segments reject and finally set the counters to 0?

viktorerlingsson: We don't outright delete the active `@wfile` or `@rfile` (i think to avoid potential segfaults if consumers/publishers or replication has work in-flight, and we need at least one segment), so we need to ack every message in those. `shift?` will fail if numbers are not correct (with `EOF but @size=x`), so we need to decrement `@size` before `shift?`. I guess we could rescue that exception also, but that doesn't feel as clean to me. 

Setting the counters explicitly at the end here shouldn't really be needed in most cases, but I think might be good as a defensive "belt and suspenders" thing (if `shift?` returns early or numbers are wrong somehow).

spuun: Yeah, i think it's good to set them explicitly. But then `@requeued.clear` should be enough instead of the loop to clear it with `#shift?, as long as we're doing it before `@segments.reject!`?

viktorerlingsson: >But then @requeued.clear should be enough instead of the loop to clear it with #shift?, as long as we're doing it before @segments.reject!`?

That would leave `@size` out of sync with reality, since `@segments.reject!` doesn't decrement `@size` for read/partially read segmenst, and then causes `EOF but @size > 0`  from `shift?`

We could do something like this I guess? 
```
   @size -= @requeued.size
   @requeued.clear
```

viktorerlingsson:
This should probably be combined with #1957

viktorerlingsson:
Superseded by #1957

PR #1946, opened 2026-05-20
refactor: remove unused methods and stdlib patches

viktorerlingsson:
Removes methods and stdlib monkey-patches with no callers in `src/` or `spec/`.

- `Array`/`Hash`/`Set#capacity` stdlib patches (callers were on `MFile`, which has its own).
- Various unused public methods: `Queue#consumers_any?`/`each_consumer`/`first_consumer`, `UserStore#has_key?`, `ParameterStore#any?`, `VHostStore#first_value`, `Server#listen_clustering`, `SNIConfig#each`, `DataDirLock#poll`, `Etcd#get_prefix`/`update_endpoints`, `Message#delay`, `MFile#msync` (no-arg), `PriorityQueue#store_for(msg, &)`, `TopicTree#[]`.
- Unused `to_s`/`inspect` overrides on GIS filters, MQTT subscription/topic trees, and `StringTokenIterator` (no log line or string interpolation referenced them).

## Test plan

- [x] `make test`
- [x] `make lint`
- [x] `crystal tool format`
- [x] Release build
- [x] Frontend specs pass
- [x] Cross-compile: Linux aarch64, FreeBSD x86_64, macOS aarch64
- [x] `lavinmqperf` smoke
- [x] Management UI click-through

### DIFF
--- a/src/lavinmq/amqp/queue/priority_queue.cr
+++ b/src/lavinmq/amqp/queue/priority_queue.cr
@@ -118,12 +118,6 @@ module LavinMQ::AMQP
         end
       end
 
-      private def store_for(msg, &)
-        store_for(msg.properties.priority || 0u8) do |store|
-          yield store
-        end
-      end
-
       def size
         @stores.sum(&.size)
       end
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -82,18 +82,6 @@ module LavinMQ::AMQP
       @consumers.empty?
     end
 
-    def consumers_any?(& : Client::Channel::Consumer -> Bool) : Bool
-      @consumers.any? { |c| yield c }
-    end
-
-    def each_consumer(& : Client::Channel::Consumer ->) : Nil
-      @consumers.each { |c| yield c }
-    end
-
-    def first_consumer : Client::Channel::Consumer?
-      @consumers.first?
-    end
-
     # BasicGet unacked accessors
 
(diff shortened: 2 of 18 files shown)


kickster97 (approved):
just curious: did you run the unreachable tool with this? :)

spuun:
IMO we can keep `#to_s` and `#inspect`.

viktorerlingsson:
> just curious: did you run the unreachable tool with this? :)

Yes, I had Claude run that and check each entry one-by-one

viktorerlingsson:
> IMO we can keep `#to_s` and `#inspect`.

@spuun even if they are never used? As far as I can tell we never use them in logs or anything.

spuun:
> @spuun even if they are never used? As far as I can tell we never use them in logs or anything.

Yeah, i think it's nice to have them in place for debugging and such. Not gonna die on the hill though :)

viktorerlingsson:
> Yeah, i think it's nice to have them in place for debugging and such. Not gonna die on the hill though :)

Sure, but one can also add them back for debugging if/when needed 🤷

carlhoerberg:
I think I agree that `to_s` and `inspect` can stay.

I'm a bit surprised that `Message#delay` is unused, what do we use instead?

viktorerlingsson:
> I think I agree that `to_s` and `inspect` can stay.

> IMO we can keep #to_s and #inspect.

Okay, added them back. 

> I'm a bit surprised that `Message#delay` is unused, what do we use instead?

It's just inlined (in `segment_position.cr:24` and `exchange.cr:300`)

PR #1947, opened 2026-05-21
Stress tester

carlhoerberg:
`extras/stress.cr` is a multi-fiber AMQP+HTTP load generator that runs
against an externally-started lavinmq. It deliberately interleaves
publishes, consumes, and basic_get against churning topology — bind/
unbind, exchange/queue declare-delete, connection kills, vhost churn,
policy add/remove, purges on live queues — plus a dedicated stream
workload that forces segment rollover and consumer-offset cleanup. 

Two extra vhosts (stress-fed-a, stress-fed-b) each carry fed.ex/fed.q/fed.qf
with exchange and queue federation upstreams pointing at the other side.
Producers/consumers run on both vhosts so traffic flows both directions, and
a churn worker tears down and re-PUTs upstream parameters under load to
exercise link teardown/reconnect.

```sh
make stress

### DIFF
--- a/extras/stress.cr
+++ b/extras/stress.cr
@@ -1,0 +1,1620 @@
+# Stress driver for LavinMQ.
+#
+# Spawns many concurrent AMQP and HTTP workers against a running server,
+# deliberately interleaving traffic with topology churn (bind/unbind,
+# declare/delete, connection kills, vhost churn, dead-lettering, TTLs) to
+# surface multi-threading bugs.
+#
+# Usage:
+#   crystal run --release extras/stress.cr -- --duration 60 --concurrency 8
+#
+# The server must be running separately, e.g.:
+#   bin/lavinmq --data-dir ./tmp/stress --debug
+
+require "amqp-client"
+require "http/client"
+require "option_parser"
+require "uri"
+require "json"
+require "base64"
+require "log"
+
+count = Fiber::ExecutionContext.default_workers_count
+Fiber::ExecutionContext.default.resize(count)
+
+module Stress
+  alias Arguments = AMQ::Protocol::Table
+  alias Properties = AMQ::Protocol::Properties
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: extras/benchmark.sh [hunk 1]
```diff
@@ -0,0 +1,142 @@
...
+
+PERF=${PERF:-bin/lavinmqperf}
+CTL=${CTL:-bin/lavinmqctl}
+DURATION=${DURATION:-10}
+URI=${URI:-amqp://guest:guest@localhost}
+
+if [ ! -x "$PERF" ]; then
+    echo "$PERF not found or not executable. Build it with: make $PERF" >&2
+    exit 1
+fi
```
viktorerlingsson: maybe do the same check for lavinmqctl?

extras/stress.cr
```diff
@@ -0,0 +1,1620 @@
...
+  # offsets file approaches its capacity and `consumer_offset_file_full?`
+  # returns true, forcing `cleanup_consumer_offsets`. Combined with segment
+  # rollover this exercises both cleanup paths.
+  def self.stream_consumer_churn(id : Int32)
+    cycle = 0_u64
+    until STOP.get
+      cycle &+= 1
+      tag = "#{STREAM_TAG_PREFIX}churn-#{id}-#{cycle}-#{Random.rand(1_000_000_000)}"
+      begin
+        AMQP::Client.start(CFG.amqp_url) do |conn|
```
viktorerlingsson, line 924: A second `with_amqp` helper that takes some arguments could de-dup some boilerplate here and in `stream_varied_offset_consumer` & `stream_filtered_consumer`

viktorerlingsson (approved):
LGTM 👍  
Doesn't touch any actual LavinMQ code so should be very safe.

PR #1951, opened 2026-05-24
fix: resolve -q flag collision between QoS and quiet in mqttperf

### LINKED ISSUE #1932: lavinmqperf mqtt throughput `-q` for QoS doesn't work
It's not possible to use `-q` for QoS, it's overridden by `-q` for quiet. This results in that if you run with `-q 1` lavinmqperf will exit and print usage because `1` is an invalid argument for quiet.

SAY-5:

The mqttperf throughput command registered `-q qos` for QoS and `-q`/`--quiet` for quiet output. The two share the short `-q` flag, so the later `--quiet` definition wins and `-q 1` is parsed as quiet plus an invalid `1` argument, printing usage and exiting (issue #1932).

This drops the colliding short alias from the QoS option, leaving `--qos=level`. `-q` now unambiguously means `--quiet`, matching the AMQP throughput tool, which uses `-q` for quiet only.

Before: `lavinmqperf mqtt throughput -q 1` exits with usage. After: `--qos 1` sets QoS 1 and `-q` toggles quiet. Verified the OptionParser behavior in isolation: `--qos 1` => qos=1, `-q` => quiet=true, no invalid-option error.

Closes #1932

### DIFF
--- a/src/lavinmqperf/mqtt/throughput.cr
+++ b/src/lavinmqperf/mqtt/throughput.cr
@@ -41,7 +41,7 @@ module LavinMQPerf
         @parser.on("-V", "--verify", "Verify the message body") do
           @verify = true
         end
-        @parser.on("-q qos", "--qos=level", "QoS level (0 or 1)") do |v|
+        @parser.on("--qos=level", "QoS level (0 or 1)") do |v|
           @qos = v.to_i
         end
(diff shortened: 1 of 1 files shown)


kickster97 (approved):
Thanks for the contribution! I think this makes sense

viktorerlingsson:
We could also change it to `-Q` for `QoS` maybe? if we want a shorthand. I think removing is fine as well though!

PR #1955, opened 2026-05-25
Detach Stream envelope body off mmap to survive segment drop

carlhoerberg:
## Summary
- `StreamMessageStore#read` and `#shift?` returned `BytesMessage` envelopes whose body slice pointed directly into the segment's mmap.
- Stream consumers release `@msg_store_lock` before delivering, so `drop_segments_while` could `munmap` the segment while a consumer was still copying body bytes — leading to SIGBUS or garbled deliveries.
- Detach the body onto a heap-allocated buffer before returning the envelope so it survives a concurrent segment drop.

## Test plan
- [x] Regression spec added in `spec/stream_reader_spec.cr` reproducing the segment-drop race
- [ ] `make test SPEC=spec/stream_reader_spec.cr`
- [ ] `make test`

### DIFF
--- a/spec/stream_reader_spec.cr
+++ b/spec/stream_reader_spec.cr
@@ -54,6 +54,53 @@ describe LavinMQ::AMQP::StreamReader do
       end
     end
   end
+  it "envelope body survives concurrent segment drop" do
+    # Regression: shift? / read returned BytesMessage whose body pointed
+    # directly into the segment's mmap. drop_segments_while munmaps the
+    # segment, and a slow consumer that's still copying bytes into the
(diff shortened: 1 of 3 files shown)


carlhoerberg:
I have an optional implementation where each stream consumer has it's own read-only `File` that it reads from instead.

carlhoerberg:
Superseded by #1965 — per-consumer FD approach makes the mmap-detach workaround unnecessary, since body bytes are now read straight from a per-consumer FD into the socket. Both regression specs from this PR are retained in #1965 and pass against the new implementation.

PR #1956, opened 2026-05-25
Fix Stream queue crash when delivery-limit policy is applied

carlhoerberg:
## Summary
- A `delivery-limit` policy matching a stream queue fell through to `AMQP::Queue#apply_policy_argument`, which spawned the `drop_redelivered` fiber.
- That fiber called the inherited `MessageStore#first?` and dereferenced the legacy `@rfile` — which streams don't maintain through `drop_segments_while` — pointing at a closed mfile from a long-dropped segment. The spawn died with `IO::Error("Closed mfile")`.
- Lift Stream's invalid-args set to a single `INVALID_ARGUMENTS` constant shared between declare-time `validate_arguments!` and `apply_policy_argument`. What's forbidden as a queue argument is now also a no-op as a matching policy key.
- Also adds `MessageStore` diagnostic logging (`state_snapshot` in `first?`/`shift?` rescues, debug logs around `select_next_read_segment`).

## Test plan
- [x] Regression spec added in `spec/stream_queue_spec.cr`
- [ ] `make test SPEC=spec/stream_queue_spec.cr`
- [ ] `make test`

### DIFF
--- a/spec/stream_queue_spec.cr
+++ b/spec/stream_queue_spec.cr
@@ -349,6 +349,42 @@ describe LavinMQ::AMQP::Stream do
       end
     end
 
+    # Regression: a `delivery-limit` policy matching a stream queue used to
+    # fall through to AMQP::Queue#apply_policy_argument, which spawned a
+    # drop_redelivered fiber. That fiber called the inherited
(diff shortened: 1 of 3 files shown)


carlhoerberg:
Or _allowlist_ allowed policy arguments instead?

PR #1957, opened 2026-05-25
Fix MessageStore @size leaks in purge_all and requeued shift?

carlhoerberg:
## Summary

Three accounting bugs in `MessageStore` that surfaced as `EOF but @size=N` crashes during `queue_purge` under the MT stress driver.

### 1. `purge_all` under-counted acked-then-purged segments
Phase 1 subtracted `@segment_msg_count[seg]` (total writes, including acks) from `@size`, but `@size` only tracks ready (unshifted) messages. Any segment with partial acks under-counted `@size`, and the next shift would underflow.

Initial fix subtracted `msg_count - acks_in_segment` and added a safety-net reset of `@size`/`@bytesize`/`@empty` at the end.

### 2. `purge_all` still underflowed on in-flight and requeued msgs
The `msg_count - ack_count` formula still underflows `@size` when past segments contain in-flight msgs (`shift?` already decremented `@size`, but they're not in the ack file) or requeued msgs (counted in `@size` via `@requeued` **and** in `msg_count`).

`purge_all` now drains `@requeued` first, then only decrements `@size` for unread segments (`seg_id > @rfile_id`). Past segments don't contribute to `@size`: msgs there are either acked, in-flight, or were drained above.

### 3. `shift?` from `@requeued` leaked counters on segment errors
`shift?` removed the `SegmentPosition` from `@requeued` before resolving the segment, so a `KeyError` (segment gone) or a parse error in `BytesMessage.from_bytes` left `@size`/`@bytesize` counting an entry that could never be re-read — a permanent leak. The failure branch now decrements both counters.

Incorporates changes from #1945.

## Test plan
- [x] Regression specs cover partial-ack purge, in-flight msgs in past segments, requeued sps pointing to deleted segments, and the safety-net reset
- [x] `make test SPEC=spec/message_store_spec.cr`
- [ ] `make test`

### DIFF
--- a/src/lavinmq/message_store.cr
+++ b/src/lavinmq/message_store.cr
@@ -108,15 +108,22 @@ module LavinMQ
     def shift?(consumer = nil) : Envelope? # ameba:disable Metrics/CyclomaticComplexity
       raise ClosedError.new if @closed
       if sp = @requeued.shift?
-        segment = @segments[sp.segment]
         begin
+          segment = @segments[sp.segment]
           msg = BytesMessage.from_bytes(segment.to_slice + sp.position)
           @bytesize -= sp.bytesize
           @size -= 1
           @empty.set true if @size.zero?
           return Envelope.new(sp, msg, redelivered: true)
         rescue ex
-          raise Error.new(segment, cause: ex)
+          # sp has already been removed from @requeued; drop its accounting too
+          # so @size/@bytesize don't leak when the segment is gone or the
+          # payload can't be parsed. Without this, future shift? calls can hit
+          # EOF with @size > 0 because the requeued entry can never be re-read.
+          @bytesize -= sp.bytesize
+          @size -= 1
+          @empty.set true if @size.zero?
+          raise Error.new(@segments[sp.segment]? || @rfile, cause: ex)
         end
       end
 
@@ -208,8 +215,13 @@ module LavinMQ
         next false if seg_id == @rfile_id || seg_id == @wfile_id
 
         delete_file(file, including_meta: true)
+        # @segment_msg_count tracks total writes (including acked); @size only
+        # counts ready (unshifted) messages, so subtract msg_count - acks_count
+        # instead of msg_count or we underflow @size on segments with acks.
+        ack_count = ((af = @acks[seg_id]?) ? af.size // sizeof(UInt32) : 0).to_u32
         if msg_count = @segment_msg_count.delete(seg_id)
-          @size -= msg_count
+          ready_count = msg_count > ack_count ? msg_count - ack_count : 0_u32
+          @size -= ready_count
         end
         if afile = @acks.delete(seg_id)
           delete_file(afile)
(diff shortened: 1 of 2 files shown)


viktorerlingsson:
I worked on something similar with #1945 last week, and I think we should combine this and that to get the best of both worlds (the `shift?` accounting fix here I think is good and something I missed). 

I tried my specs from that PR against this, and two of them fail with `Arithmetic overflow`, because `ready_count` doesn't consider unacked messages (they have already decremented `@size`, but are not `acked`), and requeued messages in "old" segments will decrement `@size` twice, once because it's included in `ready_count`, and then again during the `@requeued` drain in `shift?`. 

I don't think we have a good way to count `unacked` or `requeued` per segment, so I think the safest way to handle that is by not decrementing `@size` at all for segments that have been read (`seg_id <= @rfile_id`), so not having to do any of the calculations with `ack_count` and `ready_count`. And that means we also have to handle `@requeued` separately in `purge_all`.

Something like: 

````crystal
  def purge_all
    # shift out all requeued messages and decrement @size
    while sp = @requeued.shift?
      @size -= 1
      @bytesize -= sp.bytesize
    end

    @segments.reject! do |seg_id, file|
      next false if seg_id == @rfile_id || seg_id == @wfile_id
      delete_file(file, including_meta: true)
      if msg_count = @segment_msg_count.delete(seg_id)
        # only decrement @size if segment has not been read yet
        @size -= msg_count if seg_id > @rfile_id
      end
      # ... rest unchanged
    end
    # ... rest unchanged
  end
  ````

With those changes, both yours and my specs pass. 

WDYT?


Edit: Incorporated this fix in 22bbe5a

File: src/lavinmq/message_store.cr [hunk 1]
```diff
@@ -108,15 +108,22 @@ module LavinMQ
...
         begin
+          segment = @segments[sp.segment]
           msg = BytesMessage.from_bytes(segment.to_slice + sp.position)
           @bytesize -= sp.bytesize
           @size -= 1
           @empty.set true if @size.zero?
           return Envelope.new(sp, msg, redelivered: true)
         rescue ex
-          raise Error.new(segment, cause: ex)
+          # sp has already been removed from @requeued; drop its accounting too
```
carlhoerberg, line 119: Or should we actually try to put it back into `@requeued` again?

src/lavinmq/message_store.cr
```diff
@@ -203,13 +210,22 @@ module LavinMQ
...
+        @bytesize -= sp.bytesize
+      end
+
       # Delete all segments except the current rfile and wfile
       @segments.reject! do |seg_id, file|
         next false if seg_id == @rfile_id || seg_id == @wfile_id
 
         delete_file(file, including_meta: true)
+        # Only decrement @size for unread segments. Read segments don't
+        # contribute to @size: their msgs are either acked, in-flight, or
```
carlhoerberg, line 225: all of them? what if we've only read/delivered the first X messages out of Y messeages total in the segmnet?

carlhoerberg: oh, that comes next, were we `shift?` out the rest

PR #1958, opened 2026-05-25
Handle reject-publish overflow in HTTP publish and multi-queue routes

carlhoerberg:
## Summary
- Fix the original bug: `HTTP POST /api/exchanges/:vhost/:name/publish` returned 500 when a bound queue rejected a publish due to a reject-publish overflow policy. Now reported as `{routed: false}`, matching what publisher-confirms clients see (NACK).
- Same treatment for the fan-out case where one bound queue overflows but others accept.
- Refactor: replace the `Queue::RejectOverFlow` exception with explicit return types so overflow is no longer signalled via raise/rescue.
  - `Queue#publish` now returns a `Queue::PublishResult` enum (`Ok` / `Dropped` / `Overflow`).
  - `Exchange#publish` and `VHost#publish` return an `Exchange::PublishResult` `@[Flags]` enum (`Routed`, `Overflowed`) — the two are independent, so a publish can be both `Routed` and `Overflowed`.
  - `AMQP::Channel#confirm` now NACKs based on `result.overflowed?` instead of relying on `RejectOverFlow` propagating up.
  - HTTP, federation, MQTT, and dead-lettering call sites branch on the new fields directly.

## Test plan
- [x] Regression spec in `spec/api/exchanges_spec.cr` (HTTP overflow + fan-out)
- [x] `make test` — 1675 examples, 0 failures
- [x] `make lint`

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -236,14 +236,20 @@ module LavinMQ
         end
 
         count = 0u32
+        overflow = false
         queues.each do |queue|
-          if queue.publish(msg)
-            count += 1
-            msg.body_io.seek(-msg.bodysize.to_i64, IO::Seek::Current) # rewind
+          begin
+            if queue.publish(msg)
+              count += 1
+              msg.body_io.seek(-msg.bodysize.to_i64, IO::Seek::Current) # rewind
+            end
+          rescue Queue::RejectOverFlow
+            overflow = true
           end
         end
         @publish_out_count.add(count, :relaxed)
-        @unroutable_count.add(1, :relaxed) if count.zero?
+        @unroutable_count.add(1, :relaxed) if count.zero? && !overflow
+        raise Queue::RejectOverFlow.new if overflow
         count.positive?
       end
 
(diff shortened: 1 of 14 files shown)


src/lavinmq/amqp/exchange/exchange.cr
```diff
@@ -236,14 +236,20 @@ module LavinMQ
...
+              msg.body_io.seek(-msg.bodysize.to_i64, IO::Seek::Current) # rewind
+            end
+          rescue Queue::RejectOverFlow
+            overflow = true
           end
         end
         @publish_out_count.add(count, :relaxed)
-        @unroutable_count.add(1, :relaxed) if count.zero?
+        @unroutable_count.add(1, :relaxed) if count.zero? && !overflow
+        raise Queue::RejectOverFlow.new if overflow
```
spuun: Do we even need to raise this? Reading through the code (not only this PR) we could probably make `Queue#publish_internal` rescue the `RejectOverFlow` and just return false?

As it is today, where `RejectOverFlow` is rescued we just "ignore" it.

carlhoerberg: it is used, once, to Nack publish confirm messages when a queue has overflowed. But it's indeed not pretty to use exceptions as control flow, let's try return an enum instead.

PR #1961, opened 2026-05-25
Skip delete_message after queue close

carlhoerberg:
## Summary
- Avoid `MessageStore::ClosedError` when a dead-letter routed callback (or any in-flight `delete_message`) races with `Queue#close`.
- The source store is being torn down, so the per-sp bookkeeping is a no-op.
- Surfaced under heavy DLX churn (max-length + DLX configs) where the channel logged "Unexpected error in dead letter routed callback".

## Test plan
- [x] Regression spec added in `spec/queue_dead_lettering_spec.cr`
- [ ] `make test SPEC=spec/queue_dead_lettering_spec.cr`
- [ ] `make test`

### DIFF
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -558,7 +558,7 @@ module LavinMQ::AMQP
     end
 
     protected def publish_internal(msg : Message, dlx_tasks : Argument::DeadLettering::Tasks? = nil) : Bool
-      return false if @deleted || @closed
+      return false if @closed
       if d = @deduper
         if d.duplicate?(msg)
           @dedup_count.add(1, :relaxed)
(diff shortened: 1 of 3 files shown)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -836,6 +836,11 @@ module LavinMQ::AMQP
     end
 
     protected def delete_message(sp : SegmentPosition) : Nil
+      # Close tears down @msg_store under @msg_store_lock; a dead-letter routed
+      # callback (or any other in-flight delete) racing with close would hit
+      # MessageStore::ClosedError on @msg_store.delete. The store is gone
+      # anyway, so the per-sp bookkeeping is a no-op.
+      return if closed?
```
spuun, line 843: Shouldn't this check be in `#ack`? We should probably have this guard in all (most) public methods?

carlhoerberg: dead lettering don't call `ack` directly, hence why we need it here. but made it more consistent overall (just check `closed`, which implies `deleted` etc)

PR #1966, opened 2026-05-25
ci: pin @stoplight/spectral-rulesets to 1.22.2

viktorerlingsson:
## Summary
- Pin `@stoplight/spectral-rulesets` to `1.22.2` in the `lint-openapi` target.
- `1.22.3` introduced a regression in the `duplicated-entry-in-enum` rule that crashes Spectral with `Cannot read properties of null (reading 'enum')`. See [stoplightio/spectral#2959](https://github.com/stoplightio/spectral/issues/2959).
- Fix is in flight upstream ([#2960](https://github.com/stoplightio/spectral/pull/2960) / [#2963](https://github.com/stoplightio/spectral/pull/2963))

## Test plan
- [x] `make lint-openapi` passes locally with the pin (0 errors, pre-existing warnings only).
- [x] Lint OpenAPI passes in CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ lint-js:
 
 .PHONY: lint-openapi
 lint-openapi:
-	npx --package=@stoplight/spectral-cli spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
+	npx --yes -p @stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
 
 .PHONY: test
 test: lib views
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ lint-js:
 
 .PHONY: lint-openapi
 lint-openapi:
-	npx --yes -p @stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
+	npx --yes --package=@stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
 
 .PHONY: test
 test: lib views
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ lint-js:
 
 .PHONY: lint-openapi
 lint-openapi:
-	npx --yes --package=@stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
+	npx --yes --package=@stoplight/spectral-cli --package=@stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
 
 .PHONY: test
 test: lib views


dentarg:
> * Fix is in flight upstream ([#2960](https://github.com/stoplightio/spectral/pull/2960) / [#2963](https://github.com/stoplightio/spectral/pull/2963)); remove the pin once released.

No, move the pin forward :)  ...as we talked about in Slack

Makefile
```diff
@@ -108,7 +108,7 @@ lint-js:
 
 .PHONY: lint-openapi
 lint-openapi:
-	npx --package=@stoplight/spectral-cli spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
+	npx --yes -p @stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
```
dentarg: IMHO we should prefer long flags, easier to understand

```suggestion
	npx --yes --package=@stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
```

Makefile
```diff
@@ -108,7 +108,7 @@ lint-js:
 
 .PHONY: lint-openapi
 lint-openapi:
-	npx --package=@stoplight/spectral-cli spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
+	npx --yes --package=@stoplight/spectral-cli -p @stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
```
dentarg: Maybe my suggestion was bad? Diff does not look quite right now :)

dentarg: Or is it two packages we need?

viktorerlingsson:
> > * Fix is in flight upstream ([#2960](https://github.com/stoplightio/spectral/pull/2960) / [#2963](https://github.com/stoplightio/spectral/pull/2963)); remove the pin once released.
> 
> No, move the pin forward :)
> 

Oops, removed that part of the description.

Not sure how we should handle moving the pin forward. Would be nice if that is handled almost automatically (like dependabot?)

dentarg:
> Not sure how we should handle moving the pin forward. Would be nice if that is handled almost automatically (like dependabot?)

No, I think it is a feature that it is not handled automatically. We should upgrade when there's a need, not because there's a new version. In this area, I think that is a completely fine approach.

Makefile
dentarg: This must be what we want?

```suggestion
	npx --yes --package=@stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
```

viktorerlingsson: Ah, yes, both spectral-cli and spectral-rulesets

viktorerlingsson: I think it should be 
```
npx --yes --package=@stoplight/spectral-cli --package=@stoplight/spectral-rulesets@1.22.2 spectral --ruleset openapi/.spectral.json lint static/docs/openapi.yaml
```

viktorerlingsson:
> No, I think it is a feature that it is not handled automatically. We should upgrade when there's a need, not because there's a new version. In this area, I think that is a completely fine approach.

Okay! Yeah I guess there's no need to be bleeding edge on lint rules :)

Makefile
dentarg: Ah right, it is the `@stoplight/spectral-cli` dependency `@stoplight/spectral-rulesets` we need to pin. That makes me wonder how many other dependencies `@stoplight/spectral-cli` have? That we aren't pinning?

...19 in total it seems: https://www.npmjs.com/package/@stoplight/spectral-cli?activeTab=dependencies 😐

PR #1975, opened 2026-05-28
ci: add docs check action

kickster97:
Adds a GitHub Action that runs on every PR and decides whether the code changes require accompanying updates to `docs/`. Publishes a 3-state check run (success / neutral / failure) with a one-line reason in the summary. Non-blocking at rollout.

Pieces:

- `.github/workflows/docs-check.yml` runs `anthropics/claude-code-action@v1` with `claude-opus-4-7`, short-circuits docs-only PRs, and creates a check run via `gh api`.
- `.claude/docs-check-criteria.md` is the prompt Claude follows to judge user-impact and emit a JSON verdict at `/tmp/docs-verdict.json`.
- `.review-config` gains a `DOCS_CRITERIA` line so the workflow can locate the criteria file. The existing review continues to use `CRITERIA`. `MODEL` is shared.

Draft for now so we can validate accuracy against real PRs before flipping it to required.

### DIFF
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -21,7 +21,6 @@ jobs:
       pull-requests: read
       checks: write
       id-token: write
-      actions: read
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -1,0 +1,127 @@
+name: Docs Check
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - "src/**/*.cr"
+      - "docs/**/*.md"
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to check'
+        required: true
+        type: string
+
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  docs-check:
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+      checks: write
+      id-token: write
+      actions: read
```
viktorerlingsson: Not sure if we need these?
```suggestion
```

viktorerlingsson (approved):

.github/workflows/docs-check.yml
kickster97: the read we can probably drop cause we dont need to read the CI run state, but we need to keep `id-token` i think because the `claude-code-action` uses `OIDC` to exchange for its app token, the other claude workflow uses it as well (and others use it for depot)

PR #1982, opened 2026-05-28
Timeout all specs within 15s

carlhoerberg:
## What

Make every spec example time out on its own instead of being able to hang the whole suite, then fix the latent timing-fragile specs that a hard per-example cap exposes on CI.

### Per-example timeout
An `around_each` hook runs each example in its own fiber and fails it if it doesn't finish within `SPEC_TIMEOUT` (15s; 4× for examples tagged `slow`). A second, stacked `around_each` gives every example its own data dir, nested *inside* the timeout hook so a timed-out example's cleanup is skipped rather than colliding with the next example's dir.

### Hardening (all spec-level — no production code changes)
With a hard cap in place, several latent races started failing on the multi-threaded macOS CI runner. None reproduced on Linux, even under heavy CPU contention; they were ground-truthed by reading the macOS CI logs.

- **wait_for/should_eventually poll with `sleep`** instead of busy-spinning `Fiber.yield`, which under `-Dpreview_mt` pins a scheduler thread and starves the IO fibers it's waiting on.
- **Per-example timeout is the real cap.** `wait_for`/`should_eventually` and the `be_receiving`/`be_sending` matchers default to 15s, so no inner deadline fires before — or outlives — the example timeout. `slow` examples that genuinely need longer use explicit sleeps/`select` timeouts.
- **Read-fiber deadlock.** Buffer the channels written from `subscribe` callbacks in the shovel "forever"/"reconnect" specs: an unbuffered send blocked the connection's read fiber, deadlocking a `publish_confirm` issued on the same connection.
- **Shovel is at-least-once.** Wait for the source to durably ack moved messages before pausing/terminating in the pause-resume and "ack all messages" specs. Disrupting the shovel before destination confirms land (correctly) requeues in-flight messages, which the specs misread as duplication/leftover.
- **Leaked shovels.** Terminate spawned shovels in an `ensure` block so a still-running shovel doesn't leak reconnect-loop fibers into later examples.
- **Federation races.** Wait on the `consumers_empty` `when_false` edge (a consumer *is* present) and poll `link.state.running?` instead of a bare `@state_changed.receive` that blocks forever if the transition already happened.
- **Dead-letter TTL.** Widen the TTL margin in the dead-letter-while-unacked spec so the message isn't expired in the queue before `q.get` retrieves it under load.
- **Slow publish loops** use batch publish + `wait_for_confirms` instead of per-message `publish_confirm`.
- `make lint` pulls down ameba if it isn't present.

## Verification
`make test` and `make lint` clean locally; full suite green on both Linux and macOS CI (1691 examples, 0 failures). The shovel suite was also run 180+ times under deliberate CPU contention to confirm the contention-induced failures are gone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,12 @@ js: $(JS)
 .PHONY: deps
 deps: js lib views
 
+lib/ameba/bin/ameba:
+	shards install
+
 .PHONY: lint
-lint: lib
-	lib/ameba/bin/ameba src/ spec/
+lint: lib/ameba/bin/ameba
+	$< src/ spec/
 
 .PHONY: lint-js
 lint-js:
--- a/spec/etcd_spec.cr
+++ b/spec/etcd_spec.cr
@@ -94,7 +94,7 @@ describe LavinMQ::Etcd, tags: "etcd" do
     end
   end
 
-  it "will lose lease when loosing quorum" do
+  it "will lose lease when loosing quorum", tags: "slow" do
     cluster = EtcdCluster.new
     cluster.run do |etcds|
       etcd = LavinMQ::Etcd.new(cluster.endpoints)
@@ -108,7 +108,7 @@ describe LavinMQ::Etcd, tags: "etcd" do
     end
   end
 
-  it "signals expired channel with the underlying error" do
+  it "signals expired channel with the underlying error", tags: "slow" do
     cluster = EtcdCluster.new
     cluster.run do |etcds|
       etcd = LavinMQ::Etcd.new(cluster.endpoints)
@@ -125,7 +125,7 @@ describe LavinMQ::Etcd, tags: "etcd" do
     end
   end
 
-  it "signals expired channel when all lease is revoked" do
+  it "signals expired channel when all lease is revoked", tags: "slow" do
     cluster = EtcdCluster.new(1)
     cluster.run do
       etcd = LavinMQ::Etcd.new(cluster.endpoints)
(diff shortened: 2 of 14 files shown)


viktorerlingsson:
I'm getting `Invalid memory access (signal 7) at address...` and some spec failures when trying this locally (same as the macOS specs). I can pretty reliably trigger it with `make test SPEC=spec/msg_segment_meta_files_spec.cr`. 

I guess the `publish_confirm` changes in #1891 made those specs run much slower, causing them to timeout with this change. And since the spawned fiber (and the amqp_server that runs in it) isn't cancelled on timeout, the next spec starts up in the same datadir before the old one exited, causing the invalid memory. 


For reference, I did some quick benchmarks on my machine, and I'm getting 
`bin/lavinmqperf amqp throughput -c 1 -z 30`
```
Average publish rate: 1202 msgs/s
Average consume rate: 1202 msgs/s
```
 as compared to before #1891
```
Average publish rate: 20671 msgs/s
Average consume rate: 20670 msgs/s
``` 

So those specs are running way slower.

carlhoerberg:
Thank you! Good find!

carlhoerberg:
No segfaults anymore, but new timeouts

spuun:
Is the idea to find slow specs and fix them? If so we could abort as soon as one specs times out. 

I'm thinking that's a high risk that a timed out spec has dangling fibers doing stuff and may ruin other specs, no?

carlhoerberg:
you mean by adding `--fail-fast` to `crystal spec` in the `make test` target? Maybe, the problem this is solving is that `claude` et al don't properly detect hang specs, it most often waits way longer than a human would. I'm introducing the same "fix" as crystal core got. By introducing a unique data dir per test stray fibers don't interfer with data files from other specs.

PR #1983, opened 2026-05-28
ci: graceful shutdown of LavinMQ after stress test

carlhoerberg:
## Summary
- Send SIGTERM to LavinMQ after the stress test, poll for up to 10s, then SIGKILL and fail the step if it didn't exit gracefully — so we catch shutdown regressions instead of silently masking them.
- Run both `Print LavinMQ log` and `Check LavinMQ log for errors` on `!cancelled()` so we still get the log when SIGKILL was needed.
- Tighten the log error grep to anchored level names (`Unhandled exception|ERROR|FATAL`) to avoid matching the word "error" in unrelated lines.

## Test plan
- [ ] CI passes on this PR with normal stress-test run (graceful shutdown path).
- [ ] Confirm that if LavinMQ were to hang, the SIGKILL fallback triggers and the log is still printed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -552,7 +552,7 @@ jobs:
       - name: Check LavinMQ log for errors
         if: ${{ !cancelled() }}
         run: |
-          if grep -E 'Unhandled exception|ERROR|FATAL' lavinmq.log; then
+          if grep -E '^[^ ]+ +(ERROR|FATAL) |^Unhandled exception' lavinmq.log; then
             echo "::error::Found errors in LavinMQ log"
             exit 1
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -526,20 +526,33 @@ jobs:
...
+
       - name: Print LavinMQ log
         if: ${{ !cancelled() }}
         run: cat lavinmq.log
 
       - name: Check LavinMQ log for errors
+        if: ${{ !cancelled() }}
         run: |
-          if grep -E -i 'error|unhandled exception|fatal|panic|backtrace' lavinmq.log; then
+          if grep -E 'Unhandled exception|ERROR|FATAL' lavinmq.log; then
```
viktorerlingsson: Maybe even anchor the `ERROR|FATAL` check so we don't get any log lines that happen to include "ERROR" in the message? 
```suggestion
          if grep -E '^[^ ]+ +(ERROR|FATAL) |^Unhandled exception' lavinmq.log; then
```

PR #1987, opened 2026-05-29
Add --no-sync option to disable syncfs on publish confirm

carlhoerberg:
## What

Adds a `sync` config option (default `true`). When disabled, the `Persister`'s `sync`/`syncfs` calls become noops, so publish confirms and tx-commit acks are sent without first flushing to disk — durability is left to the OS.

Can be set three ways:
- CLI: `--no-sync`
- INI `[main]`: `sync = false`
- Env: `LAVINMQ_SYNC=false`

## Why

Useful for CI and local dev to speed up runs where on-disk durability of confirmed messages doesn't matter. **Unsafe for production** — a confirmed message can be lost on crash/power loss.

## Notes

`Persister#sync` is reached from both the publish-confirm loop and `vhost.sync` → `tx_commit`, so a single guard covers both client-facing durability points.

## Tests

- Extended `spec/config_spec.cr` to cover the new option via INI and CLI parsing.

### DIFF
--- a/spec/config_spec.cr
+++ b/spec/config_spec.cr
@@ -96,6 +96,7 @@ describe LavinMQ::Config do
           socket_buffer_size = 32768
           tcp_nodelay = true
           segment_size = 16777216
+          sync = false
           tcp_keepalive = 120:20:5
           tcp_recv_buffer_size = 65536
           tcp_send_buffer_size = 65536
@@ -179,6 +180,7 @@ describe LavinMQ::Config do
       config.socket_buffer_size.should eq 32768
       config.tcp_nodelay?.should be_true
       config.segment_size.should eq 16777216
+      config.sync?.should be_false
       config.tcp_keepalive.should eq({120, 20, 5})
       config.tcp_recv_buffer_size.should eq 65536
       config.tcp_send_buffer_size.should eq 65536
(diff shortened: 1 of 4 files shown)


spec/config_spec.cr
viktorerlingsson: For completeness
```suggestion
      ENV["LAVINMQ_SYNC"] = "false"
```

viktorerlingsson:
Since this is an "unsafe flag", maybe there should be a warning on startup if set to false? Like 
>Warning: You are running LavinMQ without FS sync on publish_confirm which could lead to message loss on crash/power loss.

carlhoerberg:
On the other hand rabbit never syncs and don't log anything like that. This could also be used for incompatible clients, that don't accurately support ack with the multiple flag.

spec/config_spec.cr
```diff
@@ -96,6 +96,7 @@ describe LavinMQ::Config do
           socket_buffer_size = 32768
           tcp_nodelay = true
           segment_size = 16777216
+          sync = false
```
spuun, line 99: `sync` feels kind of generic, maybe it be named something more descriptive? `syncfs`, `sync_before_ack`, …

spuun (approved):

spec/config_spec.cr
carlhoerberg: yes, i've gone back and forth on that. like `--no-persist-before-publish-confirm` but it gets a bit long.. postgres got `--no-sync`

PR #1989, opened 2026-05-29
Simplify message store meta specs

spuun:
The specs in spec/msg_segment_meta_files_spec.cr were spinning up full AMQP servers (with_amqp_server), declaring vhosts, opening channels, and publishing through with_channel — all just to exercise the on-disk segment metadata files of a message store. None of that machinery was actually under test.

This change rewrites every example to construct the relevant message store directly:
- Regular/durable queues → LavinMQ::MessageStore.new(dir, nil)
- Stream queues → LavinMQ::AMQP::StreamMessageStore.new(dir, nil)

Messages are pushed straight into the store with store.push(LavinMQ::Message.new(...)) — the same push the AMQP path ultimately calls — so segment roll-over and metadata-file writing behave identically.

Behavioral coverage is unchanged: meta-file creation on segment completion, stream-specific metadata format (count + offset + timestamp), regeneration of missing meta files on load, fallback message scanning, and meta-file cleanup on delete/purge.

Pros
- No fiber spawning — the AMQP listener fibers (and their Fiber.yield settling loops) are gone.
- No sockets — no TCP listener, no client connection, no channel handshake per test.
- No server/vhost/queue lifecycle — skips server boot, vhost creation, queue declaration, and the closed-queue teardown checks.
- Faster — the suite now runs in ~140ms instead of relying on network round-trips and fiber scheduling.
- More direct and readable — each spec talks to the exact component it tests (store.@segments, store.@segment_msg_count, store.purge, store.delete) with no protocol indirection.
- More deterministic — fewer moving parts means fewer timing-sensitive failures and no dependency on connection setup ordering.
- Tighter failures — a broken metadata file now fails in the store layer rather than surfacing as an opaque AMQP/queue error.

Run specs


carlhoerberg (approved):
Fantastic

PR #1993, opened 2026-05-29
fix: strip CC/BCC headers from dead-lettered message when dlrk is set

### LINKED ISSUE #1990: Dead-letter: CC header not stripped from message when x-dead-letter-routing-key is set
## Bug

When a message is dead-lettered and an `x-dead-letter-routing-key` (dlrk) is configured, the `CC` header should be removed from the dead-lettered message's properties. Instead, LavinMQ preserves the `CC` header in the message.

## Failing test

`com.rabbitmq.client.test.functional.DeadLetterExchange.deadLetterNewRK` in the RabbitMQ Java client test suite:

```
expected: <null> but was: <[foo]>
```

The test publishes a message with a `CC` header of `foo`, dead-letters it with a new routing key, and asserts that the dead-lettered message has `CC = null`.

## Root cause

In `src/lavinmq/amqp/argument/dead_lettering.cr`, the `route` method clones the headers and removes `CC`/`BCC` from the clone — but only for routing lookup purposes. The actual message `props` still contains the original `CC` header:

```crystal
# CC/BCC removed from clone used for routing, but NOT from the message itself
if dlrk && (rk = routing_headers.try &.clone)
  rk.delete("CC")
  rk.delete("BCC")
  routing_headers = rk
end
```

## Expected behaviour

when `x-dead-letter-routing-key` is set, `CC` and `BCC` headers should be stripped from the dead-lettered message's properties (not just excluded from the routing lookup).

## Fix

Delete `CC`/`BCC` from `props.headers` directly when `dlrk` is set:

```crystal
if dlrk
  props.headers.try do |h|
    h.delete("CC")
    h.delete("BCC")
  end
  routing_headers = props.headers
end
```

annieblomgren:
## Summary

- When `x-dead-letter-routing-key` is set, `CC` and `BCC` headers are now stripped from the dead-lettered message's properties entirely
- Previously the code only removed them from a cloned headers table used for routing lookup, leaving them intact in the actual message — contrary to RabbitMQ's behaviour
- Adds a regression spec

## Root cause

In `src/lavinmq/amqp/argument/dead_lettering.cr`, the old code cloned `props.headers` and deleted `CC`/`BCC` from the clone for routing purposes, with an explicit comment saying "the header should be maintained". RabbitMQ does not maintain them — the new routing key supersedes CC routing entirely.

## Test plan
- [ ] `DeadLetterExchange.deadLetterNewRK` passes in the RabbitMQ Java client test suite
- [ ] New spec `strips CC header from dead-lettered message when x-dead-letter-routing-key is set` passes

Closes #1990

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -1116,5 +1116,18 @@ module DeadLetteringSpec
         end
       end
     end
+
+    describe "CC header handling" do
+      it "strips CC header from dead-lettered message when x-dead-letter-routing-key is set" do
+        with_dead_lettering_setup do |q, dlq, ch, _|
+          props = AMQP::Client::Properties.new(headers: AMQP::Client::Arguments.new({"CC" => ["extra-rk"]}))
+          ch.default_exchange.publish_confirm("msg", q.name, props: props)
+          q.get(no_ack: false).try &.reject(requeue: false)
+          msg = wait_for { dlq.get(no_ack: true) }
+          msg.should_not be_nil
+          msg.not_nil!.properties.headers.try(&.[]?("CC")).should be_nil
+        end
+      end
+    end
   end
 end
--- a/spec/queue_dead_lettering_spec.cr
+++ b/spec/queue_dead_lettering_spec.cr
@@ -927,6 +927,38 @@ module DeadLetteringSpec
         end
       end
 
+      it "should not route to BCC in a later dead lettering hop when dead-letter-routing-key was set" do
+        # BCC is stripped from delivered messages, so a surviving BCC header is
+        # only observable through a second dead lettering hop without a routing
+        # key, where it would fan out to bcc_q.
+        qargs = {
+          "x-dead-letter-exchange"    => "",
+          "x-dead-letter-routing-key" => "dlq2",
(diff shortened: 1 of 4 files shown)


spec/queue_dead_lettering_spec.cr
```diff
@@ -1111,5 +1111,18 @@ module DeadLetteringSpec
...
+        with_dead_lettering_setup do |q, dlq, ch, _|
+          props = AMQP::Client::Properties.new(headers: AMQP::Client::Arguments.new({"CC" => ["extra-rk"]}))
+          ch.default_exchange.publish_confirm("msg", q.name, props: props)
+          q.get(no_ack: false).try &.reject(requeue: false)
+          msg = wait_for { dlq.get(no_ack: true) }
+          msg.should_not be_nil
+          msg.not_nil!.properties.headers.try(&.[]?("CC")).should be_nil
+        end
+      end
+    end
```
spuun: Instead of adding this spec, I think `should only be dead lettered to dead-letter-routing-key with preserved CC` should be renamed and modified?

kickster97 (approved):

PR #1994, opened 2026-05-29
Consistency: use `:vhost` named parameter for `DELETE` vhost-limits route

walro:
I was looking at out OpenAPI docs, to see what our practices were regarding which endpoints are documented and which aren't and Claude brought this discrepancy to my attention.

The `DELETE` `/api/vhost-limits` route used a `:name` named parameter and compensated with `vhost_key: "name"`, while the sibling `GET` and `PUT` routes use `:vhost`. 

The OpenAPI spec already documented this path as `/vhost-limits/{vhost}/{type}`, so the route was out of step with its own docs; this brings the code in line with the spec.

Aligns the placeholder with the OpenAPI documentation and siblong routes and drop the override. Adds a regression spec for the endpoint.

Specs

### DIFF
--- a/spec/api/http_api_spec.cr
+++ b/spec/api/http_api_spec.cr
@@ -199,6 +199,17 @@ describe LavinMQ::HTTP::Server do
     end
   end
 
+  describe "DELETE /api/vhost-limits/vhost/type" do
+    it "clears the limit" do
+      with_http_server do |http, s|
+        s.vhosts["/"].max_connections = 100
+        response = http.delete("/api/vhost-limits/%2f/max-connections")
+        response.status_code.should eq 204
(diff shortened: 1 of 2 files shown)


spec/api/http_api_spec.cr
```diff
@@ -199,6 +199,17 @@ describe LavinMQ::HTTP::Server do
     end
   end
 
+  describe "DELETE /api/vhost-limits/vhost/type" do
```
walro, line 202: Not sure we need this spec, or well, rather I think we do, but it's kinda "random" as we don't exhaustively spec the other endpoints.

viktorerlingsson (approved):

carlhoerberg (approved):

PR #1995, opened 2026-05-29
fix: prevent OverflowError in StreamMessageStore#cleanup_consumer_offsets

### LINKED ISSUE #1953: OverflowError in `StreamMessageStore#cleanup_consumer_offsets` prevents server startup
lavinmq fails to load vhosts on startup with an `Arithmetic overflow (OverflowError)`:

```
ERROR lmq.vhost_store Failed to load vhosts
Arithmetic overflow (OverflowError)
  from src/lavinmq/amqp/stream/stream_message_store.cr:196:7 in 'cleanup_consumer_offsets'
  from src/lavinmq/amqp/stream/stream_message_store.cr:327:7 in 'drop_overflow'
  from src/lavinmq/amqp/stream/stream_message_store.cr:23:7 in 'initialize:metadata'
  ...
  from src/lavinmq/vhost_store.cr:116:29 in 'load!'
  from src/lavinmq/server.cr:57:7 in 'initialize'
```

Once the failure mode is reached, the broker can no longer start. The on-disk
data itself is not corrupted — the crash happens entirely in memory during
metadata setup — so the data is recoverable once the bug is fixed.

## Affected file

`src/lavinmq/amqp/stream/stream_message_store.cr:188-196`

```crystal
def cleanup_consumer_offsets
  return if @consumer_offsets.size.zero?

  offsets_to_save = Hash(String, Int64).new
  lowest_offset_in_stream, _seg, _pos = offset_at(@segments.first_key, 4u32)
  capacity = 0                                              # <- Int32
  @consumer_offset_positions.each do |ctag, _pos|
    if offset = last_offset_by_consumer_tag(ctag)
      offsets_to_save[ctag] = offset if offset >= lowest_offset_in_stream
      capacity += ctag.bytesize + 1 + 8                     # <- Int32 ops
    end
  end
  @consumer_offset_positions = Hash(String, Int64).new
  replace_offsets_file(capacity * 1000) do                  # <- overflows
    ...
  end
end
```

## Root cause

`capacity` is initialised as the `Int32` literal `0` and is incremented with
other `Int32` values, so it stays `Int32`. Line 196 evaluates `capacity * 1000`
in `Int32` arithmetic and overflows when:

```
capacity > Int32::MAX / 1000  ≈  2,147,483 bytes
```

The `consumer_offsets` mfile is created with `Config.instance.segment_size`
(8 MiB by default) as its capacity. `consumer_offset_file_full?` only triggers
cleanup when the file is nearly full, so by the time cleanup runs the local
`capacity` (the size of still-valid entries) is often close to 8 MiB. Then
`capacity * 1000` evaluates to roughly 8 GiB and overflows `Int32`.

`drop_overflow` calls `cleanup_consumer_offsets` unconditionally from
`StreamMessageStore#initialize` (line 23), so every startup that touches a
stream queue with a near-full `consumer_offsets` file hits this — a fairly bad
failure mode.

## Reproducer (conceptually)

1. Open a stream queue.
2. Track offsets for enough distinct consumer tags that the `consumer_offsets`
   mfile fills toward `segment_size` (8 MiB).
3. Cleanly shut down lavinmq.
4. Start lavinmq again — `drop_overflow` → `cleanup_consumer_offsets` →
   overflow → vhost load fails.

A more direct unit-level reproducer: instantiate a `StreamMessageStore`, store
enough consumer offsets to push the local `capacity` accumulator past
`Int32::MAX / 1000` (≈ 2.1 MB of tag+offset data), and call
`cleanup_consumer_offsets`.

viktorerlingsson:
Fixes #1953.

On startup, `drop_overflow` calls `cleanup_consumer_offsets`, which computed the rewritten file size as `capacity * 1000` where `capacity` was an `Int32`. When a stream's `consumer_offsets` file is near full (~8 MiB), `capacity * 1000` overflows `Int32::MAX` and raises `OverflowError`, which fails vhost load and prevents the server from starting. The on-disk data is fine; the crash is purely in-memory during metadata setup.

## Changes
- `entry_size` returns `Int64`, so the size math no longer overflows.
- The rewritten file is bounded by `MIN`/`MAX_CONSUMER_OFFSETS_FILE_SIZE` (8 KiB / 64 MiB). When the live set exceeds the max, the oldest (least recently committed) offsets are dropped and a warning is logged.
- Size/encoding helpers (`entry_size`, `trim_to_size`) are extracted into a new `ConsumerOffsets` module so the size-cap logic is unit-testable without a store instance.

## Test plan
- New regression spec stores enough offsets to push the accumulator past `Int32::MAX // 1000` and asserts `cleanup_consumer_offsets` no longer raises.
- New unit spec for `ConsumerOffsets.trim_to_size` covering the drop-oldest behaviour.
- `make test` and `make lint` pass.

---

Pulling the consumer-offset encoding into its own file is a first step: a follow-up PR could move the rest of the consumer-offset handling out of `StreamMessageStore` and into `ConsumerOffsets` - #1998

### DIFF
--- a/src/lavinmq/amqp/stream/consumer_offsets.cr
+++ b/src/lavinmq/amqp/stream/consumer_offsets.cr
@@ -1,0 +1,29 @@
+module LavinMQ::AMQP
+  # Stateless helpers for the consumer-offset entry encoding and size-capping
+  module ConsumerOffsets
+    ENTRY_OVERHEAD = 1 + 8                # ShortString length byte + Int64 offset
+    MAX_ENTRY_SIZE = 255 + ENTRY_OVERHEAD # Largest possible entry: 255-byte tag + Int64
+
+    # Encoded size of one {consumer_tag, offset} entry in the offsets file.
+    def self.entry_size(consumer_tag : String) : Int64
+      (consumer_tag.bytesize + ENTRY_OVERHEAD).to_i64
+    end
+
+    # Drops the oldest (lowest position) offsets until the set fits in
+    # `max_size`. Survivors are returned oldest first so rewriting them keeps
+    # file position reflecting commit recency. Usually drops nothing.
+    def self.trim_to_size(offsets : Array(Tuple(String, Int64, Int64)),
+                          max_size : Int64) : Array(Tuple(String, Int64, Int64))
+      offsets.sort_by! { |_ctag, _offset, pos| pos } # oldest (lowest position) first
+
+      total_size = offsets.sum { |ctag, _offset, _pos| entry_size(ctag) }
+      drop_count = 0
+      # `>=` so the kept set stays strictly below the cap, leaving headroom.
+      while total_size >= max_size
+        total_size -= entry_size(offsets[drop_count][0])
+        drop_count += 1
+      end
+      offsets[drop_count..]
+    end
+  end
+end
--- a/spec/stream_queue_spec.cr
+++ b/spec/stream_queue_spec.cr
@@ -943,6 +943,44 @@ describe LavinMQ::AMQP::Stream do
       end
     end
 
+    it "cleanup_consumer_offsets does not overflow with many consumer offsets" do
+      queue_name = Random::Secure.hex
+      with_amqp_server do |s|
+        StreamSpecHelpers.publish(s, queue_name, 1)
+        data_dir = File.join(s.vhosts["/"].data_dir, Digest::SHA1.hexdigest queue_name)
+        msg_store = LavinMQ::AMQP::StreamMessageStore.new(data_dir, nil)
+
+        # Enough fixed-width entries that the summed size exceeds
+        # Int32::MAX // 1000 (~2.1 MB), which overflowed the old `capacity * 1000`.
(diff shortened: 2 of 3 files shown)


src/lavinmq/amqp/stream/consumer_offsets.cr
carlhoerberg, line 1: Couldn't these be in `StreamMessageStore` instead?

carlhoerberg: or do the reverse and extract all ConsumerOffset handling from `StreamMessageStore` to this class

viktorerlingsson: Yes, I was planning to move the rest of the consumer offset code to this new file in another PR.

viktorerlingsson: #1998

carlhoerberg:
@viktorerlingsson you want to make this one up to date and merge now that we are out of the freeze?

PR #1997, opened 2026-06-01
Fix MQTT perf packet ID overflow

### LINKED ISSUE #1942: lavinmqperf mqtt throughput will log `Arithmetic overflow`
Running `lavinmqperf mqtt throughput --qos 1` will eventually log `Arithmetic overflow`

It seems to be the publishing connection that hits it. The publisher is disconnected, but will reconnect and perf continues.

```bash
# ./bin/lavinmqperf mqtt throughput --qos 1
Connected to broker with --uri=mqtt://localhost:1883
Connected to broker with --uri=mqtt://localhost:1883
Publish rate: 37010 msgs/s Consume rate: 37011 msgs/s
Arithmetic overflow
Publish rate: 28524 msgs/s Consume rate: 28524 msgs/s
Connected to broker with --uri=mqtt://localhost:1883
Publish rate: 8914 msgs/s Consume rate: 8914 msgs/s
Publish rate: 38111 msgs/s Consume rate: 38110 msgs/s
Arithmetic overflow
Publish rate: 18508 msgs/s Consume rate: 18508 msgs/s
Connected to broker with --uri=mqtt://localhost:1883
Publish rate: 17746 msgs/s Consume rate: 17746 msgs/s
Publish rate: 38410 msgs/s Consume rate: 38410 msgs/s
```

kickster97:
## What

Fixes #1942.

`lavinmqperf mqtt throughput --qos=1` could eventually log `Arithmetic overflow` and reconnect the publisher.

The MQTT publisher generated packet IDs from an unbounded `UInt16` range. After packet ID `65535`, Crystal raised on overflow instead of wrapping. MQTT packet IDs are valid in the range `1..65535`, so after `65535` the publisher should wrap back to `1`.

This adds a small packet ID generator for the MQTT throughput publisher that wraps from `65535` to `1`, plus a regression spec for the wraparound case.

## Testing

- `crystal tool format --check src/lavinmqperf/mqtt/throughput.cr spec/lavinmqperf_spec.cr`
- `crystal spec spec/lavinmqperf_spec.cr:156 -Dpreview_mt -Dexecution_context`
- `make bin/lavinmqperf`
- Manual QoS 1 throughput run against a local broker

### DIFF
--- a/src/lavinmqperf/mqtt/throughput.cr
+++ b/src/lavinmqperf/mqtt/throughput.cr
@@ -8,16 +8,6 @@ require "openssl"
 module LavinMQPerf
   module MQTT
     class Throughput < Perf
-      class PacketIdGenerator
-        @next = 1_u16
-
-        def next : UInt16
-          id = @next
-          @next = id == UInt16::MAX ? 1_u16 : id + 1_u16
-          id
-        end
-      end
-
       @publishers = 1
       @consumers = 1
       @size = 16
@@ -243,11 +233,11 @@ module LavinMQPerf
 
         start = Time.instant
         pubs_this_second = 0
-        packet_id_generator = PacketIdGenerator.new
+        packet_id_generator = (1_u16..UInt16::MAX).cycle
         wait_until_all_are_connected(connected)
         until @stopped
           @random.random_bytes(data) if @random_bodies
-          packet_id = @qos > 0 ? packet_id_generator.next : nil
+          packet_id = @qos > 0 ? packet_id_generator.next.as(UInt16) : nil
 
           publish = LavinMQ::MQTT::Publish.new(
             topic: @topic,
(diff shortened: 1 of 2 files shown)


src/lavinmqperf/mqtt/throughput.cr
```diff
@@ -233,11 +243,11 @@ module LavinMQPerf
 
         start = Time.instant
         pubs_this_second = 0
-        packet_id_generator = (1_u16..).each
+        packet_id_generator = PacketIdGenerator.new
```
viktorerlingsson: We could use `.cycle` instead I think, then there's no need for the `PacketIdGenerator` class.  (then we also need to revert back to `packet_id = @qos > 0 ? packet_id_generator.next.as(UInt16) : nil`)

```suggestion
        packet_id_generator = (1_u16..UInt16::MAX).cycle
```

viktorerlingsson (approved):

PR #1999, opened 2026-06-01
Return unexpected frame for heartbeat on non-zero channel

### LINKED ISSUE #1991: UnexpectedFrames: heartbeat on non-zero channel returns FRAME_ERROR (501) instead of UNEXPECTED_FRAME (505)
## Bug

When a heartbeat frame is received on a non-zero channel, LavinMQ closes the connection with error code 501 (FRAME_ERROR) instead of 505 (UNEXPECTED_FRAME) as required by the AMQP spec.

## Failing test

`com.rabbitmq.client.test.functional.UnexpectedFrames.heartbeatOnChannel` in the RabbitMQ Java client test suite:

```
expected: <505> but was: <501>
```

## Root cause

In `lib/amq-protocol/src/amq/protocol/frames.cr`, `Heartbeat.from_io` raises `Protocol::Error::FrameDecode` when the channel is non-zero:

```crystal
def self.from_io(channel, size, io, format)
  unless channel.zero?
    raise Protocol::Error::FrameDecode.new("Heartbeat frame channel must be 0, got #{channel}")
  end
  ...
end
```

This is caught in `src/lavinmq/amqp/client.cr`'s read loop and mapped to FRAME_ERROR (501):

```crystal
rescue ex : AMQ::Protocol::Error::FrameDecode | AMQ::Protocol::Error::InvalidFrameEnd
  send_frame_error(ex.message)  # sends 501 FRAME_ERROR
```

## Expected behaviour

Per the AMQP 0-9-1 spec, a heartbeat on a non-zero channel is an *unexpected* frame (505), not a malformed one (501). Both are connection-level errors, but the code must be 505.

## Suggested fix

Two-part fix across `cloudamqp/amq-protocol.cr` and lavinmq:

**1. In `cloudamqp/amq-protocol.cr` — pass the channel through instead of raising:**

```crystal
def initialize(@channel = 0_u16)
  @bytesize = 0_u32
end

def self.from_io(channel, size, io, format)
  unless size.zero?
    raise Protocol::Error::FrameDecode.new("Heartbeat frame size must be 0, got #{size}")
  end
  new(channel)
end
```

**2. In `src/lavinmq/amqp/client.cr` — handle the non-zero channel in `process_frame` with the correct code:**

```crystal
when AMQP::Frame::Heartbeat
  unless frame.channel.zero?
    close_connection(frame, ConnectionReplyCode::UNEXPECTED_FRAME, "Heartbeat frame must be on channel 0")
  end
```

The library change needs to land in `cloudamqp/amq-protocol.cr` first, then `shard.lock` updated in lavinmq.

kickster97:
## Summary

- close the connection with `UNEXPECTED_FRAME` when a heartbeat frame is received on a non-zero channel
- add a regression spec that sends a raw heartbeat on channel `1` and expects reply code `505`

Fixes #1991.

## Merge blocker

This depends on cloudamqp/amq-protocol.cr#38 being released first. After that release, update `shard.lock` in this PR to the released `amq-protocol` version before marking it ready for review or merging.

## Testing

- `crystal tool format --check src/lavinmq/amqp/client.cr spec/connection_spec.cr`
- `git diff --check`
- `env CRYSTAL_PATH=/Users/christinadahlen/84codes/amq-protocol.cr/src:lib:/opt/homebrew/Cellar/crystal/1.20.2/share/crystal/src crystal spec -Dpreview_mt -Dexecution_context spec/connection_spec.cr`


carlhoerberg (approved):

carlhoerberg:
Release a new version of amq-protocol.cr and then we can merge this. include the new `shards.lock` in this PR

snichme (approved):
Correct, minimal fix for #1991 — a heartbeat on a non-zero channel now closes the connection with 505 UNEXPECTED_FRAME (AMQP 0-9-1 §4.2.7: heartbeats are connection-level, channel 0 only), while channel-0 heartbeats still no-op as before. Solid raw-frame regression spec.

Verified the amq-protocol 1.2.1 bump is necessary and valid: PR cloudamqp/amq-protocol.cr#38 ("Preserve heartbeat frame channel during decode") is merged and released in v1.2.1, so `frame.channel` now carries the client's value. The merge-blocker noted in the PR body is therefore cleared.

Nit (non-blocking): the `return` after `close_connection` is redundant since nothing follows it in the `when` branch, but it's harmless and reads defensively.

carlhoerberg:
@kickster97 let's fix merge conflict and the merge this

PR #2002, opened 2026-06-01
Acknowledge publishes and definitions only when durable on all in-sync replicas

carlhoerberg:
## What

In a clustered setup, an acknowledged operation must survive leader failover. This PR establishes that guarantee end to end: **a publish confirm or definition acknowledgment (`Declare-Ok` etc.) is only sent once the data is durable on the leader *and* on every node that could be elected leader in its place.** Concretely, an acknowledgment now implies:

- the data is fsynced on the leader (`syncfs(2)`, batched);
- every follower listed in the etcd ISR has received it, **synced it to disk**, and acked it;
- any ISR shrink (a follower that disconnected or stopped acking) is **committed to etcd** first, so a leader crash right after the ack can't elect a replica that lacks the data.

Getting this right surfaced and fixed several pre-existing replication bugs (follower corruption on join, torn records, ISR staleness windows, election against a stale ISR).

## Leader: acknowledgment path

- The `Persister`'s confirm loop flushes the replication stream, runs the local `syncfs`, then waits for every in-sync follower's ack (followers persist in parallel with the leader's syncfs). The wait is event-driven off the existing `ack_loop` fiber — no polling, no per-message fibers (`Follower#wait_for_confirm`).
- The confirm loop runs on an isolated thread (for the blocking `syncfs`) and never writes follower sockets itself: flushes are delegated to a follower-owned fiber on the default execution context (`Follower#request_flush`/`flush_loop`). The sockets' fds belong to the default context's event loop (where `ack_loop` keeps a read pending), so a blocking cross-context write would raise `RuntimeError` instead of waiting and kill the confirm loop permanently.
- Durable definition changes (queue/exchange declare/delete/bind/unbind, via AMQP or the HTTP API) go through the same fence: `DefinitionsStore` waits for follower acks after its local fsync, before the caller sends the `*-Ok`. The definitions file is written through to disk (`sync = true`): a joining follower's cut and `full_sync` read the file through separate fds, so a frame sitting in a user-space write buffer at dispatch time would be invisible to the join yet skipped by the change stream as below the cut — the follower would enter the ISR permanently missing it.
- A confirm may only be delivered against an ISR that is **committed to the coordinator**. A dirty ISR (follower connected/disconnected since the last write) is flushed — retrying until the etcd write succeeds — before any acknowledgment goes out, both in the confirm loop and after every replicated dispatch (`each_follower`). Acknowledgments stall rather than being issued against a stale ISR; if etcd stays unreachable the lease expires and the process exits.

## Follower: an ack means durable and applied

- The follower runs `syncfs` before sending each coalesced ack batch (`File#sync=true` alone never fsynced); on sync failure it exits so the leader drops it from the ISR.
- Large actions are acked incrementally per 64 KiB chunk as they're persisted, so the leader's ack deadline measures genuine stall rather than action size (a 128 MiB message at 100 Mbit/s streams >10 s).
- A graceful close (e.g. the follower being promoted to leader) waits for the ack-sending fiber to finish its pending sync-and-send before closing the data dir fd, so a late `syncfs` can't hit a closed fd and take the process down with exit 1 mid-promotion.
- A delete is acked only after the unlink is applied, and a replace's final ack only after the `.tmp` file is renamed into place — an acked byte can't describe a change that hasn't happened yet.

## ISR maintenance and election

- A follower that disconnects while **behind** is removed from the etcd ISR immediately; a **caught-up** follower stays listed as a valid failover candidate and is removed before the next acknowledged operation.
- A connected follower that stops acking outstanding data for 3 s (`ACK_TIMEOUT`) is disconnected and re-syncs on reconnect; the deadline tracks data-outstanding time, so idle periods and healthy-but-behind followers never trip it.
- A join that fails at its ISR commit (etcd error right after the follower was marked synced) drops the follower like any other disconnect, instead of leaving a zombie in the replica set — synced, listed in the ISR, but with no ack loop — that every confirm and definition fence would block on forever. `Follower#close` also marks a follower dead even when its ack loop never ran, unblocking any waiter already parked on it.
- A node that wins the etcd election while no longer in the ISR releases the lease and exits instead of serving (its candidacy outlived its ISR membership), preventing confirmed data from being lost cluster-wide via a full_sync from the stale node.
- ISR writes are **fenced on the leader's etcd election key**: the write is a `kv/txn` conditioned on the election key still being held by the leader's lease, instead of a plain `kv/put`. A deposed leader (e.g. paused past its lease TTL while a follower was promoted) that resumes can no longer overwrite the ISR — its write is rejected (`StaleLeadership`), so the stalled confirm waits out the lease expiry and the process exits, rather than shrinking the ISR to exclude departed followers and confirming data that exists only on the old leader (which the new leader would then delete on the old leader's rejoin/full_sync). The election win is captured as the fencing token in the same step as the campaign, so a node cannot serve as leader without it and can't forget to arm.

## Join correctness (pre-existing corruption, fixed here)

- A follower joining during active publishing could duplicate bytes in segment files: the local segment write races `full_sync`, which could send a message whose replication dispatch was still pending. The join now snapshots a per-file byte **cut** under the lock, caps the final `full_sync` to it, and records it as the follower's synced baseline; appends below the baseline are skipped for that follower.
- The cut is a live file size and can land *inside* a record; an append straddling it now streams only the unsynced tail instead of being skipped whole (which tore the record and corrupted the segment on promotion). Baselines are forgotten when a file is replaced (e.g. definitions compaction) and freed once the follower catches up.
- With joins cut-clean, an in-sync follower provably has every prior write, so confirms wait on the *current* in-sync set — no batch snapshots or sync generations.

## `--no-sync` preserves the clustered fences

`--no-sync` skips local `syncfs` work; it must not weaken clustered publish-confirm safety. Previously a clustered leader with sync disabled bypassed the `Persister` confirm loop entirely, so persistent publishes could be confirmed without waiting for in-sync follower acks or for a dirty ISR shrink to commit — failover-unsafe, since etcd could still list a replica lacking an already-confirmed write.

- Clustered confirms now stay on the normal batching path even with sync disabled; no-sync skips only the leader's local `syncfs` call (`Persister#enqueue_ack`/`close`).
- The follower client follows the same rule: it applies streamed changes before acking them, and no-sync skips only the follower-side `syncfs` before sending the ack (`Client#sync_to_disk`/`sync_data_dir`).
- The controller's self-election signal is split from leader startup readiness (`@elected_leader` vs the old `@is_leader`): the election observer recognizes this node's advertised URI immediately after the campaign wins, while ISR validation is still running, avoiding a false duplicate-URI shutdown if the validation read is slow.

## Deprecated

- `clustering_max_unsynced_actions` is a no-op (still accepted with a warning): it only ever sized the follower's local ack buffer, never the ISR-lag bound its docs claimed. Follower lag is governed by the leader's ack deadline.

## Tests

- `spec/clustering/isr_spec.cr` — ISR removal timing on disconnect (behind vs caught-up); dirty ISR committed before durable operations return and before confirms; confirms stall while etcd is unreachable; declares held until followers ack or the silent follower's eviction is committed; no-sync publish confirms still wait on follower acks and flush a dirty ISR; the self-leader observer recognizes this node during ISR validation.
- `spec/clustering/join_race_spec.cr` — follower joining during concurrent publishing ends byte-identical to the leader, plus an oracle control.
- `spec/clustering/follower_spec.cr` — `wait_for_confirm` blocks until acked, unblocks on disconnect, supports concurrent waiters and is safe to call from an isolated execution context (regression: no cross-context socket write); `close` marks a follower whose ack loop never ran as dead and unblocks waiters; ack deadline drops a stalled follower but not an idle or catching-up one; baseline straddling/reset.
- `spec/clustering/client_sync_spec.cr` — incremental acks for large actions; deletes/replaces acked only after applied (including deterministic never-acked-when-unappliable cases); `close` waits for the ack loop's pending sync before closing the data dir fd; no-sync follower still acks replicated changes after applying them.
- `spec/definitions_replication_spec.cr` — definition frames are readable from disk at their dispatched offset at the moment they are dispatched (regression for the buffered-definitions join hole).
- `spec/clustering/isr_spec.cr` also covers the failed-join path: a follower whose ISR commit raises is dropped and the next durable operation commits an ISR without it instead of hanging.
- `spec/etcd_spec.cr` — a stale election holder's fenced ISR write is rejected (`StaleLeadership`) once a new node wins the election, while the new leader's write succeeds and overwrites it.
- `spec/clustering_spec.cr`, `spec/clustering/server_spec.cr` — end-to-end confirmed publishes against real followers; no deadlock on disconnect under load; no-sync clustered publish confirms against real followers.

`make test` (1904 examples, 0 failures, 10 pending) and `make lint` (0 failures) pass. Each regression spec was verified to fail against the code before its fix.


src/lavinmq/clustering/client.cr
```diff
@@ -283,29 +304,38 @@ module LavinMQ
       end
 
       private def stream_changes(socket, lz4)
-        acks = Channel(Int64).new(@config.clustering_max_unsynced_actions)
-        spawn send_ack_loop(acks, socket), name: "Send ack loop"
+        @acks = Channel(Int64).new(ACK_BUFFER_CAPACITY)
+        @ack_loops.add # before spawn, so a concurrent #close can't miss the fiber
+        spawn send_ack_loop(@acks, socket), name: "Send ack loop"
```
spuun: Should be equivalent, and no need for `@ack_loops.done` in `#send_ack_loop`.
```suggestion
        @ack_loops.spawn(name: "Send ack loop") { send_ack_loop(@acks, socket) }
```

src/lavinmq/clustering/follower.cr
```diff
@@ -45,30 +62,124 @@ module LavinMQ
...
+      # the baseline agree and in-flight writes aren't duplicated.
+      def full_sync(caps : Hash(String, Int64)? = nil) : Nil
+        send_file_list(caps: caps)
+        send_requested_files(caps: caps)
       end
 
-      def ack_loop
+      def ack_loop(ack_timeout : Time::Span = ACK_TIMEOUT)
         @running.add
+        spawn(name: "Clustering follower flush loop") { flush_loop }
```
spuun: I guess, in theory, that the counter in `@running` may hit zero and any waiter is woken before the flush_loop starts? Should this be 
```suggestion
        @running.spawn(name: "Clustering follower flush loop") { flush_loop }
```
instead and the `@running.add` and `@running.done` in `#flush_loop` are removed?

src/lavinmq/clustering/follower.cr
```diff
@@ -45,30 +62,124 @@ module LavinMQ
...
+      # spawned by ack_loop on the default execution context: the publish
+      # confirm loop runs on an isolated thread and must not write the socket
+      # itself — the socket's fd belongs to the default context's event loop
+      # (ack_loop keeps a read pending on it), and a write that blocks from
+      # another context raises instead of waiting.
+      private def flush_loop
+        @running.add
+        while @flush_requested.receive?
+          flush
+        end
```
spuun, line 146: This loop will exit on the first flush request and flush will never happen, no?

Change `@flush_requested` to `::Channel(Bool)` and send `true` instead.

src/lavinmq/clustering/server.cr
```diff
@@ -99,23 +106,43 @@ module LavinMQ
...
+      # heap allocation. `offset` is where the value is written on the leader,
+      # used to skip bytes a just-joined follower already got via full_sync.
+      def append_value(path : String, value : UInt32 | Int32, offset : Int64)
         path = strip_datadir path
         @file_index.lock { |_files, checksums| checksums.delete(path) }
-        each_follower &.append(path, value)
+        each_follower do |f|
+          case skip = f.already_synced(path, offset, 4i64)
+          when 0 then f.append(path, value)
+          when 4 then next # entirely within the follower's full_sync snapshot
```
spuun: 4?

src/lavinmq/clustering/server.cr
```diff
@@ -184,15 +225,40 @@ module LavinMQ
...
+            end
+          end
+        end
+        sizes
+      end
+
       def followers : Array(Follower)
         @lock.synchronize do
-          @followers.select(&.synced?).dup # for thread safety
+          @followers.select(&.synced?) # for thread safety
```
spuun: The comment made sense with the `#dup` call but is a little bit ambiguous without it
```suggestion
          @followers.select(&.synced?) # select returns new array => thread safe
```

src/lavinmq/clustering/server.cr
```diff
@@ -184,15 +225,40 @@ module LavinMQ
...
         end
       end
 
       def syncing_followers : Array(Follower)
         @lock.synchronize do
-          @followers.select(&.syncing?).dup # for thread safety
+          @followers.select(&.syncing?) # for thread safety
```
spuun: Ditto
```suggestion
          @followers.select(&.syncing?) # select returns new array => thread safe
```

src/lavinmq/clustering/server.cr
```diff
@@ -237,25 +303,57 @@ module LavinMQ
...
-        # @sync_lock is always acquired before @lock to avoid deadlock
-        @sync_lock.synchronize do
-          follower.full_sync # sync the bulk
-          @lock.synchronize do
-            follower.full_sync    # sync the last
-            follower.mark_synced! # Change state to Synced
-            update_isr
-          end
-        end
         begin
```
spuun: Maybe break out this block to it's own method? At first I thought the `ensure` belonged to `#handle_socket`.

spuun (commented):
I think i grasp what it does 😅 

A few comments...

PR #2005, opened 2026-06-02
fix: delete empty queue dirs on followers during streaming

### LINKED ISSUE #1854: Orphaned queue directories left on follower nodes after queue deletion

viktorerlingsson:
Part of #1854.

The leader streams individual file deletes to followers but never a directory delete, so an emptied queue dir lingered on followers until the next full sync (which already cleans up empty dirs).

After a streamed file delete, walk up from the file's parent removing now-empty dirs, stopping at the first non-empty dir. Paths are relative to the data dir and the walk stops when `File.dirname` reaches `"."`, so it never touches the data dir root. A non-force `Dir.delete` (plus `rescue`) keeps it safe; appends/replaces recreate the full path via `mkdir_p`.

## Test plan

- `removes empty queue dir from follower when queue is deleted`
- `keeps the queue dir on follower when a segment is deleted but the queue isn't`

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -338,12 +338,12 @@ module LavinMQ
       private def delete_empty_dirs(dir)
         while dir != "."
           path = File.join(@data_dir, dir)
-          Dir.delete(path)
+          Dir.delete?(path) || break
           Log.debug { "Deleted empty dir #{dir}" }
           dir = File.dirname(dir)
         end
       rescue ex : File::Error
-        Log.debug { "Stopped deleting empty dirs at #{dir}: #{ex.message}" }
+        Log.error(exception: ex) { "Could not delete #{dir}: #{ex.message}" }
       end
 
       private def replace(filename, len, lz4)
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -324,6 +324,27 @@ module LavinMQ
         end
         @checksums.delete(filename)
         @file_digests.delete(filename)
+        delete_empty_dirs File.dirname(filename)
+      end
+
+      # Removes now-empty parent directories (e.g. an emptied queue dir) after a
+      # file delete. The leader only streams file deletes, not directory deletes,
+      # so without this empty queue dirs would linger until the next full sync.
+      #
+      # We walk up one level per iteration until a non-empty dir, stopping when
+      # File.dirname reaches ".". A non-force Dir.delete and the rescue stop the
+      # walk safely if a dir still has files or is already gone. Both append and
+      # replace file re-create the full path if needed.
+      private def delete_empty_dirs(dir)
+        while dir != "."
+          path = File.join(@data_dir, dir)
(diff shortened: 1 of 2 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -324,6 +324,27 @@ module LavinMQ
...
+      # so without this empty queue dirs would linger until the next full sync.
+      #
+      # We walk up one level per iteration until a non-empty dir, stopping when
+      # File.dirname reaches ".". A non-force Dir.delete and the rescue stop the
+      # walk safely if a dir still has files or is already gone. Both append and
+      # replace file re-create the full path if needed.
+      private def delete_empty_dirs(dir)
+        while dir != "."
+          path = File.join(@data_dir, dir)
+          break unless Dir.empty?(path)
```
carlhoerberg: I think we should avoid that and relay on the non-recursive `Dir.delete` that will raise if there are other files in the dir.

viktorerlingsson: just skip the `Dir.empty?`? sure!

src/lavinmq/clustering/client.cr
```diff
@@ -324,6 +324,26 @@ module LavinMQ
...
+      # so without this empty queue dirs would linger until the next full sync.
+      #
+      # We walk up one level per iteration until File.dirname reaches ".". The
+      # non-recursive Dir.delete raises File::Error if the dir still has files
+      # (or is already gone), and the rescue stops the walk safely. Both append
+      # and replace file re-create the full path if needed.
+      private def delete_empty_dirs(dir)
+        while dir != "."
+          path = File.join(@data_dir, dir)
+          Dir.delete(path)
```
carlhoerberg: Avoiding using exceptions as control flow
```suggestion
          Dir.delete?(path) || break
```

src/lavinmq/clustering/client.cr
```diff
@@ -324,6 +324,26 @@ module LavinMQ
...
+          Log.debug { "Deleted empty dir #{dir}" }
+          dir = File.dirname(dir)
+        end
+      rescue ex : File::Error
+        Log.debug { "Stopped deleting empty dirs at #{dir}: #{ex.message}" }
```
carlhoerberg: Do some error logging instead, if `Dir.delete?` is used, because it will not raise if 
```suggestion
        Log.error(exception: ex) { "Could not delete #{dir}: #{ex.message}" }
```

carlhoerberg (approved):

PR #2011, opened 2026-06-04
Remove badges from README.md

jage:
GitHub UI shows this in other places nowadays, in the past it made sense to have these buttons but IMO mostly noise now.

License is a tab just above the README preview:

Screenshot 2026-06-04 at 10 47 01

Latest release is on the right side:

Screenshot 2026-06-04 at 10 47 03


viktorerlingsson (approved):

kickster97 (approved):

kickster97:
I looked around at a few popular repositories, and it feel like its a bit 50/50 if projects have these or not. But the build status link is very ugly and the failing ci is more scary than it has to be..

carlhoerberg:
Thanks!

PR #2012, opened 2026-06-05
Anchor server uptime and CPU counters to OS process lifetime

weistrand:

Anchors `lavinmq_uptime` and the counters `@user_time` / `@sys_time` / `@blocks_in` / `@blocks_out (which feed both `/metrics` and `/api/nodes/{name}` with e.g. the metrics lavinmq_cpu_user_time_total` and `lavinmq_cpu_system_time_total`) to the process lifetime instead of the `Server` instance lifetime.

`launcher.start` constructs a fresh `Server` on every leader election. Before this change, that reset:

- `@start = Time.instant` → `Server#uptime` dropped to ~0 on every leader transition.
- `@user_time = 0_i64` (macro default) → cached counter dropped to 0 until `stats_loop`'s first tick repopulated it.

Which also reset prometheus cpu counters `lavinmq_cpu_user_time_total` and `lavinmq_cpu_system_time_total`. 

Noticed this while calculating `rate()` of the cpu counters (`lavinmq_cpu_user_time_total` and `lavinmq_cpu_system_time_total`) which produced huge spikes that went far above 100 % on some LavinMQ multi-node clusters. All of them happening a few seconds after a leader transition. 

**Fix:**

- New `src/lavinmq/process_start.cr` defines `LavinMQ::PROCESS_START = Time.instant`, captured once at static-init and immune to `Server.new`. `Server#uptime` now reads it instead of the instance var `@start`.
- `Server#initialize` seeds `@user_time` / `@sys_time` / `@blocks_in` / `@blocks_out` from `System.resource_usage`, so a fresh `Server` instance already reflects the OS process's accumulated rusage.
- The `METRICS` macro changed from `getter foo = 0_i64` to `getter foo : Int64` (no default). Crystal refuses to compile if a future change removes the seeding from `initialize` — the type system is the regression check.

**Spec** for the uptime fix:

```
make test SPEC=spec/api/prometheus_spec.cr
```

**Compile-time check** (regression for the cpu/blocks seeding):

Removing the seeding lines from `Server#initialize` produces:

```
Error: instance variable '@user_time' of LavinMQ::Server was not initialized directly in all of the 'initialize' methods, rendering it nilable.
```

### DIFF
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -113,6 +113,13 @@ describe LavinMQ::HTTP::PrometheusController do
         mfile_metric.not_nil![:value].should be >= 0
       end
     end
+
+    it "should report uptime anchored to PROCESS_START, surviving Server re-creation" do
+      with_metrics_server do |_, server|
+        expected = Time.instant - LavinMQ::PROCESS_START
+        (server.uptime - expected).abs.should be < 100.milliseconds
+      end
+    end
   end
 
   describe "vhost access control" do
--- a/spec/api/prometheus_spec.cr
+++ b/spec/api/prometheus_spec.cr
@@ -115,9 +115,12 @@ describe LavinMQ::HTTP::PrometheusController do
     end
 
     it "should report uptime anchored to PROCESS_START, surviving Server re-creation" do
+      uptime1 = uninitialized Time::Span
       with_metrics_server do |_, server|
-        expected = Time.instant - LavinMQ::PROCESS_START
-        (server.uptime - expected).abs.should be < 100.milliseconds
+        uptime1 = server.uptime
+      end
+      with_metrics_server do |_, server|
+        server.uptime.should be > uptime1
       end
     end
   end
--- a/src/lavinmq/process_start.cr
+++ b/src/lavinmq/process_start.cr
@@ -1,0 +1,4 @@
+module LavinMQ
+  # Anchors Server#uptime to OS process; survives leader transitions.
+  PROCESS_START = Time.instant
+end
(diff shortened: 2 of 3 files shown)


weistrand:

spec/api/prometheus_spec.cr
```diff
@@ -113,6 +113,16 @@ describe LavinMQ::HTTP::PrometheusController do
...
       end
     end
+
+    it "should report uptime anchored to PROCESS_START, surviving Server re-creation" do
+      uptime1 = uninitialized Time::Span
+      with_metrics_server do |_, server|
+        uptime1 = server.uptime
+      end
+      with_metrics_server do |_, server|
+        server.uptime.should be > uptime1
```
jage, line 123: Lets test user_time and sys_time as well?

weistrand: It's hard to actually get that right, user_time and sys_time gets updated by stats_loop and stats_loop runs on another thread and almost always overwrites the buggy value with a current rusage value before the spec reads which masks the bug anyways.

weistrand: We could of course add tests that at least checks that they are returning some values > 0, doesn't help this bug but makes sure they are there at least

jage: Ah ok, yeah those values also have a different source so not as "bug prone" I guess. We can probably skip it, unless we can make some actually good by stopping stats_loop etc, but I'm ok with skipping it.

src/lavinmq/process_start.cr
```diff
@@ -0,0 +1,4 @@
+module LavinMQ
+  # Anchors Server#uptime to OS process; survives leader transitions.
+  PROCESS_START = Time.instant
```
carlhoerberg: Should not be in the Server class?

carlhoerberg (approved):

src/lavinmq/process_start.cr
weistrand: https://github.com/cloudamqp/lavinmq/pull/2012/commits/71395cf42ea6ba96bc2b647199a5e398dbf9c6c2

PR #2013, opened 2026-06-05
Raft-backed clustering (replaces etcd for metadata consensus)

antondalgren:
## Summary

Replaces etcd with embedded [raft](https://github.com/84codes/raft.cr) consensus for the cluster's **metadata** layer — leader election and the in-sync replica (ISR) set. The AMQP/MQTT **data plane** is unchanged: it still streams through `Clustering::Server` (leader) / `Clustering::Client` (follower).

Clustering is now backend-pluggable behind two interface modules — `Clustering::Elector` (leader election lifecycle) and `Clustering::Coordinator` (ISR + replication secret) — with three symmetric implementations selected by `clustering_backend`:

- **`raft`** — embedded raft (`Raft::Backend` + `Raft::Server`), no external dependency.
- **`etcd`** — the existing etcd-backed path (`EtcdBackend`), kept working.
- **`standalone`** — single node, no clustering.

## What's in this branch

**Raft backend (`src/lavinmq/raft/`)**
- `Raft::Server` — tick-loop wrapper around `raft.cr`'s `Node`, marshalling all node mutations onto the tick fiber; exposes `propose_committed` (blocks until an entry commits under the proposing term), a consistent ISR/state snapshot, and a tick-fiber peer snapshot.
- `Raft::Backend` — the Elector+Coordinator: campaign/bootstrap/join, the follow-leader **reconcile loop** (idempotent: follow the current leader at its current address; coalesces bursts; re-resolves on leader *and* configuration change), ISR commit via `propose_committed(SetIsr)`, and split HTTP handlers (status on the metrics port, admin behind auth).
- `ClusterStateMachine` / `ClusterCommand` — only the ISR is replicated through consensus (`SetIsr`).
- `BootstrapDecision` — pure resume/bootstrap/join seam; `.join_target` takes precedence over seed discovery.

**Replication secret** — a pre-shared `<data_dir>/.clustering_password` file (not in the raft log/snapshots). The leader generates it on bootstrap; it is excluded from follower file-sync. (See "Known limitations".)

**Cluster formation** — declarative via `clustering_seed_uris` (lowest advertised host bootstraps, the rest join) or explicit `lavinmqctl raft_join`.

**`lavinmqctl` (`src/lavinmqctl/raft.cr`)** — `raft_status`, `raft_reset`, `raft_join`. Disaster-recovery guards: reset/join fail closed on a multi-peer or unverifiable node without `--force`; `.clustering_id` is preserved across resets.

**Metrics** — every node (not just the leader) exposes LavinMQ- and raft.cr-side gauges on its always-on metrics endpoint.

**Merged `main` (incl. #2002 "ack only when durable on all in-sync replicas").** Reconciled onto a single `Coordinator#update_isr : Bool` contract (raft's `propose_committed` returns committed/rejected directly; etcd returns false on `StaleLeadership`/transient errors), so the leader's publish-confirm fence stalls-and-retries rather than acknowledging against a stale ISR. etcd's ISR-fencing (election-key txn) is ported into `EtcdBackend`.

## Known limitations / follow-ups

- **`seed_uris` formation is not fully zero-touch on the raft backend yet:** the `.clustering_password` secret must be present on every node (it isn't replicated), so it must be pre-distributed (or, planned, injected via a `clustering_secret` env/config). Docs to be updated accordingly.
- Rolling-upgrade unknown-`ClusterCommand`-tag handling and a force-ISR disaster-recovery runbook remain.
- Cross-thread `Node` access redesign is deferred; current cross-fiber reads go through mutex-protected snapshots.

## Test plan

- [x] `make test` — 2046 examples, 0 failures, 0 errors, 9 pending
- [x] `make lint` — 390 inspected, 0 failures
- [x] `crystal tool format` clean
- [x] Jepsen — in progress (restarted-follower ISR promotion fix just landed)
- [ ] CI green


src/lavinmq/clustering/server.cr
```diff
@@ -203,10 +187,11 @@ module LavinMQ
       end
 
       def password : String
-        @coordinator.password
+        @password ||= @coordinator.password
```
spuun: I think this is more or less equivalent with
```crystal
getter password { @coordinator.password }
```

src/lavinmq/clustering/coordinator.cr
spuun: Working against interfaces instead of concrete types is always nice
```suggestion
    abstract def update_isr(synced_node_ids : Enumerable(Int32)) : Nil
```

src/lavinmq/raft/server.cr
```diff
@@ -0,0 +1,208 @@
...
+    end
+
+    private def tick_loop : Nil
+      # FIXME: `timeout(50.ms)` in `select` only fires when no other case is
+      # ready. Under sustained inbox/actions load (e.g. a future per-queue
+      # data-plane Node), @node.tick would be starved and the node would miss
+      # heartbeats / election timeouts. For the current low-frequency
+      # metadata inbox this is harmless. When a high-frequency Raft node
+      # lands, replace this with a dedicated tick-timer fiber sending on a
+      # buffered(1) channel that competes fairly with the other select cases.
```
spuun, line 218: Could calculate timeout from time since last tick. If the calculated value is less or equal to zero, do a tick and set value to 50ms.

src/lavinmq/raft/server.cr
```diff
@@ -0,0 +1,208 @@
...
+      end
+    rescue ::Channel::ClosedError
+    ensure
+      @node.close
+      @tick_done.close
+    end
+
+    private def wire_callbacks : Nil
+      @node.on_role_change do |_old_role, new_role|
+        @is_leader.set(new_role == ::Raft::Role::Leader)
```
spuun: I guess it's an enum? I like the "question methods":
```suggestion
        @is_leader.set(new_role.leader?)
```

src/lavinmq/clustering/server.cr
carlhoerberg: don't we want to check this each time so it hasn't diverged?

src/lavinmq/clustering/server.cr
carlhoerberg, line 406: can `@followers` include duplicate IDs? And if it does, what's the downside of including that in the ISR?

src/lavinmq/launcher.cr
```diff
@@ -40,10 +42,18 @@ module LavinMQ
...
 
       if @config.clustering?
-        etcd = Etcd.new(@config.clustering_etcd_endpoints)
-        @runner = controller = Clustering::Controller.new(@config, etcd)
-        coordinator = Clustering::EtcdCoordinator.new(@config, etcd)
-        @replicator = Clustering::Server.new(@config, coordinator, controller.id)
+        case @config.clustering_backend
+        when "etcd"
+          etcd = Etcd.new(@config.clustering_etcd_endpoints)
+          coordinator = Clustering::EtcdCoordinator.new(@config, etcd)
```
carlhoerberg: Maybe we should take a while and rethink this, now we got Launcher, Runner, Server, Clustering Server, Clustering Controller and Coordinator 😅

carlhoerberg (commented):
Do we need all the example/ files in the repo?

src/lavinmq/clustering/server.cr
antondalgren: Adding a check would work but I don't think it adds value. The clustering secret is a cluster-wide constant set exactly once, at bootstrap, and both backends enforce write-once at the coordinator level. `Raft::Coordinator` proposes `SetSecret` only when no secret is in the state machine, after which the proposal is a no-op. Every node — leader or follower — reads the same value from the coordinator, so there's no path where the cached `@password` on the leader's `Clustering::Server` could diverge from `@coordinator.password`.

src/lavinmq/clustering/server.cr
antondalgren: No, the code in `handle_socket` ensures that `@followers` doesn't include duplicates. It will remove any stale follower before inserting the new follower with the same id. 

Even if a duplicate id would slip through, the ISR wouldn't include that since its represented by a `Set`.

src/lavinmq/launcher.cr
antondalgren: Yeah, could probably be changed! 🤖 suggested following renames 
```md
  - `Clustering::Server` → `Clustering::Replicator` (matches `@replicator` in the launcher) — clears the Server ambiguity.
  - `Clustering::Controller` → `Clustering::EtcdElector` — names what it actually does (etcd-based leader election).
  - `Raft::Runner` → `Raft::Elector` — symmetric with the above.
  - Common parent abstraction `Clustering::Elector` for both — what the launcher actually treats them as.
  ```

File: src/lavinmq/clustering/server.cr (hunk 1 above)
Comment (antondalgren) [reply src/lavinmq/clustering/server.cr]: I went with the current approach to not have to synchronize the mutex for every read. I guess it doesn't really matter since it's not a part of the hotpath anyway... Do we have a preference?

Comment (carlhoerberg) [reply src/lavinmq/clustering/server.cr]: ok!

File: src/lavinmq/raft/cluster_command.cr [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+
+      def body_bytesize : Int32
+        4 + @secret.bytesize
+      end
+
+      protected def self.read_body(io : IO, format : IO::ByteFormat) : SetSecret
+        len = io.read_bytes(UInt32, format)
+        buf = Bytes.new(len)
+        io.read_fully(buf)
+        new(String.new(buf))
```
carlhoerberg: single allocation:
```suggestion
        str = io.read_string(len)
        new(str)
```

src/lavinmq/raft/cluster_state_machine.cr
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def restore(io : IO) : Nil
+      fmt = IO::ByteFormat::LittleEndian
+      version = io.read_bytes(UInt8, fmt)
+      raise InvalidSnapshotVersion.new(version) unless version == SNAPSHOT_VERSION
+      secret_len = io.read_bytes(UInt32, fmt)
+      buf = Bytes.new(secret_len)
+      io.read_fully(buf)
+      secret = String.new(buf)
```
carlhoerberg: Avoids double allocation:
```suggestion
      secret = io.read_string(secret_len)
```

src/lavinmq/raft/cluster_state_machine.cr
```diff
@@ -0,0 +1,91 @@
...
+      isr_count = io.read_bytes(UInt32, fmt)
+      isr = Set(Int32).new(initial_capacity: isr_count.to_i32)
+      isr_count.times { isr.add(io.read_bytes(Int32, fmt)) }
+      @mutex.synchronize do
+        @secret = secret
+        @isr = isr
+        signal_isr_changed
```
carlhoerberg: could this not be outside the mutex? now you signal, wake up another fiber, that will try to read, but will be blocked when trying to lock the mutex?

src/lavinmq/raft/coordinator.cr
```diff
@@ -0,0 +1,28 @@
...
+
+    def password : String
+      existing = @server.secret
+      return existing unless existing.empty?
+      new_secret = Random::Secure.base64(32)
+      @server.propose(ClusterCommand::SetSecret.new(new_secret))
+      deadline = Time.instant + 5.seconds
+      while @server.secret.empty?
+        raise "timed out waiting for SetSecret apply" if Time.instant > deadline
+        Fiber.yield
```
carlhoerberg: Aggressive, maybe:
```suggestion
        sleep 1.milliseconds
```

src/lavinmqctl/raft.cr
```diff
@@ -0,0 +1,157 @@
...
+  rescue ex : IO::Error | Socket::Error
+    @io.puts "raft_status: cannot reach local node: #{ex.message}"
+    raise CtlExit.new(1)
+  end
+
+  # Resolves data_dir using the same precedence as the lavinmq server:
+  # CLI --data-dir → ENV LAVINMQ_DATADIR → [main] data_dir in the INI →
+  # default /var/lib/lavinmq. Lets `lavinmqctl raft_*` work without
+  # repeating --data-dir when a config file is present.
+  private def resolved_data_dir : String
```
carlhoerberg, line 55: Could not both join and reset be called through the HTTP API? Then we could avoid doing file manipulation from cli.

src/lavinmqctl/raft.cr
```diff
@@ -0,0 +1,157 @@
...
+        # ctl command would mask the actual user intent.
+      end
+    end
+    "/var/lib/lavinmq"
+  end
+
+  def raft_reset
+    data_dir = resolved_data_dir
+    # Running-node detection: Task 10 implements maybe_signal_running_node. For now, no-op when no pidfile.
+    if pidfile = @options["pidfile"]?
```
carlhoerberg: Use `DataDirLock` instead

src/lavinmq/raft/server.cr
```diff
@@ -0,0 +1,214 @@
+require "raft"
+require "../bool_channel"
+require "./cluster_command"
+require "./cluster_state"
+require "./cluster_state_machine"
+
+module LavinMQ::Raft
+  class Server
```
carlhoerberg, line 9: Couldn't this be part of raft.cr as some kind of abstract class that could be extended? A lot seems like generic boilerplate code, eg. `tick_loop`

src/lavinmq/raft/elector.cr
```diff
@@ -0,0 +1,256 @@
...
+        ::Raft::HTTP::StatusHandler(ClusterCommand).new(@server.node, @transport, advertised_address))
+    end
+
+    # Mutating POST /raft/admin/* surface. Mount only behind authentication.
+    # Hands the raw Node to the handler for now; mutations should be marshaled
+    # through Server#run_on_tick once the handler accepts an interface.
+    def admin_handler : LavinMQ::HTTP::RaftHandlerWrapper
+      LavinMQ::HTTP::RaftHandlerWrapper.new(
+        ::Raft::HTTP::AdminHandler(ClusterCommand).new(@server.node, @transport))
+    end
```
carlhoerberg: Should AdminHandler really live in `raft.cr`? I'm thinking it's a implementation detail. For `ClusterCommand` it works well, but thinking of when each queue is a raft state machine, does the same API works well then? Maybe, but then shouldn't the wrapper, which is always needed(?), also live in `raft.cr`? 

And if `AdminHandler` was a non generic class we wouldn't need the RaftHandlerWrapper, right?

src/lavinmq/raft/backend.cr
```diff
@@ -0,0 +1,256 @@
...
+      Log.info { "Fresh node with no peers — bootstrapping single-node cluster" }
+      unless @server.bootstrap
+        Log.warn { "Bootstrap rejected (node already has peers); continuing as follower" }
+      end
+    end
+
+    JOIN_MAX_ATTEMPTS   = 30
+    JOIN_RETRY_INTERVAL = 1.second
+
+    def perform_join(leader_uri : String) : Nil
```
carlhoerberg: Here we call the raft admin handler, which lives in `raft.cr`, but this pretty generic and very coupled to that class, is implemented here?

src/lavinmqctl/raft.cr
antondalgren: Ideally we would need to have a solution that allows a node to reset and rejoin. The current architecture could allow for that by deleting the data dir. In turn, that would generate a new clusterid and with the current approach, potentially land in a split-brain solution if the resetted node has the lowest lexicographic hostname. 

This could be avoided by asking all the configured peers if there already is a cluster available among them. But I feel like this PR is just growing in scope rather than having this as an experimental solution that works and improve on that along the way.

src/lavinmq/raft/server.cr
antondalgren: Yeah that's a potential solution indeed. Is it required?

PR #2016, opened 2026-06-06
Replace periodic GC.collect with on-demand GC

carlhoerberg:
## Why

The 30-second periodic `GC.collect` (in `StandaloneRunner` and `Clustering::Controller`) caused high CPU spikes on servers using a lot of memory — e.g. hundreds of thousands of bindings or queues. This removes the timers and makes garbage collection an explicit, on-demand operation instead.

## Changes

**Removed periodic GC**
- `StandaloneRunner#run` now just blocks on the stop channel.
- `Clustering::Controller` post-leadership loop just waits on lease expiry.
- (Left the event-driven `SystemD::MemoryPressure` collection and the `USR2` signal trap in `launcher.cr` untouched — those aren't periodic.)

**HTTP API** (`NodesController`, admin only)
- `GET /api/nodes/gc_stats` — returns `GC.prof_stats` as JSON.
- `POST /api/nodes/gc_collect` and `POST /api/nodes/:name/gc_collect` — trigger a collection, return `204 No Content`.
- GC stats are *not* part of the regular `/api/nodes` payload.

**UI** (nodes page)
- A **Garbage collection** card at the bottom (admin only) showing the prof_stats, each field labelled with its Boehm `GC_prof_stats_s` description via tooltip.
- A **Collect** button and a refresh icon button on the title row. GC stats are fetched once on load and then only on demand (Refresh / after Collect) — they do *not* join the 5s auto-poll.

**lavinmqctl**
- New `gc_collect` command: triggers a collection then prints the resulting GC stats.

## Tests
- `spec/api/node_spec.cr` — `gc_stats` (200), `gc_collect` (204 + `gc_no` increments), `:name` variant (204 / 404), and that `/api/nodes` no longer carries `gc_stats`.
- `spec/lavinmqctl_spec.cr` — `gc_collect` increments `gc_no` and prints stats.
- OpenAPI docs updated for the new paths and the `gc_stats` schema.

### DIFF
--- a/src/lavinmq/http/controller/nodes.cr
+++ b/src/lavinmq/http/controller/nodes.cr
@@ -83,6 +83,24 @@ module LavinMQ
         }
       end
 
+      private def gc_stats
+        ps = GC.prof_stats
+        {
+          gc_no:                     ps.gc_no,
+          heap_size:                 ps.heap_size,
+          free_bytes:                ps.free_bytes,
+          unmapped_bytes:            ps.unmapped_bytes,
+          bytes_since_gc:            ps.bytes_since_gc,
+          bytes_before_gc:           ps.bytes_before_gc,
+          non_gc_bytes:              ps.non_gc_bytes,
+          markers_m1:                ps.markers_m1,
+          bytes_reclaimed_since_gc:  ps.bytes_reclaimed_since_gc,
+          reclaimed_bytes_before_gc: ps.reclaimed_bytes_before_gc,
+          expl_freed_bytes_since_gc: ps.expl_freed_bytes_since_gc,
+          obtained_from_os_bytes:    ps.obtained_from_os_bytes,
+        }
+      end
+
       private def stats(context)
         current_user = user(context)
         my_vhosts = vhosts(current_user)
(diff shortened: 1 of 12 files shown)


carlhoerberg:
image

src/lavinmq/http/controller/nodes.cr
```diff
@@ -101,6 +119,14 @@ module LavinMQ
           context
         end
 
+        # Garbage collector profiling stats for the current node.
+        # Registered before the :name route so it isn't matched as a node name.
+        get "/api/nodes/gc_stats" do |context, _params|
```
spuun, line 124: Why not `/api/nodes/:name/gc_stats` to avoid route collision?

src/lavinmq/http/controller/nodes.cr
```diff
@@ -109,6 +135,25 @@ module LavinMQ
           end
           context
         end
+
+        # Run GC on the current node, without having to know its name
+        post "/api/nodes/gc_collect" do |context, _params|
```
spuun, line 140: Maybe only go with the below (`/api/nodes/:name/gc_collect`), which currently isn't used.

static/js/nodes.js
```diff
@@ -33,6 +34,83 @@ function start (cb) {
...
+      .finally(() => { gcBtn.disabled = false })
+  })
+  // Only admins may read GC stats; the gc-btn element is in the DOM for all
+  // users (require-administrator only hides it via CSS), so fetching here
+  // unconditionally would 403 for non-admins and pop up an error. Gate the
+  // initial load on the (persisted, synchronously applied) admin role class;
+  // afterwards fetch only on demand.
+  if (Helpers.stateClasses.has('user-is-administrator')) {
+    refreshGCStats()
+  }
```
spuun, line 111:
```suggestion
    refreshGCStats()
```
Button should already be hidden for non-administrators

spuun (commented):
I think @bengtmagnus should take a look on UI to make sure it's aligned with everything else. The inlined svg seems unnecessary.

static/js/nodes.js
carlhoerberg: button yes, but on first load the gc stats where loaded regardless

src/lavinmq/http/controller/nodes.cr
carlhoerberg: yes, but that requires two http calls then, first one to overview to fetch node name. And i'm afraid we allow duplicate nodenames/hostnames. So probable remove that route and only go with this one

src/lavinmq/http/controller/nodes.cr
carlhoerberg: because `name` might not be unique

static/js/nodes.js
spuun: I see I misread the code. I thought this was in the click handler 🤦

PR #2017, opened 2026-06-06
Pause UI polling when the browser tab is hidden

carlhoerberg:
## Motivation

The management UI polls the API every 5s on every open page (overview, queues/connections/exchanges lists, and the queue/exchange/connection/channel/stream/node detail pages). A browser tab left open in the background keeps generating this load indefinitely — for a user with several idle tabs, or many users who never close the dashboard, this is pure waste on the backend.

This PR uses the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to **stop polling while the page is hidden** and **poll immediately when it becomes visible again** (so a returning user sees fresh data without waiting for the next interval).

## What changed

Two distinct polling mechanisms existed, so there are two fixes:

1. **`setInterval`-based polling** (detail pages + overview/nodes). New reusable helper `Helpers.pollWhileVisible(fn, interval)`:
   - runs the interval only while visible
   - on hide → `clearInterval`
   - on show → calls `fn()` immediately, then restarts the interval

   Applied in `channel.js`, `exchange.js`, `connection.js`, `stream.js`, `queue.js`, `overview.js`, `nodes.js`.

2. **`setTimeout`-chain auto-reload** in the table `DataSource` (`datasource.js`, used by all list pages). `_enqueueReload()` now skips scheduling while hidden, and a `visibilitychange` listener clears the pending timer on hide / reloads immediately on show.

## Tests

Playwright frontend specs (`overview.spec.js` for the `pollWhileVisible` path, `queues.spec.js` for the `DataSource` path) assert that:
- advancing the virtual clock while hidden fires **no** requests, and
- becoming visible triggers an immediate request and resumes the interval.

A `setPageVisibility` helper was added to `spec/frontend/helpers.js`.

> Note: I verified locally by running a debug server + Chromium. The `overview.spec.js` tests (and a throwaway DataSource spec) pass. The full `queues.spec.js` suite needs Node 24 locally (the shared `apimap` fixture uses `RegExp.escape`); CI runs Node 24 so it should be fine there.

## 🗣️ Discussion points

The current behavior is **stop completely as soon as the tab is hidden**. Two open questions before marking ready:

1. **Hard stop vs. throttle.** Should we stop entirely on hide (current), or instead **throttle** — keep polling but at a slower cadence (e.g. every 60s) after the tab has been hidden for a while? Trade-offs:
   - *Hard stop:* maximum backend savings; data can be arbitrarily stale, but we force an immediate refresh on return so the user never *sees* stale data.
   - *Throttle:* charts/rates keep a rough continuity while backgrounded (useful if someone glances back and forth between tabs), at the cost of continued — if reduced — load.

2. **Grace period?** Should hide take effect immediately, or only after the tab has been hidden for N seconds — to avoid churn from quick tab flicks (alt-tab away and back)? With the current immediate-refresh-on-show this is mostly cosmetic, but it affects how aggressively we tear down/restart timers.

My instinct is the hard stop here is the right default (simplest, biggest win, and immediate refresh on return hides the staleness), but happy to switch to a throttle or add a grace period if we think background continuity matters for the rate charts.

### DIFF
--- a/static/js/datasource.js
+++ b/static/js/datasource.js
@@ -82,6 +82,15 @@ class DataSource {
         this.reload({ updateState: false })
       })
     }
+    // Pause auto-reloading while the page is hidden (Page Visibility API) and
+    // reload immediately when it becomes visible again.
+    document.addEventListener('visibilitychange', () => {
+      if (document.hidden) {
+        clearTimeout(this._reloadTimer)
+      } else {
+        this.reload()
+      }
+    })
   }
 
   _setStateFromHash () {
@@ -164,7 +173,9 @@ class DataSource {
   }
 
   _enqueueReload () {
-    if (this._opts.autoReloadTimeout > 0) {
+    // Don't schedule another reload while the page is hidden; the
+    // visibilitychange handler reloads as soon as it becomes visible again.
+    if (this._opts.autoReloadTimeout > 0 && !document.hidden) {
       clearTimeout(this._reloadTimer)
       this._reloadTimer = setTimeout(this.reload.bind(this), this._opts.autoReloadTimeout)
     }
--- a/spec/frontend/helpers.js
+++ b/spec/frontend/helpers.js
@@ -53,4 +53,14 @@ async function trackCspViolations (page) {
   return () => page.evaluate(() => window.__cspViolations || [])
 }
 
-export { waitForPathRequest, trackCspViolations }
+// Simulate the Page Visibility API by overriding document.hidden /
+// document.visibilityState and dispatching a visibilitychange event.
+async function setPageVisibility (page, state) {
+  await page.evaluate((state) => {
+    Object.defineProperty(document, 'visibilityState', { value: state, configurable: true })
+    Object.defineProperty(document, 'hidden', { value: state === 'hidden', configurable: true })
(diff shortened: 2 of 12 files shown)


static/js/datasource.js
```diff
@@ -82,6 +82,15 @@ class DataSource {
         this.reload({ updateState: false })
       })
     }
+    // Pause auto-reloading while the page is hidden (Page Visibility API) and
+    // reload immediately when it becomes visible again.
+    document.addEventListener('visibilitychange', () => {
+      if (document.hidden) {
+        clearTimeout(this._reloadTimer)
+      } else {
+        this.reload()
```
snichme, line 91: This will change how it works today and I'm not sure if it's the right thing to do. 
Lets say you are debugging a cluster and looking at a list of consumers, if you switch tab to something else and then back again the view will reload and you might loose the data you are looking at since consumers might change. I think in those cases you want the view to be still. Today we have some views where we set `autoReloadTimeout` to 0 which means don't reload and we should probably keep that even with this feature.

carlhoerberg: true

PR #2019, opened 2026-06-08
Log kTLS=off instead of kTLS= when offload is inactive

jage:
TLS connections without kernel TLS offload logged "kTLS=" with an empty value, which reads like a missing field. Fall back to "off" in the log line so the value is always explicit.

### DIFF
--- a/spec/server_spec.cr
+++ b/spec/server_spec.cr
@@ -1,7 +1,21 @@
 require "./spec_helper"
 require "benchmark"
+require "log/spec"
 
 describe LavinMQ::Server do
+  it "logs kTLS=off for TLS connections without kernel offload" do
+    with_amqp_server(tls: true) do |s|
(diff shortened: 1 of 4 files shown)


carlhoerberg:
Or modify `def ktls_status` ? `case ... in {false, false} then "off"`

jage:
Yes, I didn't want to touch since it was in stdlib, what will happen when our "monkey patch" is removed? Will you sync this will Crystal implementation?

Since we now have a spec it should break if it is changed at least.

carlhoerberg:
`ktls_status` is not in stdlib, it's something we add: https://github.com/search?q=repo%3Acrystal-lang%2Fcrystal%20ktls_status&type=code

jage:
Yeah just assumed we would try to upstream it.

PR #2024, opened 2026-06-08
Fix: sub-second stats_interval produced NaN/Inf rates

jage:
`interval = Config.instance.stats_interval // 1000` truncates to 0 when stats_interval < 1000 ms, so the rate becomes delta / 0 = NaN/Inf, which breaks JSON serialization of the stats endpoints.

Compute interval as float seconds instead.

### DIFF
--- a/spec/config_spec.cr
+++ b/spec/config_spec.cr
@@ -565,6 +565,35 @@ describe LavinMQ::Config do
     config.mqtt_bind.should eq "0.0.0.0"
   end
 
+  describe "stats_interval" do
+    it "accepts sub-second values" do
+      config_file = File.tempfile do |file|
+        file.print <<-CONFIG
+          [main]
+          stats_interval = 500
+        CONFIG
+      end
+      config = LavinMQ::Config.new
+      config.parse(["-c", config_file.path])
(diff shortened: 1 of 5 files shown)


jage:
For me very relevant when working on stats sub-system, reducing interval to run benchmark/analysis more quickly.

carlhoerberg (commented):
why add `::LavinMQ` when you already are in that namespace?

jage:
> why add `::LavinMQ` when you already are in that namespace?

Ah that's the specs fault here, specs didn't expand correctly but will put the spec probe in the LavinMQ module instead.

viktorerlingsson:
Should this also be handled in `server.cr:377`? we do `interval = @config.stats_interval.milliseconds.to_i` which evaluates to 0 if `stats_interval < 1000`

jage:
We also allow 0 and negative numbers in the config, adding a guard for that. Then there's some hardcoded stuff in the JS, but leaving that for now, I will most likely adress that in the new graph PR.

viktorerlingsson (approved):
LGTM! 👍

snichme (approved):

PR #2025, opened 2026-06-08
Fix: HTTP API drop rate-history log from channel/connection lists

jage:
Bugfix, drop the log from the list representation; the detail endpoints (/api/channels/:name, /api/connections/:name) still return the full log.

We only show the current rate, never any history, huge resources used to serve these lists, follow the convention we have for queues etc.

GET /api/channels, 20k channels (warm logs), before vs after:

    response   59 MB  -> 8 MB    (7x smaller)
    latency    758 ms -> 120 ms  (6x faster)
    cpu/req    765 ms -> 86 ms   (9x less)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `unix_proxy_protocol` config option; Unix sockets always auto-detect PROXY protocol headers [#1601](https://github.com/cloudamqp/lavinmq/pull/1601)
 
+### Fixed
+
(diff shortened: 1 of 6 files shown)


jage:
One could argue it's a breaking change, I argue it's a bugfix since this isn't how it's supposed to work. However we could add an optional parameter that the UI passes to keep API compatibility, I don't think it's worth it.

carlhoerberg:
> One could argue it's a breaking change, I argue it's a bugfix since this isn't how it's supposed to work. However we could add an optional parameter that the UI passes to keep API compatibility, I don't think it's worth it.

Agree, not worth it

carlhoerberg (commented):
Should not like in `Queue` have `details_tuple` with only current stats and a custom `to_json` with log stats?

jage:
> Should not like in `Queue` have `details_tuple` with only current stats and a custom `to_json` with log stats?

Done.

carlhoerberg (approved):

PR #2026, opened 2026-06-08
MQTT: O(1) session binding lookup (fix O(n^2) subscribe/disconnect)

jage:
`Session#subscribe`/`#unsubscribe` called `find_binding`, which `flat_mapped` ALL
exchange bindings into a fresh `Array(BindingDetails)` and scanned it: O(n) CPU +
O(n) alloc per call, O(n^2) total (~55% GC + ~18% bindings_details at 15k
sessions).

Add a reverse index (`session` -> `{routing_key -> arguments}`) updated in
`#bind`/`#unbind`, so the lookup is O(1) via `Exchange#subscription`. `#bind`/`#unbind`
also run on definition load, so restored persistent sessions are populated.

I was working on benchmarking connections/sessions, but couldn't reach my goal of 100k durable sessions due to CPU starvation. For 20k sessions this is 3.7x faster at least.

See per commit message for subscribe and cleanup improvements, plus a refactoring to avoid having a redundant `@bindings` now.

### DIFF
--- a/spec/mqtt/integrations/subscribe_spec.cr
+++ b/spec/mqtt/integrations/subscribe_spec.cr
@@ -131,5 +131,32 @@ module MqttSpecs
         end
       end
     end
+
+    it "keeps multiple subscriptions independent across resubscribe/unsubscribe" do
+      with_server do |server|
+        with_client_io(server) do |io|
+          connect(io)
+          subscribe(io, topic_filters: mk_topic_filters({"m/1", 0}, {"m/2", 0}, {"m/3", 0}))
+          # resubscribe only m/2 at qos1, and unsubscribe m/1
+          subscribe(io, topic_filters: mk_topic_filters({"m/2", 1}))
+          unsubscribe(io, topics: ["m/1"])
+
+          # m/1 is gone: publishing to it delivers nothing
+          publish(io, topic: "m/1", payload: "x".to_slice, qos: 1u8)
+          # m/2 still bound, now at qos1
+          publish(io, topic: "m/2", payload: "x".to_slice, qos: 1u8)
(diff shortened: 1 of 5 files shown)


jage:
Need to think about how to solve the "Compile changed *.cr files independently" check, hmm.

carlhoerberg:
> Need to think about how to solve the "Compile changed *.cr files independently" check, hmm.

Just add this to `src/lavinmq/mqtt/sessionc.cr`
```crystal
require "./exchange"
```

carlhoerberg:
```patch
commit b43d4d9b83c3f680c6f7122b31395f522f998ecf (HEAD -> mqtt-find-binding-o1)
Author: Carl Hörberg <carl@84codes.com>
Date:   Mon Jun 8 19:53:37 2026 +0200

    make mqtt/session.cr compile individually

diff --git a/src/lavinmq/mqtt/session.cr b/src/lavinmq/mqtt/session.cr
index 7a8a90882..8fb7a0393 100644
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -3,6 +3,7 @@
 require "../sortable_json"
 require "./client"
 require "./consts"
+require "./exchange"

 module LavinMQ
   module MQTT

jage:
@carlhoerberg hehe, yeah now each file compiles independent but not together since session requires exchange which require session. :)

Looks like #957 is on the right track to improve this.

carlhoerberg:
It does compile?

jage:
> It does compile?

Yeah but no MQTT specs work and there seems to something about those references between the files, without that require they all pass. I'm not sure what's going on TBH.

jage:
@spuun lets close this and just do a benchmark once you're done with #957

PR #2029, opened 2026-06-08
Make the lavinmqctl control socket path configurable

dentarg:
The control socket was hardcoded to `/tmp/lavinmqctl.sock`, so two LavinMQ instances on one host collided: each rebinds that path on startup (hijacking `lavinmqctl`) and deletes it on shutdown.

Add a `control_unix_path` option — CLI `--control-unix-path`, INI `[main] control_unix_path`, env `LAVINMQ_CONTROL_UNIX_PATH` — defaulting to `/tmp/lavinmqctl.sock` for backwards compatibility. The HTTP server bind, shutdown cleanup, clustering-follower socket and the auth bypass now all read the configured path. `lavinmqctl` gains a matching `--control-unix-path` flag and `LAVINMQCTL_CONTROL_UNIX_PATH` env var.

Run multiple instances by giving each a distinct `--control-unix-path` (demonstrated in `extras/start-cluster-in-tmux.sh`).

### DIFF
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -2,9 +2,6 @@
 data_dir = /var/lib/lavinmq
 log_level = info
 ;pidfile = /var/run/lavinmq.pid
-; Socket that lavinmqctl connects to. Set a unique path per instance when
-; running multiple LavinMQ servers on the same host.
-;control_unix_path = /tmp/lavinmqctl.sock
 ;default_user_only_loopback = true
 ;tls_cert = /etc/lavinmq/cert.pem
 ;tls_key = /etc/lavinmq/key.pem
--- a/extras/lavinmq.ini
+++ b/extras/lavinmq.ini
@@ -2,6 +2,9 @@
 data_dir = /var/lib/lavinmq
 log_level = info
 ;pidfile = /var/run/lavinmq.pid
+; Socket that lavinmqctl connects to. Set a unique path per instance when
+; running multiple LavinMQ servers on the same host.
+;control_unix_path = /tmp/lavinmqctl.sock
 ;default_user_only_loopback = true
 ;tls_cert = /etc/lavinmq/cert.pem
 ;tls_key = /etc/lavinmq/key.pem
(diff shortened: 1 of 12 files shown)


carlhoerberg:
Good! In addition to this I think we should check whether an existing unix socket is in use before deleting it, with something like this: 

```crystal
  private def prepare_control_socket(path)
    return unless info = File.info?(path, follow_symlinks: false)

    unless info.type.socket?
      raise "Control socket #{path} exists and is not a socket"
    end

    begin
      UNIXSocket.open(path) { }
      raise "Control socket #{path} is already in use"
    rescue ex : Socket::ConnectError
      # ECONNREFUSED: socket inode exists, but nobody is listening.
      File.delete(path)
    rescue ex : Socket::Error
      # EACCES or anything ambiguous: fail closed, don't delete.
      raise "Cannot verify stale control socket #{path}: #{ex.message}"
    end
  end

**Comment by dentarg** [comment]
That sounds like a good idea, but also something for a follow-up PR?

File: src/lavinmq/http/constants.cr [hunk 1]
```diff
@@ -1,6 +1,6 @@
 module LavinMQ
   module HTTP
-    INTERNAL_UNIX_SOCKET = "/tmp/lavinmqctl.sock"
-    MAX_PAGE_SIZE        = 10_000
+    DEFAULT_CONTROL_UNIX_PATH = "/tmp/lavinmqctl.sock"
```
carlhoerberg, line 3: Can't the default value be in the config only instead?

carlhoerberg: no, used by the lavinmctl cli parser too.

carlhoerberg (approved):

extras/lavinmq.ini
```diff
@@ -2,6 +2,9 @@
 data_dir = /var/lib/lavinmq
 log_level = info
 ;pidfile = /var/run/lavinmq.pid
+; Socket that lavinmqctl connects to. Set a unique path per instance when
+; running multiple LavinMQ servers on the same host.
+;control_unix_path = /tmp/lavinmqctl.sock
```
carlhoerberg: this is not a very common config option, let's not pollute this config with it, unless we go the route and document ALL options we have. 
```suggestion
```

dentarg: Sure

PR #2030, opened 2026-06-08
Fix broker crashes and concurrency races surfaced by the stress test

dentarg:
## Summary

The "Stress test" CI job (`extras/stress.cr`) deliberately interleaves heavy
traffic with topology churn — bind/unbind, declare/delete, connection kills,
vhost/policy churn, dead-lettering, TTLs, streams, and federation — to flush out
multi-threading bugs under `-Dpreview_mt`. This branch fixes every crash and
spurious error it turned up. Each is an independent fix with its own regression
spec.

The common thread: an operation races a concurrent teardown/churn. The fixes
either add the missing serialization/lock, or treat a concurrently-closed
resource as a benign no-op (drop / `GetEmpty`) instead of raising or crashing.

## Fixes

**Streams / message store**
- Skip the message-expire fiber on streams, avoiding a `Closed mfile` crash when
  the last consumer disconnects after retention dropped segments (`ebdd6077`).
- Serialize per-resource policy application and run the stream's `drop_overflow`
  (which `munmap`s segments) under the message-store lock, fixing a **segfault**
  when policies are applied concurrently with publishing/consuming (`46d5841c`).

**Queue lifecycle (delete racing traffic)**
- Treat a publish to a concurrently-closed queue as dropped instead of raising,
  which previously surfaced as an HTTP publish `500` (`a486e82c`).
- Reply with `Basic.GetEmpty` instead of erroring when a `basic_get` races a
  concurrent queue delete (`61586d19`).

**AMQP channel / connection**
- Discard settlement frames (`ack`/`nack`/`reject`) for an already-closed
  channel instead of tearing down the connection with `CHANNEL_ERROR`. Channels
  a connection has opened are tracked in a set, so a frame on a never-opened
  channel (even one below the highest opened) is still a protocol error
  (`61fb1de2`).

**VHost / MQTT**
- Serialize vhost-store saves so concurrent vhost create/delete don't race on the
  shared `vhosts.json` temp file and fail with a rename error (`e7a903e8`).
- Don't crash the MQTT brokers on a vhost `Closed` event for a vhost that was
  never fully registered (e.g. one whose create didn't finish) (`3f115a08`).

**Logging**
- Stop federation/shovel links before tearing down vhosts on shutdown, and keep
  the embedded amqp-client's routine connection-teardown logging out of the
  broker log — it's already reported via the `lmq.*` layers and was also tripping
  the stress-test log check (`0e75921f`).

## Testing

- Every fix ships with a regression spec; the full suite and `ameba` pass.
- End-to-end validation: a 30-minute stress run against the release binary
  finished with **`Aliveness OK`** — ~750M publishes, ~14.5k policy
  applies/removes, **0 broker `ERROR`/`FATAL`**, and no crash or deadlock. For
  comparison, the unfixed broker segfaulted within ~15 minutes under the same
  workload.

### DIFF
--- a/spec/upstream_spec.cr
+++ b/spec/upstream_spec.cr
@@ -1,4 +1,5 @@
 require "./spec_helper"
+require "log/spec"
 require "../src/lavinmq/federation/upstream"
 
 module UpstreamSpecHelpers
@@ -991,6 +992,27 @@ describe LavinMQ::Federation::Upstream do
             v3fe.publish_in_count.should eq 0
           end
         end
+      end
+    end
+  end
+
+  describe "shutdown" do
+    it "closes federation links without logging unexpected-close errors" do
+      # A federation link's upstream connection lives on another vhost. On
+      # broker shutdown every link must be stopped first (cleanly closing those
+      # connections against still-live peers); otherwise a peer vhost can
+      # force-close the link's socket mid-flight and the amqp-client read loop
+      # logs "connection closed unexpectedly" / "Couldn't write CloseOk frame".
+      Log.capture("amqp.client.connection", :error) do |logs|
+        with_amqp_server do |s|
+          upstream, up, down = UpstreamSpecHelpers.setup_federation(s, "shutdown-test", nil, "uq")
+          up.declare_queue("uq", false, false)
+          down.declare_queue("dq", false, false)
+          link = upstream.link(down.queue("dq"))
+          wait_for { link.state.running? }
+        end
+        # with_amqp_server's teardown has now run s.close (the broker shutdown).
+        logs.empty
       end
     end
   end
--- a/src/lavinmq/amqp/client.cr
+++ b/src/lavinmq/amqp/client.cr
@@ -399,11 +399,20 @@ module LavinMQ
           when AMQP::Frame::Body
             @log.trace { "Discarding #{frame.class.name}, waiting for Close(Ok)" }
             frame.body.skip(frame.body_size)
+          when AMQP::Frame::Basic::Ack, AMQP::Frame::Basic::Nack, AMQP::Frame::Basic::Reject
+            # A settlement can race a channel close in another fiber and arrive
+            # after we've removed the channel; discard it rather than killing the
+            # connection.
+            @log.warn { "Discarding #{frame.class.name} for unknown channel #{frame.channel}" }
           else
-            @log.error { "Channel #{frame.channel} not open while processing #{frame.class.name}" }
-            close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "Channel #{frame.channel} not open")
+            reject_unknown_channel(frame)
           end
         end
+      end
+
+      private def reject_unknown_channel(frame) : Nil
+        @log.error { "Channel #{frame.channel} not open while processing #{frame.class.name}" }
+        close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "Channel #{frame.channel} not open")
       end
 
       private def open_channel(frame)
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -821,6 +821,11 @@ module LavinMQ::AMQP
       get(no_ack) do |env|
         yield env
       end.tap { ensure_expire_fiber }
+    rescue ClosedError | MessageStore::ClosedError
+      # The queue was closed/deleted concurrently with the get (its @closed flag
+      # or the message store). Report no message available (Basic.GetEmpty)
+      # rather than letting the error escape into the connection's read loop.
+      false
     end
 
     # If nil is returned it means that the delivery limit is reached
--- a/src/lavinmq/amqp/queue/queue.cr
+++ b/src/lavinmq/amqp/queue/queue.cr
@@ -605,6 +605,7 @@ module LavinMQ::AMQP
       publish_internal(msg)
     end
 
+    # ameba:disable Metrics/CyclomaticComplexity
     protected def publish_internal(msg : Message, dlx_tasks : Argument::DeadLettering::Tasks? = nil) : PublishResult
       return PublishResult::Dropped if @closed
       if d = @deduper
@@ -616,9 +617,11 @@ module LavinMQ::AMQP
       end
       return PublishResult::Overflow if reject_on_overflow?(msg)
       was_empty = false
+      pushed = false
       @msg_store_lock.synchronize do
         was_empty = @msg_store.empty?
         @msg_store.push(msg)
+        pushed = true
         drop_overflow(dlx_tasks)
       end
       @publish_count.add(1, :relaxed)
@@ -630,6 +633,11 @@ module LavinMQ::AMQP
       end
 
       PublishResult::Ok
+    rescue MessageStore::ClosedError
+      # A racing delete closed the store. If push hadn't stored the message it's
+      # dropped (avoids an HTTP 500); if it had, a later ClosedError from the
+      # post-publish expire-fiber check still means the publish succeeded.
+      pushed ? PublishResult::Ok : PublishResult::Dropped
     rescue ex : MessageStore::Error
       @log.error(ex) { "Queue closed due to error" }
       close
--- a/src/lavinmq/policy.cr
+++ b/src/lavinmq/policy.cr
@@ -6,29 +6,32 @@ module LavinMQ
     getter policy : Policy?
     getter operator_policy : OperatorPolicy?
     getter effective_policy_args = Array(String).new
+    # apply_policies is spawned per policy/parameter add & delete, so concurrent
+    # applies can hit the same resource; serialize them to protect the shared
+    # @effective_args / store state.
+    @policy_lock = Mutex.new(:reentrant)
 
     def reapply_policy
       apply_policy(@policy, @operator_policy)
     end
 
     def apply_policy(policy : Policy?, operator_policy : OperatorPolicy?)
-      clear_policy
-      effective_policy_args = Array(String).new
-      Policy.merge_definitions(policy, operator_policy).each do |key, value|
-        if apply_policy_argument(key, value)
-          effective_policy_args << key
+      @policy_lock.synchronize do
+        clear_policy
+        effective_policy_args = Array(String).new
+        Policy.merge_definitions(policy, operator_policy).each do |key, value|
+          if apply_policy_argument(key, value)
+            effective_policy_args << key
+          end
+        rescue ex
+          # Skip an invalid policy argument and carry on with the rest.
+          Log.warn(exception: ex) { "Error applying policy argument #{key}=#{value}: #{ex.message}" }
         end
-      rescue ex
-        # We rescue exceptions and ignore the argument, continuing on the next one.
-        #
-        # This log line isn't very good. Sometimes @log should be used, but we can't know that here. With the new L
-        # logging it's not as important.
-        Log.warn(exception: ex) { "Error applying policy argument #{key}=#{value}: #{ex.message}" }
+        @effective_policy_args = effective_policy_args
+        @policy = policy
+        @operator_policy = operator_policy
+        after_policy_applied
       end
-      @effective_policy_args = effective_policy_args
-      @policy = policy
-      @operator_policy = operator_policy
-      after_policy_applied
     end
 
     def clear_policy
--- a/src/lavinmq/vhost_store.cr
+++ b/src/lavinmq/vhost_store.cr
@@ -82,6 +82,20 @@ module LavinMQ
     end
 
     def close
+      # Stop outbound links (federation upstreams and shovels) on every vhost
+      # before tearing any vhost down. A link holds an AMQP client connection to
+      # another vhost, so force-closing one vhost's connections while another
+      # vhost's link is still live would close that link's socket from under it —
+      # which the amqp-client read loop logs as "connection closed unexpectedly".
+      # Stopping them first lets each link close cleanly against a live peer.
+      WaitGroup.wait do |wg|
+        @vhosts.each_value do |vhost|
+          wg.spawn do
+            vhost.stop_shovels
+            vhost.stop_upstream_links
+          end
+        end
+      end
       WaitGroup.wait do |wg|
         @vhosts.each_value do |vhost|
           wg.spawn do
--- a/src/lavinmq/vhost_store.cr
+++ b/src/lavinmq/vhost_store.cr
@@ -17,6 +17,7 @@ module LavinMQ
 
     def initialize(@data_dir : String, @users : Auth::UserStore, @replicator : Clustering::Replicator?, @persister : Persister)
       @vhosts = Hash(String, VHost).new
+      @save_lock = Mutex.new
     end
 
     def []?(name : String) : VHost?
@@ -143,8 +144,12 @@ module LavinMQ
     def save!
       Log.debug { "Saving vhosts to file" }
       path = File.join(@data_dir, "vhosts.json")
-      File.open("#{path}.tmp", "w") { |f| to_pretty_json(f); f.fsync }
-      File.rename "#{path}.tmp", path
+      # Serialize saves so concurrent create/delete don't race on the shared
+      # `.tmp` file and fail the rename.
+      @save_lock.synchronize do
+        File.open("#{path}.tmp", "w") { |f| to_pretty_json(f); f.fsync }
+        File.rename "#{path}.tmp", path
+      end
       @replicator.try &.replace_file path
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Discard settlement frames (`ack`/`nack`/`reject`) for an already-closed channel instead of closing the connection with `CHANNEL_ERROR`
 - Treat a publish to a concurrently-closed queue as dropped instead of raising, which previously surfaced as an HTTP publish `500`
 - Stop federation and shovel links before tearing down vhosts on shutdown, and keep the embedded amqp-client's routine connection-teardown logging (already reported via the `lmq.*` federation/shovel layers) out of the broker log
+- Reply with `Basic.GetEmpty` instead of erroring when a `basic_get` races a concurrent queue delete
 
 ## [2.8.1] - 2026-05-18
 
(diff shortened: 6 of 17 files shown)


dentarg:
I force pushed this branch to address review (Claude from above + Codex), see details below for the findings and initial push vs current state diffs.

<details><summary>Details</summary>
<p>

> • Findings (Codex)
> 
>   1. Medium - src/lavinmq/amqp/client.cr:398 now drops Basic::Ack/Nack/Reject for any missing
>      channel, not just an already-closed one. A malformed client can send Basic.Ack on channel
>      0 or a never-opened channel and keep the connection alive. Previously this hit
>      CHANNEL_ERROR. The fix needs to distinguish recently closed channels from never-opened/
>      invalid channels, and add a spec for that case.
> 
>   2. Medium - src/lavinmq/amqp/queue/queue.cr:633 rescues MessageStore::ClosedError for the
>      whole publish method. That includes the post-push TTL path at src/lavinmq/amqp/queue/
>      queue.cr:627. If a per-message-TTL publish succeeds, then a concurrent close makes
>      ensure_expire_fiber hit a closed store, this now reports Dropped even though the message
>      may already be persisted. Scope the rescue to the push race or track whether push
>      succeeded.

> ---
> 
> ### Commit 1 — `Skip message-expire fiber on streams` (`06737455` → `ebdd6077`)
> **Unchanged** — `git diff 06737455 ebdd6077` is empty. No review items touched this commit.
> 
> ---
> 
> ### Commit 2 — `Discard settlement frames on a closed channel` (`6f3bb339` → `9927e557`)
> Addresses **Codex #1** (don't drop frames for channel 0 / never-opened channels):
> 
> ```diff
> diff --git a/src/lavinmq/amqp/client.cr b/src/lavinmq/amqp/client.cr
> @@ -35,6 +35,10 @@ module LavinMQ
>        @connected_at = RoughTime.unix_ms
>        @channels = Hash(UInt16, Client::Channel).new
> +      # Highest channel number ever opened on this connection. Used to tell a
> +      # settlement frame racing a channel close (a channel we opened, now gone)
> +      # apart from one for a never-opened channel, which is a protocol error.
> +      @highest_channel_opened = 0_u16
>        @actual_channel_max : UInt16
> @@ -400,15 +404,25 @@ module LavinMQ
>              # unacked messages are gone, so there's nothing to act on — discard it
> -            # instead of failing the whole connection with a CHANNEL_ERROR.
> -            @log.debug { "Discarding #{frame.class.name} for closed channel #{frame.channel}" }
> +            # instead of failing the whole connection. Only for a channel this
> +            # connection actually opened, though; a settlement frame on channel 0
> +            # or a never-opened channel is still a protocol error.
> +            if frame.channel.zero? || frame.channel > @highest_channel_opened
> +              reject_unknown_channel(frame)
> +            else
> +              @log.debug { "Discarding #{frame.class.name} for closed channel #{frame.channel}" }
> +            end
>            else
> -            @log.error { "Channel #{frame.channel} not open while processing #{frame.class.name}" }
> -            close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "Channel #{frame.channel} not open")
> +            reject_unknown_channel(frame)
>            end
>          end
>        end
>  
> +      private def reject_unknown_channel(frame) : Nil
> +        @log.error { "Channel #{frame.channel} not open while processing #{frame.class.name}" }
> +        close_connection(frame, ConnectionReplyCode::CHANNEL_ERROR, "Channel #{frame.channel} not open")
> +      end
> +
>        private def open_channel(frame)
> @@ -418,6 +432,7 @@ module LavinMQ
>            @channels[frame.channel] = AMQP::Channel.new(self, frame.channel)
> +          @highest_channel_opened = frame.channel if frame.channel > @highest_channel_opened
>            @vhost.event_tick(EventType::ChannelCreated)
>            send AMQP::Frame::Channel::OpenOk.new(frame.channel)
> ```
> ```diff
> diff --git a/spec/connection_spec.cr b/spec/connection_spec.cr
> @@ -139,6 +139,24 @@ describe LavinMQ::Server do
> +    it "closes the connection for a settlement frame on a never-opened channel" do
> +      with_amqp_server do |s|
> +        with_raw_amqp_connection(s) do |io, stream|
> +          io.write_bytes AMQ::Protocol::Frame::Channel::Open.new(1_u16), IO::ByteFormat::NetworkEndian
> +          io.flush
> +          stream.next_frame.as(AMQ::Protocol::Frame::Channel::OpenOk)
> +          # Channel 5 was never opened on this connection, so an ack there is a
> +          # protocol violation, not a close race — the connection must be closed
> +          # with CHANNEL_ERROR rather than silently dropping the frame.
> +          io.write_bytes AMQ::Protocol::Frame::Basic::Ack.new(5_u16, 1_u64, false), IO::ByteFormat::NetworkEndian
> +          io.flush
> +          close = stream.next_frame.as(AMQ::Protocol::Frame::Connection::Close)
> +          close.reply_code.should eq 504
> +        end
> +      end
> +    end
> ```
> 
> ---
> 
> ### Commit 3 — `Drop publish to a concurrently-closed queue` (`25d44f01` → `d38bd58d`)
> Addresses **Codex #2** (don't report a persisted message as dropped) and the **Claude-bot** missing-spec note:
> 
> ```diff
> diff --git a/src/lavinmq/amqp/queue/queue.cr b/src/lavinmq/amqp/queue/queue.cr
> @@ -605,6 +605,7 @@ module LavinMQ::AMQP
>        publish_internal(msg)
>      end
>  
> +    # ameba:disable Metrics/CyclomaticComplexity
>      protected def publish_internal(msg : Message, dlx_tasks : Argument::DeadLettering::Tasks? = nil) : PublishResult
> @@ -616,9 +617,11 @@ module LavinMQ::AMQP
>        return PublishResult::Overflow if reject_on_overflow?(msg)
>        was_empty = false
> +      pushed = false
>        @msg_store_lock.synchronize do
>          was_empty = @msg_store.empty?
>          @msg_store.push(msg)
> +        pushed = true
>          drop_overflow(dlx_tasks)
>        end
> @@ -631,11 +634,13 @@ module LavinMQ::AMQP
>        PublishResult::Ok
>      rescue MessageStore::ClosedError
> -      # The queue was closed/deleted concurrently, between the @closed check
> -      # above and acquiring the store lock. Treat it as dropped, just like the
> -      # early @closed return, so a racing delete doesn't fail delivery to the
> -      # other routed queues or surface as an HTTP 500.
> -      PublishResult::Dropped
> +      # The queue was closed/deleted concurrently with this publish. If the
> +      # message wasn't stored yet (the @closed check above races the store
> +      # close) it's dropped — so a racing delete doesn't fail delivery to the
> +      # other routed queues or surface as an HTTP 500. If it was already stored,
> +      # the failure is in best-effort post-publish bookkeeping (e.g. the
> +      # expire-fiber check) and the publish still succeeded.
> +      pushed ? PublishResult::Ok : PublishResult::Dropped
>      rescue ex : MessageStore::Error
> ```
> ```diff
> diff --git a/src/lavinmq/amqp/stream/stream.cr b/src/lavinmq/amqp/stream/stream.cr
> @@ -154,6 +154,7 @@ module LavinMQ::AMQP
>      rescue MessageStore::ClosedError
>        # Closed/deleted concurrently after the @state.closed? check; treat as
>        # dropped instead of surfacing the race as an error (see Queue#publish_internal).
> +      # push is the only call here that can raise it, so nothing was stored.
>        PublishResult::Dropped
> ```
> ```diff
> diff --git a/spec/stream_queue_spec.cr b/spec/stream_queue_spec.cr
> @@ -461,6 +461,21 @@ describe LavinMQ::AMQP::Stream do
> +  it "drops a publish when the store is closed concurrently" do
> +    with_amqp_server do |s|
> +      with_channel(s) do |ch|
> +        ch.queue("stream-closed-store-publish", args: stream_queue_args)
> +        stream = s.vhosts["/"].queue("stream-closed-store-publish").as(LavinMQ::AMQP::Stream)
> +        # Simulate a delete racing publish_internal: the store is closed after
> +        # the @state.closed? check but before push, so push raises ClosedError.
> +        # The publish must report Dropped, not raise (mirrors the queue spec).
> +        stream.@msg_store.close
> +        msg = LavinMQ::Message.new("", "stream-closed-store-publish", "body", LavinMQ::AMQP::Properties.new)
> +        stream.publish(msg).dropped?.should be_true
> +      end
> +    end
> +  end
> ```
> 
> ---
> 
> Two heads-ups for the PR note:
> - I also lightly **reworded the commit messages** for commits 2 and 3 to match the refined behavior (`git range-diff` shows these under "Commit message"). Commit 1's message is untouched.
> - `queue_spec.cr` and `CHANGELOG.md` are byte-identical between old and new commit 3 (their content didn't change, only moved within the same commit), so they don't appear in the diff.

</p>
</details>

src/lavinmq/amqp/client.cr
```diff
@@ -35,6 +35,10 @@ module LavinMQ
 
       @connected_at = RoughTime.unix_ms
       @channels = Hash(UInt16, Client::Channel).new
+      # Highest channel number ever opened on this connection. Used to tell a
+      # settlement frame racing a channel close (a channel we opened, now gone)
+      # apart from one for a never-opened channel, which is a protocol error.
+      @highest_channel_opened = 0_u16
```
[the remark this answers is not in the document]
dentarg: > Good catch. Replaced the `@highest_channel_opened` high-water mark with a `Set(UInt16)` of channels opened at least once (`@opened_channels`). Now a gap like opening 1 and 5 no longer makes a settlement frame on the never-opened 2 look like a benign close race. Fixed in `61fb1de2`.

src/lavinmq/amqp/client.cr
```diff
@@ -395,13 +399,30 @@ module LavinMQ
...
+            # arrive after we've already removed the channel. The channel and its
+            # unacked messages are gone, so there's nothing to act on — discard it
+            # instead of failing the whole connection. Only for a channel this
+            # connection actually opened, though; a settlement frame on channel 0
+            # or a never-opened channel is still a protocol error.
+            if frame.channel.zero? || frame.channel > @highest_channel_opened
+              reject_unknown_channel(frame)
+            else
+              @log.debug { "Discarding #{frame.class.name} for closed channel #{frame.channel}" }
+            end
```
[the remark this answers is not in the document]
dentarg: > Agreed. The check is now `frame.channel.zero? || !@opened_channels.includes?(frame.channel)`, so a settlement on a never-opened channel below the highest opened is correctly rejected as a protocol error. Added a regression test for exactly this case (open 1 and 5, ack on 2 → `CHANNEL_ERROR`), which fails on the old high-water-mark logic. `61fb1de2`.

src/lavinmq/amqp/client.cr
```diff
@@ -411,6 +432,7 @@ module LavinMQ
           close_connection(frame, ConnectionReplyCode::NOT_ALLOWED, reply_text)
         else
           @channels[frame.channel] = AMQP::Channel.new(self, frame.channel)
+          @highest_channel_opened = frame.channel if frame.channel > @highest_channel_opened
```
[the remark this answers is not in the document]
dentarg: > Done — `open_channel` now records it with `@opened_channels << frame.channel`. `61fb1de2`.

src/lavinmq/policy.cr
```diff
@@ -6,29 +6,38 @@ module LavinMQ
...
+        effective_policy_args = Array(String).new
+        Policy.merge_definitions(policy, operator_policy).each do |key, value|
+          if apply_policy_argument(key, value)
+            effective_policy_args << key
+          end
+        rescue ex
+          # We rescue exceptions and ignore the argument, continuing on the next one.
+          #
+          # This log line isn't very good. Sometimes @log should be used, but we can't know that here. With the new L
+          # logging it's not as important.
```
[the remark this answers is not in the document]
dentarg: > Rewrote it as a single complete sentence: `# Skip an invalid policy argument and carry on with the rest.` `46d5841c`.

spec/upstream_spec.cr
```diff
@@ -994,4 +995,25 @@ describe LavinMQ::Federation::Upstream do
...
+        with_amqp_server do |s|
+          upstream, up, down = UpstreamSpecHelpers.setup_federation(s, "shutdown-test", nil, "uq")
+          up.declare_queue("uq", false, false)
+          down.declare_queue("dq", false, false)
+          link = upstream.link(down.queue("dq"))
+          wait_for { link.state.running? }
+        end
+        # with_amqp_server's teardown has now run s.close (the broker shutdown).
+        logs.empty
+      end
```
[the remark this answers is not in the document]
dentarg: > This is actually the intended assertion, not a no-op. `Log::EntriesChecker#empty` (Crystal's stdlib `log/spec`) is the verb form that **fails the example if any entries were captured** — see `crystal/src/log/spec.cr:142`, which calls `fail(...)` when `@entries.first?` is present. It's the negative counterpart to `.check`, and `empty?` isn't defined on that type. So the example does fail if unexpected `ERROR` logs appear during shutdown. Leaving it as-is.

https://github.com/crystal-lang/crystal/blob/d2369ac2ea8eca1468ea138c308e73f64cd2357e/src/log/spec.cr#L141-L149

src/lavinmq/amqp/client.cr
```diff
@@ -35,6 +35,12 @@ module LavinMQ
 
       @connected_at = RoughTime.unix_ms
       @channels = Hash(UInt16, Client::Channel).new
+      # Channel numbers this connection has opened at least once. Used to tell a
+      # settlement frame racing a channel close (a channel we opened, now gone)
+      # apart from one for a never-opened channel, which is a protocol error.
+      # A set, not a high-water mark: clients may open channels out of order or
+      # with gaps (e.g. 1 and 5), so a frame on an unopened 2 must still fail.
+      @opened_channels = Set(UInt16).new
```
carlhoerberg: The memory requirement for this is up to 2 MiB for a client that opens all channels at least once, 32 KiB for 2000 channels. I would consider just Log warn stray ack/reject et al.

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -630,6 +633,14 @@ module LavinMQ::AMQP
...
 
       PublishResult::Ok
+    rescue MessageStore::ClosedError
+      # The queue was closed/deleted concurrently with this publish. If the
+      # message wasn't stored yet (the @closed check above races the store
+      # close) it's dropped — so a racing delete doesn't fail delivery to the
+      # other routed queues or surface as an HTTP 500. If it was already stored,
+      # the failure is in best-effort post-publish bookkeeping (e.g. the
+      # expire-fiber check) and the publish still succeeded.
+      pushed ? PublishResult::Ok : PublishResult::Dropped
```
carlhoerberg, line 640: Isnt `MessageStore::ClosedError` always raised before pushed is set to true? Or maybe it can be raised when dropping overflow too?

src/lavinmq/vhost_store.cr
```diff
@@ -129,9 +144,17 @@ module LavinMQ
...
+      # Serialize concurrent saves: vhost create/delete (e.g. under churn) run in
+      # separate fibers and share the same `.tmp` path, so without this two saves
+      # can race and one's `File.rename` finds the tmp already moved by the other
+      # ("No such file or directory"). A failed save in `create` leaves the vhost
+      # half-registered (added to @vhosts but no Added event), which later crashes
+      # observers like the MQTT brokers.
+      @save_lock.synchronize do
+        File.open("#{path}.tmp", "w") { |f| to_pretty_json(f); f.fsync }
+        File.rename "#{path}.tmp", path
+        @replicator.try &.replace_file path
```
carlhoerberg: This is good, but writes to the replicator is already serialized, so no need to include this in this lock too

src/lavinmq/vhost_store.cr
```diff
@@ -129,9 +144,17 @@ module LavinMQ
...
-      File.open("#{path}.tmp", "w") { |f| to_pretty_json(f); f.fsync }
-      File.rename "#{path}.tmp", path
-      @replicator.try &.replace_file path
+      # Serialize concurrent saves: vhost create/delete (e.g. under churn) run in
+      # separate fibers and share the same `.tmp` path, so without this two saves
+      # can race and one's `File.rename` finds the tmp already moved by the other
+      # ("No such file or directory"). A failed save in `create` leaves the vhost
+      # half-registered (added to @vhosts but no Added event), which later crashes
+      # observers like the MQTT brokers.
+      @save_lock.synchronize do
```
carlhoerberg, line 149: Do we need this in the other Stores as well? UserStore, ParameterStore, etc?

viktorerlingsson:
I worked on something pretty similar at the end of last week but didn't have time to finish it & push before other things got in the way 🙂 Anyway, the fixes here are pretty similar to what I came up with 👍 

In general I think most of the new comments can either be more concise or aren't needed at all. I think the code explains itself well enough in many places. For example, I don't think this comment in `brokers.cr` is needed: 

```
          # The broker may be absent if the vhost never finished registering
          # (e.g. its create raised after adding it to the store but before the
          # Added event fired), so don't assume it's present.
          @brokers[vhost]?.try &.close
```

src/lavinmq/amqp/client.cr
viktorerlingsson: yeah, I think just logging warn instead of error is good enough

src/lavinmq/amqp/queue/queue.cr
dentarg: Here's Claude speaking:

> Not from `drop_overflow` — that runs inside the same `@msg_store_lock` block as `push`, and `close` needs that lock too, so the store can't close between them; it won't raise there.
>
> The post-`pushed` source is the **unlocked bookkeeping after the block**: `ensure_expire_fiber → should_start_expire_fiber?` calls `@msg_store.first?` (queue.cr:277), which can hit a store that a racing delete closed *after* the message was already pushed. That's the case the `pushed` flag turns into `Ok`. So it's reachable, just not via `drop_overflow`.
>
> If you want to point at it precisely: `publish_internal` releases `@msg_store_lock` at the end of the `synchronize` block, then runs `ensure_expire_fiber` if the message has a TTL/expiration — and `should_start_expire_fiber?` reads `@msg_store.first?` without the lock, so a concurrent `delete` can close the store right there.

This is now covered with a spec in b3ee462a

dentarg:
Thanks for reviewing @carlhoerberg and @viktorerlingsson, I've addressed the feedback and force pushed the branch, but you can view the diff before vs now at https://gist.github.com/dentarg/c25e098478c2b5c85b8817a870388d9a

src/lavinmq/amqp/client.cr
dentarg: Thanks for catching this, it was addressed in 2e22b823

src/lavinmq/vhost_store.cr
dentarg: Got it, corrected in e094a595

src/lavinmq/vhost_store.cr
dentarg: Yes, we do, I expanded the fix to include them, see e094a595

dentarg:
This was just a rebase with main (solved conflict with the changelog)

carlhoerberg (approved):

PR #2032, opened 2026-06-09
stats: Store object stats as packed column buffers

jage:
TL;DR: Overall efficient, huge improvements on idle exchanges/channels,
i.e. no problem handling a 100k channel leak now.

Complete rewrite of metrics storage, it's API compatible[1] with no
reduction in precision. There's a steady overall improvement in
memory efficiency but a dramatic improvement for idle objects mostly
made up of stats buffers (channels, exchanges).

This replaces a previous optimisation attempt by replacing `Deque` with
`RingBuffer` PR #1353. This is much more focused on memory efficiency
but also has an overall reduction in CPU time spent handling stats.

Instead of objects (Channel, Queue etc) each having N rolling Deques
(12-14 per queue/channel) each object has one buffer that stores all
logs with a shared head/size index.

By using a single `GC.malloc_atomic` buffer (Float64 rates + UInt32
counts) there's not only less memory overhead there's also reduced GC
work as the whole region is marked as "no pointers" and the GC never scans it.

To further optimize the case with "idle" objects, i.e. objects without
any activity with the set stats log interval (default 10 minutes) we can
completely remove this buffer (0 allocations). This gives very
impressive numbers on a large number of idle channels and exchanges
which use most of the memory to handle stats. (86% less memory)

For an active object, we push the allocation to the stats loop fiber
as all objects are "idle by default". Channel churn benchmark shows a
20% improvement.

Reads are handled via `StatLogView`, which manages the buffer base pointer and
column offset. Designed for a concurrent use and keeps whole buffer
alive while any reader iterates even though a concurrent `update_rates`
may compact and free it (no interior pointers).

This does not touch the "hot path" for message handling, nonetheless
benchmarked to ensure no regressions.

Benchmark results:

Idle RSS, 100k objects (MiB, peak-base)

    object                       before   after  reduction
    --------------------------- ------- ------- ----------
    exchanges                       507      72       -86%
    channels                       1416     200       -86%
    queues                         1878     393       -79%
    streams                        2425     944       -61%
    priority queues (P=5, 30k)      887     455       -49%
    connections (AMQP)             3364    3097        -8%
    MQTT connect                   3294    3017        -8%
    MQTT subscribe (100k)          6434    4622       -28%

Active RSS, 100k objects (MiB, peak-base)

    object                       before   after  reduction
    --------------------------- ------- ------- ----------
    exchanges                       535     523        -2%
    channels                       1510    1477        -2%
    queues                         1917    1716       -10%
    streams                        2486    2325        -6%
    priority queues (P=5, 30k)      908     843        -7%
    connections (AMQP)             5984    5552        -7%
    MQTT connect                   3744    3646        -3%
    MQTT subscribe (100k)          6399    5197       -19%

Idle means there's no activity the last 10 minutes, active has some sort
of activity that needs a buffer allocated.

Benchmarks ran on a Ubuntu 26.04 Linux 7.0
Crystal 1.20.2 `-Dpreview_mt -Dexecution_context` and `GC_UNMAP_THRESHOLD=1`

Specs for the stats API already existed, but added specs that should
cover this implementation. Existing specs should reduce risk of a
regression.

[1]: add_logs! no longer mutates its second argument, so slight change

### DIFF
--- a/spec/stat_log_compact_race_spec.cr
+++ b/spec/stat_log_compact_race_spec.cr
@@ -1,0 +1,91 @@
+require "./spec_helper"
+
+# Memory-safety stress for the compact path: a writer repeatedly materializes
+# then compacts an object's stats buffer (compact nulls the buffer; GC.collect +
+# GC_UNMAP_THRESHOLD=1 can unmap it) while reader fibers iterate StatLogViews to
+# JSON (the HTTP read path). If a view did not pin the buffer via its BASE
+# pointer, a compacted+unmapped buffer would be dereferenced -> SIGSEGV.
+#
+# Uses 3 rate columns so columns 1 and 2 sit at INTERIOR offsets (base + i*cap),
+# the exact case the base-pointer storage protects. Run the adversarial control
+# to prove no dependence on interior-pointer recognition:
+#   GC_UNMAP_THRESHOLD=1 GC_FREE_SPACE_DIVISOR=1 crystal spec ... (must pass)
+#   GC_UNMAP_THRESHOLD=1 GC_ALL_INTERIOR_POINTERS=0 crystal spec ... (must also pass)
+module LavinMQ
+  class RaceProbe
+    include Stats
+    rate_stats({"a", "b", "c"}, {"d"})
+
+    @d = 0_u32
+
+    def d
+      @d
+    end
+
+    def bump(n : UInt64)
+      @a_count.add(n)
+      @b_count.add(n)
+      @c_count.add(n)
+      @d &+= 1
+    end
+
+    def constant?
+      @_stats_rate_buffer.null?
+    end
+  end
+end
+
+describe "StatLogView compact/read race" do
+  it "never frees a buffer out from under a concurrent reader" do
+    cap = LavinMQ::Config.instance.stats_log_size
+    probe = LavinMQ::RaceProbe.new
+    stop = Atomic(Int32).new(0)
+    errors = Atomic(Int32).new(0)
+    done = Channel(Nil).new
+    readers = 8
+
+    readers.times do
+      spawn do
+        until stop.get(:relaxed) == 1
+          begin
+            io = IO::Memory.new
+            jb = JSON::Builder.new(io)
+            jb.document do
+              jb.array do
+                probe.a_log.to_json(jb) # column 0 (base)
+                probe.b_log.to_json(jb) # column 1 (interior)
+                probe.c_log.to_json(jb) # column 2 (interior)
+                probe.d_log.to_json(jb) # count column
+              end
+            end
+            probe.b_log.to_a
+            probe.c_log.each { |v| v }
+          rescue
+            errors.add(1)
+          end
+          Fiber.yield
+        end
+        done.send(nil)
+      end
+    end
+
+    materializes = 0
+    compacts = 0
+    300.times do |c|
+      probe.bump(c.odd? ? 7_u64 : 131_u64) # changing delta -> rate diverges -> materialize
+      probe.update_rates
+      materializes += 1 unless probe.constant?
+      (cap + 1).times { probe.update_rates } # constant (no bump) -> rate holds -> compact
+      compacts += 1 if probe.constant?
+      GC.collect # with GC_UNMAP_THRESHOLD=1 may unmap a just-freed buffer
+      Fiber.yield
+    end
+
+    stop.set(1, :relaxed)
+    readers.times { done.receive }
+
+    errors.get(:relaxed).should eq 0
+    (materializes > 100).should be_true # really exercised buffered logs
+    (compacts > 100).should be_true     # ...and really compacted them
+  end
+end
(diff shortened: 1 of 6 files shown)


jage:
My goal here was to build a low risk improvement of the metrics system. There are many paths one might take, we could skip resolution (Float32) or explore writing to disk more instead of using memory but that would require significant more change in behaviour, which I wanted to avoid right now. 

One could say this is a follow-up on the RingBuffer in #1353 which was not memory efficient. 

Will definitely research more into how we can make sure it's even more efficient, but there's also a fine line not re-inventing Prometheus here.

jage:
More benchmark data

### Test environment

| | |
|---|---|
| CPU | Intel Xeon Silver 4210R @ 2.40 GHz, 4 vCPU |
| RAM | 15 GiB |
| OS / kernel | Ubuntu 26.04 LTS / 7.0.0-15-generic |
| Crystal | 1.20.2, release, `-Dpreview_mt -Dexecution_context` |
| GC | `GC_UNMAP_THRESHOLD=1` |
| fsync | `libeatmydata` (no-op) on durable runs |

### Memory: RSS at 100k (MiB, peak - base)

| object | idle before | idle after | idle Δ | active before | active after | active Δ |
|---|--:|--:|--:|--:|--:|--:|
| exchanges | 507 | 72 | -86% | 535 | 523 | -2% |
| channels | 1416 | 200 | -86% | 1510 | 1477 | -2% |
| queues | 1878 | 393 | -79% | 1917 | 1716 | -10% |
| streams | 2425 | 944 | -61% | 2486 | 2325 | -6% |
| priority queues (P=5, 30k) | 887 | 455 | -49% | 908 | 843 | -7% |
| AMQP connections | 3364 | 3097 | -8% | 5984 | 5552 | -7% |
| MQTT connections | 3294 | 3017 | -8% | 3744 | 3646 | -3% |
| MQTT subscriptions (100k) | 6434 | 4622 | -28% | 6399 | 5197 | -19% |

### Object create throughput (ops/s)

| operation | before | after | Δ |
|---|--:|--:|--:|
| channel-churn | 9.05k | 10.88k | +20.2% |
| connection-churn | 2.97k | 3.17k | +6.7% |
| queue-churn (durable) | 77.4 | 81.2 | +4.9% |
| queue-churn (transient) | 261.6 | 262.1 | +0.2% |
| consumer-churn | 24.34 | 24.35 | parity |
| MQTT session onboard | ~243/s | ~240/s | parity |

### Throughput and latency

I would say it's "the same", but some measurement errors.

| metric | before | after |
|---|--:|--:|
| publish | 648k msg/s | 640k msg/s |
| consume | 647k msg/s | 638k msg/s |
| latency median | 19.80 ms | 19.49 ms |
| latency p99 | 26.998 ms | 26.651 ms |

### Stats sweep CPU: update_rates per 5s, idle (ms)

| operation | before | after |
|---|--:|--:|
| queues | 74 | 16 |
| channels | 102 | 23 |
| exchanges | 21 | 4 |
| streams | 77 | 18 |
| rate arithmetic (idle-counter skip) | 7.44 | 2.90 |

jage:
Next thing to explore here is if we can split up the constant base per stat, e.g. a queue has a lot of rate stats but it's unlikely all are always used (publish, redeliver, ack etc).

Also running a bit more GC benchmarks to understand how this could improve that.

carlhoerberg:
Good stuff. Minor issues found by Codex: 

```
• The new pointer-backed stats logs introduce races during concurrent stats reads and no longer
  honor reloaded stats_log_size for existing objects. These can make management API statistics
  incorrect in supported scenarios.

  Full review comments:

  - [P2] Publish buffers after backfilling them — src/lavinmq/stats.cr:106-108
    When an HTTP stats request reads logs while update_rates materializes a constant run, this
    assignment exposes _stats_rate_buffer before the buffer is back-filled (and before
    _stats_count_buffer is allocated/filled). The generated *_log getters will then switch to
    buffered mode and can serialize uninitialized GC.malloc_atomic contents or a null count
    buffer instead of the previous constant history; keep the new pointers in locals and assign
    them to the instance vars only after all columns are initialized.

  - [P2] Reuse the checked buffer pointer — src/lavinmq/stats.cr:28-29
    If a stats read races with compaction, this getter checks _stats_rate_buffer and then reads
    the instance variable again to build the view. When update_rates nulls the buffer between
    those reads, the returned view has a null base with 0_f64 as its constant and the API
    reports a zero-filled history instead of the compacted value; capture the buffer in a local
    and use that same pointer for the branch and view (the UInt32 getter below needs the same
    treatment).

  - [P2] Refresh log capacity after reloads — src/lavinmq/stats.cr:7-7
    stats_log_size is reloaded on SIGHUP and the old implementation read it on every
    update_rates, but this caches the capacity for each queue/vhost/connection at construction
    time. After a config reload that changes the size, existing objects keep emitting histories
    with the old length (and overview/nodes aggregation will follow those stale lengths) until
    they are recreated; re-read/reallocate when the configured size changes.

Issue #2003, opened 2026-06-01
sni is not enabled when reloading config unless it was configured on start


hananasen:
When LavinMQ boots with no [sni:*] sections in lavinmq.ini and a [sni:<host>] section is added at runtime, a systemctl reload lavinmq (SIGHUP) re-parses the config file and populates @config.sni_manager, but TLS handshakes to still receive the default cert, not the per-SNI one.

I don't know if we should:
1. Always install the callbacks whenever TLS is configured, even if sni_manager is empty.    Then the callback can just return nil until SNI hosts appear later.  https://github.com/cloudamqp/lavinmq/blob/bb89d1adbc4c1baf614962e1ba61bc256e7cb044/src/lavinmq/launcher.cr#L349
or,
2.  Call setup_sni_callbacks again during reload, after @config.reload / TLS reload.

PR #2033, opened 2026-06-09
Enable sni hosts added on reload

hananasen:

`setup_sni_callbacks` has an early return if the `sni_manager` is empty. This is the case when you bootstrap a lavinmq server with no sni-sections in the lavinmq.ini, which is fine. However, if a user wishes to add an sni-section, they have to restart lavinmq for that change to take effect.

On reload, `reload_tls_context` will reload the sni managers hosts and exist in memory, but since the callback is not installed for the listeners, they will still serve the default certificate.

I created a spec that installs the callback first, adds an SNI host afterwards and verifies a TLS client requesting that hostname is served the correct per-host certificate.

Manual steps is to create a certificate and key, boot lavinmq with no sni host section, send SIGHUP to the lavinmq process after adding an sni host section, for example this config

```
[main]
log_level = debug
data_dir = ./tmp/data
tls_cert = /tmp/default.pem
tls_key  = /tmp/default.key

[amqp]
tls_port = 5671

[mgmt]
tls_port = 15671

[sni:broker-a.local]
tls_cert = /tmp/a.pem
tls_key  = /tmp/a.key
```

and this should be the result of the openssl command:
`
echo | openssl s_client -connect localhost:15671 -servername broker-a.local 2>/dev/null | openssl x509 -noout -subject
subject=CN=broker-a.local
`

before this change it shows the default certificate.

### DIFF
--- a/src/lavinmq/launcher.cr
+++ b/src/lavinmq/launcher.cr
@@ -306,6 +306,7 @@ module LavinMQ
         configure_tls_context(ctx)
       end
       @config.sni_manager.reload
+      setup_sni_callbacks
     end
 
     private def configure_tls_context(ctx : OpenSSL::SSL::Context::Server)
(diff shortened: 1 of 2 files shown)


hananasen:
This is a suggestion to solve https://github.com/cloudamqp/lavinmq/issues/2003. 

Let me know if you want a video demo to demonstrate the ''problem''. its not really a bug that it works this way I think its more that its convenient not to have to restart lavinmq.

src/lavinmq/launcher.cr
```diff
@@ -348,47 +349,32 @@ module LavinMQ
...
+      end
+    end
+
+    def self.install_sni_callback(ctx : OpenSSL::SSL::Context::Server,
+                                  sni_manager : SNIManager,
+                                  protocol : Symbol)
+      ctx.on_server_name do |hostname|
+        if sni_host = sni_manager.get_host(hostname)
+          Log.debug { "SNI (#{protocol}): Using certificate for hostname '#{hostname}'" }
+          case protocol
```
spuun, line 378: One advantage to use an enum over symbol is that you can make an [exhaustive case statement](https://crystal-lang.org/reference/1.20/syntax_and_semantics/case.html#enum-values) to make sure all cases are covered (compile-time check). Besides that you also get a compile-time check that value passed to the method is valid (e.g you can't mispell `:amqp` as `:ampq`).

hananasen: nice catch, didn't think about enums. However, I see you have protocol enums in two other classes already: [server](https://github.com/cloudamqp/lavinmq/blob/1cd34970726ebeeaa3a0e5fdb99f0b5dff4a5239/src/lavinmq/server.cr#L29) and [WebsocketHandler](https://github.com/cloudamqp/lavinmq/blob/1cd34970726ebeeaa3a0e5fdb99f0b5dff4a5239/src/lavinmq/http/handler/websocket.cr#L7), should I use one of those or create a new enum?

spuun: The one in `Server` represents the different "message protocols" we have listeners for while the one in `WebsocketHandler` represents the different "message protocols" we support via websockets, so they MAY have different members. Also, you want "HTTP".

I'd say you could add another one 😅

hananasen: implemented the enum and changed from using symbols to enum. Tested it, I like using enums and I think it makes sense to add a new one, think it can live in the launcher class but no strong opinions.

carlhoerberg:
Was this fixed in another PR recently? @spuun @viktorerlingsson ?

hananasen:
As I understand it, 7ab1785  now warn-logs on reload that a restart is required to enable SNI when no SNI host was configured before.. Callbacks are still only set up at boot, so enabling SNI from scratch needs a restart.

Thats ok IMO. Worth noting the restart is only needed for the first SNI host: once a callback is installed at boot, later host changes are picked up live. So on CloudAMQP we can avoid restarting per-host by making sure an SNI host is registered at startup? or read the configuration and restart or reload depending on if its enabled.

That said, I think the cheapest win is to always install the callback at boot regardless of empty?, it reads the config dynamically, so the first-host case would then work live with no restart and no mutation of a context that's actively serving handshakes. But I might be missing a reason for the guard.

Feel free to close this if you don't agree or have another approach in mind. 😄

hananasen:
Closing in favor of https://github.com/cloudamqp/lavinmq/pull/2129

PR #2035, opened 2026-06-10
Rework connection lifecycle registration

carlhoerberg:
## Summary
- move AMQP connection registration/unregistration out of client read-loop cleanup and into the AMQP server lifecycle
- keep MQTT client lifecycle owned by the MQTT broker
- make AMQP client own channel registration/close stats
- make AMQP channel own consumer registration and unregistering

## Testing
- make test TAGS='~slow'\n- make lint

### DIFF
--- a/src/lavinmq/amqp/channel.cr
+++ b/src/lavinmq/amqp/channel.cr
@@ -20,22 +20,23 @@ module LavinMQ
       getter id, name
       property? running = true
       getter? flow = true
-      @consumers = Array(Consumer).new
+      @consumers = Array(AMQP::Consumer).new
 
-      def consumers : Array(Consumer)
-        @consumers.dup
+      def consumers : Array(LavinMQ::Client::Channel::Consumer)
+        @consumers.map &.as(LavinMQ::Client::Channel::Consumer)
       end
 
       def consumers_size : Int32
         @consumers.size
       end
 
-      def find_consumer(& : Consumer -> Bool) : Consumer?
+      def find_consumer(& : AMQP::Consumer -> Bool) : AMQP::Consumer?
         @consumers.find { |c| yield c }
       end
 
-      def delete_consumer(consumer : Consumer) : Consumer?
-        @consumers.delete(consumer)
+      def cancel_consumer(consumer : AMQP::Consumer) : Nil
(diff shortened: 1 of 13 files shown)


src/lavinmq/amqp/channel.cr
```diff
@@ -20,22 +20,23 @@ module LavinMQ
       getter id, name
       property? running = true
       getter? flow = true
-      @consumers = Array(Consumer).new
+      @consumers = Array(AMQP::Consumer).new
 
-      def consumers : Array(Consumer)
-        @consumers.dup
+      def consumers : Array(LavinMQ::Client::Channel::Consumer)
```
spuun: Do we really need this to be typed as `LavinMQ::Client::Channel::Consumer` instead of `AMQP::Consumer`? I think we want to get rid of `LavinMQ::Client::Channel::Consumer`, since consumer is a amqp thing and we're moving towards splitting protocols up more.

src/lavinmq/amqp/channel.cr
```diff
@@ -453,6 +453,11 @@ module LavinMQ
         Fiber.yield # Notify :add_consumer observers
       end
 
+      private def add_consumer(consumer : AMQP::Consumer) : Nil
+        @consumers.push(consumer)
+        consumer.queue.add_consumer(consumer)
```
spuun: This line... Should the consumer add itself to the queue instead? This looks weird.

carlhoerberg: entities registering themselves is exactly what i wanted to avoid, but this "helper" just made things uglier, agree, reverting that.

src/lavinmq/amqp/channel.cr
carlhoerberg: it's because the consumers http controller requires the generic variant https://github.com/cloudamqp/lavinmq/blob/80cd197a62b8a8365fdba637b3561874b24d586d/src/lavinmq/http/controller/consumers.cr#L56-L58

src/lavinmq/amqp/channel.cr
spuun: > entities registering themselves is exactly what i wanted to avoid,

I agree on this too.

src/lavinmq/amqp/channel.cr
carlhoerberg: fixing

PR #2036, opened 2026-06-10
Refuse to steal a live control socket

carlhoerberg:
Stacked on #2029.

Both the leader and the follower 503-server previously did an unconditional `File.delete?` before binding the lavinmqctl control socket, so a second node started on the same machine silently hijacked the socket from the running one.

Now the path is probed before binding:

- A stale socket inode (nobody listening, `ECONNREFUSED`) is deleted and the bind proceeds.
- A live socket raises `ControlSocketInUseError`: the **leader** fails hard at startup, while a **follower** logs a warning and skips its 503 server, assuming the listener is the leader (or another follower) on the same machine — so multiple nodes per machine still work.
- A path that exists but isn't a socket, or can't be verified (e.g. `EACCES`), fails closed without deleting.

The follower's 503 server is now also stored on `Clustering::Client` and closed in `#close`, so a promoted node releases the socket before binding it as leader instead of relying on the old delete-and-steal behavior.

Specs cover the four probe outcomes plus both follower behaviors (skip when in use, bind when alone).


extras/lavinmq.ini
```diff
@@ -2,6 +2,9 @@
 data_dir = /var/lib/lavinmq
 log_level = info
 ;pidfile = /var/run/lavinmq.pid
+; Socket that lavinmqctl connects to. Set a unique path per instance when
+; running multiple LavinMQ servers on the same host.
+;control_unix_path = /tmp/lavinmqctl.sock
```
carlhoerberg:
```suggestion
```

PR #2037, opened 2026-06-10
Fix lost exchange federation bindings during link startup

dentarg:
I noticed this in CI: https://github.com/cloudamqp/lavinmq/actions/runs/27281857240/job/80578254772#step:5:4300

--

`ExchangeLink#setup` registered the link as a bind/unbind observer on the downstream exchange, but `@consumer_ex` was only assigned after `setup` returned, one network round-trip later. Bind events are delivered synchronously and `with_consumer_ex` silently drops them while `@consumer_ex` is nil, so a binding created in that window was never propagated to the upstream exchange and stayed missing until the link reconnected.

This made the spec "should continue after upstream restart" flaky on slow runners (macOS CI): `downstream_q.bind` raced into the window, the `#` binding never reached the upstream exchange and `msg1` was unroutable, timing out the spec.

Fix by assigning `@consumer_ex` before registering the observer. Bindings made before registration are still covered by the snapshot replay, which runs after registration (exchanges store bindings before notifying observers), so every binding now takes at least one of the two paths.

The regression spec stretches the snapshot replay with pre-existing bindings, waits for the link to register its observer and then binds mid-startup; every binding must be reflected upstream. It fails 5/5 without this fix.

### DIFF
--- a/spec/upstream_spec.cr
+++ b/spec/upstream_spec.cr
@@ -692,6 +692,37 @@ describe LavinMQ::Federation::Upstream do
       end
     end
 
+    it "should reflect bindings made while link is starting" do
+      with_amqp_server do |s|
+        upstream, upstream_vhost, downstream_vhost =
+          UpstreamSpecHelpers.setup_federation(s, "ef test bindings during start", "upstream_ex")
+        with_channel(s, vhost: "downstream") do |downstream_ch|
+          downstream_ch.exchange("downstream_ex", "topic")
+          downstream_q = downstream_ch.queue("downstream_q")
+          # Pre-existing bindings stretch the binding replay the link does
+          # during startup, so that the binds below land mid-replay.
+          before = 200
+          before.times { |i| downstream_q.bind("downstream_ex", "before.link.#{i}") }
+
+          UpstreamSpecHelpers.start_link(upstream)
+          link = wait_for { upstream.links.first? }
+          downstream_ex = downstream_vhost.exchange("downstream_ex").as(LavinMQ::AMQP::Exchange)
+          # The link starts observing the downstream exchange while it is
+          # still replaying the bindings above to the upstream exchange.
+          wait_for { downstream_ex.@__lavinmq_exchangeevent_observers.includes?(link) }
+          # Binds observed during startup must also be reflected upstream.
+          # (Regression: they were dropped if observed before the link had
+          # an upstream channel.)
+          during = 10
+          during.times { |i| downstream_q.bind("downstream_ex", "during.link.#{i}") }
+
+          upstream_ex = wait_for { upstream_vhost.exchange?("upstream_ex") }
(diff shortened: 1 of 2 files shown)


dentarg:
Some words from a 🤖 when I asked `Is there any recent changes that made this problem more likely to happen?`

> Yes — and the answer is interesting. The bug itself is ancient, but a recent spec-harness change shifted the timing to expose it. Post-rebase, all 30 upstream specs still pass, including the regression spec.
> 
> ## The bug is 5+ years old
> 
> The dropped-event path (`with_consumer_ex` silently discarding bind events while `@consumer_ex` is nil) and the observer-registered-before-`@consumer_ex` ordering both date to `177c4aee` "Improve federation reconnect reliability" (Feb 2021) — the same commit that added the now-flaky "should continue after upstream restart" spec. The ordering survived the Dec 2025 federation refactor (#1492) unchanged.
> 
> ## What changed recently: #1982 "Timeout all specs within 15s" (f9dddcf8, May 29 2026)
> 
> This commit changed `wait_for`/`should_eventually` from a busy `Fiber.yield` polling loop to `sleep 1.millisecond`, and that plausibly flipped the typical interleaving:
> 
> - **Before:** the spec fiber re-checked `upstream.links.first?` on every scheduler pass. The link object lands in `@ex_links` before its run fiber has even opened the TCP connection, so the spec resumed almost instantly, and its `queue_declare` + `bind` (2 round-trips) completed while the link was still in its AMQP handshake (several round-trips) — comfortably **before** `register_observer`, i.e. on the safe snapshot path. The busy-spin also actively starved the server's IO fibers (the reason #1982 removed it), slowing link startup relative to the spec and pushing the bind even further ahead of the window.
> - **After:** the spec sleeps ≥1ms per poll. On loopback, 1ms covers many round-trips — the link can finish its handshake and most of `setup` during that first sleep, so the bind now lands right around the `register_observer` → `@consumer_ex` window. Add macOS-runner scheduling jitter and it occasionally lands exactly inside it.
> 
> #1982 also raised `be_receiving`'s default timeout from 5s to 15s and added the 15s per-example timeout — that changed how the failure *reports* (the duplicated failure entries we saw), not whether it happens.
> 
> ## Things that are *not* recent factors
> 
> - `-Dpreview_mt -Dexecution_context` on `make test` — added April 2025, and the spec runner's default execution context is single-threaded anyway.
> - Your policy-serialization PR (#2030), whose CI run this was: the link *is* created inside `apply_policy_argument("federation-upstream", ...)`, which now runs under the new `@policy_lock` — but that lock is uncontended in this spec, so it adds only negligible delay. Any perturbation can flip a knife-edge race, but there's no mechanism in #2030 that systematically widens this window. I'd call it bad luck that the latent flake fired on that run.
> 
> So: #1982 made the five-year-old race likely enough to observe on a loaded macOS runner; our fix removes the window entirely.

spuun (approved):

Issue #1272, opened 2025-09-14
MQTT Permissions issue


petyots:
**Describe the bug**
Permissions not respected in MQTT

**Describe your setup**
Running LavinMQ in Docker
root@orbstack:/var/lib/lavinmq# lavinmq --build-info
LavinMQ 2.4.1
Crystal 1.17.0 (2025-07-16)
LLVM: 18.1.3
Default target: aarch64-unknown-linux-gnu
Build flags: --release --debug

**How to reproduce**

- Open http://localhost:15672
- Create new user
- Edit the user and add `Configure regexp: ^$`, `Write regexp: ^$`, `Read regexp: ^$`(This should disallow subscribing or publishing to any topic)
- Subscribe to test topic and publish a message
- Message received by other clients successfully

<img width="754" height="374" alt="Image" src="https://github.com/user-attachments/assets/a854bef6-1896-4709-8cfc-cf56c9910468" />

**Expected behavior**
This user should not be able to publish or subscribe to any topics at all, however it can successfully subscribe and/or publish to any topic

<img width="866" height="783" alt="Image" src="https://github.com/user-attachments/assets/8b390328-6db1-438e-a308-b77f756ab08f" />

kickster97:
Hello, thanks for your report. Will look into this asap!

spuun:
The permissions in LavinMQ isn't "topic permissions", but "resource permissions". What you can configure is if a user can read from or write to an exchange, not specific topics (or routing keys), however not for MQTT.

We'll probably look into adding topic permissions.

petyots:
> The permissions in LavinMQ isn't "topic permissions", but "resource permissions". What you can configure is if a user can read from or write to an exchange, not specific topics (or routing keys), however not for MQTT.
> 
> We'll probably look into adding topic permissions.

Can't the "resource permissions" be reused for MQTT too for the current time being? IMO they will be able to cover most of the use cases.

spuun:
> Can't the "resource permissions" be reused for MQTT too for the current time being? IMO they will be able to cover most of the use cases.

Ah, you just want to block publish or consuming for clients? Then yes!

petyots:
> > Can't the "resource permissions" be reused for MQTT too for the current time being? IMO they will be able to cover most of the use cases.
> 
> Ah, you just want to block publish or consuming for clients? Then yes!

Pretty much yes, client A should not be able to listen/publish on client B topics. I guess it will be tough to support mqtt wildcard?

[...]

PR #2038, opened 2026-06-10
Add MQTT client_id validation tied to the username

kickster97:
Adds a new config option, client_id_validation in the [mqtt] section, with two modes: none (default, current behavior) and username (client_id must equal the authenticated username). Connections that violate the mode get CONNACK return code 2 (identifier rejected) and are closed. Empty client_ids are auto-assigned a conforming id. Validation uses the username after the optional vhost: prefix has been stripped.

This is groundwork for topic permissions (#1272): rules like devices/{client_id}/# only give real isolation when the client_id is as trustworthy as the login, since plain MQTT clients can claim any client_id.

Worth noting: username mode implies one connection per user, since a second connection with the same client_id takes over the session per MQTT 3.1.4-2.

A username_prefix mode for sharing one login across several devices was considered but dropped from this PR. A plain prefix match ties the namespace separator to a character that is legal in usernames, so a user whose name is a prefix of another (fleet1 vs fleet1-eu) can claim ids in that user's namespace. It needs a safer namespace design before shipping.

Connect rejections (auth failures and invalid client_ids) now raise MQTT::Error subclasses handled by the existing Connect rescue, so #start is a single happy path with all connack-on-error logic in one place.

### DIFF
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -93,6 +93,7 @@ Internally, MQTT is implemented on top of LavinMQ's AMQP infrastructure:
 | `max_packet_size` | `[mqtt]` | `268435455` | Max MQTT packet size in bytes |
 | `default_vhost` | `[mqtt]` | `/` | Default vhost for MQTT connections |
 | `permission_check_enabled` | `[mqtt]` | `false` | Enable ACL checks on MQTT publish/subscribe |
+| `client_id_validation` | `[mqtt]` | `none` | Validate client_id against the username: `none` or `username` |
 
 ## Permissions
 
@@ -108,6 +109,16 @@ When disabled, any authenticated MQTT client can publish and subscribe to any to
 MQTT clients authenticate using the CONNECT packet's username and password fields. These are validated against the same authentication chain as AMQP (local users, OAuth2). For OAuth2, the password field carries the JWT token.
 
 The username field can include a vhost using the format `vhost:username`. If no colon is present, `default_vhost` is used.
+
+### Client ID Validation
+
+By default any client_id is accepted. Since the client_id is chosen freely by the client, it cannot be trusted for identity purposes on its own. The `client_id_validation` setting ties it to the authenticated username:
+
+- `username`: the client_id must be equal to the username
+
+A CONNECT with a non-conforming client_id is rejected with return code 2 (identifier rejected) and the connection is closed. An empty client_id is automatically assigned a conforming one. When the username includes a vhost (`vhost:username`), the client_id is validated against the username part only.
+
+Note that connecting with a client_id already in use takes over that session, so `username` mode limits each user to one connection at a time.
 
 ## Limitations
 
--- a/src/lavinmq/mqtt/client_id_validation.cr
+++ b/src/lavinmq/mqtt/client_id_validation.cr
@@ -1,0 +1,10 @@
+module LavinMQ
+  module MQTT
+    # Controls how the MQTT client_id is validated against the
+    # authenticated username at connect.
+    enum ClientIdValidation
+      None     # any client_id accepted
+      Username # client_id must equal the username
+    end
+  end
+end
--- a/src/lavinmq/mqtt/connection_factory.cr
+++ b/src/lavinmq/mqtt/connection_factory.cr
@@ -23,16 +23,12 @@ module LavinMQ
           io = Protocol::IO.new(socket, @config.mqtt_max_packet_size)
           if packet = io.read_packet.as?(Protocol::Connect)
             logger.trace { "recv #{packet.inspect}" }
-            if user_and_broker = authenticate(io, packet)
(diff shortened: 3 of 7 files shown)


src/lavinmq/mqtt/client_id_validation.cr
```diff
@@ -0,0 +1,11 @@
+module LavinMQ
+  module MQTT
+    # Controls how the MQTT client_id is validated against the
+    # authenticated username at connect.
+    enum ClientIdValidation
+      None           # any client_id accepted
+      Username       # client_id must equal the username
+      UsernamePrefix # client_id must start with the username
```
spuun:
```suggestion
      UsernamePrefix # client_id must start with the username-
```

docs/mqtt.md
```diff
@@ -109,6 +110,17 @@ MQTT clients authenticate using the CONNECT packet's username and password field
 
 The username field can include a vhost using the format `vhost:username`. If no colon is present, `default_vhost` is used.
 
+### Client ID Validation
+
+By default any client_id is accepted. Since the client_id is chosen freely by the client, it cannot be trusted for identity purposes on its own. The `client_id_validation` setting ties it to the authenticated username:
+
+- `username`: the client_id must be equal to the username
+- `username_prefix`: the client_id must be the username, or the username followed by `-` and a device name, so one set of credentials can be shared by several devices (user `fleet1` can connect as `fleet1-sensor-a` and `fleet1-sensor-b`). Avoid usernames containing `-` with this mode, since a username like `fleet1-eu` overlaps with the id namespace of a user named `fleet1`.
```
spuun: > Avoid usernames containing `-` with this mode, since a username like `fleet1-eu` overlaps with the id namespace of a user named `fleet1`.

This is an interesting observation... I wonder if we can rethink something to avoid this. Use another separator that isn't allowed to be part of usernames?

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -25,7 +25,12 @@ module LavinMQ
             logger.trace { "recv #{packet.inspect}" }
             if user_and_broker = authenticate(io, packet)
               user, broker = user_and_broker
-              packet = assign_client_id(packet) if packet.client_id.empty?
+              packet = assign_client_id(packet, user.name) if packet.client_id.empty?
+              unless valid_client_id?(packet.client_id, user.name)
+                logger.warn { "Client id \"#{packet.client_id}\" rejected for user \"#{user.name}\" (client_id_validation=#{@config.mqtt_client_id_validation})" }
+                connack io, false, Connack::ReturnCode::IdentifierRejected
+                return
+              end
```
spuun: To reduce the conditional complexity of `#start`, `valid_client_id?` could raise `MQTT::Error::Connect` (or maybe custom exceptions defined in LavinMQ, but inheriting `MQTT::Error::Connect`) with the return code set. Logging and connack would then be handled in the rescue handler for `MQTT::Error::Connect`.
```suggestion
              validate_client_id!(packet.client_id, user.name)
```
With custom exceptions the error message may be more clear, saying that "client id must be prefixed with `<username>-`" etc.

Dunno, but I think it may be cleaner?

spuun (commented):
Maybe we should make sure to think an extra round on how to handle the UsernamePrefix case?

docs/mqtt.md
kickster97: I think it will be hard to find a separator that is safe.. I'm not 100% sure about this solution.. i think i will drop username prefix for now.. its a nice feature but i need to work it a bit more to make it a good code solution and also natural to use for the user

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -65,23 +63,36 @@ module LavinMQ
...
       end
+
+      private def validate_client_id!(client_id : String, username : String) : Nil
+        valid = case @config.mqtt_client_id_validation
+                in .none?     then true
+                in .username? then client_id == username
+                end
+        return if valid
+        raise Protocol::Error::IdentifierRejected.new(
+          "client_id \"#{client_id}\" rejected for user \"#{username}\" (client_id_validation=#{@config.mqtt_client_id_validation})")
```
spuun: Maybe the `(client_id_validation=#{@config.mqtt_client_id_validation})` could be unclear to some users? I think it's better if it says explicit that client id must be same as username.

carlhoerberg:
```suggestion
        case @config.mqtt_client_id_validation
        in .none?
          return true
        in .username?
          return true client_id == username
          raise Protocol::Error::IdentifierRejected.new(
            %(client_id "#{client_id}" rejected for user "#{username}" \
              (client_id_validation=username)))
        end
        
```

carlhoerberg (approved):

spuun (approved):

PR #2039, opened 2026-06-10
Fix flaky delayed-exchange deduplication spec by pausing RoughTime

dentarg:
Noticed this in CI: https://github.com/cloudamqp/lavinmq/actions/runs/27284026515/job/80586115289?pr=2037#step:6:4261

--

`should not reset x-cache-ttl on expire when using delayed exchange` kept failing in CI: the first `wait_for { !delay_q.empty? }` timed out even though the publish was confirmed.

The spec relied on *observing* the message resting in the internal delayed queue for its `x-delay` of 190ms of real time. But `RoughTime.unix_ms` — used both to stamp `msg.timestamp` on publish and as "now" in `DelayedExchangeQueue#has_expired?` — has 100ms granularity and is refreshed every ~100ms, so a stamp can lag real time by up to ~200ms. When an updater tick lands between the publish and the expire fiber's check (which `delay()` pings synchronously on every push), the clock can jump past the whole 190ms delay and the message is expired out of the delayed queue before the spec's first poll — after which `delay_q` stays empty forever. The same miss happens without any clock jump if the spec fiber is descheduled for >190ms between `publish_confirm` and the first poll on a loaded runner. #2010 replaced the edge-triggered `select`s here with polling, which fixed the missed-edge hang but cannot observe a transient state that has already passed.

Make the spec deterministic with `RoughTime.paused`/`travel` (added in #1160 for exactly this) instead of racing the wall clock:

- While time is paused a delayed message cannot expire, so the "message is delayed" observations cannot be missed.
- Travel in multiples of 100ms (the `unix_ms` granularity) to first expire the delayed message *inside* the cache ttl, then move *past* the cache ttl, replacing the `x-delay = cache_ttl - 10` + 20ms-sleep timing dance.
- Also assert that the expired message actually reaches the bound queue, which catches the expire-path-deduplicates-itself variant of the issue #1153 regression as well.

Also fix `RoughTime.travel` in spec support: it *added* the truncated unix timestamp to `@@paused_unix_ms` instead of assigning it, so any spec that traveled and then read `unix_ms` saw a garbage value (~2x epoch). Latent so far because existing `travel` users only exercised `instant`-based paths or accidentally passed because "everything looks expired".

Verified: the spec fails when the issue #1153 fix is reverted (`expire_msg` routed via `publish` instead of `route_msg`), and passes 30/30 repeated runs including 10 under full CPU contention. Full suite green.

### DIFF
--- a/spec/deduplication_spec.cr
+++ b/spec/deduplication_spec.cr
@@ -162,7 +162,12 @@ describe LavinMQ::Deduplication::Deduper do
     end
 
     it "should not reset x-cache-ttl on expire when using delayed exchange" do
-      cache_ttl = 200
+      # RoughTime.unix_ms has 100ms granularity and can lag real time by up to
+      # ~200ms, so a message with a short x-delay can be expired out of the
+      # delayed queue (near) instantly in real time. Pause RoughTime and travel
+      # explicitly (in multiples of 100ms) instead of racing the wall clock.
+      cache_ttl = 500
+      delay = 100
       x_args = AMQP::Client::Arguments.new({
         "x-delayed-exchange"      => true,
         "x-delayed-type"          => "topic",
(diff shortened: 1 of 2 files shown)


dentarg:
From the 🤖 

> ## Root cause
> 
> The spec published a message with `x-delay = 190ms` and then polled `wait_for { !delay_q.empty? }` to catch it resting in the delayed queue. That's a race it can lose permanently: both the message timestamp and the expiry check use `RoughTime.unix_ms`, which is truncated to 100 ms resolution and refreshed only every ~100 ms (`src/lavinmq/rough_time.cr`), so the stamp can lag real time by up to ~200 ms. When a RoughTime tick lands between the publish and the expire fiber's check — which runs immediately, because `DelayedExchangeQueue#delay` pings `@message_ttl_change` on every push — "now" can jump past the entire 190 ms delay and the message is expired out of the delayed queue *before the spec's first poll*. The same miss happens with no clock jump at all if the spec fiber is descheduled >190 ms before its first poll on a loaded CI runner. Either way the message is already in the bound queue, `delay_q` stays empty forever, and the 1 ms-interval polling can never see a state that has passed — hence the 15 s spec timeout plus the "Execution expired" at line 192 (the abandoned example fiber keeps running after the timeout hook fires, which is why you get two reported failures for one example).
> 
> ## The fix
> 
> - Rewrote the spec to use `RoughTime.paused` / `travel` (the helper added in #1160 for exactly this). With time frozen, a delayed message *cannot* expire, so the "it's delayed" observations are race-free; the spec then travels in multiples of 100 ms (the clock granularity) to expire the delayed message inside the cache TTL, and again past the TTL. I also added an assertion that the expired message actually reaches the bound queue, so the spec now catches both variants of an issue-#1153 regression (cache-TTL refresh *and* expire-path self-deduplication).
> - Fixed a real bug this exposed in `spec/support/rough_time.cr`: `travel` did `@@paused_unix_ms += ...` instead of `=`, producing a ~2× epoch garbage value after any travel. It was latent because existing users only exercised `instant`-based paths.
> 
> Verified: the new spec fails when I temporarily reverted the #1153 fix (`expire_msg` routed via `publish` instead of `route_msg`), passes 30/30 repeated runs including 10 under full CPU contention, and now runs in ~100 ms instead of depending on ~400 ms of real-time sleeps. Note for local runs: the full suite needs an `etcd` binary, which I installed from GitHub releases in this container.
> 
> ## Did recent changes make this more likely?
> 
> Yes, though the root race is as old as the spec (added Aug 2025 in `24f74bd7`):
> 
> - **`cf484e41` (Jun 3, #2010)** is the most direct contributor — it rewrote this exact spec from edge-triggered `select` on `empty.when_false` to level-triggered `wait_for` polling. That fixed the missed-edge *hang* it was chasing, but polling every 1 ms (with a 1 ms sleep before the first check) is strictly blind to a transient state that's already gone, whereas the old subscription could at least catch a brief episode if armed in time. Same root cause, new guaranteed-miss window — and the failed runs cited in that commit show this same spec flaking since May.
> - **`f9dddcf8` (May 29, #1982)** added the 15 s per-example timeout and switched polling from `Fiber.yield` busy-spin to 1 ms sleeps — it made the failure *visible* as this two-failure signature and reduced poll frequency ~1000×.
> - **`aa34e633` (Jan 2026, #1600)** made `delay()` wake the expire fiber synchronously on every push, which is what enables the instant-expiry path when the coarse clock jumps.
> - The newest commits on main (policy serialization `7f4b8399` etc.) are unrelated — this spec uses no policies.
> 
> One product-side observation, out of scope but worth a look sometime: `Exchange#publish` ignores `delay()`'s return value (`exchange.cr:232`), so a publish to a closed delayed queue is silently dropped yet confirmed as `Routed`. And more broadly, `x-delay` values under ~200 ms are effectively best-effort given RoughTime's granularity — messages can be delivered early.

snichme (approved):

PR #2040, opened 2026-06-10
Don't recreate removed bindings during federation link startup

dentarg:
At startup an exchange federation link registers itself as a bind/unbind observer on the downstream exchange, then replays a snapshot of its current bindings to the upstream exchange, one round-trip per binding. An unbind observed mid-replay was sent upstream immediately, where it was a no-op because the replay had not created that binding yet — and the replay then recreated it from its stale snapshot. The result was an upstream binding with no downstream counterpart, so the upstream kept forwarding messages the downstream no longer wanted, until the link reconnected.

Fixed by queueing bind/unbind events observed while the replay is running and applying them in order once it finishes. The handoff back to direct handling happens atomically with observing an empty queue, so no event is lost; if the replay fails, the queue is discarded and the next setup resyncs from a fresh snapshot. Queueing also stops concurrent RPCs on the consumer channel during startup, which could mix up replies (`BindOk`/`UnbindOk` share one reply queue per channel).

The regression spec stretches the replay with 200 pre-existing bindings, unbinds tail-end snapshot bindings mid-replay, and requires the upstream exchange to settle on exactly the remaining bindings once the link is running. It fails 5/5 without this fix.

### DIFF
--- a/spec/upstream_spec.cr
+++ b/spec/upstream_spec.cr
@@ -723,6 +723,42 @@ describe LavinMQ::Federation::Upstream do
       end
     end
 
+    it "should not recreate bindings removed while link is starting" do
+      with_amqp_server do |s|
+        upstream, upstream_vhost, downstream_vhost =
+          UpstreamSpecHelpers.setup_federation(s, "ef test unbindings during start", "upstream_ex")
+        with_channel(s, vhost: "downstream") do |downstream_ch|
+          downstream_ch.exchange("downstream_ex", "topic")
+          downstream_q = downstream_ch.queue("downstream_q")
+          # Pre-existing bindings stretch the binding replay the link does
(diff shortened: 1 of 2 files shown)


dentarg:
This wasn't a CI failure, it was something spotted by codex gpt-5.5 when reviewing https://github.com/cloudamqp/lavinmq/pull/2037

> Findings
> `/app/src/lavinmq/federation/link.cr:460`: `setup` still registers the observer before replaying bindings_details, so unbinds can race with the replay. If a binding is present in the snapshot, then removed downstream before the loop reaches it, on(Unbind) sends an idempotent upstream unbind before that upstream binding exists, and the replay later recreates the stale binding. The new spec only covers binds during startup, not removals, so this startup sync can still leave upstream exchange bindings that no longer exist downstream.

PR #2042, opened 2026-06-11
fix: guard empty destination in Move messages UI

### LINKED ISSUE #2041: "Move messages" in management interface queue view deletes messages if no destination queue is selected
When you have messages in a queue and go to the "move messages" section but do not select a destination queue, it seems like the messages are moved to a non-existing queue, so they are effectively deleted.

## Expected behaviour
You have to select a queue for the move to be possible, since this is a destructive action.

## Steps to reproduce
0. Use lavinmq v2.6.10
1. Open management interface
2. Select "queues" in the left side bar
3. Select a queue with at least one message
4. Scroll down to the "Move messages" box
5. Do *NOT* choose a destination queue in the drop-down
6. Press the "Move messages" button
7. The messages have now been moved from queue but with seemingly no destination.

Image

viktorerlingsson:
## Summary

The Move messages form accepted an empty destination, which published to the default exchange with no routing key - silently dropping the messages. Added a guard in both the HTML (`required`) and JS (trim check + toast).

## Test plan

- Open a queue, click `Move messages`, leave the destination empty and submit -> form now blocks submission and shows an error toast instead of silently dropping messages.

## Follow-up

While investigating this, we found related issues in shovel destination routing and `validate_config!` (destination required when RabbitMQ treats it as optional, contradictory dest-queue/dest-exchange combos, default-exchange permission handling). Those are out of scope here and tracked in #2058.

Closes #2041

### DIFF
--- a/static/js/queue.js
+++ b/static/js/queue.js
@@ -288,6 +288,10 @@ moveMessagesForm.addEventListener('submit', function (evt) {
   const password = Auth.getPassword()
   const uri = HTTP.url`amqp://${username}:${password}@localhost/${vhost}`
   const dest = document.querySelector('[name=shovel-destination]').value.trim()
+  if (dest === '') {
+    DOM.toast.error('Please select a destination queue')
+    return
+  }
   const name = 'Move ' + queue + ' to ' + dest
   const url = HTTP.url`api/parameters/shovel/${vhost}/${name}`
   const body = {
--- a/views/queue.shtml
+++ b/views/queue.shtml
@@ -245,7 +245,7 @@
         <h3>Move messages</h3>
         <label>
           <span>Destination queue</span>
-          <input type="text" name="shovel-destination" list="queue-list">
+          <input type="text" name="shovel-destination" required list="queue-list">
           <datalist id="queue-list">
           </datalist>
         </label>


antondalgren:
I guess this can still cause trouble for API users? Should we block there and show the error in the HTML instead?

viktorerlingsson:
> I guess this can still cause trouble for API users? Should we block there and show the error in the HTML instead?

Move messages just uses the parameters API to create a normal shovel, and I don't think we should block shovels from doing this?

antondalgren:
> Move messages just uses the parameters API to create a normal shovel, and I don't think we should block shovels from doing this?

https://github.com/cloudamqp/lavinmq/blob/109a1c96afd67fb1764e04352f3b539601feef07/src/lavinmq/shovel/store.cr#L62

Shouldn't we fix the logic here instead? IMO it should not be possible to create a shovel into nothing? Or am I missing some usecase here where it's a valid thing to do?

viktorerlingsson:
> Shouldn't we fix the logic here instead? IMO it should not be possible to create a shovel into nothing? Or am I missing some usecase here where it's a valid thing to do?

Ah, yes you're right. Looking into that!

viktorerlingsson:
> Shouldn't we fix the logic here instead? IMO it should not be possible to create a shovel into nothing? Or am I missing some usecase here where it's a valid thing to do?

oh, it is actually valid - if neither destination queue nor exchange is set, `AMQPDestination::push` pushes the message to the original exchange and routing key. 

```
dest-queue
The queue to which messages should be published. 
Either this or dest-exchange (but not both) may be set. 
If neither is set then messages are republished with their original exchange and routing key.
```

I'm working on making sure all validations are correct

Edit: But that's not valid for move messages though, that would just republish them to the same queue. So I think we should still have this frontend checks to stop that.

spuun:
> Edit: But that's not valid for move messages though, that would just republish them to the same queue. So I think we should still have this frontend checks to stop that.

Maybe we should rework "move messages" to not use a shovel.

viktorerlingsson:
> > Edit: But that's not valid for move messages though, that would just republish them to the same queue. So I think we should still have this frontend checks to stop that.
> 
> Maybe we should rework "move messages" to not use a shovel.

But it pretty much is a shovel though, with some restrictions. Do we really need to invent a new wheel to handle this?

viktorerlingsson:
Found some things in regards to shovel routing and validation, but that's not related to this PR/issue. Removed that from this PR and created a new issue to track that https://github.com/cloudamqp/lavinmq/issues/2058

spuun (approved):

PR #2045, opened 2026-06-11
Eliminate allocations in the message publish/deliver hot path

baelter:
## What

Eliminates all avoidable heap allocations in the message publish→store→deliver hot path, measured with a new in-process allocation benchmark (`bench/alloc_bench.cr`, included). Every publish-path scenario the harness covers is now **0 bytes allocated per message**.

### Bytes allocated per operation (release build, 50k ops/scenario)

| Scenario | Before | After |
|---|---:|---:|
| publish: default exchange → queue | 0 | 0 |
| publish: direct exchange, bound rk | 0 | 0 |
| publish: direct exchange, unbound rk (unique keys) | 143 + permanent hash growth | **0** |
| publish: topic / fanout exchange | 0 | 0 |
| publish: headers exchange, matching message | 256 | **0** |
| publish: message with 4 headers (any exchange) | 224 | **0** |
| deliver (basic_get): plain message | 64 | 64 |
| deliver (basic_get): message with 4 headers | 608 | **160** |

The remaining deliver-path bytes are the routing-key String and headers Table object built by `BytesMessage.from_bytes`, which the outgoing `basic.deliver` frame genuinely needs — removing those would require Bytes-backed frame types in amq-protocol.

### Throughput (no regression)

End-to-end `extras/benchmark.sh`, 5 passes × 10s/scenario, 64 connections, branch vs main on the same host (isolated broker). Steady-state throughput is flat to slightly positive: median **+1.0%** across 32 publish/consume metrics, range -4.7% to +7.2%. Most per-scenario deltas sit within run-to-run noise (CV 1-8%); the two gainers consistent across all 5 passes are publish-confirm (**+7%**) and baseline 1-to-1 (**+2.8%**, CV ~1-3%). These scenarios aren't GC-bound, so the allocation cuts surface as lower GC pressure rather than raw msgs/s — the throughput numbers are here to show the change doesn't regress it.

## Fixes

1. **Direct exchange: publish to an unbound routing key allocated and permanently leaked a hash entry.** `each_destination` used `@bindings[routing_key]` on a Hash whose default block *inserts* an empty Set on miss, so every publish with a never-bound routing key grew the bindings hash forever — unbounded memory growth driven by publishers. Now uses `[]?`. The same insert-on-miss applied to `unbind` on the direct, topic, and headers exchanges (one leaked entry per failed unbind); also fixed.

2. **`SegmentPosition.make` iterated all message headers** via `Table#each`, which allocates a String per key and a Field per value — on every message-store push *and* shift/peek. Replaced with targeted `has_key?`/`[]?` lookups for the two keys it cares about (`x-dead-letter-exchange`, `x-delay`), which scan the table buffer without allocating. 224 → 0 bytes per call for a 4-header message.

3. **Headers exchange re-parsed binding arguments on every published message.** Binding args are now parsed once at bind time into a match spec (x-match mode + non-`x-` pairs); routing iterates the precompiled pairs instead of the args Table. Match semantics are unchanged, including the quirks: per-binding `x-match` falls back to the exchange's own `x-match` argument, bindings whose args contain only `x-`-prefixed keys match any non-empty headers under `all` but not empty headers, and empty-args bindings match messages without headers.

4. **Vendored `AMQ::Protocol::Table#has_entry?`** (`src/lavinmq/amqp/table_ext.cr`) — proposed upstream to amq-protocol.cr — which compares a table entry against an expected value without materializing a String for stored string fields, and used it for the headers-exchange pair comparison. This removed the last 32 bytes/op in headers matching. The file is marked for deletion once shard.yml requires an amq-protocol release containing the method.

5. **MQTT `SubscriptionTree#unsubscribe` left an empty inner hash behind** for every unique non-wildcard topic filter ever subscribed; entries are now dropped when the last subscriber goes.

## Benchmark harness

`bench/alloc_bench.cr` exercises the real server code in-process (no sockets, so numbers are attributable to LavinMQ itself): `vhost.publish` → exchange routing → queue → message store on the publish side, and `queue.basic_get` → frame encode on the deliver side, measuring `GC.stats.total_bytes` per operation. Run with:

```
crystal run --release -Dpreview_mt -Dexecution_context bench/alloc_bench.cr
```

## Verification

- Full spec suite (minus `spec/api`+`spec/http`, which need IPv6 loopback unavailable in the dev container): 1234 examples, 0 failures. The HTTP/API suites should be covered by CI.
- New regression specs: direct-exchange unbound-publish leak, `SegmentPosition.make` header semantics (x-delay extraction, overflow/negative → 0, dlx flag), headers-exchange x-match default/edge-case behavior, `Table#has_entry?` equivalence sweep, subscription-tree entry cleanup.
- `make lint` clean.

## Related

Upstream PR: cloudamqp/amq-protocol.cr#39 (`Table#has_entry?` + specs). Once released and required by shard.yml, `src/lavinmq/amqp/table_ext.cr` should be deleted.


carlhoerberg (approved):
I don't fully comprehend the Header exchange optimization, but others looks good

extras/alloc_bench.cr
carlhoerberg, line 1: Was this for this PR only or something we want to continue to use? In that case, should we move it to `extras/`, to not create another top directory?

src/lavinmq/amqp/exchange/headers.cr
```diff
@@ -3,39 +3,81 @@ require "./exchange"
...
+      # routing doesn't have to re-parse the arguments table (which allocates
+      # a String per key and a Field per value) on every published message.
+      private class Binding
+        private struct Pair
+          getter key : String
+          getter value : AMQP::Field
+
+          def initialize(@key, @value)
+          end
+        end
```
carlhoerberg:
```suggestion
        private record Pair, key : String, value : AMQP::Field
```

extras/alloc_bench.cr
baelter: It's meant to be kept — it's the regression check for these numbers (and for future hot-path changes). Moved to `extras/alloc_bench.cr` in 7ab6437, along with applying the `record Pair` suggestion.

PR #2046, opened 2026-06-12
Fix silent drop of delayed messages when the internal delayed queue is closed

dentarg:
`Exchange#publish` ignored the return value of `DelayedExchangeQueue#delay`, which returns `false` once the queue is closed (permanently so after a `MessageStore::Error` on the delayed store). Such publishes were silently dropped while still being counted as `publish_out` and reported as `Routed`, so the publisher got a confirm without a mandatory return.

Treat a closed delayed queue the same as a closed regular queue in `route_msg`: count the message as unroutable and return `PublishResult::None`, giving mandatory publishes a `basic.return` with `312 NO_ROUTE`.

### DIFF
--- a/src/lavinmq/amqp/exchange/exchange.cr
+++ b/src/lavinmq/amqp/exchange/exchange.cr
@@ -228,8 +228,7 @@ module LavinMQ
           d.add(msg)
         end
         if should_delay_message?(msg.properties.headers)
-          if q = @delayed_queue
-            q.delay(msg)
+          if (q = @delayed_queue) && q.delay(msg)
             @publish_out_count.add(1, :relaxed)
             return PublishResult::Routed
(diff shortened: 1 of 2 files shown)


src/lavinmq/amqp/exchange/exchange.cr
spuun, line 238: I wonder if we should raise from `DelayedExchangeQueue#delay` if the queue is closed, which would result in an internal error for the client.

Maybe it should be something like
```crystal
return PublishResult::Routed if message_is_delayed?(msg)
```
and `message_is_delayed?` is what `should_delay_message?` is today plus`q.delay(msg)`.

Stats should probably not be touched here, it will be handled when the message is actually being routed after the delay.

PR #2055, opened 2026-06-15
Use f_frsize for disk sizes on Linux

### LINKED ISSUE #2054: LavinMQ reports incorrect disk size when running with virtiofs mounts
**Describe the bug**

LavinMQ reports incorrect disk size when running with virtiofs mounts (Docker Desktop, Lima etc)

  Image

  On Linux, block counts from `statfs(2)` are in `f_frsize` units, but the disk size was computed with `f_bsize` (the optimal transfer size). On local filesystems the two are equal so it goes unnoticed, but virtiofs (Docker Desktop/Lima/Colima bind mounts) and NFS report a 1 MiB `f_bsize` against a 4 KiB `f_frsize`, inflating `disk_total`/`disk_free` by 256x.

**Describe your setup**

```
LavinMQ 2.8.0
Crystal 1.20.2 [2482c62c1] (2026-05-15)
LLVM: 20.1.8
Default target: aarch64-unknown-linux-gnu
```

Running in Docker via Colima (virtiofs) on macOS, data dir on a bind-mounted host path.
                                                                                                                                                              **How to reproduce**

1. Run LavinMQ in Docker on macOS/Windows (Docker Desktop, or Colima/Lima) with the data dir on a virtiofs bind mount from the host.
2. Open the management UI and look at the node's disk usage, or `curl /api/nodes`.
3. `disk_total`/`disk_free` are 256x larger than reality (e.g. a 926 GiB disk shows as 231.6 TiB).

**Expected behavior**

  Disk size matches what `df` reports for the data dir (frsize is always equal to bsize on local block storage, so those are unaffected).

jage:
statfs(2) on Linux counts blocks in f_frsize units; f_bsize is only the optimal transfer size. They are equal on local filesystems, but virtiofs (Docker bind mounts) and NFS report a 1 MiB transfer size with 4 KiB fragments, inflating disk_total/disk_free 256x in the dashboard. Darwin is unaffected, its f_bsize is the fundamental block size.

Fixes #2054

### DIFF
--- a/src/stdlib/filesystem.cr
+++ b/src/stdlib/filesystem.cr
@@ -78,8 +78,19 @@ end
 
 struct FilesystemInfo
   def initialize(statfs : LibC::Statfs)
-    @available = statfs.bavail.to_u64 * statfs.bsize
-    @total = statfs.blocks.to_u64 * statfs.bsize
+    block_size = block_size(statfs)
+    @available = statfs.bavail.to_u64 * block_size
+    @total = statfs.blocks.to_u64 * block_size
+  end
+
+  # Linux counts blocks in f_frsize units; bsize is the transfer size,
+  # 1 MiB on virtiofs/NFS, which would inflate sizes 256x.
+  private def block_size(statfs : LibC::Statfs)
+    {% if flag?(:linux) %}
+      statfs.frsize.zero? ? statfs.bsize : statfs.frsize
+    {% else %}
+      statfs.bsize
+    {% end %}
   end
 
   # Bytes available to non-privileged on disk
--- a/spec/filesystem_spec.cr
+++ b/spec/filesystem_spec.cr
@@ -1,0 +1,45 @@
+require "spec"
+require "../src/stdlib/filesystem"
+
+describe FilesystemInfo do
+  # Real statfs64 against the OS; only diverges from the bug on virtiofs/NFS
+  # mounts. The deterministic specs below guard the regression everywhere.
+  it "matches df for the current filesystem" do
+    path = Dir.current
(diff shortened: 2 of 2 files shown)


src/stdlib/filesystem.cr
```diff
@@ -78,8 +78,19 @@ end
...
-    @available = statfs.bavail.to_u64 * statfs.bsize
-    @total = statfs.blocks.to_u64 * statfs.bsize
+    block_size = self.class.block_size(statfs)
+    @available = statfs.bavail.to_u64 * block_size
+    @total = statfs.blocks.to_u64 * block_size
+  end
+
+  # Linux counts blocks in f_frsize units; bsize is the transfer size,
+  # 1 MiB on virtiofs/NFS, which would inflate sizes 256x.
+  protected def self.block_size(statfs : LibC::Statfs)
```
carlhoerberg: protected and static method due to specing? or could be instance method and private?

carlhoerberg:
```suggestion
  private def block_size(statfs : LibC::Statfs)
```

src/stdlib/filesystem.cr
```diff
@@ -78,8 +78,19 @@ end
 
 struct FilesystemInfo
   def initialize(statfs : LibC::Statfs)
-    @available = statfs.bavail.to_u64 * statfs.bsize
-    @total = statfs.blocks.to_u64 * statfs.bsize
+    block_size = self.class.block_size(statfs)
```
carlhoerberg:
```suggestion
    block_size = block_size(statfs)
```

carlhoerberg (approved):

PR #2061, opened 2026-06-16
Include ::MQTT::Protocol in LavinMQ::MQTT::Protocol to not pollute LavinMQ::MQTT

spuun:
Before this PR `::MQTT::Protocol` (from [mqtt-protocl.cr](github.com/84codes/mqtt-protocol.cr)) was included in `LavinMQ::MQTT` meaning that each packet ended upp in that namespace. One problem with this is that also `::MQTT::Protocol::Error` ends up in `LavinMQ::MQTT` blocking us from adding a "base" `LavinMQ::MQTT::Error` class.

This PR moves the include of `::MQTT::Protocol` to `LavinMQ::MQTT::Protocol`, which probably makes more sense.

Still compiles and specs passes

### DIFF
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -72,7 +72,7 @@ module LavinMQ
         @vhost.rm_connection(client)
       end
 
-      def publish(packet : MQTT::Publish)
+      def publish(packet : Protocol::Publish)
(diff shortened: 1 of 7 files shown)


antondalgren (approved):
Nothing that stands out, seems like a solid mechanical refactoring.

snichme (approved):

PR #2062, opened 2026-06-16
Raft clustering: code-review fixes (node id, shutdown, ctl safety, password file)

carlhoerberg:
Follow-up fixes on top of #2013, addressing correctness findings from a code review of the raft-backed clustering branch. Each issue is an isolated commit with a regression spec.

## Fixes

- **`.clustering_id` parse (#2)** — `load_or_generate_node_id` only rescued `File::NotFoundError`, so an empty or non-base-36 id file (e.g. a partial write from a prior crash) raised an opaque `ArgumentError` on boot. Parse with `to_i32?(36)` and fail with a clear message; do **not** regenerate (a fresh id would silently drop the node from its cluster).
- **node id never 0 (#5)** — `Random::Secure.rand(Int32::MAX)` can return `0`, which raft.cr uses as the "no node / no leader" sentinel. Generate in `1..Int32::MAX`.
- **`run_on_tick` shutdown hang (#4)** — the tick loop's `select` races `@stop_signal` against the `@actions` case, so on `stop()` it can exit without draining an enqueued action, leaving the caller blocked forever on `reply.receive`. Wait on `@tick_done` alongside the reply.
- **ctl fail-closed on an unverifiable node (#3)** — `raft_reset`/`raft_join` only ran the safety gate when a node was actively running (lock held). A *stopped* multi-peer member could be wiped without `--force`, silently dropping it from its cluster. `ensure_safe_to_stop` now runs unconditionally; an unreachable/unverifiable node requires `--force`.
- **ctl unknown peer count (#6)** — a missing/non-array `peers` in `/raft/status` mapped to `0` and passed the `peers <= 1` safe check. Treat an unknown peer count as unsafe.
- **password in a file, not the raft log (#1)** — the replication secret was proposed as a `SetSecret` command and applied into the state machine, persisting it in every node's log/snapshots. Worse, `Coordinator#password` on a follower proposed (returns false) then busy-waited up to 5s for replication; on timeout the raise stranded `follow_leader_loop` (which only retries on the next leader change), so a follower could stop replicating entirely. The secret now lives in a local `<data_dir>/.clustering_password` file (like the etcd backend reads it from etcd): a leader with no file generates one (`0600`) for single-node bootstrap and logs that it must be copied to other nodes; a non-leader with no file fails fast. `SetSecret` and the state machine's `secret` field are removed — only the ISR is replicated now.

> Note on the password file: this mirrors the `raft3` branch's `.clustering_password` file approach, with the addition that a leader generates the file if missing (so single-node bootstrap still works out of the box). For a real multi-node cluster the file must be pre-shared to followers.

## Testing

- `make test` — 1936 examples, 0 failures, 0 errors
- `make lint` — clean

### DIFF
--- a/spec/lavinmqctl/raft_spec.cr
+++ b/spec/lavinmqctl/raft_spec.cr
@@ -209,7 +209,9 @@ describe "LavinMQCtl raft_*" do
         original_argv = ARGV.dup
         begin
           ARGV.clear
-          ARGV.concat(["raft_reset", "--data-dir=#{data_dir}"])
+          # --force: no node is running here, so the safety check can't reach
+          # /raft/status and would otherwise fail closed.
+          ARGV.concat(["raft_reset", "--data-dir=#{data_dir}", "--force"])
           cli = LavinMQCtl.new(stdout)
           cli.run_cmd
         ensure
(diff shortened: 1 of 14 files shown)


antondalgren:
Cherry-picked into #2013

Issue #1835, opened 2026-03-30
Save leader checksums to disk periodically to survive unclean   shutdowns


viktorerlingsson:
1. On unclean shutdown the leader doesn't save any checksums and has to hash every segment file on restart, which takes a long time. 
2. We don't calculate checksums at all during normal operations, meaning that when a follower connects, the leader has to calculate checksums for everything that has changed since the last follower connected. This can potentially take a lot of time. 

Both things can be fixed by calculating checksum every time we are done with a segment (`open_new_segment`), and write to disk (`checksums.sha1`), so that followers can quickly connect, and that this node can quickly become a follower if needed.

PR #2065, opened 2026-06-16
Persist follower checksums incrementally during full sync (#1834)

### LINKED ISSUE #1834: Save checksums on followers periodically during hash calculations
When a follower calculates file checksums during full sync, it currently starts from scratch if interrupted (unclean shutdown, leader switch, etc). This can cause a loop where it never manages to calculate all checksums. Persist completed hash results to disk periodically so sync can resume from where it left off rather than hashing all files.

carlhoerberg:
Fixes #1834.

## Problem

During clustering full sync a follower hashes each pre-existing local file while comparing against the leader (`Client#sync_files`), but kept the results only in memory until a graceful `close` wrote `checksums.sha1`. If the process dies mid-sync (crash, leader switch), that work is lost and the next start re-hashes everything from scratch — which on a large data dir can loop and never finish.

## Fix

Persist each computed hash *as it is produced* by appending one line to `checksums.sha1`. `restore` (already called in `initialize`, then it deletes the file — a one-shot read) loads the partial progress so hashing resumes where it left off.

- `Checksums#append(path, hash)` sets the hash in memory **and** appends a `<sha1> *<path>` line (lazily opened append handle, flushed, no fsync). `store` closes that handle before its full clean rewrite.
- Only the `sync_files` compare loop appends — there `sha1.file` hashes the *entire* current file and that file is not being mutated. Streaming hashes (`@file_digests`, `file_from_socket`) stay in memory and are flushed by the close-time full rewrite, so a still-growing file can never drift onto disk.

## Safety

A persisted follower hash is only ever compared against the leader's authoritative hash, so a stale entry triggers at most a redundant re-fetch, never silent data loss. `restore`'s delete-on-read keeps it one-shot (a stale line never outlives one restart). No fsync needed — the page cache survives a process crash (the #1834 scenario) and the cache is purely an optimization.

## Tests

- `spec/clustering/checksums_spec.cr` (new): append persists immediately; restore loads then deletes (one-shot); `store` produces a clean snapshot.
- `spec/clustering/client_sync_spec.cr`: drives `sync_files` against a pre-existing local file and asserts the hash lands in `checksums.sha1` **before** any close.

Full suite: 1948 examples, 0 failures. Lint clean.

> Note: the leader side (#1835) is intentionally **not** in this PR; it's handled separately.

### DIFF
--- a/spec/clustering/checksums_spec.cr
+++ b/spec/clustering/checksums_spec.cr
@@ -14,7 +14,7 @@ describe LavinMQ::Clustering::Checksums do
     end
   end
 
-  it "restores appended hashes and deletes the file (one-shot)" do
+  it "restores appended hashes once, then discards them (one-shot)" do
     with_datadir do |data_dir|
       hash = Digest::SHA1.digest("hello")
       written = LavinMQ::Clustering::Checksums.new(data_dir)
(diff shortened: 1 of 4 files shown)


spuun:
Should `Client#file_from_socket` use `Checksums#append` instead of `Checksums#[]=`?
https://github.com/cloudamqp/lavinmq/blob/999a38b5e20d34eb49c571386843ffdcedde3023/src/lavinmq/clustering/client.cr#L303

carlhoerberg:
Yes!

carlhoerberg:
Fixed

PR #2087, opened 2026-06-17
Fix: Validate definitions import entries up front

### LINKED ISSUE #2073: Batched definitions import drops the deferred store flush on a mid-loop exception (unpersisted params/policies)
**Describe the bug**

Definitions import was changed to a batched-flush model: each entry is added with `save: false`
and the affected store is persisted once at the end of the loop (`touched_parameters.each(&.save_parameters!)`
etc., `definitions.cr` import_parameters ~200-228, import_policies ~267-288,
import_global_parameters ~253-265). A per-entry parse can raise mid-loop with no `rescue`/`ensure`:
`p["value"].as_h` (TypeCastError, ~line 210) or `value["priority"].as_i.to_i8` (OverflowError for
priority > 127, ~247/279). The exception aborts the loop before the flush, so entries added
before the bad one are never written to disk. But `add_parameter` (`vhost.cr:419-424`) calls
`apply_parameters` unconditionally, so any shovel/federation upstream added earlier is already
running in memory. On restart the unpersisted config silently vanishes.

2.8.1 added each entry with `save: true`, so a mid-loop failure still left earlier entries durable.
Regressed by commit `5a82c961` (#2014).

**Describe your setup**

- LavinMQ 2.9.0.
- Admin imports definitions (`POST /api/definitions`, `/upload`, or `/api/definitions/:vhost`)
  with a valid entry followed by a malformed one.

`bin/lavinmq --build-info`

**How to reproduce**

1. `POST /api/definitions` with `parameters: [<valid shovel s1>, <operator_policy with priority 999>]`.
2. Shovel s1 starts running; the second entry raises OverflowError before the flush.
3. `parameters.json` is never updated; the request returns 400/500.
4. Restart the broker: shovel s1 is gone, though it was live until restart.

**Expected behavior**

Durable definitions added before a failing entry are persisted (or none are applied); in-memory
and on-disk state stay consistent.

**Solution plan**

Flush the `touched_*` stores in an `ensure`, or validate all entries before applying any.

kickster97:
Fixes #2073.

Batched definitions import added each entry with `save: false` and flushed the touched stores once at the end of the loop. A per-entry parse could raise mid-loop (`value.as_h` on a non-object, `priority.as_i.to_i8` for priority > 127), aborting before the flush, so entries added earlier were applied in memory but never written to disk and vanished on restart.

Following the discussion below, this parses and validates every entry up front in `import_parameters`, `import_policies`, and `import_global_parameters`, and only applies them once all entries are valid. A malformed entry now makes the whole import a clean no-op instead of partially applying, so memory and disk stay consistent.

### DIFF
--- a/spec/api/definitions_spec.cr
+++ b/spec/api/definitions_spec.cr
@@ -460,6 +460,77 @@ describe LavinMQ::HTTP::Server do
       end
     end
 
+    it "applies no parameters when one entry is malformed (issue #2073)" do
+      with_http_server do |http, s|
+        # A valid shovel followed by an operator_policy with priority > 127, which
+        # overflows Int8 and raises while parsing. Entries are validated up front,
+        # so a malformed entry makes the whole import a clean no-op: the shovel
+        # before it is neither applied in memory nor written to disk.
+        body = %({ "parameters": [
+          {
+            "name": "regression_shovel_2073",
+            "component": "shovel",
+            "vhost": "/",
+            "value": {
+              "src-uri": "#{s.amqp_url}",
+              "src-queue": "regression_q_2073",
+              "dest-uri": "#{s.amqp_url}",
+              "dest-queue": "regression_q_2073"
+            }
+          },
+          {
+            "name": "regression_op_2073",
+            "component": "operator_policy",
+            "vhost": "/",
+            "value": {
+              "pattern": "^.*",
+              "apply-to": "queues",
+              "priority": 999,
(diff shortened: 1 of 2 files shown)


kickster97:
@carlhoerberg The alternative is a pre-check that validates all entries up front so a bad import is a clean no-op, what do you think?

carlhoerberg:
I think validate!

Issue #2083, opened 2026-06-17
proxy_protocol_trusted_sources fails open when all CIDR entries are invalid


viktorerlingsson:
## Summary

If every entry in `proxy_protocol_trusted_sources` fails to parse, the trusted
list ends up empty and the broker trusts PROXY headers from all sources. A config
typo silently disables the IP allow-list.

## Mechanism

- `IPMatcher.parse_list` (`ip_matcher.cr:54-66`) drops invalid CIDR/IP entries
  with a logged error and continues. If every entry is malformed, the result is
  an empty array.
- `trusted_proxy_source?` (`server.cr:182-186`) treats an empty list as "trust
  all sources" (intended as backward-compat for the unconfigured case).

The code cannot distinguish "unconfigured" from "configured but all entries
failed to parse". A single typo (e.g. `10.0.0.0/33`) silently reverts a
locked-down config to trust-all, letting any peer spoof its source IP via a
forged PROXY header. The startup warning (`server.cr:71-72`) only checks
`.empty?`, so it does not flag this case distinctly.

## Impact

Security-relevant: source-IP-based ACLs and logging can be bypassed by an
untrusted client when the operator believes the source is locked down. Requires
operator misconfiguration to trigger, so this is hardening rather than an active
exploit.

## Suggested fix

A malformed entry in a non-empty `proxy_protocol_trusted_sources` should fail
loudly (abort at startup) or fail closed, rather than collapsing to trust-all.
Reserve the trust-all backward-compat path for the genuinely unconfigured case
(detected from the raw config value, not inferred from an empty parsed list).

## Test plan

Configure `proxy_protocol_trusted_sources` with only invalid entries and assert
startup is rejected (or the source is treated as untrusted), not trust-all.
Existing `spec/proxy_protocol_spec.cr` only tests that valid entries survive
alongside invalid ones.

Issue #2079, opened 2026-06-17
PROXY auto-detect peek blocks pre-auth with no read timeout (slowloris / FD exhaustion)


kickster97:
**Describe the bug**

`ProxyProtocol.parse`'s first statement is `peeked = io.peek` (`proxy_protocol.cr:18`), with no
read timeout set beforehand. The `read_timeout = 15.seconds` is set only later, inside
`V1.parse`/`V2.parse` (`proxy_protocol.cr:34,108`), after the peek has already blocked. `server.cr`
sets no read timeout before this point. A peer that completes the TCP/unix handshake then sends
zero bytes parks the accept fiber in `io.peek` forever, holding the FD. There is no accept-level
connection timeout in config.

Reachable paths:
1. TCP listeners with `tcp_proxy_protocol = true` (`server.cr:166-167`).
2. ALL unix-socket connections, unconditionally (`accept_unix`, `server.cr:206`) - default config.
3. Clustering follower-address TCP connections (`server.cr:174-175`).

Pre-2.9.0 the configured-proxy path set the timeout before the first read; for unix sockets it
did not read at all unless explicitly configured. Regressed by commit `ad80d79c` (auto-detect).

**Describe your setup**

- LavinMQ 2.9.0.
- `tcp_proxy_protocol = true` (TCP variant), or any unix-socket client (default config).

`bin/lavinmq --build-info`

**How to reproduce**

1. Set `tcp_proxy_protocol = true`. Open a TCP connection to the AMQP port and send nothing.
2. The accept fiber blocks in `io.peek` forever instead of timing out after 15s.
3. Repeat to exhaust file descriptors. (Unix variant needs no config: connect to the AMQP/MQTT
   unix socket and send nothing.)

**Expected behavior**

An idle pre-auth connection times out (as pre-2.9.0, 15s); the initial peek is bounded.

**Solution plan**

Set `io.read_timeout = 15.seconds` before the peek in `ProxyProtocol.parse` (reset in `ensure`),
or set a handshake read timeout in `accept_tcp`/`accept_unix` before calling parse.

PR #2099, opened 2026-06-21
fix: PROXY header TCP segment splits (#2082) and IPv4-mapped IPv6 trusted sources (#2070)

### LINKED ISSUE #2070: PROXY trusted-source check fails for IPv4 LB on dual-stack (::) listeners (::ffff: mapped addresses)
**Describe the bug**

On a dual-stack listener (`bind = ::`), an IPv4 load balancer arrives as `::ffff:10.0.0.1`. A
`proxy_protocol_trusted_sources` IPv4 entry builds a `Type::IPv4` matcher, and
`IPMatcher#matches?` converts the peer with `ip_to_bytes_v4`, which returns nil for a mapped
address (`parse_v4_fields?("::ffff:...")` is nil) and has no `::ffff:` fallback
(`ip_matcher.cr:69-76,96-100`). So `trusted_proxy_source?` returns false
(`server.cr:182-186`), the PROXY header is dropped, and every client is attributed to the LB's
IP.

Note: `spec/ip_matcher_spec.cr:330-338` currently asserts this failing behavior as correct, so
the fix must flip those specs. Shipped default (`amqp_bind = 127.0.0.1`) is unaffected; this
only triggers when an operator sets `bind = ::`.

Introduced by `ad80d79c` (#1601).

**Describe your setup**

- LavinMQ 2.9.0 RC, `bind = ::` (dual-stack).
- PROXY protocol enabled, `proxy_protocol_trusted_sources` with an IPv4 address/CIDR.
- An IPv4 load balancer in front.

`bin/lavinmq --build-info`

**How to reproduce**

1. Set `amqp_bind = ::`, enable PROXY, set `proxy_protocol_trusted_sources = <LB IPv4/CIDR>`.
2. Connect through the IPv4 LB (broker sees `::ffff:<LB-IPv4>` + PROXY header).
3. Log shows "PROXY protocol from untrusted source ... ignoring header"; real client IP is lost.

**Expected behavior**

An IPv4 trusted source matches an IPv4-mapped IPv6 peer, so the PROXY header is honored and
real client IPs are preserved.

**Solution plan**

Separate PR, no existing work to cherry-pick. Fix: in `ip_to_bytes_v4`, when `parse_v4_fields?`
is nil, fall back to `parse_v6_fields?` and extract the embedded v4 from an `::ffff:` mapped
address. Flip the two specs at `spec/ip_matcher_spec.cr:330-338` and add a
`trusted_proxy_source?` dual-stack regression spec. Difficulty 2.

### LINKED ISSUE #2082: PROXY protocol header can be misread on a TCP segment split (wrong source IP / dropped connection)
## Summary

PROXY protocol detection and the V2 header read assume a single read returns
enough bytes. On a TCP segment split they can short-read, leading to a wrong
source IP and/or a dropped connection.

## Mechanism

- `ProxyProtocol.parse` (`proxy_protocol.cr:17-26`) detects the protocol from a
  single `io.peek`. `IO::Buffered#peek` performs at most one `fill_buffer` (one
  `read` syscall) and returns whatever is buffered, with no minimum-fill
  guarantee.
- If the first TCP segment carries fewer than 12 bytes (V2) or 5 bytes (V1),
  `peek` returns short, both signature checks are skipped, and `parse` returns
  `nil`.
- `extract_conn_info` (`server.cr:163-179`) then falls back to the peer address
  (the load balancer's IP), and the unconsumed PROXY bytes remain in the stream
  and are fed to the AMQP/MQTT parser, breaking the handshake.
- `V2.parse` (`proxy_protocol.cr:110`) has the same class of bug: `io.read` of
  the 16-byte header ignores a short read (line 128 correctly uses `read_fully`).

## Impact

Low likelihood: a load balancer on a LAN sends the ~28-byte header in one
segment. A split is possible under PMTU edges or LBs that emit the header in a
separate small write. When it happens the result is a wrong source IP (which
feeds IP-based ACLs and logging) and/or a dropped connection, intermittent and
hard to diagnose.

## Suggested fix

Add a `peek_at_least(n)` helper that loops `fill_buffer` until `n` bytes are
buffered or EOF, and use it for signature detection. Use `read_fully` for the V2
header read.

## Test plan

Feed a V2 header in two chunks (write part of it, parse in a fiber, write the
rest) and assert detection still succeeds. Existing `spec/proxy_protocol_spec.cr`
always writes the full header into the pipe, so the split case is uncovered.

carlhoerberg:
## Fixes #2082 — PROXY header misread on TCP segment split
- Detection now uses a **prefix match** on the peeked bytes. A single `peek` can return fewer bytes than the full signature when the header is split across TCP segments, so we compare only the bytes we have against the *start* of each signature. No supported protocol (AMQP `AMQP\0\0\9\1`, MQTT, HTTP) begins with `"PROXY"` or the V2 signature, so the prefix match is unambiguous; the V1/V2 parsers then read the rest of the header.
- `V2.parse` uses `read_fully` for the 16-byte header instead of `io.read`, which silently ignored short reads on a split.

## Fixes #2070 — IPv4 trusted source vs IPv4-mapped IPv6 on dual-stack listeners
On a `bind = ::` listener an IPv4 load balancer arrives as `::ffff:10.0.0.1`. `IPMatcher#ip_to_bytes_v4` now falls back to extracting the embedded IPv4 from an `::ffff:` mapped address, so an IPv4 trusted-source entry still matches. The two `ip_matcher_spec` assertions that asserted the old (broken) behavior are flipped, with a non-mapped IPv6 guard added.

## Regression coverage for already-merged fixes
- #2079 (handshake timeout, merged in #2095): spec that the initial peek times out and `read_timeout` is reset.
- #2083 (fail-loud trusted sources, merged in #2096): spec that an invalid entry raises rather than collapsing to trust-all.

## Tests
- `spec/proxy_protocol_spec.cr` — V1/V2 segment-split detection, peek timeout, fail-loud, IPv4-mapped dual-stack match (25 examples, 0 failures)
- `spec/ip_matcher_spec.cr` — flipped 2 mapped-address assertions, added non-mapped IPv6 guard (51 examples, 0 failures)
- `make lint` clean

### DIFF
--- a/spec/ip_matcher_spec.cr
+++ b/spec/ip_matcher_spec.cr
@@ -327,14 +327,22 @@ describe LavinMQ::IPMatcher do
         matcher.matches?("2001:db8:0:2::").should be_false
       end
 
-      it "IPv4 matcher rejects IPv4-mapped IPv6" do
+      it "IPv4 matcher matches IPv4-mapped IPv6 (dual-stack listener)" do
         matcher = LavinMQ::IPMatcher.parse("192.168.1.1")
-        matcher.matches?("::ffff:192.168.1.1").should be_false
+        matcher.matches?("::ffff:192.168.1.1").should be_true
+        matcher.matches?("::ffff:192.168.1.2").should be_false
       end
 
-      it "IPv4 CIDR rejects IPv4-mapped IPv6" do
+      it "IPv4 CIDR matches IPv4-mapped IPv6 (dual-stack listener)" do
         matcher = LavinMQ::IPMatcher.parse("192.168.1.0/24")
-        matcher.matches?("::ffff:192.168.1.50").should be_false
+        matcher.matches?("::ffff:192.168.1.50").should be_true
+        matcher.matches?("::ffff:192.168.2.50").should be_false
+      end
+
+      it "IPv4 matcher doesn't match a real (non-mapped) IPv6 address" do
+        matcher = LavinMQ::IPMatcher.parse("192.168.1.1")
+        matcher.matches?("2001:db8::c0a8:0101").should be_false
+        matcher.matches?("::1").should be_false
       end
 
       it "IPv6 matcher rejects raw IPv4" do
--- a/spec/proxy_protocol_spec.cr
+++ b/spec/proxy_protocol_spec.cr
@@ -156,6 +156,42 @@ describe "ProxyProtocol" do
       # read_timeout must be reset so it doesn't leak into the connection handler
       r.read_timeout.should be_nil
     end
+
+    it "detects a V2 header split across TCP segments" do
+      pp_bytes = UInt8.static_array(
+        0x0d, 0x0a, 0x0d, 0x0a, 0x00, 0x0d, 0x0a, 0x51,
+        0x55, 0x49, 0x54, 0x0a, 0x21, 0x11, 0x00, 0x0c,
+        0x7f, 0x00, 0x00, 0x01, 0x7f, 0x00, 0x00, 0x01,
+        0x92, 0x30, 0x16, 0x27
+      )
+      r, w = IO.pipe
+      ch = Channel(LavinMQ::ConnectionInfo?).new
+      spawn { ch.send LavinMQ::ProxyProtocol.parse(r) }
+
+      # First segment carries fewer than the 12 signature bytes
+      w.write pp_bytes.to_slice[0, 6]
+      Fiber.yield
+      w.write pp_bytes.to_slice[6..]
+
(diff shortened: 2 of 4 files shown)


viktorerlingsson:
A couple of these issues already had PRs: 

#2095 (peek timeout) is functionally a duplicate of the fix here, we can close that in favor of this. 

#2096 however takes a stricter line on trusted sources - it fails loudly on any malformed entry in the config. I'd argue that bad config = crash is the safer default, rather than swallowing it and just logging an error. Which way do we want to go?

PR #2100, opened 2026-06-21
Read the full AMQP protocol header in confirm_header

eandersson:
A single IO#read into an uninitialized buffer could return a short count
when the 8-byte preamble is split across TLS records / TCP segments under
load, leaving the tail uninitialized; the protocol comparison then failed
and the server dropped the connection mid-handshake. Use read_fully? to
read all 8 bytes (bounded by the existing read_timeout), and add a
regression test for a fragmented header.

I ran into this issue when adding e2e testing to amqpstorm for lavinmq.
https://github.com/eandersson/amqpstorm/commits/backup_20260601/
https://github.com/eandersson/amqpstorm/actions/runs/27906654414/job/82576716040

### DIFF
--- a/spec/connection_spec.cr
+++ b/spec/connection_spec.cr
@@ -315,4 +315,26 @@ describe LavinMQ::Server do
       end
     end
   end
+
+  describe "protocol header" do
+    it "accepts a protocol header delivered in fragments" do
+      with_amqp_server do |s|
+        io = TCPSocket.new("localhost", amqp_port(s))
+        io.read_timeout = 5.seconds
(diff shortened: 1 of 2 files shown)


eandersson:

eandersson:
Still working on this, trying to get my tests to pass 100% of the time.

eandersson:
This ended up being on my side. Fixed it here.
https://github.com/eandersson/amqpstorm/commit/81790370f9307eb5c99390c40281fcfb4fedf6b4

PR #2102, opened 2026-06-21
fix: route all publisher confirms through the persister loop (#2078)

### LINKED ISSUE #2078: SIGHUP sync-mode flip reorders publisher confirms on a standalone node (cumulative Ack regression)
**Describe the bug**

`Persister#enqueue_ack` picks the confirm path per call via `if Config.instance.sync? || @replicator`
(`persister.cr:33-44`). `sync` is a runtime-mutable INI option (`config/options.cr:194`; SIGHUP ->
reload, `launcher.cr:283`). On a standalone node (`@replicator` nil) the path depends solely on
`sync?`:

- `sync? true`: batched, msgid queued in `@pending_acks`, confirmed only after a blocking `syncfs`
  on the isolated publish-confirm thread (can be many ms).
- `sync? false`: direct, immediate `channel.enqueue_confirm_ack` on the publishing fiber.

Both feed one 1-slot mailbox whose writer sends `Basic.Ack(msgid, multiple: true)`
(`channel.cr:331-351`). Timeline: msgid 5 is queued batched (awaiting syncfs), the operator
SIGHUPs `sync` off, msgid 6 is published direct -> `Ack(6, multiple: true)` goes out first
(cumulatively confirming 5), then the persister loop sends `Ack(5, multiple: true)` after it,
re-acking already-confirmed tags out of order. Strict clients treat an ack for an already-settled
tag as a protocol violation and close the channel. (A secondary mailbox-coalescing race exists
when both producers run in parallel.)

Commits `b3041a46` / `5aa37e4c` / `24c83ff3`.

**Describe your setup**

- LavinMQ 2.9.0, standalone node (no clustering).
- A publisher in confirm mode.
- Operator toggles `sync` via INI + SIGHUP while publishing.

`bin/lavinmq --build-info`

**How to reproduce**

1. Standalone broker with `sync = true` (default); `channel.confirm_select`.
2. Publish a steady stream of small persistent messages so acks queue behind the syncfs.
3. Edit the INI to `sync = false` under `[main]` and send SIGHUP.
4. Capture AMQP frames: observe `Basic.Ack(multiple:true)` for a higher delivery tag arriving
   before one for a lower tag on the same channel.

**Expected behavior**

Confirms stay in delivery-tag order across a runtime sync-mode change; no duplicate cumulative ack.

**Solution plan**

Snapshot the confirm path per channel at `confirm_select` time (don't re-evaluate `sync?` per
call), or always route confirms through the persister loop (skipping `syncfs` under no-sync) so
each channel has one ordered producer.

carlhoerberg:
## Problem

`Persister#enqueue_ack` picked the confirm path per call based on the runtime-mutable `Config.instance.sync?`:

- `sync? false`: direct, immediate `channel.enqueue_confirm_ack` on the publishing fiber.
- `sync? true` (or clustered): msgid queued in `@pending_acks`, confirmed only after the publish confirm loop drains.

`sync` is reloaded on SIGHUP. On a standalone node, flipping `sync` off mid-stream produces two producers of ack frames for the same channel. The classic timeline: msgid 5 is queued batched (awaiting syncfs), the operator SIGHUPs sync off, msgid 6 is published direct → `Ack(6, multiple: true)` goes out first (cumulatively confirming 5), then the loop sends `Ack(5, multiple: true)` after it — re-acking an already-settled tag out of order.

Strict clients reject this. Confirmed by reading the 84codes clients:

- **amqp-client.rb** — `Channel#confirm` does `@unconfirmed.index(delivery_tag) || raise("Delivery tag not found")`; the `RuntimeError` escapes the read loop's `rescue` (not a `READ_EXCEPTIONS`), so the `ensure` block **closes the socket** → connection drop.
- **amqp-client.cr / amqp-client.js** — tolerant: silently drop / log a warning for the unknown tag.

## Fix

Always route confirms through the publish confirm loop, so each channel has exactly one ordered producer of ack frames. `drain_pending_acks` already gates the actual `syncfs` on `Config.instance.sync?`, so no-sync only pays a single hop to the loop rather than a disk flush. This also removes the secondary mailbox-coalescing race between the two producers.

(This is option 2 from the issue's solution plan; it's simpler than snapshotting the path per channel and removes the dual path entirely.)

## Tests

- Added regression specs in `spec/publish_confirm_persistence_spec.cr` that toggle `sync` mid-stream (both directions) on a confirm-mode channel and assert every publish is confirmed and counts are correct.
- The existing "confirms a single publish quickly (no batching delay)" (<100ms) test still passes, confirming the loop hop is cheap.
- `make test` (1946 examples, 0 failures) and `make lint` green.

Fixes #2078

### DIFF
--- a/spec/publish_confirm_persistence_spec.cr
+++ b/spec/publish_confirm_persistence_spec.cr
@@ -82,11 +82,9 @@ describe "Publish Confirm Persistence" do
   end
 
   describe "runtime sync-mode flip (regression for #2078)" do
-    # `sync` is a runtime-mutable INI option (INI + SIGHUP). All confirms are
-    # routed through the publish confirm loop regardless of sync state, so a
-    # mid-stream flip can never let a later ack overtake an earlier one and send
-    # cumulative Basic.Ack frames out of delivery-tag order. Toggle sync while a
-    # confirm-mode publisher is in flight and assert every publish is confirmed.
+    # Smoke test for #2078: confirms keep flowing across a mid-stream `sync` flip
+    # (see Persister#enqueue_ack for why). Asserts confirms + counts only, not
+    # Basic.Ack ordering - the Crystal client tolerates out-of-order tags.
     it "keeps confirming when sync is disabled mid-stream" do
       LavinMQ::Config.instance.sync = true
       with_amqp_server do |s|
--- a/spec/publish_confirm_persistence_spec.cr
+++ b/spec/publish_confirm_persistence_spec.cr
@@ -81,6 +81,47 @@ describe "Publish Confirm Persistence" do
     end
   end
 
+  describe "runtime sync-mode flip (regression for #2078)" do
+    # `sync` is a runtime-mutable INI option (INI + SIGHUP). All confirms are
+    # routed through the publish confirm loop regardless of sync state, so a
+    # mid-stream flip can never let a later ack overtake an earlier one and send
+    # cumulative Basic.Ack frames out of delivery-tag order. Toggle sync while a
+    # confirm-mode publisher is in flight and assert every publish is confirmed.
+    it "keeps confirming when sync is disabled mid-stream" do
+      LavinMQ::Config.instance.sync = true
+      with_amqp_server do |s|
+        with_channel(s) do |ch|
+          ch.confirm_select
+          q = ch.queue("flip_sync_off")
+          50.times { q.publish "msg" }
+          LavinMQ::Config.instance.sync = false # operator SIGHUPs sync off mid-stream
+          50.times { q.publish "msg" }
+          ch.wait_for_confirms.should be_true
+          s.vhosts["/"].queue(q.name).message_count.should eq 100
+        end
+      end
+    ensure
+      LavinMQ::Config.instance.sync = true
+    end
+
+    it "keeps confirming when sync is enabled mid-stream" do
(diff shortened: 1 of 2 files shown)


spec/publish_confirm_persistence_spec.cr
```diff
@@ -81,6 +81,47 @@ describe "Publish Confirm Persistence" do
     end
   end
 
+  describe "runtime sync-mode flip (regression for #2078)" do
+    # `sync` is a runtime-mutable INI option (INI + SIGHUP). All confirms are
+    # routed through the publish confirm loop regardless of sync state, so a
+    # mid-stream flip can never let a later ack overtake an earlier one and send
+    # cumulative Basic.Ack frames out of delivery-tag order. Toggle sync while a
+    # confirm-mode publisher is in flight and assert every publish is confirmed.
```
viktorerlingsson: The specs don't actually test the ordering, I think we should leave that part out here (It's already covered in the comment in `persister.cr` anyway)
```suggestion
    # Smoke test for #2078: confirms keep flowing across a mid-stream `sync` flip
    # (see Persister#enqueue_ack for why). Asserts confirms + counts only, not
    # Basic.Ack ordering - the Crystal client tolerates out-of-order tags.
```

viktorerlingsson (approved):

PR #2104, opened 2026-06-21
fix: restart delayed queue expire loop on store error (#2074)

### LINKED ISSUE #2074: DelayedExchangeQueue expire loop dies on a store error and is never restarted (delayed messages stranded)
**Describe the bug**

`DelayedExchangeQueue#message_expire_loop` (`delayed_exchange_queue.cr:65-90`) rescues only
`::Channel::ClosedError`, and its `ensure` just sets `@message_expire_fiber_active.set(false)`
with no restart. The base `Queue#message_expire_loop` (`queue.cr:188-195`) by contrast rescues
`MessageStore::Error` (close) and calls `ensure_expire_fiber` to restart.

`expire_messages -> first_delayed?/shift_delayed?` can raise `MessageStore::Error` on a
corrupt/truncated segment (`delayed_message_store.cr:78,93`). That propagates out of the override
uncaught, the only release fiber dies, and nothing restarts it (`publish`/`basic_get`/`ack`/`reject`
are no-ops on a delayed queue; `delay()` never calls `ensure_expire_fiber`). All pending delayed
messages are stranded forever. Worse, the next `delay()` does an unbuffered
`@message_ttl_change.send nil` (`delayed_exchange_queue.cr:50`) with no receiver, hanging the
publishing fiber.

Commit `8cd0a48f`.

**Describe your setup**

- LavinMQ 2.9.0.
- A durable `x-delayed-message` exchange with pending delayed messages.
- A corrupt/truncated delayed segment (or any store read error in the expire path).

`bin/lavinmq --build-info`

**How to reproduce**

1. Declare a durable `x-delayed-message` exchange and publish messages with `x-delay`.
2. Induce a `MessageStore::Error` in the expire path (e.g. a corrupt/truncated delayed segment).
3. The expire fiber dies and is never restarted; pending delayed messages are never released.
4. Publish another delayed message: the producer connection hangs on `@message_ttl_change.send`.

**Expected behavior**

A transient store error does not permanently kill the release fiber; the loop restarts like the
base queue, and `delay()` never blocks the publisher.

**Solution plan**

Mirror the base loop: add `rescue MessageStore::Error` (close) and `ensure_expire_fiber` to the
`ensure`, and switch `delay()`'s `send` to `try_send?`.

carlhoerberg:
Fixes #2074.

## Problem

`DelayedExchangeQueue#message_expire_loop` only rescued `::Channel::ClosedError`, and its `ensure` just flipped `@message_expire_fiber_active` to `false` with no restart. The base `Queue#message_expire_loop` by contrast rescues `MessageStore::Error` (closing the queue) and calls `ensure_expire_fiber`.

`expire_messages -> first_delayed?/shift_delayed?` can raise `MessageStore::Error` on a corrupt/truncated segment. In the override that propagated out uncaught, killing the only release fiber with nothing to restart it (`publish`/`basic_get`/`ack`/`reject` are no-ops on a delayed queue). All pending delayed messages were stranded forever. Worse, the next `delay()` did an unbuffered `@message_ttl_change.send nil` with no receiver, hanging the publishing fiber.

## Fix

- Mirror the base loop: `rescue MessageStore::Error` (close) and `ensure_expire_fiber` in the `ensure`.
- Switch `delay()`'s `send` to `try_send?` and call `ensure_expire_fiber` so the publisher never blocks on a dead fiber.

## Test

Added a regression spec that seeds the delayed index with an entry pointing past the segment data (so the expire loop's `BytesMessage.from_bytes` raises a `MessageStore::Error`, simulating a corrupt segment), wakes the loop, and asserts:
- the queue closes cleanly instead of silently stranding messages, and
- a subsequent `delay()` returns promptly rather than blocking the publisher.

The spec fails with "Execution expired" against the unfixed source and passes with the fix.

### DIFF
--- a/spec/delayed_message_exchange_spec.cr
+++ b/spec/delayed_message_exchange_spec.cr
@@ -315,8 +315,8 @@ describe "Delayed Message Exchange" do
       wait_for { queue.closed? }
       queue.closed?.should be_true
 
-      # delay() must return promptly (no unbuffered send to a dead fiber) even
-      # after the release fiber is gone.
+      # After the store error closes the queue, delay() must refuse promptly
+      # instead of blocking the publisher.
       msg = LavinMQ::Message.new("", queue.name, "after-error")
       done = Channel(Bool).new
       spawn { done.send(queue.delay(msg)) }
--- a/spec/delayed_message_exchange_spec.cr
+++ b/spec/delayed_message_exchange_spec.cr
@@ -286,6 +286,49 @@ describe "Delayed Message Exchange" do
     end
   end
 
+  it "closes cleanly without hanging the publisher when the expire loop hits a store error" do
+    with_amqp_server do |s|
+      with_channel(s) do |ch|
+        x = ch.exchange(x_name, "topic", args: x_args)
+        # Long delay so the message stays parked in the internal delayed queue
+        hdrs = AMQP::Client::Arguments.new({"x-delay" => 600_000})
+        x.publish_confirm "test message", "rk", props: AMQP::Client::Properties.new(headers: hdrs)
+      end
+      queue = s.vhosts["/"].queue(delay_q_name).as(LavinMQ::AMQP::DelayedExchangeQueue)
+      queue.message_count.should eq 1
+
+      store = queue.@msg_store.as(LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore)
+      seg_id = store.@segments.first_key
+      requeued = store.@requeued.as(LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore::DelayedRequeuedStore)
+
+      # Insert an index entry pointing past the segment data with an already-elapsed
+      # expire_at, so the expire loop reads it first and BytesMessage.from_bytes raises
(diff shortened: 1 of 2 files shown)


spec/delayed_message_exchange_spec.cr
```diff
@@ -286,6 +286,49 @@ describe "Delayed Message Exchange" do
...
+      # Wake the expire loop so it processes the (now-expired) bogus entry
+      queue.@message_ttl_change.send(nil)
+
+      # The store error must close the queue rather than silently killing the only
+      # release fiber and stranding all delayed messages forever.
+      wait_for { queue.closed? }
+      queue.closed?.should be_true
+
+      # delay() must return promptly (no unbuffered send to a dead fiber) even
+      # after the release fiber is gone.
```
viktorerlingsson: This doesn't actually test the `try_send` change, it returns from the closed guard `return false if @closed`. Still valid to test IMO, but I think the comment alludes to testing the `try_send` path. 
```suggestion
      # After the store error closes the queue, delay() must refuse promptly
      # instead of blocking the publisher.
```

viktorerlingsson (approved):

PR #2108, opened 2026-06-22
Deduplication looks at queue content

snichme:

Deduplication currently works by looking at the separate cache that gets populated on each publish and removed after a configured time, same for exchange deduplication. 
But on queue level it makes more sense to look at the content in a queue to decide if the message is a duplicate or not. This PR changes the deduplication behavior to look at current messages in the queue. 

New specs for the new changes.


snichme:
Some concurrency concerns: 

1. Check-then-add race — duplicate? + add as two calls let two concurrent publishes of the same key both slip through. Fixed with an atomic insert? (returns false if already present) as the single gate under the lock.
2. Phantom key on overflow — the old code added the key before the overflow check, so a rejected message left a permanent phantom. Reordered so insert? happens after the overflow decision, inside the lock with push.
3. Rebuild-vs-delivery race — in background mode, a consumer can ack a message mid-scan. We deliberately chose Cheap-A (deleted-aware MessageStore#each, no chunked-lock discipline), accepting a rare transient phantom (cleared on next restart) rather than gating delivery. Also noted: holding the lock for the whole scan would block that queue's publishes (not the broker) — currently the rebuild does take the lock for the scan, which is correct and simple; chunked locking is the future refinement if big queues need it.


Also, I've taken the easy approach to block the queue until done by taking the \@msg_store_lock while rebuilding the index. It can be slow for long queue but should only affect the queue that rebuilding not the broker or the other queues. 
I had some ideas on a "starting" state but too complicated for this change but can improve this feature I think.

PR #2109, opened 2026-06-22
AMQP 1.0

carlhoerberg:
Adds AMQP 1.0 support on the existing AMQP listener while keeping AMQP 0-9-1 on its current path.

Includes:

- SASL PLAIN negotiation, AMQP 1.0 transport detection, and connection/session/link handling
- AMQP 1.0 frame, value, message, transfer, and performative codecs
- Queue/exchange v2 addressing, `open.hostname` vhost selection, anonymous sender routing, and dynamic source/target queues
- Publish and consume support with link credit, settlements, outcomes, transfer fragmentation, max-frame-size handling, and flow/window management
- Interop mappings for application-properties, AMQP headers, `amqp-value`/data bodies, timestamps, URI-decoded addresses, and UUID IDs
- Delivery/lifecycle behavior for single-active-consumer, priority consumers, paused queues, vhost flow, consumer timeout, and idle management close
- Regression coverage in `spec/amqp10_spec.cr`


carlhoerberg:
Will maybe extract the protocol implementation to a separate repo.

src/lavinmq/amqp10/client.cr
```diff
@@ -0,0 +1,433 @@
...
+      if @vhost.max_queues.try { |max| @vhost.queues_size >= max }
+        raise ProtocolError.new("queue limit in vhost '#{@vhost.name}' is reached")
+      end
+
+      name = LavinMQ::AMQP::Queue.generate_name
+      unless @user.can_config?(@vhost.name, name)
+        raise ProtocolError.new("User '#{@user.name}' does not have permissions to queue '#{name}'")
+      end
+      frame = LavinMQ::AMQP::Frame::Queue::Declare.new(0_u16, 0_u16, name, false, false, true, true, false, LavinMQ::AMQP::Table.new)
+      @vhost.apply(frame)
```
carlhoerberg: is there a `VHost#declare_queue` method?

PR #2116, opened 2026-06-24
Persist definitions with JSON snapshots and WAL

carlhoerberg:
Summary:
- Persist definitions as per-vhost JSON snapshots plus a WAL in the vhost directory.
- Load snapshots plus WAL on boot and compact on request/idle or once the WAL reaches 8 MiB.
- Minimize snapshot JSON by omitting implicit defaults and empty arguments.

Tests:
- make test SPEC=spec/vhost_spec.cr
- make lint
- make test

### DIFF
--- a/spec/clustering/follower_spec.cr
+++ b/spec/clustering/follower_spec.cr
@@ -254,6 +254,37 @@ module FollowerSpec
       end
     end
 
+    it "writes a File range with a negative size header" do
+      with_datadir do |data_dir|
+        follower_socket, client_socket = FakeSocket.pair
+        file_index = FakeFileIndex.new(data_dir)
+        follower = LavinMQ::Clustering::Follower.new(follower_socket, data_dir, file_index)
+        file = File.open(File.join(data_dir, "append_source"), "w+")
+        file.write "0123456789".to_slice
+        file.flush
+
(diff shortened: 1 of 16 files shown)


carlhoerberg:
If we implement this it won't be possible to downgrade, unless we keep writing to the legacy file in parallel of course.

carlhoerberg:
But the biggest problem is that arguments (AMQP::Table) won't be byte identical.

carlhoerberg:
10.000 queues, 300.000 bindings takes about 100ms and uses 22 MiB to compact/snapshot. Pretty printed json takes 150 ms and uses 29 MiB.

carlhoerberg:
Integers and float in tables/arguments is not a problem actually, `AMQP::Table#==` compares them over widths, regardless of underlaying storage format. Only timestamp and bytes can't be deciphered correctly from json. But we can augument the json output with `{"@type": "timestamp", "value": 11234000}`

carlhoerberg:
If we implement this it won't be possible to downgrade, unless we keep writing to the legacy file in parallel of course.

carlhoerberg:
We now keep writing to `definitions.amqp` as long as it exists (but won't read it, only if you downgrade again it will be read).

PR #2123, opened 2026-06-26
Advertise UI version via LavinMQ-Version header instead of build-time injection

carlhoerberg:
## Why

The web UI showed the server version in the header by **baking it into the static HTML at build time**: the Makefile computed `VERSION` and passed it to `views/render.sh`, which replaced an SSI `#echo var="VERSION"` directive in the header partial. The compiled value was shipped in every `static/*.html` and went stale unless rebuilt — coupling the front-end build to git/version state.

## What

Advertise the version at runtime **with no extra request**:

- **`ApiDefaultsHandler`** sets a `LavinMQ-Version: <version>` header on every API response. (That handler only sees `/api/*` requests — HTML/static are served upstream — so the header is scoped to API responses.) A custom header is used rather than `Server` because proxies (nginx, …) commonly rewrite or strip `Server`.
- **`static/js/http.js`** reads it off each API response, caches it in `sessionStorage`, and updates the `#version` span. Since every authenticated page makes API calls through `http.js`, the version refreshes on each page load (a rebuilt/upgraded server's version shows up without a hard refresh).
- **`static/js/layout.js`** paints the last-known cached value on load for instant display.
- **Removed the VERSION plumbing** from the `Makefile` and `views/render.sh`, and emptied the header span. `render.sh` still handles `TITLE` and `#include` directives.

Reading a header off the page's own HTML navigation isn't possible from JS, but reading it from a same-origin `fetch()` response is — and works in every browser.

## Tests

- `spec/api/http_api_spec.cr`: the `LavinMQ-Version` header is present on API responses (`/api/whoami`) and absent on static/HTML responses (`/login`).
- `spec/frontend/layout.spec.js`: the version from the header is shown in `#version`, and cached in `sessionStorage`. Uses the `/queues` page (because `nodes.js` overwrites `#version` from `/api/nodes`) and reads `/api/queues` (not `/api/vhosts`, which the test fixture mocks without the header).

Verified locally: `npx standard` clean on changed files, `http_api_spec` green, full `layout.spec.js` green and stable on **chromium and firefox**, and `curl` confirms the `LavinMQ-Version` header on `/api/*` responses and its absence on HTML.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ CTL_SOURCES := $(shell find src/lavinmqctl -name '*.cr' 2> /dev/null)
 VIEW_SOURCES := $(wildcard views/*.shtml)
 VIEW_TARGETS := $(patsubst views/%.shtml,static/%.html,$(VIEW_SOURCES))
 VIEW_PARTIALS := $(wildcard views/partials/*.shtml)
-VERSION := $(patsubst v%,%,$(or $(version),$(shell git describe --tags 2>/dev/null || shards version)))
 JS := static/js/lib/chunks/helpers.segment.js static/js/lib/chart.js static/js/lib/luxon.js static/js/lib/chartjs-adapter-luxon.esm.js static/js/lib/elements-8.2.0.js static/js/lib/elements-8.2.0.css $(wildcard static/js/*.js)
 CRYSTAL_FLAGS := --release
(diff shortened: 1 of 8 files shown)


kickster97 (approved):

snichme (commented):
Why not just do a `make clean views` to update the views, no need for an additional endpoint and less runtime logic

carlhoerberg:
Reload the page now if `http.js` fetches a new version.

PR #2124, opened 2026-06-26
Allow keys with no alg parameter when fetching jwks

hananasen:
[RFC 7517 4.4](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) states that this parameter is optional. Notice that we still must ensure its in the header. But we do that here:
https://github.com/cloudamqp/lavinmq/blob/01e9b7aea634f0abc8759cbb066ceb73e5af581a/src/lavinmq/auth/jwt/token_verifier.cr#L65-L66

Compiled and tested with entra, logs:
`Jun 26 06:49:57 dev-cold-kiwi-01 lavinmq[21671]:  lmq.jwks_fetcher Refreshed JWKS with 6 key(s), TTL=1.00:00:00`

Compiling main it outputs 0 key(s)

### DIFF
--- a/src/lavinmq/auth/jwt/jwks_fetcher.cr
+++ b/src/lavinmq/auth/jwt/jwks_fetcher.cr
@@ -163,8 +163,8 @@ module LavinMQ
             next unless key.kty == "RSA"
             # Skip keys not intended for signatures (RFC 7517 Section 4.2)
             next if key.use != "sig"
-            # Skip keys for other algorithms (RFC 7517 Section 4.4)
-            next if key.alg != "RS256"
+            # skip keys with unsupported algorithms, but allow keys with no alg (RFC 7517 Section 4.4)
+            next if (alg = key.alg) && alg != "RS256"
             next unless (n = key.n) && (e = key.e)
             kid = key.kid || "unknown-#{idx}"
             public_keys[kid] = to_pem(n, e)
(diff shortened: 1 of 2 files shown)


hananasen:
Entra omits it:
https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference

kickster97 (approved):
nice catch! a spec, and maybe docs that clarify what is accepted/allowed could accompany this

hananasen:
ok I added a spec now, I wonder if it might be too technical to add docs for this in the authentication.md. users shouldn't really have to care about it, its the IdPs

kickster97:
> ok I added a spec now, I wonder if it might be too technical to add docs for this in the authentication.md. users shouldn't really have to care about it, its the IdPs

yeah you're right, then i'm happy with this as is :)

src/lavinmq/auth/jwt/jwks_fetcher.cr
```diff
@@ -163,8 +163,8 @@ module LavinMQ
             next unless key.kty == "RSA"
             # Skip keys not intended for signatures (RFC 7517 Section 4.2)
             next if key.use != "sig"
-            # Skip keys for other algorithms (RFC 7517 Section 4.4)
-            next if key.alg != "RS256"
+            # skip keys with unsupported algorithms, but allow keys with no alg (RFC 7517 Section 4.4)
+            next if (alg = key.alg) && alg != "RS256"
```
carlhoerberg:
```suggestion
            next if key.alg.try { |alg| alg != "RS256" }
```

carlhoerberg (approved):

src/lavinmq/auth/jwt/jwks_fetcher.cr
hananasen: sure :)

PR #2126, opened 2026-06-26
MQTT topic permissions

kickster97:

Adds per-topic authorization for MQTT clients through permission groups.

A permission group belongs to a vhost. It has members and topic rules. A member is a username, so a user with many devices is one member. A rule has an identifier, a topic filter pattern, and read and write flags. A pattern uses MQTT subscription semantics, so `a/#` also matches `a`. A pattern can contain the `{client_id}` variable. Its value is the client id of the connection.

A topic is allowed only when a rule grants it. A vhost with no `mqtt_permissions.json` gets a group named `default` in memory. Its member is `*` and its rule `#` grants read and write, so an upgraded server keeps every topic open. The operator locks a vhost down when it deletes or narrows the `default` group through the API. The first change writes `mqtt_permissions.json`, and from then on the groups on disk decide.

The server drops a denied publish, but it sends the ack and keeps the connection open. The server accepts all subscribes and filters the messages when they enter the session. The server drops a denied will message. It logs each denial at the debug level.

A session stores the username of its last client in its `.metadata` file. A durable session keeps its rules after a restart while its device is offline. When a connection with a different username takes over a session, the server checks new messages against the new user. It still delivers the messages already in the queue.

The HTTP API is at `/api/mqtt/permission-groups`. Definitions export and import the groups under the `mqtt_permissions` key. An import into a vhost with no `mqtt_permissions.json` replaces the automatic `default` group. `lavinmqctl definitions export` emits the same `default` group for such a vhost.

The `mqtt.permission_check_enabled` option is unchanged. When it is set, the AMQP exchange check runs in front of the topic check. A client that fails the AMQP check is disconnected as before.

Follow-ups: a `{username}` variable, removal of a deleted user from all groups, an opt-in refusal of session takeover, a management UI, and a cache for the rule matcher.

Specs cover the matcher, the service, MQTT integration with takeover and restart, the HTTP API, and definitions. For a manual test, delete the `default` group with `DELETE /api/mqtt/permission-groups/:vhost/default`. Then PUT a group, a member, and a rule. Connect as that user. The server drops publishes and deliveries outside the rules, and the connection stays open.


docs/mqtt.md
```diff
@@ -104,6 +104,29 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
...
+## Topic permissions
+
+Enable with `topic_permissions = true` in the `[mqtt]` config section. When enabled, MQTT publish and receive are authorized per topic using permission groups, and the feature is default-deny: a connection may only publish to or receive on topics granted by a matched rule.
+
+Permission groups are global objects managed via `/api/permission-groups`. A group has a protocol (`mqtt`), an optional `apply_to_all` flag, a member list, and rules. Each rule is an MQTT topic filter with `read` and `write` booleans:
+
+```json
+{
+  "protocol": "mqtt",
+  "apply_to_all": false,
```
carlhoerberg: What does this mean? What's "all" and what's being applied?

docs/mqtt.md
```diff
@@ -104,6 +104,29 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
 
 When disabled, any authenticated MQTT client can publish and subscribe to any topic.
 
+## Topic permissions
+
+Enable with `topic_permissions = true` in the `[mqtt]` config section. When enabled, MQTT publish and receive are authorized per topic using permission groups, and the feature is default-deny: a connection may only publish to or receive on topics granted by a matched rule.
```
carlhoerberg: Are normal ACL rules still applied when this is in effect or does it replace it?

carlhoerberg: I think we could expand on what "topic permissions" are before going into configuration elements. Doesn't have to be long, but something that sells the feature, rather than assume everybody know what it is and how it works.

docs/mqtt.md
```diff
@@ -104,6 +104,29 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
...
+
+Permission groups are global objects managed via `/api/permission-groups`. A group has a protocol (`mqtt`), an optional `apply_to_all` flag, a member list, and rules. Each rule is an MQTT topic filter with `read` and `write` booleans:
```
carlhoerberg: So you have to enable a config under `[mqtt]` AND specify protocol when creating groups, is the idea that we in the future can have permission groups for other protocols?

docs/mqtt.md
```diff
@@ -104,6 +104,29 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
...
+
+```json
+{
+  "protocol": "mqtt",
+  "apply_to_all": false,
+  "members": ["alice"],
```
carlhoerberg: Maybe clarify somewhere that this is usernames, and not client id's?

docs/mqtt.md
```diff
@@ -104,6 +104,29 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
...
+  "rules": [
+    { "pattern": "chat/{client_id}/#", "read": true, "write": true }
+  ]
+}
+```
+
+Patterns use MQTT wildcards (`+`, `#`) and support the substitution variables `{username}` and `{client_id}`, expanded per connection.
+
+`{client_id}` only provides isolation when client ids are trustworthy. Set `client_id_validation = username` (see below) so a client cannot claim another client's id; otherwise `{client_id}` rules give no real isolation.
```
carlhoerberg: When is there a case where you want to use client_id rather than username? either client id can be anything, and anyone can impersonate it, or it is strictly the username?

docs/mqtt.md
kickster97: yeah.. unsure about this one, maybe just needs a name change.. but its a tool so that we can create a PermissionGroup that is automatically applied to all users.. something me and Jon discussed would be nice to have

docs/mqtt.md
kickster97: yes, that is the idea.. but should be able to do it more elegantly than needing to set the value twice

docs/mqtt.md
kickster97: hm.. no, this is a leftover from when i planned to include `username_prefix` in the username as `client_id` feature.. then a single user could have many connections with their own `client_ids`.. and you want per-device topic isolation, like `devices/{client_id}/#` gives each device its own subtree while `devices/{username}/#` would merge them all.

docs/mqtt.md
kickster97: No, they are replaced by topic rules if configured. However, on connect we still have a resource permission check, a user must have a permission entry for the vhost to establish an MQTT connection at all.
But input is ofc welcome!

src/lavinmq/mqtt/client.cr
```diff
@@ -185,7 +211,22 @@ module LavinMQ
...
+            if read.overlaps?(tf.topic)
+              allowed << tf
+              Protocol::SubAck::ReturnCode.from_int(tf.qos)
+            else
+              Log.debug { "Subscribe refused: user '#{user.name}' not allowed to read filter '#{tf.topic}'" }
+              Protocol::SubAck::ReturnCode::Failure
+            end
+          end
+          @broker.subscribe(self, allowed) unless allowed.empty?
+          send(Protocol::SubAck.new(return_codes, packet.packet_id))
```
spuun: I don't think we want this.
If a client is only allowed to read from `path/#` and tries to subscribe to `#` the subscribe will be denied, right? I think one should be able to subscribe broad but only receive what's within the clients permissions.

src/lavinmq/mqtt/client.cr
```diff
@@ -74,6 +77,21 @@ module LavinMQ
...
 
+      # Compiled topic permissions for this connection, recompiled lazily when
+      # the global permission group store revision changes.
+      def topic_permissions : TopicPermissions
+        store = @broker.permission_groups
+        rev = store.revision
+        tp = @topic_permissions
+        if tp.nil? || rev != @topic_permissions_revision
+          groups = store.for_mqtt_user(@user.name)
+          tp = TopicPermissions.build(groups, @user.name)
```
spuun: So each client has a copy of all permissions, but with e.g. username replaced? Won't this consume a lot of memory with many clients?

Wouldn't it be possible for clients to reference the same permissions objects? I'm thinking that one could create structures like how we do with `Segment` in the topic exchange.

src/lavinmq/auth/permission_group.cr
```diff
@@ -0,0 +1,36 @@
...
+        getter? write : Bool = false
+
+        def initialize(@pattern : String, @read : Bool = false, @write : Bool = false)
+        end
+      end
+
+      getter name : String
+      getter protocol : String
+      getter? apply_to_all : Bool
+      getter members : Array(String)
```
spuun: On disk I guess it may make sense to store usernames, but once in memory isn't a list of `User` more reasonable?

src/lavinmq/mqtt/session.cr
```diff
@@ -35,6 +36,7 @@ module LavinMQ
       @closed = Atomic(Bool).new(false)
       @deleted = false
       @client : MQTT::Client? = nil
+      property topic_read : TopicFilterSet? = nil
```
spuun: Why is this needed? Only for when no client is set?

src/lavinmq/mqtt/client.cr
kickster97: Subscribing to `#` isn't denied. `overlaps?` treats `#` as matching anything, so `overlaps?` returns true. Then delivery-time filtering trims what the client actually receives down to `path/#`. A subscribe only gets a SUBACK failure when the filter overlaps nothing in the read set, like reading `path/#` and subscribing to `other/#`, and I kept that so the client gets immediate feedback that the subscription can never deliver anything.

src/lavinmq/mqtt/client.cr
kickster97: static rules like `sensors/+/temp` that compile identically for every user and can be shared by reference the way you describe, and `{username}` rules like `devices/{username}/#` that expand differently per user, where sharing the compiled tree doesn't help since each result is unique. 

I'm hoping we can optimize for both cases, so I'm leaning toward compiling each group once and binding `{username}` at match time, so a connection just holds references to its groups plus the username with no per-connection tree at all.

src/lavinmq/auth/permission_group.cr
kickster97: maybe, but why? current solution really only references a username

src/lavinmq/mqtt/session.cr
kickster97: yes, so when a client is offline but the session is unclean we need a way to still be able to have the right topic permissions

docs/configuration.md
```diff
@@ -94,6 +94,7 @@ Not every setting takes effect on reload. The log level and TLS certificates are
 | `default_vhost` | — | — | String | `/` | Default vhost for MQTT connections |
 | `permission_check_enabled` | — | — | Bool | `false` | Enable MQTT permission checks |
```
carlhoerberg: when this is false, acl rules are ignored? i don't think the name plays well with `topic_permissions`, like do you have to enable both to get topic permission checks?

docs/mqtt.md
```diff
@@ -104,6 +105,35 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
...
+
+Enable with `topic_permissions = true` in the `[mqtt]` config section. When enabled, topic permissions replace the per-operation ACL check for MQTT publish and receive: authorization is decided entirely by the permission groups below, and the feature is default-deny, so a connection may only publish to or receive on topics granted by a matched rule. A user still needs a permission entry on the vhost to establish the connection in the first place.
+
+Because it is default-deny, define the permission groups before enabling the flag. Otherwise every MQTT publish and subscribe is denied until at least one matching group exists, and there is no administrator bypass. The broker logs a warning at startup when the flag is on but no permission groups are defined.
+
+Permission groups are global objects managed via `/api/permission-groups`. A group has a member list and a set of rules, where each rule is an MQTT topic filter with `read` and `write` booleans:
+
+```json
+{
+  "protocol": "mqtt",
```
carlhoerberg: is this premature generalization? what other protocol can we imagine where permission groups like this would fit? can `mqtt` be default if not explicitly written?

docs/mqtt.md
carlhoerberg: i think `members` should be used for that. now, if you say `true` here and still assign it `members`, how does that interact? 

`"members": "*"`, `"members": ["*"]` or something like that

docs/mqtt.md
```diff
@@ -104,6 +105,35 @@ By default, MQTT permission checks are disabled. When `permission_check_enabled`
 
 When disabled, any authenticated MQTT client can publish and subscribe to any topic.
 
+## Topic permissions
+
+Topic permissions give MQTT clients fine-grained, per-topic authorization. Instead of a single read/write grant on the whole MQTT exchange (the standard ACL described above), each client is authorized against a set of MQTT topic filters, so a user can be allowed to publish and subscribe only within their own topic subtree.
+
+Enable with `topic_permissions = true` in the `[mqtt]` config section. When enabled, topic permissions replace the per-operation ACL check for MQTT publish and receive: authorization is decided entirely by the permission groups below, and the feature is default-deny, so a connection may only publish to or receive on topics granted by a matched rule. A user still needs a permission entry on the vhost to establish the connection in the first place.
```
carlhoerberg: i'm wondering if `topic_permissions = true` shouldn't be the default? all other options seems  worse.

docs/configuration.md
kickster97: maybe we don't even need the topic_permission flag.. we could just let it "activate" when the first permission group is added..

docs/mqtt.md
kickster97: Defaulting to true would deny all MQTT publish/subscribe on upgrades and fresh installs until groups are defined, since the feature is default-deny. So instead of a flag we could let the feature activate when the first permission group is created. No groups would mean current behavior, any group means topic rules govern.

docs/mqtt.md
kickster97: fair point, only mqtt exists and the rule syntax is MQTT specific anyway, I'll drop it and we can add it with a default later without breaking stored files.

kickster97:
I think I have now addressed most of the comments..

The `topic_permissions` flag is removed entirely. Since we can detect whether permission groups are in use, the feature can activate itself when the first mqtt group is created and deactivate when the last one is deleted. This also means we don't have to touch or deprecate the `permission_check_enabled` flag we already have, it simply keeps applying when no groups exist, and it acts as a safety net: if it is enabled, deleting the last group falls back to the coarse ACL checks instead of open access. The tradeoff is that adding or removing a permission group toggles the feature, so this needs to be well documented (it is) and maybe some UI down the line.

`apply_to_all` is gone, a group applies to everyone with `"members": ["*"]`. protocol is optional and defaults to mqtt, and the leftover `{client_id}` patterns are cleaned out of the specs.

For the memory concern: each group is compiled once per store revision and shared by every connection it applies to. `{username}` rules are bound to the connection's username at match time, so a connection just holds references to its groups plus the username, no per-connection tree.

src/lavinmq/mqtt/client.cr
spuun: I wonder if we should remove this check completely? It's kind of unintuitive, and with topic permissions more or less unnecessary?

src/lavinmq/auth/permission_group.cr
```diff
@@ -0,0 +1,43 @@
...
+
+      getter name : String
+      getter protocol : String = "mqtt"
+      getter members : Array(String) = [] of String
+      getter rules : Array(Rule) = [] of Rule
+
+      def initialize(@name : String,
+                     @protocol : String = "mqtt",
+                     @members : Array(String) = [] of String,
+                     @rules : Array(Rule) = [] of Rule)
```
spuun:
```suggestion
                     @members = Array(String).new,
                     @rules = Array(Rule).new)
```

src/lavinmq/auth/permission_group.cr
```diff
@@ -0,0 +1,43 @@
...
+        getter? write : Bool = false
+
+        def initialize(@pattern : String, @read : Bool = false, @write : Bool = false)
+        end
+      end
+
+      getter name : String
+      getter protocol : String = "mqtt"
+      getter members : Array(String) = [] of String
+      getter rules : Array(Rule) = [] of Rule
```
spuun:
```suggestion
      getter members = Array(String).new
      getter rules = Array(Rule).new
```

src/lavinmq/auth/permission_service.cr
```diff
@@ -0,0 +1,154 @@
...
+        rules.each do |rule|
+          next unless write ? rule.write : rule.read
+          return true if MQTT::TopicRuleSegment.matches?(rule.chain, topic, client_id)
+        end
+        false
+      end
+
+      private def rebuild : Nil
+        by_member = Hash(String, Array(CompiledRule)).new
+        wildcard_rules = [] of CompiledRule
```
spuun: At first I thought wildcard rules meant a rule where the pattern contain a wildcard. Maybe we can find a better name?

```suggestion
        global_rules = [] of CompiledRule
```

src/lavinmq/mqtt/session.cr
```diff
@@ -35,6 +36,10 @@ module LavinMQ
       @closed = Atomic(Bool).new(false)
       @deleted = false
       @client : MQTT::Client? = nil
+      property permission_service : Auth::PermissionService? = nil
```
spuun: This will always be set, right? If so
```suggestion
      property! permission_service : Auth::PermissionService? = nil
```
to not have to nil-check it.

src/lavinmq/mqtt/session.cr
```diff
@@ -174,6 +180,9 @@ module LavinMQ
       end
 
       def publish(msg : Message) : Bool
+        if (permissions = @permission_service) && !permissions.can_read?(@client_id, msg.routing_key)
```
spuun: If the above change about `property!` is done, then
```suggestion
        return true if !permission_service.can_read?(@client_id, msg.routing_key)
```

src/lavinmq/http/controller/permission_groups.cr
spuun, line 1: What about making the API better and "complete" right away? Right now it's really cumbersome to work with.

Id like to be able to create an empty group, then add members and rules one at the time without having to read out the entire group, update it and put it back. 

Same for removing members and rules.

src/lavinmq/mqtt/topic_filter.cr
```diff
@@ -0,0 +1,28 @@
...
+      # Compiling walks one level per segment, so cap how deep a filter can be.
+      MAX_FILTER_SEPARATORS = 200
+
+      # A malformed filter would otherwise silently widen when compiled
+      # (e.g. `secret/#/x` behaving as `secret/#`).
+      def valid_filter?(filter : String) : Bool
+        return false if filter.empty?
+        levels = filter.split('/')
+        return false if levels.size - 1 > MAX_FILTER_SEPARATORS
+        last = levels.size - 1
```
spuun:
```suggestion
        last = levels.size - 1
        return false if last > MAX_FILTER_SEPARATORS
```

src/lavinmq/mqtt/permission_service.cr
spuun, line 1: `PermissionGroup` is protocol agnostic as it's designed to be able to support multiple protocols in the future, but `PermissionService` is MQTT specific but with generic names (`PermissionService`, `CompiledRule`, `Compiled`). I wonder if this could be improved without too much hassle.

src/lavinmq/auth/permission_service.cr
```diff
@@ -0,0 +1,154 @@
...
+                       @mqtt_in_use : Bool)
+        end
+      end
+
+      @save_lock = Mutex.new
+      @compiled : Compiled
+
+      def initialize(@data_dir : String, @replicator : Clustering::Replicator?)
+        @groups = Hash(String, PermissionGroup).new
+        @compiled = Compiled.new(Hash(String, Array(CompiledRule)).new, [] of CompiledRule, false)
```
spuun:
```suggestion
        @compiled = Compiled.new(Hash(String, Array(CompiledRule)).new, Array(CompiledRule).new , false)
```

src/lavinmq/auth/permission_service.cr
```diff
@@ -0,0 +1,154 @@
...
+
+      record CompiledRule,
+        chain : MQTT::TopicRuleSegment,
+        read : Bool,
+        write : Bool
+
+      class Compiled
+        getter by_member : Hash(String, Array(CompiledRule))
+        getter wildcard_rules : Array(CompiledRule)
+        getter? mqtt_in_use : Bool
```
spuun: I think this is unclear. Basically it means that `Compiled` is "non-empty" (as in it has at least one rule?), right?
So what about renaming it to `empty?` (or `has_rules?`) and "calculate it" from `@by_member` and `@wildcard_rules` in the constructor and remove the `@mqtt_in_use` argument?

`PermissionService#mqtt_in_use?` could just return `!@compiled.empty?`

src/lavinmq/http/controller/permission_groups.cr
```diff
@@ -0,0 +1,57 @@
...
+          refuse_unless_administrator(context, user(context))
+          name = params["name"]
+          body = parse_body(context)
+          begin
+            protocol = (raw = present(body, "protocol")) ? String.from_json(raw.to_json) : "mqtt"
+            members = (raw = present(body, "members")) ? Array(String).from_json(raw.to_json) : [] of String
+            rules = (raw = present(body, "rules")) ? Array(Auth::PermissionGroup::Rule).from_json(raw.to_json) : [] of Auth::PermissionGroup::Rule
+            is_update = @server.permission_service[name]?
+            @server.permission_service.put(Auth::PermissionGroup.new(name, protocol, members, rules))
+            context.response.status_code = is_update ? 204 : 201
```
spuun: We should do this in existing code too, so one doesn't have to remember http codes
```suggestion
            context.response.status = is_update ? :no_content : :created
```
(or use `HTTP::Status::NO_CONTENT` instead of symbols)

src/lavinmq/http/controller/permission_groups.cr
```diff
@@ -0,0 +1,57 @@
...
+            bad_request(context, "Invalid permission group: #{ex.message}")
+          end
+          context
+        end
+
+        delete "/api/permission-groups/:name" do |context, params|
+          refuse_unless_administrator(context, user(context))
+          group = @server.permission_service.delete(params["name"])
+          not_found(context) unless group
+          context.response.status_code = 204
```
spuun:
```suggestion
          context.response.status = :no_content
```

spuun (commented):
I'm just adding comments for now, about to do some manual testing too.

spuun:
Two things after discussions:
- `PermissionService` should be per vhost
- Let's take a step back and make this mqtt only

kickster97:
Went through all the review comments and made some style fixes.

Topic permissions are now moved into the MQTT namespace and live on vhost level instead of server level. The protocol field is gone, sessions get the permission service at construction so it's no longer nilable, and `permission_check_enabled` is removed in favor of topic permissions.

The API is granular now: create a group by name, then add or remove members and rules through their own endpoints. Rules have an identifier so they can be deleted by path. I guess we could discuss adding an update rule endpoint as well, but imo is future work. 

ping @spuun for visibility

src/lavinmq/http/controller/permission_groups.cr
kickster97: the :no_content / :created symbols don't compile through a ternary, so I used the HTTP::Status constants.

src/lavinmq/http/controller/permission_groups.cr
```diff
@@ -0,0 +1,131 @@
+require "../controller"
+
+module LavinMQ
+  module HTTP
+    class PermissionGroupsController < Controller
+      # ameba:disable Metrics/CyclomaticComplexity
+      private def register_routes
+        get "/api/permission-groups" do |context, _params|
```
spuun: Maybe we should prefix all routes with `mqtt` since it's a mqtt thing only?
```suggestion
        get "/api/mqtt/permission-groups" do |context, _params|
```

src/lavinmq/http/controller/permission_groups.cr
```diff
@@ -0,0 +1,131 @@
...
+          refuse_unless_administrator(context, user(context))
+          JSON.build(context.response) do |json|
+            json.array do
+              @server.vhosts.each_value do |vhost|
+                vhost.mqtt_permission_service.values.each(&.to_json(json))
```
spuun: Dunno if we should list all groups in a flat list with full content here?

Would it make sense to group them on vhost and just list group names? Maybe also number of rules and members per group.

src/lavinmq/http/controller/permission_groups.cr
```diff
@@ -0,0 +1,131 @@
...
+              end
+            end
+          end
+          context
+        end
+
+        get "/api/permission-groups/:vhost" do |context, params|
+          refuse_unless_administrator(context, user(context))
+          with_vhost(context, params) do |vhost|
+            vhost.mqtt_permission_service.values.to_json(context.response)
```
spuun: Same as above: name, number of members, number of rules

src/lavinmq/http/controller/permission_groups.cr
spuun, line 1: All routes requires administrator, but since this will be a per vhost, maybe we can make all routes including `:vhost` require some other tag? Policy maker (not a perfect fit, but what else do we have)?

spuun, line 1: When we do `group.to_json` we kind of assume that we always want to send all public fields in the group. What if we introduce more fields that isn't supposed to be exposed?

I think that we should use "views" (and not only here, but in all controllers) to control the data. I.e. introduce e.g. `struct PermissionGroupView` that accepts a `PermissionGroup` in the constructor and has a `#to_json` that renders what we want. This will introduce a few structs for this controller, but I think it's worth it.

src/lavinmq/mqtt/topic_rule_segment.cr
```diff
@@ -0,0 +1,87 @@
...
+      abstract def match?(it : BytesTokenIterator, client_id : String) : Bool
+
+      def self.compile(pattern : String) : TopicRuleSegment?
+        return unless TopicFilter.valid_filter?(pattern)
+        chain = nil.as(TopicRuleSegment?)
+        pattern.split('/').reverse_each do |level|
+          chain = case level
+                  when "#"           then HashRuleSegment.new
+                  when "+"           then PlusRuleSegment.new(chain)
+                  when "{client_id}" then ClientIdRuleSegment.new(chain)
```
spuun, line 19: Note for future: We can probably make `ClientIdRuleSegment` more dynamic in the future, so the client id can be part of the segment, e.g. `abc{client_id}`.

spuun (commented):
I wonder if we'll see any issues with always rewriting the json file, but we can handle that in upcoming PRs. I'm thinking that having a WAL file for could be useful.

src/lavinmq/http/controller/permission_groups.cr
kickster97: hm, because a server admin is too broad of a requirement for vhost? 

for me, setting policymaker instead breaks consistency.. 
policy maker controls behaviour objects, like policy, shovel, parameters etc.. 
admin controls access-control objects, like users and permissions... 

i guess we need to have a feel for if a mqtt permission group is more like a policy or like a permission 

another thought could be to introduce a new tag, MQTTPermissionManager or something..

src/lavinmq/mqtt/permission_service.cr
```diff
@@ -0,0 +1,174 @@
...
+      def size : Int32
+        @groups.size
+      end
+
+      def each_value(& : PermissionGroup ->) : Nil
+        @groups.each_value { |group| yield group }
+      end
+
+      def in_use? : Bool
+        !@compiled.empty?
```
snichme, line 61: If I understand this correctly, if `@compiled` is empty `can_read` and `can_write` will return true which means that if we have no rules we allow all? 

What happens if we have some rules to enforce permissons on a vhost and then the users delete all rules, wont everyone get access then?

src/lavinmq/mqtt/permission_service.cr
```diff
@@ -0,0 +1,177 @@
...
+      end
+
+      def put(group : PermissionGroup, save = true) : PermissionGroup
```
snichme, line 64: I think we should remove the argument `save` never call `save!` from within this method. A nicer API would be to call `s.put(group1); s.put(group2); s.save!` so the caller has to call the `save!` method. 

Same api as many buffered interfaces has, which I feel this reassemble.

kickster97: The save argument is gone since #2232, `put` and `delete` always write `mqtt_permissions.json` before they change the active groups. The definitions import already commits a whole vhost as one batch. `save!` is public only because the vhost close uses it to persist the automatic default group.

snichme (approved):

PR #2127, opened 2026-06-29
Make management UI OAuth2 scope configurable

### LINKED ISSUE #2125: Make OAuth2 authorization scope configurable
The scope requested during the OAuth2/OIDC management UI login flow is currently hardcoded to "openid profile" in src/lavinmq/http/controller/oauth.cr. This should be configurable via the [oauth] config section so operators can adapt the login flow to their identity provider.

Background

When a user logs in to the management UI via OAuth, LavinMQ builds an authorization request and sends a fixed scope=openid profile parameter to the IdP's authorization endpoint:

https://github.com/cloudamqp/lavinmq/blob/01e9b7aea634f0abc8759cbb066ceb73e5af581a/src/lavinmq/http/controller/oauth.cr#L85

openid profile is a reasonable default, but it does not work for every deployment. For instance, Entra ID is stricter than other Idps in which you can get around this.

Suggest to make them configurable via a setting that can be augmented to the scopes paramameter. RabbitMQ uses `oauth_scopes` for this, but that can be confused with the `additional_scopes_claims`. 
`p.add "scope", "openid profile" + <oauth scopes>`

kickster97:
The scope requested during the management UI OAuth2/OIDC login flow was hardcoded to "openid profile" in the authorize controller, which does not work for every identity provider (Entra ID, for example, is stricter about requested scopes).

This adds an [oauth] mgmt_scopes setting that replaces the requested scope verbatim. It defaults to "openid profile" so existing deployments are unaffected, and operators can set it to whatever their IdP requires. The name avoids confusion with the existing additional_scopes_keys and scope_prefix settings, which apply to parsing scopes out of a received token rather than the outbound authorization request.

Closes #2125

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -68,6 +68,7 @@ jwks_cache_ttl = 3600
 | `issuer` | (required) | OIDC issuer URL. Used to discover JWKS endpoint. |
 | `client_id` | (none) | OAuth2 client ID. Required to enable management UI SSO (see below) |
 | `mgmt_base_url` | (none) | Public base URL of the management UI. Enables SSO login when set together with `issuer` and `client_id` |
+| `mgmt_scopes` | `openid profile` | Space-separated scopes requested in the management UI SSO authorization request. Override to match the requirements of your identity provider |
 | `resource_server_id` | (none) | Resource server identifier |
 | `preferred_username_claims` | `sub,client_id` | JWT claims to extract the username from (tried in order) |
 | `additional_scopes_keys` | `[]` | Additional JWT claims to check for scope strings |
--- a/src/lavinmq/http/controller/oauth.cr
+++ b/src/lavinmq/http/controller/oauth.cr
@@ -82,7 +82,7 @@ module LavinMQ
           p.add "client_id", client_id
           p.add "redirect_uri", build_redirect_uri
           p.add "response_type", "code"
-          p.add "scope", "openid profile"
+          p.add "scope", config.oauth_mgmt_scopes
           p.add "code_challenge", challenge
           p.add "code_challenge_method", "S256"
           p.add "state", state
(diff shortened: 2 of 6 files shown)


docs/authentication.md
```diff
@@ -68,6 +68,7 @@ jwks_cache_ttl = 3600
 | `issuer` | (required) | OIDC issuer URL. Used to discover JWKS endpoint. |
 | `client_id` | (none) | OAuth2 client ID. Required to enable management UI SSO (see below) |
 | `mgmt_base_url` | (none) | Public base URL of the management UI. Enables SSO login when set together with `issuer` and `client_id` |
+| `mgmt_scopes` | `openid profile` | Space-separated scopes requested in the management UI SSO authorization request. Override to match the requirements of your identity provider |
```
hananasen, line 71: love it!

src/lavinmq/http/controller/oauth.cr
```diff
@@ -82,7 +82,7 @@ module LavinMQ
           p.add "client_id", client_id
           p.add "redirect_uri", build_redirect_uri
           p.add "response_type", "code"
-          p.add "scope", "openid profile"
+          p.add "scope", config.oauth_mgmt_scopes
```
hananasen, line 85: nice! I like that we don't just augment the openid profile as well and that the user must explicitly set openid profile <my-scopes>, I think this looks great.

hananasen (approved):
awesome 😍

I rebased this work to the work in https://github.com/cloudamqp/lavinmq/pull/2124 and tested on a lavinmq server and I am able to use entra to this 🎉

kickster97:
@carlhoerberg we said yesterday that we were going to work through a few more things before merging this, did not mark this as draft so thats my bad.

carlhoerberg:
Oh, Sorry! Yes, ready and approved was merge ready for me. You can either revert it, or open a new PR with the additions. Does it work as is?

kickster97:
> Oh, Sorry! Yes, ready and approved was merge ready for me. You can either revert it, or open a new PR with the additions. Does it work as is?

Yes, Hanna did some more testing and seems like we are all good and no more changes needed, so its no worries :)

PR #2128, opened 2026-06-29
Shovel error handling improvments

### LINKED ISSUE #1357: Overflow mode `reject-publish` not respected by shovel

snichme:

Reworks how shovels settle messages and fixes #1357 (reject-publish nack was acking/dropping the upstream message) and #1612 (HTTP errors dropped the message). A destination now classifies each attempt into one of four dispositions — Confirmed, Retry, Reject, Abort — via a pure classify, and the Runner is the single place mapping each to a source action: ack, requeue + backoff, reject-without-requeue (DLX), or keep-and-error-out.

How each destination handles errors and what happens on each failure is documented here:  https://github.com/cloudamqp/lavinmq/blob/shovel/failure-handle-improvments/docs/shovel-delivery-outcomes.md

Specs cover the new code. 

There are scripts in the `extras/` folder that lets you setup shovels and test them out quickly: https://github.com/cloudamqp/lavinmq/pull/2128/changes/4157a2d23f07b29aebe1c7135af93215ffae854f

Using the script from the issue #1357 https://github.com/user-attachments/files/22912898/lmq_reject-overflow.sh

### DIFF
--- a/src/lavinmq/shovel/amqp_source.cr
+++ b/src/lavinmq/shovel/amqp_source.cr
@@ -58,7 +58,9 @@ module LavinMQ
       def stop
         # If we have any outstanding messages when closing, ack them first.
         @ch.try &.basic_cancel(@tag, no_wait: true)
-        @last_unacked.try { |delivery_tag| ack(delivery_tag, batch: false) }
+        @settle.synchronize do
+          @last_unacked.try { |delivery_tag| ack_locked(delivery_tag, batch: false) }
+        end
         @conn.try &.close(no_wait: false)
         @q = nil
         @ch = nil
@@ -73,20 +75,29 @@ module LavinMQ
       end
 
       @done = WaitGroup.new(1)
+      # Serializes settlement (ack/reject/timeout-flush/stop). @last_unacked is
+      # written from the confirm fiber and the ack-timeout fiber, which run on
+      # separate threads under -Dpreview_mt; the read-decide-emit-update must be
+      # indivisible, or a flush could double-settle a tag.
+      @settle = Mutex.new
 
       def ack(delivery_tag, batch = true)
-        if ch = @ch
-          return if ch.closed?
+        @settle.synchronize { ack_locked(delivery_tag, batch) }
+      end
 
-          # We batch ack for faster shovel
-          batch_full = delivery_tag % ack_batch_size == 0
-          if !batch || batch_full || at_end?(delivery_tag)
-            @last_unacked = nil
-            ch.basic_ack(delivery_tag, multiple: true)
-            @done.done if at_end?(delivery_tag)
-          else
-            @last_unacked = delivery_tag
-          end
+      private def ack_locked(delivery_tag, batch)
+        ch = @ch
+        return unless ch
+        return if ch.closed?
+
+        # We batch ack for faster shovel
+        batch_full = delivery_tag % ack_batch_size == 0
+        if !batch || batch_full || at_end?(delivery_tag)
+          @last_unacked = nil
+          ch.basic_ack(delivery_tag, multiple: true)
+          @done.done if at_end?(delivery_tag)
+        else
+          @last_unacked = delivery_tag
         end
       end
 
@@ -94,16 +105,20 @@ module LavinMQ
       # throughput, so before rejecting tag T we must flush any pending batched
       # ack of earlier tags — otherwise a later multiple-ack would settle T too.
       def reject(delivery_tag, requeue)
-        if ch = @ch
-          return if ch.closed?
-          if last = @last_unacked
-            if last < delivery_tag
-              @last_unacked = nil
-              ch.basic_ack(last, multiple: true)
-            end
+        @settle.synchronize { reject_locked(delivery_tag, requeue) }
+      end
+
+      private def reject_locked(delivery_tag, requeue)
+        ch = @ch
+        return unless ch
+        return if ch.closed?
+        if last = @last_unacked
+          if last < delivery_tag
+            @last_unacked = nil
+            ch.basic_ack(last, multiple: true)
           end
-          ch.basic_reject(delivery_tag, requeue: requeue)
         end
+        ch.basic_reject(delivery_tag, requeue: requeue)
       end
 
       def started? : Bool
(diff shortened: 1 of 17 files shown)


carlhoerberg (commented):
Can we split the outcome handling and the UI/stats improvements to two different PRs? (They can be staged, one targets the other instead of both targeting main)

A lot of new, not so generic, files in `extras/`. 

Not sure I love the `on_outcome` proc, I wish it was just a method instead.

snichme:
> Can we split the outcome handling and the UI/stats improvements to two different PRs? (They can be staged, one targets the other instead of both targeting main)

Yes, thats good 

The files in `extras/` will be removed, I kept them if anyone want to test this out 

> Not sure I love the `on_outcome` proc, I wish it was just a method instead.

It was a design to decouple source from destination, before we had `source` as an argument to `Destination#push` but it's not 100% clear where it comes from.

snichme:
@carlhoerberg `on_outcome` is removed in favor of an interface now instead.

snichme:
Synced with @lukas8219 that this PR covers changes in #1667 as well.

kickster97:
Now you create a full new .md file for `Shovel Delivery Outcomes`, maybe we could tighten it and make it a part of the standard shovel documentation? Unless you feel like delivery outcomes are that much of a stand alone feature? 

Another option might be to start structuring the docs into component-folders to avoid it just being random files in `/docs`.. for example we could have a `/docs/shovels` folder, that contain one shovels.md that docs the base of the feature and then one .md for each individual feature in a shovel.

src/lavinmq/shovel/http_destination.cr
```diff
@@ -45,12 +62,89 @@ module LavinMQ
...
+        reused = @reused
+        resp = c.post(path, headers: headers, body: body)
+        @reused = true
+        resp
+      rescue ex : IO::Error | OpenSSL::SSL::Error
+        c.close
+        @reused = false
+        if reused && stale_connection?(ex)
+          Log.debug { "shovel=#{@name} stale keep-alive (#{ex.message}), retrying on a fresh connection" }
+          post(c, path, headers, body)
```
kickster97, line 106: A broken connection does not prove that the server failed to process the request. Repeating the post could send a duplicate? If that is ok then its no problem, but flagging just in case

src/lavinmq/shovel/multi_destination.cr
```diff
@@ -2,35 +2,164 @@ require "./destination"
...
+      # with backoff.
       def start
         return if started?
-        next_dest = @destinations.sample
-        return unless next_dest
-        next_dest.start
-        @current_dest = next_dest
+        reset_streaks
+        error = nil
+        each_index_from(@index) do |i|
```
kickster97, line 43: This is a nice touch.. it enables you to have a prefferred destination.. but i wonder if it might make things messy for someone who has built their system on it being picked at random? 

say we have 30 shovels and 3 destinations, before that load would be somewhat balanced, now it will all go to the first destination 

what do you think?

src/lavinmq/shovel/runner.cr
```diff
@@ -45,34 +62,164 @@ module LavinMQ
           @retries = 0
           @source.each do |msg|
             @message_count += 1
-            @destination.push(msg, @source)
+            # Paused/terminated/aborted: start no new delivery. An already
+            # in-flight push can't be interrupted, but we don't begin another one.
+            next if should_stop_loop?(run_generation)
+            next if abort_if_unusable
+            backoff_if_failing
```
kickster97, line 69: Hm, during `backoff_if_failing`, a user could pause and resume their shovel, which opens a new source channel. But the old run still holds `msg.delivery_tag` from the previous channel.

When it wakes up, it calls `@destination.push(msg)` without checking `run_generation` again. That delivery tag could now belong to a different message on the new channel, so `@source.ack(delivery_tag)` could remove a message that hasn't been confirmed at the destination.

| Delivery | Old channel | New channel |
  |---|---|---|
  | A, first attempt | Tag 1 | |
  | A, retried | Tag 2 | |
  | A, delivered again after resume | | Tag 1 |
  | B, next message | | Tag 2 |

src/lavinmq/shovel/runner.cr
```diff
@@ -87,16 +234,24 @@ module LavinMQ
...
-          error:         @error,
-          message_count: @message_count,
+          name:                 @name,
+          vhost:                @vhost.name,
+          state:                @state.to_s,
+          error:                @error,
+          message_count:        @message_count,
+          confirmed:            @confirmed_total.get,
+          retried:              @retried_total.get,
+          dead_lettered:        @rejected_total.get,
```
kickster97, line 244: Why not call this `rejected`?
```suggestion
          rejected:        @rejected_total.get,
```

kickster97 (commented):
Have not gotten through everything yet, but some comments in the mean time

src/lavinmq/shovel/amqp_source.cr
```diff
@@ -58,38 +74,127 @@ module LavinMQ
...
           end
+        elsif delivery_tag > @frontier
+          @settled_above << delivery_tag
         end
       end
 
+      # Ack everything settled so far in one cumulative ack. A flush, not a
+      # settlement: each tag was counted when its ack or reject came in.
+      private def flush_ack(ch)
+        return if @frontier <= @flushed
```
kickster97, line 149: Not sure I grasp the counters 100%, but should we compare `@frontier` with `@flushed` if `@frontier` can include rejected deliveries? Wouldn't `flush_ack` then also ack deliveries we've already rejected?

kickster97:
Overall i think this is a good design, just the comments above that might need attention. Otherwise i would say approve!

src/lavinmq/shovel/http_destination.cr
snichme: Yes it will but there is no way to tell if the request has been handled or not if we don't get a reply so sending it again is the safest way to ensure no message loss. The receiver then needs to handle possible duplicates. But it's a good thing to flag for

src/lavinmq/shovel/multi_destination.cr
snichme: Agree, using just random is easier and less state to worry about. It was like that before so I will restore.

src/lavinmq/shovel/runner.cr
snichme: Good catch, fixed now.

kickster97 (approved):

PR #2129, opened 2026-06-29
Apply SNI changes on config reload

hananasen:

setup_sni_callbacks has an early return if the sni_manager is empty. This is the case when you bootstrap a lavinmq server with no sni-sections in the lavinmq.ini, which is fine. However, if a user wishes to add an sni-section, they have to restart lavinmq for that change to take effect.

On reload, reload_tls_context will reload the sni managers hosts and exist in memory, but since the callback is not installed for the listeners, they will still serve the default certificate.

I added https://github.com/cloudamqp/lavinmq/pull/2033 first but was cleaner to create a new PR rather than rebasing this to main.

Dropped the refactoring of reloading sni contexts, as that changed the dynamic behavior.

Specs.

Manually:

create a certificate and key,
boot lavinmq with no sni host section, 
send SIGHUP to the lavinmq process after adding an sni host section, for example this config

```
[main]
log_level = debug
data_dir = ./tmp/data
tls_cert = /tmp/default.pem
tls_key  = /tmp/default.key

[amqp]
tls_port = 5671

[mgmt]
tls_port = 15671

[sni:broker-a.local]
tls_cert = ./tmp/a.pem
tls_key  = ./tmp/a.key
```

and this should be the result of the openssl command:
`echo | openssl s_client -connect localhost:15671 -servername broker-a.local 2>/dev/null | openssl x509 -noout -subject subject=CN=broker-a.local
`

before this change it shows the default certificate.

### DIFF
--- a/spec/sni_spec.cr
+++ b/spec/sni_spec.cr
@@ -1,5 +1,14 @@
 require "./spec_helper"
 require "../src/stdlib/openssl_on_server_name"
+require "../src/lavinmq/launcher"
+
+class LavinMQ::Launcher
+  getter amqp_tls_context
+
+  def reload_for_spec
+    reload_config
+  end
+end
 
 describe LavinMQ::SNIHost do
   it "creates TLS contexts with default settings" do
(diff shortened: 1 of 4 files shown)


spec/sni_spec.cr
```diff
@@ -455,3 +464,70 @@ describe "SNI end-to-end" do
     server_done.receive
   end
 end
+
+describe LavinMQ::Launcher do
+  it "serves a per-host certificate for an SNI host added on config reload" do
```
viktorerlingsson: I think this spec could live in `config_spec.cr` instead, and that way hopefully re-use some of the helpers there

src/lavinmq/launcher.cr
```diff
@@ -338,10 +335,6 @@ module LavinMQ
         Log.warn { "Disabling TLS requires a restart to take effect" }
         return
       end
-      if (amqp_ctx = @amqp_tls_context) && !amqp_ctx.sni_callback? && !@config.sni_manager.empty?
```
viktorerlingsson, line 341: I don't think we're using `OpenSSL::SSL::Context::Server::sni_callback?` anywhere after removing this, so we could remove that method from `openssl_on_server_name.cr` as well!

viktorerlingsson (approved):

carlhoerberg (approved):

PR #2135, opened 2026-07-01
[CSS] Remove zIndex of table links

ThomasSarlin:
Removes zIndex of `table tbody a` as it overlaps other UI elements:
image

:eyes:

image

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -1173,7 +1173,7 @@ footer .logo {
       margin-top: 1rem;
       border-top: 1px solid var(--color-border-separator);
       a {
-        background-image: var(--icon-http-api); 
+        background-image: var(--icon-http-api);
         margin-top: 0.5rem;
       }
     }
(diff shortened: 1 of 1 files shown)


kickster97:
does this have any other implications, or does this only touch the value for the columns menu?

ThomasSarlin:
> does this have any other implications, or does this only touch the value for the columns menu?

No, this does not affect the columns menu, it affects links inside tables which due to the high z-index value proceeds to be displayed on top of the menu. I do not know the reason of why setting a z-value to the table links was deemed necessary, but I cannot find a reason why there should be one.

ThomasSarlin:
Oh found now: https://github.com/cloudamqp/lavinmq/commit/83095cd1
It was added when a hover effect was added to links, the whole hover state has been dropped in recent design changes so this is now outdated and can be safely removed in my opinion. 

Edit:
Also asked claude nicely to sanity check me:
  _**Verdict**: z-index:1000 went dead the moment d4d3ad3d killed the hover-reveal — its whole job was lifting expanded hover text above neighbor cells, and that mechanism no longer exists. Your
  removal in 9e79944a = pure cleanup, no regression. Original reason valid in 2021, obsolete now._

viktorerlingsson (approved):

PR #2139, opened 2026-07-02
Report actual MQTT protocol version in connection details

walro:

The MQTT CONNECT handler accepts both protocol level 3 (MQTT 3.1, "MQIsdp") and level 4 (MQTT 3.1.1, "MQTT"), but the connection details hardcoded the protocol as "MQTT 3.1.1", mislabelling 3.1 clients.

Thread the negotiated protocol version through to MQTT::Client as a ProtocolVersion enum and report "MQTT 3.1" for level 3 and "MQTT 3.1.1" for level 4. Also preserve the version when the CONNECT packet is rebuilt for an empty client id, which previously reset it to level 4.

Specs and manually:

```
$ mosquitto_sub -u guest -P guest -t test -V mqttv31
$ curl -s http://guest:guest@localhost:15672/api/connections | jq '.[].protocol'
"MQTT 3.1"

$ mosquitto_sub -u guest -P guest -t test -V mqttv311
$ curl -s http://guest:guest@localhost:15672/api/connections | jq '.[].protocol'
"MQTT 3.1.1"
```

### DIFF
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -49,6 +49,7 @@ module LavinMQ
           user,
           self,
           packet.client_id,
+          ProtocolVersion.from_value(packet.version),
           packet.clean_session?,
           packet.keepalive,
           packet.will)
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -11,6 +11,13 @@ require "../stats"
 
 module LavinMQ
   module MQTT
+    # Protocol level from the CONNECT packet:
+    # level 3 is MQTT 3.1 (MQIsdp), level 4 is MQTT 3.1.1 (MQTT).
+    enum ProtocolVersion : UInt8
+      V3_1   = 3
+      V3_1_1 = 4
+    end
+
     class Client < LavinMQ::Client
       include Stats
       include SortableJSON
@@ -20,6 +27,7 @@ module LavinMQ
       @connected_at = RoughTime.unix_ms
       @channels = Hash(UInt16, Client::Channel).new
       @session : MQTT::Session?
+      @protocol : String
       rate_stats({"send_oct", "recv_oct"})
       Log = LavinMQ::Log.for "mqtt.client"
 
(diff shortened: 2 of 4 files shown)


src/lavinmq/mqtt/broker.cr
```diff
@@ -49,6 +49,7 @@ module LavinMQ
           user,
           self,
           packet.client_id,
+          ProtocolVersion.from_value(packet.version),
```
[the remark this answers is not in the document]
walro: I don't think that is possible, see: https://github.com/84codes/mqtt-protocol.cr/blob/739428e1cd4ea9dab1bb6cf101fb18ea54a099e5/src/mqtt/protocol/packets/connect.cr#L46-L56 so we can trust his value to be set here.

I guess the problem could arise if we add MQTT 5 support in the protocol shard and not update here, but OTOH what should we default to then? MQTT 3.1.1 is hardly correct either :)

src/lavinmq/mqtt/client.cr
```diff
@@ -11,6 +11,13 @@ require "../stats"
 
 module LavinMQ
   module MQTT
+    # Protocol level from the CONNECT packet:
+    # level 3 is MQTT 3.1 (MQIsdp), level 4 is MQTT 3.1.1 (MQTT).
+    enum ProtocolVersion : UInt8
```
walro, line 16: Not needed if we would do `case/when` instead if `case/in`, not sure it's worth this `enum` but I liked the added compiler safety we get and the typing instead of UInt8.

carlhoerberg: it's good

carlhoerberg (approved):

snichme (approved):

viktorerlingsson (approved):
I've done something pretty similar to this for `3.1`/`3.1.1` during my work for implementing mqtt5 (but mostly in `mqtt-protocol.cr`). But I say we merge this now and I will adapt my PR to this 👍

src/lavinmq/mqtt/client.cr
```diff
@@ -49,9 +57,14 @@ module LavinMQ
...
                      @broker : MQTT::Broker,
                      @client_id : String,
+                     protocol_version : ProtocolVersion,
                      @clean_session : Bool = false,
                      @keepalive : UInt16 = 30,
                      @will : Protocol::Will? = nil)
+        @protocol = case protocol_version
+                    in .v3_1?   then "MQTT 3.1"
+                    in .v3_1_1? then "MQTT 3.1.1"
+                    end
```
spuun: What about adding 
```crystal
enum ProtocolVersion : UInt8 
  V3_1   = 3
  V3_1_1 = 4

  def name
    case self
    in .v3_1?   then "MQTT 3.1"
    in .v3_1_1? then "MQTT 3.1.1"
    end
  end
end
```

and change this to 

```suggestion
        @protocol = protocol_version.name
```

walro: Ah enums can have methods, stuff you learn. Thanks for the help Christina!

PR #2142, opened 2026-07-03
Add non-destructive peek endpoint for queue messages

oskgu360:
Adds POST /api/queues/:vhost/:name/peek that reads messages directly from disk without consuming them via AMQP. Supports offset/count windowing and a state parameter selecting either the ready messages (requeued first, then from the read head, in delivery order) or the unacked messages currently held by clients. Works for both AMQP queues (including priority queues) and MQTT sessions.

The message store lock is only held while copying one message at a time (PeekedMessage, a heap-only copy of body and headers, which otherwise are slices into the mmap:ed segments that can be unmapped by a concurrent ack). Serialization to the HTTP response happens outside the lock, so a slow client cannot stall publishes or deliveries on the queue.

- MessageStore#peek_requeued snapshots requeued SegmentPositions, MessageStore#peek_step scans segments from the read head in bounded steps (max 100 positions per lock hold), skipping acked positions, mirroring shift? order.
- The ready-message walk lives in the Peekable module, shared by AMQP::Queue and MQTT::Session (which stores messages in the same MessageStore). It walks requeued then segments with a cursor, yielding the store lock between copies.
- Queue#peek_unacked discovers channels through the vhost's connections, so it covers both consumer deliveries and basic_get:s. Session#peek_unacked snapshots the packet-id => SegmentPosition map of in-flight QoS 1 messages. (QoS 0 messages to a connected client never touch the store, so peek only shows pending/in-flight messages.)
- Payload copies are capped at 1 MiB per message (client truncate can lower it; the UI sends 50 kB).
- PriorityMessageStore delegates peek to its substores, highest priority first, tracked by the store index in the PeekStep cursor.
- Stream queues are rejected (they have /stream already)
- In the UI, peek is a mode in the existing Get messages card; the peek-specific fields (offset, state) appear when the Peek mode is selected. Results show a state badge (ready / unacked).

localhost_15672_queue (4)

### DIFF
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -377,7 +377,15 @@ module LavinMQ
       def peek_unacked(offset : Int32, count : Int32, max_body : Int32, &block : PeekedMessage -> Nil) : Nil
         return if count <= 0
 
-        sps = @msg_store_lock.synchronize { @unacked.values }
+        max = offset.to_i64 + count
+        sps = Array(SegmentPosition).new
+        @msg_store_lock.synchronize do
+          @unacked.each_value do |sp|
+            break if sps.size >= max
+            sps << sp
+          end
+        end
+
         yielded = 0
         sps.each.skip(offset).each do |sp|
           break if yielded >= count
--- a/src/lavinmq/mqtt/session.cr
+++ b/src/lavinmq/mqtt/session.cr
@@ -136,15 +136,14 @@ module LavinMQ
         return if closed?
         @last_get_time = RoughTime.instant
 
-        unless clean_session?
-          @msg_store_lock.synchronize do
+        @msg_store_lock.synchronize do
+          unless clean_session?
             @unacked.values.each do |sp|
               @msg_store.requeue(sp)
             end
           end
+          @unacked.clear
         end
-
-        @unacked.clear
         @unacked_count.set(0, :release)
         @unacked_bytesize.set(0, :release)
         @has_capacity.set(true)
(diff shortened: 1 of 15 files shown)


src/lavinmq/mqtt/session.cr
spuun, line 1: The changes related to protecting `@unacked` with `@msg_store_lock` can probably be a separate PR.

views/queue.shtml
```diff
@@ -281,7 +282,18 @@
             <label>
               <span>Messages</span>
               <input type="number" name="messages" value=1 min="1" max="1000">
-              <small id="dataTags">Maximum 1000 messages</small>
+              <small class="form-help">Maximum 1000 messages</small>
```
spuun, line 285: Dunno if it's worth adding this help which adds more CSS if we're not using it in more places? Maybe @bengtmagnus should figure out a nice form help html/css/js?

static/js/queue.js
```diff
@@ -251,32 +251,59 @@ document.querySelector('#publishMessage').addEventListener('submit', function (e
...
       for (let i = 0; i < messages.length; i++) {
         const message = messages[i]
         const msgNode = template.cloneNode(true)
         msgNode.removeAttribute('id')
-        msgNode.querySelector('.message-number').textContent = i + 1
+        msgNode.querySelector('.message-number').textContent = i + 1 + offset
         msgNode.querySelector('.messages-remaining').textContent = message.message_count
+        if (message.state) {
+          msgNode.querySelector('.message-state-row').classList.remove('hide')
+          msgNode.querySelector('.message-state').textContent = message.state
```
spuun, line 305: I would be good NOT to use classes for selectors here. Maybe use data attributes instead? Or id, if we can be sure it's unique.
(I see that this PR adopts to what's already being used though.)

spuun (commented):
What's the motivation of this? :)

oskgu360:
> What's the motivation of this? :)

Debug tool we believe is useful for having in lavin basically :) 

Today we achieve the same by `pause queue -> stop consumers -> pull message to check contents -> start consumers -> resume queue`. This would be convienent to avoid interfering with the clients and a bit less cumbersome, especially if you want to see a message deeper into the queue

Issue #2093, opened 2026-06-18
Counter metrics decrease when a queue or vhost is deleted


jage:
**Describe the bug**

Counter-typed metrics decrease when a queue or vhost is deleted, so Prometheus rate()/increase()
reads each drop as a reset and fabricates a spike.

- Queue delete: `global_messages_*_total` re-sum per-queue counters (`vhost.message_details`,
  vhost.cr:277), so a deleted queue's deliveries vanish from the total.
- Vhost delete: churn counters (`connections/channels/queues/consumers_*_total`) and the
  `/api/overview` + `/api/nodes` count fields sum per-vhost atomics over live vhosts only
  (`vhost_stats`, prometheus.cr:447), with no compensation. The lone accumulator
  (`deleted_vhosts_messages_*`, 4 message stats) is written in the HTTP handler (vhosts.cr:60),
  not the canonical `VHostStore#delete` (vhost_store.cr:72).

**Describe your setup**

Current main, `bin/lavinmq --build-info`: 2.8.0-85-g279028c4, default config.
(`queues_declared_total` is server-wide cumulative; run each block on a fresh broker.)

**How to reproduce**

Queue delete:

```Sh
curl -u guest:guest -XPUT localhost:15672/api/vhosts/t
curl -u guest:guest -H 'content-type: application/json' -XPUT -d '{"durable":true}' localhost:15672/api/queues/t/q
for i in 1 2 3 4 5; do
  curl -u guest:guest -H 'content-type: application/json' -XPOST -d '{"routing_key":"q","payload":"x","payload_encoding":"string","properties":{}}'
localhost:15672/api/exchanges/t/amq.default/publish
done
curl -u guest:guest -H 'content-type: application/json' -XPOST -d '{"count":5,"ack_mode":"get"}' localhost:15672/api/queues/t/q/get
curl -s localhost:15692/metrics | grep '^lavinmq_global_messages_delivered_total '  # 5
curl -u guest:guest -XDELETE localhost:15672/api/queues/t/q
curl -s localhost:15692/metrics | grep '^lavinmq_global_messages_delivered_total '  # 0
```

Vhost delete:

```Sh
curl -u guest:guest -XPUT localhost:15672/api/vhosts/u
curl -u guest:guest -H 'content-type: application/json' -XPUT -d '{"durable":true}' localhost:15672/api/queues/u/q1
[...]

PR #2143, opened 2026-07-03
Fix message_stats over-count and counter resets on queue/vhost deletion

### LINKED ISSUE #2092: Per-vhost message_stats over-counts published messages (sums per-queue counters)
**Describe the bug**

`/api/vhosts/:name` builds `message_stats` by summing each queue's counters
(`vhost.message_details`, vhost.cr:277), so a publish routed to N queues counts as N. The vhost's
own atomic counter holds the true client-publish count and is exposed by `/api/overview` and
`detailed_vhost_messages_published_total`, so the surfaces disagree.

**Describe your setup**

Current main, `bin/lavinmq --build-info`: 2.8.0-85-g279028c4, default config.

**How to reproduce**

```Sh
curl -u guest:guest -XPUT localhost:15672/api/vhosts/t
curl -u guest:guest -H 'content-type: application/json' -XPUT -d '{"type":"fanout","durable":true}' localhost:15672/api/exchanges/t/fx
for q in q1 q2 q3; do
  curl -u guest:guest -H 'content-type: application/json' -XPUT -d '{"durable":true}' localhost:15672/api/queues/t/$q
  curl -u guest:guest -H 'content-type: application/json' -XPOST -d '{"routing_key":""}' localhost:15672/api/bindings/t/e/fx/q/$q
done
curl -u guest:guest -H 'content-type: application/json' -XPOST -d '{"routing_key":"","payload":"x","payload_encoding":"string","properties":{}}' localhost:15672/api/exchanges/t/fx/publish
curl -u guest:guest localhost:15672/api/vhosts/t                  # message_stats.publish = 3
curl -u guest:guest -H 'x-vhost: t' localhost:15672/api/overview  # message_stats.publish = 1
```

**Expected behavior**

One client publish counts as one publish, and `/api/vhosts/:name` agrees with `/api/overview`.
(Fix: source `message_stats` from the vhost atomics.)

jage:

- Per-vhost `message_stats` now reads the vhost's own cumulative counters instead of summing per queue, so a fan-out publish is counted once and the value survives queue deletion (#2092).
- Server-wide counters (`global_messages_*`, churn `*_total`, `/api/overview`, `/api/nodes`) stay monotonic across queue/vhost deletion. 
 
A `DeletedVHostStats` accumulator, folded in the canonical `VHostStore#delete`, keeps the totals from dropping, previously a delete made the sum go backwards and Prometheus read it as a counter reset (#2093). It derives from `VHost::STATS_KEYS` (single source) and uses atomic counters.

Closes #2092, #2093.

- `make test SPEC=spec/api/prometheus_spec.cr` New regression specs: fan-out counted once; counters monotonic across queue delete and vhost delete (via the store path); `/api/overview` and `/api/nodes` monotonic.
- Manual: Curl as described in #2092 and we'll now get the correct result

### DIFF
--- a/src/lavinmq/http/controller/main.cr
+++ b/src/lavinmq/http/controller/main.cr
@@ -42,6 +42,16 @@ module LavinMQ
           {{ name.id }}_rate = 0_f64
           {% end %}
 
+          unless x_vhost
+            deleted_stats = @server.vhosts.deleted_stats
+            {% for name in OVERVIEW_STATS %}
+            {{ name.id }}_count += deleted_stats.{{ name.id }}
+            {% end %}
+            {% for name in CHURN_STATS %}
+            {{ name.id }} += deleted_stats.{{ name.id }}
+            {% end %}
+          end
+
           vhosts(user(context)).each do |vhost|
             next if x_vhost && vhost.name != x_vhost
             vhost.each_connection do |c|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The management UI version is advertised via the `LavinMQ-Version` response header instead of being injected at build time [#2123](https://github.com/cloudamqp/lavinmq/pull/2123)
 - CC and BCC headers are removed from dead-lettered messages when `x-dead-letter-routing-key` is set, instead of being preserved, matching RabbitMQ [#1993](https://github.com/cloudamqp/lavinmq/pull/1993)
 
+### Fixed
+
+- Per-vhost `message_stats` no longer over-counts published messages by the fan-out factor; cumulative message counters are read from the vhost's own counters instead of summing per-queue [#2092](https://github.com/cloudamqp/lavinmq/issues/2092)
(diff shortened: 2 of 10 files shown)


src/lavinmq/http/controller/main.cr
```diff
@@ -42,6 +42,16 @@ module LavinMQ
           {{ name.id }}_rate = 0_f64
           {% end %}
 
+          unless x_vhost
+            deleted_stats = @server.vhosts.deleted_stats
+            {% for name in OVERVIEW_STATS %}
```
snichme, line 47: Is there a way to keep `OVERVIEW_STATS` better linked to `Vhost.STATS_KEYS`? 

Because now if I add a key to `OVERVIEW_STATS` I must make sure to add it to `STATS_KEYS` as well and that's not so easy to understand by just looking at the code.

spuun: You'll get a compile time error because the property wont exist in `DeletedVHostStats` or in `VHost`. Maybe not super clear why, but at least you won't ship broken code.

spuun (approved):

src/lavinmq/http/controller/main.cr
jage: Yeah, agree it could be better but not sure what a better would look like right now.

PR #2146, opened 2026-07-07
shovel ui: render runtime view + client-side message rate

snichme:

Introduce a 'Msg rate' column showing per-shovel msgs/s, derived client-side from the delta in cumulative message_count between successive 5s polls, cached outside the row objects by vhost+name.

state as a colored badge (Running/Starting green/amber, Error red, Paused/Stopped/Terminated grey); the error is promoted from a hover tooltip to an inline red sub-line; a degraded sub-line surfaces 'retrying (backoff)' (consecutive_failures>0) and 'aborts n/threshold' (consecutive_aborts>0).

:eyes:

### DIFF
--- a/static/js/shovels.js
+++ b/static/js/shovels.js
@@ -7,19 +7,78 @@ import { UrlDataSource } from './datasource.js'
 
 Helpers.addVhostOptions('createShovel')
 
+const STATE_BADGE_MODIFIER = {
+  Running: 'badge-state--running',
+  Starting: 'badge-state--starting',
+  Paused: 'badge-state--paused',
+  Stopped: 'badge-state--stopped',
+  Terminated: 'badge-state--terminated',
+  Error: 'badge-state--error'
+}
+
+// State as a colored badge, with the error promoted inline (not just on hover)
+// and a degraded sub-line for shovels that are alive but backing off or
+// progressing toward an error-out.
 function renderState (item) {
-  if (item.error) {
-    const state = document.createElement('a')
-    state.classList.add('arg-tooltip')
-    state.appendChild(document.createTextNode(item.state))
-    const tooltip = document.createElement('span')
-    tooltip.classList.add('tooltiptext')
-    tooltip.textContent = item.error
-    state.appendChild(tooltip)
-    return state
-  } else {
-    return item.state
+  const container = document.createElement('div')
+  container.classList.add('state-cell-inner')
+
+  const badge = document.createElement('span')
+  badge.classList.add('badge-state')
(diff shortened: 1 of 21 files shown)


static/js/shovels.js
```diff
@@ -7,19 +7,78 @@ import { UrlDataSource } from './datasource.js'
...
-    state.appendChild(document.createTextNode(item.state))
-    const tooltip = document.createElement('span')
-    tooltip.classList.add('tooltiptext')
-    tooltip.textContent = item.error
-    state.appendChild(tooltip)
-    return state
-  } else {
-    return item.state
+  const container = document.createElement('div')
+  container.classList.add('state-cell-inner')
```
spuun: Is this wrapper necessary?

static/js/shovels.js
```diff
@@ -7,19 +7,78 @@ import { UrlDataSource } from './datasource.js'
...
+  }
+
+  if (item.consecutive_aborts > 0) {
+    const degraded = document.createElement('small')
+    degraded.classList.add('state-degraded')
+    degraded.textContent = `aborts ${item.consecutive_aborts}/${item.abort_threshold}`
+    container.appendChild(degraded)
   }
+
+  return container
```
spuun: Is it possible for the container to contain multiple badges?

static/main.css
spuun, line 1: Can we make a "base" badge (the common rules from tiny-badge and badge-state) and reuse that?

static/js/shovels.js
```diff
@@ -7,19 +7,78 @@ import { UrlDataSource } from './datasource.js'
...
-    state.appendChild(tooltip)
-    return state
-  } else {
-    return item.state
+  const container = document.createElement('div')
+  container.classList.add('state-cell-inner')
+
+  const badge = document.createElement('span')
+  badge.classList.add('badge-state')
+  const modifier = STATE_BADGE_MODIFIER[item.state]
```
spuun, line 28: Why not use `item.state.toLowerCase()` as modifier, then nest the classes like 
```css
.badge-state {
  …
  &.running { background: var(--color-success); color: #141414; }
  …
}
```
That would get rid of `STATE_BADGE_MODIFIER`.

Also, IMO, it's a nice usage of nested css rules.

spuun (commented):
@bengtmagnus should have a look at the css and markup. 

I didn't bother to figure out how to get real errors, so I just hard coded some values in shovels.js to test.
﻿﻿
Image

Dunno if we ever will have multiple rows of texts?
And to me something looks off with a badge and text under it.

PR #2147, opened 2026-07-20
Return 404 instead of 403 for non-existent vhosts when user is admin

hananasen:
`GET`/`DELETE /api/vhosts/:vhost` returned 403 for a non-existent vhost regardless of the requesting user's role. For administrators this is misleading.
- Administrators now get 404 for a non-existent vhost. Non-administrators still get 403.

Specs.

Manually build & run then
```
curl -XDELETE -u guest:guest localhost:15672/api/vhosts/nonexistingvhost

{"error":"Object Not Found","reason":"Vhost nonexistingvhost does not exist"}
```


kickster97 (approved):
Nice, makes sense to me!

viktorerlingsson:
Maybe we should patch/fix the RabbitMQ HTTP specs to accept the 404s?

PR #2149, opened 2026-07-21
Report negotiated MQTT protocol version in connections

johanrhodin:
## What

The management API/UI hardcoded every MQTT connection's `protocol` field as `"MQTT 3.1.1"`, so MQTT 3.1.0 clients (protocol level 3, `MQIsdp`) were mislabelled.

This threads the CONNECT packet's version byte through to `MQTT::Client` and maps it to a human-readable name.

## How

- `broker.cr` passes `packet.version` when constructing the client.
- `client.cr` gains a `@protocol_version` param and an extensible `PROTOCOL_NAMES` lookup table (`0x03 => "MQTT 3.1.0"`, `0x04 => "MQTT 3.1.1"`). Adding MQTT 5.0 later is a one-line change; unknown levels fall back to `"MQTT <n>"` rather than crashing.

## Testing

- Added regression specs: a level-4 CONNECT reports `"MQTT 3.1.1"`, a level-3 CONNECT reports `"MQTT 3.1.0"`.
- `make test SPEC=spec/mqtt/integrations/connect_spec.cr` → 21 examples, 0 failures.

### DIFF
--- a/spec/mqtt/integrations/connect_spec.cr
+++ b/spec/mqtt/integrations/connect_spec.cr
@@ -45,6 +45,28 @@ module MqttSpecs
       end
     end
 
+    describe "reports the negotiated protocol version" do
+      it "as MQTT 3.1.1 for protocol level 4" do
+        with_server do |server|
+          with_client_io(server) do |io|
+            connect(io, version: 0x04u8)
+            client = wait_for { server.vhosts["/"].connections.select(LavinMQ::MQTT::Client).first? }.not_nil!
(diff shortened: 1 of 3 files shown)


johanrhodin:
This is what it looks like in the management interface `/connections`
/connections showing two MQTT connections with different versions and one AMQP 0-9-1 connection

johanrhodin:
This is already fixed in #2139, which is better than this one (reviewed and all). This one has a screenshot though!

PR #2155, opened 2026-07-28
refactor: Split LavinMQ::BindingDetails into AMQP::BindingDetails and MQTT::SubscriptionDetails

spuun:
### WHAT is this pull request doing?
This is a part of #2111.

### HOW can this pull request be tested?
Specs

PR #2156, opened 2026-07-28
refactor: Split LavinMQ::BindingKey into AMQP::BindingKey and MQTT::SubscriptionKey

spuun:
Part of #2111, depends on #2155

QoS 2 will now be downcasted to QoS 1 "for real".

Specs


src/lavinmq/mqtt/subscription_key.cr
```diff
@@ -0,0 +1,43 @@
...
+      def routing_key : String
+        @topic_filter
+      end
+
+      def arguments : AMQP::Table?
+        if @qos.zero?
+          QOS0_ARGUMENTS
+        else
+          QOS1_ARGUMENTS
+        end
```
kickster97, line 34: Hm, since we can store qos2 in the subscription tree, on line session.cr:164 we can end up fetching the hardcoded QOS 1 and comparing QOS 1 to QOS 2. So we unbind and rebind on a repeated subscribe that should be a no op. 

like if a client subscribes with qos 2, and then send in another identical subscribe, it can be a no-op.. but since we grab a harddcoded QOS 1 here, we will compare it to that instead and rebind it.

kickster97 (changes requested):

kickster97 (approved):

PR #2159, opened 2026-07-30
Bugfix: sessions' bindings are lost on restart

spuun:

This pr fixes two related bugs, each with its own cause:

  - The MQTT Exchange was created AFTER definitions was loaded on startup, causing subscription (bindings) to be dropped.
  - Subscriptions (bindings) of persistent MQTT sessions was never written to definitions file because bindings from non-durable exchange isn't presisted, and the MQTT Exchange is non-durable

## Changes
  - The definition store now creates the MQTT exchange to make sure it's created before definitions are loaded
  - Add a condition when applying Bind and Unbind frames if the exchange is the MQTT exchange
  - Retained messages moved from the exchange to the broker, leaving the router to only route. 

Clean-session subscriptions are still not persisted, and the router itself is still never
  written to the definitions file.

Specs

### DIFF
--- a/src/lavinmq/definitions_store.cr
+++ b/src/lavinmq/definitions_store.cr
@@ -4,6 +4,7 @@ require "./event_type"
 require "./queue_factory"
 require "./amqp/exchange/*"
 require "./amqp/queue"
+require "./mqtt/exchange"
 
 module LavinMQ
   class DefinitionsStore
@@ -11,10 +12,18 @@ module LavinMQ
 
     @definitions_file : File
 
+    # The vhost's MQTT exchange. Created with the store rather than declared, so
+    # that it's always there when frames are applied — bindings with it as
+    # source would be dropped otherwise. It isn't durable, which is what keeps
+    # it out of the definitions file.
+    getter mqtt_exchange : MQTT::Exchange
+
     def initialize(@vhost : VHost, @data_dir : String, @replicator : Clustering::Replicator?, @log : Logger)
       @exchanges = Hash(String, Exchange).new
       @queues = Hash(String, AMQP::Queue).new
       @sessions = Hash(String, MQTT::Session).new
+      @mqtt_exchange = MQTT::Exchange.new(@vhost, MQTT::EXCHANGE)
+      @exchanges[MQTT::EXCHANGE] = @mqtt_exchange
       @definitions_lock = Mutex.new(:reentrant)
       @definitions_file_path = File.join(@data_dir, "definitions.amqp")
       # Unbuffered (sync) writes: replication offsets (store_definition) and a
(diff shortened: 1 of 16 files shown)


src/lavinmq/mqtt/exchange.cr
```diff
@@ -15,10 +14,12 @@ module LavinMQ
         "mqtt"
       end
 
-      def initialize(vhost : VHost, name : String, @retain_store : MQTT::RetainStore)
+      def initialize(vhost : VHost, name : String)
         super(vhost, name, false, false, true)
       end
 
+      # Routes a packet to the subscribed sessions. Retaining the message is the
+      # `Broker`'s responsibility, it owns the retain store.
```
kickster97: Do we need this comment?

src/lavinmq/definitions_store.cr
```diff
@@ -216,18 +225,25 @@ module LavinMQ
           x = @exchanges[f.exchange_name]? || return false
           q = @queues[f.queue_name]? || @sessions[f.queue_name]? || return false
           return false unless x.bind(q, f.routing_key, f.arguments)
```
kickster97, line 227: Since `mqtt.default` now exists already during definitions load , a definitions file with a binding from `mqtt.default` to a regular AMQP queue hits `MQTT::Exchange#bind(Destination)`, which raises `AccessRefused`.

`import_bindings` has no per-binding rescue and the launcher exits on any load error, so input that was previously silently skipped (exchange lookup failed and returned `false`) now prevents the server from booting. We should skip or rescue that case here so a bad binding can't fail boot.

src/lavinmq/mqtt/exchange.cr
kickster97: This cast breaks on definitions import. JSON numbers are parsed as `Int64`, so a binding argument like `"mqtt.qos": 1` coming from `POST /api/definitions` or a boot-time definitions file raises `TypeCastError` here, which crashes boot or 500s the API. It only works today because bindings are written as `UInt8` internally.

src/lavinmq/mqtt/exchange.cr
```diff
@@ -28,11 +29,6 @@ module LavinMQ
         bodysize = packet.payload.bytesize.to_u64
         body = ::IO::Memory.new(packet.payload, writable: false)
 
-        if packet.retain?
-          @retain_store.retain(packet.topic, body, bodysize)
-          body.rewind
```
snichme, line 34: this `body.rewind` got lost I think, now we only to `@retain_store..retain` in Broker.

src/lavinmq/definitions_store.cr
```diff
@@ -4,17 +4,26 @@ require "./event_type"
...
+    # that it's always there when frames are applied — bindings with it as
+    # source would be dropped otherwise. It isn't durable, which is what keeps
+    # it out of the definitions file.
+    getter mqtt_exchange : MQTT::Exchange
+
     def initialize(@vhost : VHost, @data_dir : String, @replicator : Clustering::Replicator?, @log : Logger)
       @exchanges = Hash(String, Exchange).new
       @queues = Hash(String, AMQP::Queue).new
       @sessions = Hash(String, MQTT::Session).new
+      @mqtt_exchange = MQTT::Exchange.new(@vhost, MQTT::EXCHANGE)
```
snichme, line 25: I understand why this was created here, but does it make sense that the `DefinitionStore` own the mqtt exchange? I would much rather it was a property on the vhost and you reach it from `@vhost.mqtt_exchange`

src/lavinmq/mqtt/exchange.cr
spuun: We get rid of the IO so no need to rewind!

snichme (approved):

src/lavinmq/definitions_store.cr
spuun: I think this very rare, we've never allowed bindings from mqtt exchange to amqp queues, so there shouldn't be any definitions with these bindings. In the future the mqtt exchagne will most likely be gone too.

kickster97 (approved):

PR #2161, opened 2026-07-30
Clear stale clustering follower file handles before resync

### LINKED ISSUE #2160: Clustering follower keeps stale file handles across reconnects
## Symptom

A follower repeatedly falls out of sync and can enter a persistent disconnect loop:

```
lmq.clustering.client Disconnected from server <leader>:5679 (Error deleting file: '<data-dir>/<vhost-hash>/<queue-hash>/msgs.0000000001': No such file or directory), retrying...
```

Each occurrence forces a full resync, which on data directories with many segment files takes a long time. Only restarting the follower process clears the loop.

## Root cause

`Clustering::Client` caches an open append-mode file handle per replicated file in `@files` (src/lavinmq/clustering/client.cr:49). The cache is never invalidated when the connection to the leader drops: the `follow` loop (client.cr:115-129) rescues `IO::Error` and reruns `sync_files` with `@files` intact.

`sync_files` deletes files on disk behind those handles, either because the local hash mismatches (client.cr:231) or because the file no longer exists on the leader (client.cr:250). Deleting a path does not affect an open handle: the handle keeps pointing at the now-unlinked inode.

After the resync, streaming resumes and appends look up `@files[filename]`, hitting the stale handle. Two failure modes follow:

1. Silent divergence: appends are written into the unlinked inode, fsynced, and acked to the leader as durable, but the visible file never changes. The follower immediately drifts out of sync again, and a promotion in this window loses publisher-confirmed messages.
2. Crash loop: when a deleted path is reused (a recreated queue restarts at msgs.0000000001), the appends for the new file also go into the ghost inode, so no file is created on disk. When the leader later streams a delete for that segment, `f.delete` (client.cr:383) raises `File::Error: Error deleting file: ... No such file or directory`, which disconnects the follower and starts the next full resync, with `@files` still poisoned.

`@file_digests` has the same lifecycle problem: the running checksums no longer describe the on-disk files after a resync, so wrong hashes are persisted on close.

## Reproduction

1. Stream an append (opens the handle), disconnect, resync against a diverged leader copy (file is deleted and re-fetched), reconnect, stream another append. The append is acked but the file on disk does not contain it.
2. Same, but the resync deletes the file (queue gone from the leader) and the path is then reused. No file is ever created on disk, and a streamed delete raises the exact error above.

## Suggested fix

At the start of each connection in the `follow` loop, before `sync`, close all handles in `@files` and clear it, and clear `@file_digests`. Defensively, `Client#delete` should close the cached handle and use `File.delete?` on the path so a missing file can never kill the connection. Reconnects are rare, so reopening handles lazily has no performance impact.

## Environment

Observed on 2.9.1; code unchanged on main.

kickster97:
Fixes #2160.

The follower caches an open file handle per replicated file in @files and never invalidates the cache on reconnect. But the resync deletes or replaces files on disk behind those handles. Appends after the resync then write into the old unlinked inode: they are acked as durable but never reach the visible file, so the follower drifts out of sync again. If a deleted path is reused (a recreated queue starts over at msgs.0000000001), the file is never created on disk and the eventual delete record raises "Error deleting file: No such file or directory", crashing the connection in a loop that only a process restart clears.

Fix: sync_files closes and clears @files and @file_digests before comparing files, and Client#delete tolerates a missing file. Two regression specs follow the reconnect sequence and fail without this change.


src/lavinmq/clustering/client.cr
```diff
@@ -180,6 +180,9 @@ module LavinMQ
       end
 
       private def sync_files(socket, lz4)
+        @files.each_value &.close
+        @files.clear
+        @file_digests.clear
```
viktorerlingsson: should this be moved to `sync` instead? or is there a need to clear these again before the second `sync_file` pass?

viktorerlingsson:
The spec duplicates a bunch of "helper" code from `client_sync_spec.cr`,  that could be moved to a helper and re-used instead

src/lavinmq/clustering/client.cr
spuun: I see that the same code is added in #2167, but placed in `sync`.

kickster97: Youre right! i'll move it to sync

spec/support/clustering_helper.cr
```diff
@@ -1,7 +1,78 @@
...
+
+    def stream_changes_public(socket, lz4)
+      stream_changes(socket, lz4)
+    end
+
+    # what #sync does on a reconnect
+    def resync_files_public(socket, lz4)
+      reset_file_state
+      sync_files(socket, lz4)
+    end
```
viktorerlingsson: To make the testing more future proof and not rely on correct duplicate ordering here, maybe we could move the part after `set_socket_opts(socket)` in  `sync` to a new method, like `full_sync`, that can then be called from both `sync` and the spec? 

Something like
```crystal
      private def full_sync(socket, lz4)
        reset_file_state
        sync_files(socket, lz4)
        Log.info { "Bulk synchronised" }
        sync_files(socket, lz4)
        Log.info { "Fully synchronised" }
      end
```

Then we can have `full_sync_public` in the spec that just calls the private `full_sync` (`simulate_leader` would need to serve two rounds of files as well). WDYT?

kickster97: yeah, fine by me! check out commit 19b8b9047c900f8fa07ddf7ad75dfe9870b8dafa

spec/clustering/client_reconnect_spec.cr
```diff
@@ -0,0 +1,150 @@
...
+        run_sync(client, {filename => "AAAABBBB"})
+        File.read(path).should eq "AAAABBBB"
+
+        with_stream_session(client) do |w, io|
+          send_append(w, filename, "CCCC")
+          read_acks(io, framing(filename) + 4)
+        end
+        File.read(path).should eq "AAAABBBBCCCC"
+
+        client.close
```
viktorerlingsson, line 112: The clustering client is stuck waiting for `@follower_done.receive` here, so the spec stalls for 5s. 

```suggestion

        spawn { client.@follower_done.send(nil) }
        client.close
```

spec/clustering/client_reconnect_spec.cr
```diff
@@ -0,0 +1,150 @@
...
+        with_stream_session(client) do |w, io|
+          send_append(w, filename, "BBBB")
+          read_acks(io, framing(filename) + 4)
+          File.read(path).should eq "BBBB"
+          send_delete(w, filename)
+          read_acks(io, framing(filename))
+          File.exists?(path).should be_false
+        end
+
+        client.close
```
viktorerlingsson, line 148: (same as above)

```suggestion
        spawn { client.@follower_done.send(nil) }
        client.close
```

viktorerlingsson (approved):

PR #2164, opened 2026-08-03
Pre-calculate and persist follower checksums incrementally during full sync

spuun:
A follower hashed its local files *while* comparing them against the leader's file list. This means that the lock the leader holds during full sync is held even for all hash calculations. 

By pre-calculating all hashes before connecting to the leader the time the lock is held is shorten. However, the follower will now calculate hashes for all its local files, even files that should be deleted, which means that the connection to the leader is delayed.

### DIFF
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -121,6 +121,7 @@ module LavinMQ
         end
         loop do
           hash_local_files
+          return if @closed
           @socket = socket = TCPSocket.new(host, port)
           socket.sync = true
           socket.read_buffering = false # use lz4 buffering
--- a/src/lavinmq/clustering/client.cr
+++ b/src/lavinmq/clustering/client.cr
@@ -16,11 +16,8 @@ module LavinMQ
       # LZ4::Reader's internal 64 KiB buffer.
       BUFFER_SIZE = 64 * 1024
 
-      # How many files #hash_local_files hashes between Fiber.yields. Hashing is
-      # CPU bound, and although it runs before we connect, the proxies are
-      # already forwarding clients to the leader, so it can't hog the scheduler.
-      # Yielding per file would add a scheduler round trip per (often small)
-      # file; a batch amortizes that without starving anyone.
+      # Files #hash_local_files hashes between Fiber.yields. Hashing is CPU
+      # bound, but a yield per (often tiny) file costs more than it gives back.
       HASH_YIELD_INTERVAL = 32
 
       # Capacity of the channel buffering acks from the stream-reading fiber to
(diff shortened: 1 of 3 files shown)


src/lavinmq/clustering/client.cr
```diff
@@ -274,6 +289,61 @@ module LavinMQ
         Log.info { "Received all #{requested_files.size} files" } unless requested_files.empty?
       end
 
+      # Hash every local file before connecting to the leader. Comparing our
```
kickster97: can we make this comment more direct, i feel like its kind of hard to understand if you dont have historical context of this PR

src/lavinmq/clustering/client.cr
```diff
@@ -164,10 +172,13 @@ module LavinMQ
...
-        Log.info { "Bulk synchronised" }
-        sync_files(socket, lz4)
-        Log.info { "Fully synchronised" }
+        full_sync_time = Time.measure do
+          bulk_time = Time.measure { sync_files(socket, lz4) }
+          Log.info { "Bulk synchronised in #{bulk_time.total_seconds} seconds" }
+          rest_time = Time.measure { sync_files(socket, lz4) }
+          Log.info { "Changes since bulk synchronized in #{rest_time.total_seconds} seconds" }
+        end
+        Log.info { "Fully synchronised in #{full_sync_time.total_seconds} seconds" }
```
kickster97, line 179: nice

src/lavinmq/clustering/client.cr
```diff
@@ -113,6 +120,7 @@ module LavinMQ
           spawn unix_mqtt_proxy.forward_to(host, @config.mqtt_port), name: "MQTT proxy"
         end
         loop do
+          hash_local_files
           @socket = socket = TCPSocket.new(host, port)
```
kickster97, line 122: If the follower is closed while it's hashing local files, close gives up waiting after 5 seconds and releases the data dir lock and fd. The follow loop then finishes hashing, sees nothing, and connects to the leader anyway, starting a full sync on a data dir it no longer owns. On leader change this means the old client and the new one may sync the same data dir at once, or the process exits 1 when the ack loop hits the closed fd. The race existed before but was only a 1 second window; now it spans the whole hashing pass, which can be minutes, so worth considering doing an early return somewhere maybe if closed?

src/lavinmq/clustering/client.cr
```diff
@@ -274,6 +289,61 @@ module LavinMQ
...
+            # doesn't have (they're about to be deleted), so one unreadable file
+            # must not abort the pass and wedge us in the reconnect loop. Left
+            # uncached: if the leader does have the file, the compare loop retries
+            # the hash and fails the sync there, as it always did.
+            Log.warn(exception: ex) { "Failed to calculate checksum for #{path}" }
+          end
+          # #restore truncates checksums.sha1, so until something rewrites it a
+          # crash would throw away hashes that were on disk at boot. Snapshot the
+          # full set now, while we're not holding up the leader.
+          @checksums.store if computed > 0
```
kickster97, line 313: Should we also snapshot when restore loads entries from disk? restore truncates checksums.sha1 after loading it, and hash_local_files only stores when it computed at least one new hash. So on a healthy restart with an unchanged data dir the file stays empty until a clean shutdown, and an unclean shutdown in that window means the next boot re-hashes the whole data dir. If we store once after restore has loaded entries we keep the cache across unclean shutdowns too, without rewriting the file on every retry iteration.

kickster97 (approved):

src/lavinmq/clustering/client.cr
```diff
@@ -248,6 +260,10 @@ module LavinMQ
         files_to_delete.each do |path|
           Log.debug { "File not on leader: #{path}" }
           File.delete path
+          # #hash_local_files hashes every local file, including ones the leader
+          # doesn't have; drop the entry with the file so the checksum map (and
+          # every snapshot written from it) doesn't accumulate dead paths.
+          @checksums.delete(relative_path(path))
```
viktorerlingsson, line 258: should we persist the checksums to disk here as well (after actually deleting the files a few lines down) with `@checksums.store` so we don't accumulate entries for files that don't exist on crashes?

spuun: Hm, yeah. That could make sense! I'll think this through...

spuun: Made this change. Also made sure to remove the checksums for files that had a mismatching hash (which are also deleted locally).

viktorerlingsson (approved):

PR #2165, opened 2026-08-03
Reuse cached checksums in the capped full_sync pass

### LINKED ISSUE #2162: Leader re-hashes every file while holding the replication lock during follower sync
## Symptom

While a follower is completing its sync, the whole cluster appears to stall: publishes and publish confirms hang, and the management UI barely responds. On a data directory with tens of thousands of segment files this lasts minutes. Reproduced on an idle cluster with no connections or messages: a follower restart took over 8 minutes to sync against a leader with many small files.

## Cause

Syncing a follower ends with a second, capped pass: the leader snapshots each file's size (the follower's synced baseline) and hashes each file up to that cut, inside `@lock` (`Server#sync_and_serve`, src/lavinmq/clustering/server.cr:346-360). `files_with_hash` deliberately bypasses the checksum cache whenever caps are given, since cached hashes are full-size and the cache does not record how many bytes a hash covers (server.cr:201). So the capped pass re-reads and re-hashes every file from disk.

Every replicated operation dispatches through `each_follower`, which needs the same `@lock` (server.cr:474). All publishing therefore queues behind the hashing for the duration of the pass.

## Suggested fix

Store the covered byte count alongside each hash in `Checksums`. The capped pass can then reuse a cached hash exactly when its covered size equals the file's cap, which is the case for all but the handful of files actively written between the two passes. Comparing the cap against the file's *current* size instead is not safe: a local write can grow a file after the size snapshot but before cache invalidation (that write path does not hold `@lock`), and a reused full-size hash would then cover bytes the follower does not have.

Moving the remaining hashing out of `@lock` entirely is a natural follow-up but a separate change, since it alters locking.

kickster97:
Fixes #2162.

Builds on @lukas8219's 80cd1992, which spotted the capped pass never using the checksum cache. That commit is cherry-picked here with a corrected reuse condition on top: it compared the digest's byte length against the caps hash, so it could never be true and the capped pass still recomputed everything.

During the final full_sync pass the leader hashes every file up to a snapshotted cut while holding the lock that all replicated operations need, so publishing stalls until the pass is done. The checksum cache was bypassed because entries did not record how many bytes a hash covers, so a cached hash could not be trusted for a capped size.

Every checksum write now records the covered byte count, on the leader as well as on the follower, so a size may only be recorded by code that just hashed exactly that many bytes. Lookups hand out the whole entry, hash and size together, and the caller decides whether the recorded coverage fits its use. The capped pass reuses a cached hash exactly when it covers the requested cut, which is every file not written to between the two passes, so it no longer re-reads tens of thousands of unchanged files under the lock. Comparing the cap against the file's current size instead would race local writes that have not invalidated the cache yet, which is why the covered size is stored rather than inferred.

The size is persisted in checksums.sha1 as "hash size *path". Files written by older versions parse as entries without a size; the uncapped pass recomputes those and records sizes, so an old cache heals in one pass and the capped pass never trusts a sizeless entry.

Specs: the reuse spec deletes the file from disk before the capped pass, so the hash can only come from the cache, and it fails without this change. Further specs cover recompute on cap mismatch, healing of old-format entries, size persistence across append/store/restore, and that follower-written checksums match both content and size on disk.

### DIFF
--- a/src/lavinmq/clustering/checksums.cr
+++ b/src/lavinmq/clustering/checksums.cr
@@ -3,10 +3,8 @@ module LavinMQ
     class Checksums
       Log = LavinMQ::Log.for "clustering.checksums"
 
-      # `size` is the number of bytes the hash covers. Only entries restored
-      # from an old-format checksums.sha1 lack it; callers must treat those
-      # as unusable until recomputed.
-      record Entry, hash : Bytes, size : Int64?
+      # `size` is the number of bytes the hash covers.
+      record Entry, hash : Bytes, size : Int64
 
       @checksums = Hash(String, Entry).new
       # Always-open handle to checksums.sha1, kept open across rewrites so
@@ -52,11 +50,12 @@ module LavinMQ
           loop do
             hash = f.read_string(40).hexbytes
             rest = f.read_line
-            if rest.starts_with?(" *") # sizeless format: "<hash> *<path>"
-              @checksums[rest[2..]] = Entry.new(hash, nil)
-            elsif idx = rest.index(" *", 1) # sized format: "<hash> <size> *<path>"
-              size = rest[1...idx].to_i64?
-              @checksums[rest[idx + 2..]] = Entry.new(hash, size)
+            # Line format: "<hash> <size> *<path>". Old-format lines without a
+            # size are dropped; a hash with unknown coverage is unusable.
+            if idx = rest.index(" *", 1)
+              if size = rest[1...idx].to_i64?
+                @checksums[rest[idx + 2..]] = Entry.new(hash, size)
+              end
             end
           rescue IO::EOFError
             break
@@ -94,11 +93,7 @@ module LavinMQ
       end
 
       private def line(path : String, entry : Entry) : String
-        if size = entry.size
-          "#{entry.hash.hexstring} #{size} *#{path}"
-        else
-          "#{entry.hash.hexstring} *#{path}"
-        end
+        "#{entry.hash.hexstring} #{entry.size} *#{path}"
       end
 
       private def checksums_path : String
--- a/src/lavinmq/clustering/checksums.cr
+++ b/src/lavinmq/clustering/checksums.cr
@@ -6,29 +6,7 @@ module LavinMQ
       # `size` is the number of bytes the hash covers, when known. Entries
       # without a size (older checksums.sha1 files, follower-written entries)
       # can never be served for a sized lookup (#hash_for?).
-      record Entry, hash : Bytes, size : Int64? do
-        # One line per entry: "<hash> <size> *<path>", or "<hash> *<path>"
-        # when the size is unknown. The path is the caller's hash key, so it
-        # travels alongside the entry rather than being stored in it.
-        def to_io(io : IO, path : String) : Nil
-          io << hash.hexstring
-          if s = size
-            io << ' ' << s
-          end
-          io << " *" << path << '\n'
-        end
-
-        # Returns nil for malformed lines. Raises IO::EOFError at end of file.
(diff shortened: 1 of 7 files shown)


src/lavinmq/clustering/checksums.cr
```diff
@@ -34,19 +40,24 @@ module LavinMQ
...
-            f.skip(2) # " *"
-            path = f.read_line
-            @checksums[path] = hash
+            rest = f.read_line
+            if rest.starts_with?(" *") # sizeless format: "<hash> *<path>"
+              @checksums[rest[2..]] = Entry.new(hash, nil)
+            elsif idx = rest.index(" *", 1) # sized format: "<hash> <size> *<path>"
+              size = rest[1...idx].to_i64?
+              @checksums[rest[idx + 2..]] = Entry.new(hash, size)
+            end
```
spuun, line 59: Maybe add `#to_io` and `.from_io` to `Entry`? I think that's cleaner and keeps the format of a serialized Entry in one place.

viktorerlingsson:
Here's some thoughts on an improvement from me and Claude 🤖 

At `server.cr:209` the uncapped pass reuses a cached hash without checking whether the entry records a covered size, and it
  doesn't call `set` on the hit, so a sizeless entry gets served as-is and never upgraded. The capped pass only reuses entries
  that have a size, so those files keep getting re-hashed under `@lock` on every follower join.

  That bites in two routine cases. Any `checksums.sha1` written before this PR is entirely in the old format, so every entry
  restored at startup is sizeless. And every entry a follower writes (`client.cr:225`, `:335`, `:569`) is sizeless too, and
  `Clustering::Server#listen` restores that same file at `server.cr:300`, so a node promoted from follower to leader starts
  with nothing but sizeless entries.

  A file only picks up a size once a write invalidates its entry and the uncapped pass recomputes it, so actively written files
  heal on their own while untouched files keep their sizeless entry indefinitely. Untouched files are exactly what the capped
  pass spends its time on, which means on a cluster upgraded to this version that's most files at the first follower join, and
  after a failover it's all of them. I think that makes the improvement hard to see on a real cluster as it stands.

  Treating a sizeless entry as a miss fixes it. Beside `hash_for?`:

  ```crystal
  # The hash for `path` only if the entry records how many bytes it covers.
  # A sizeless entry is deliberately a miss: recomputing records the size,
  # which is what makes the entry reusable by a later capped pass.
  def sized_hash?(path) : Bytes?
    if entry = @checksums[path]?
      entry.hash if entry.size
    end
  end
  ```

  and then at `server.cr:209`:

  ```crystal
  cached_hash = @file_index.shared do |_files, checksums|
    if caps
      checksums.hash_for?(path, caps[path]? || 0i64)
    else
      checksums.sized_hash?(path)
    end
  end
  ```

  Nothing else needs to change since `set` at `server.cr:226` already records the size, and because every sized entry is a
  full-file hash (only the uncapped pass calls `set`, and the capped pass never writes) this is really just the uncapped
  counterpart to `hash_for?`. The extra recompute lands in the uncapped pass, which runs outside `@lock` (`sync_and_serve` only
  holds it around the capped pass), so it costs sync time rather than publish latency. Worth a spec as well: seed a sizeless
  entry the way a follower would, run the uncapped pass, then assert the capped pass reuses the hash, deleting the file first
  so it can only have come from the cache.

  Passing sizes at the `client.cr` write sites would help too, but I'd leave that out of here: once the above lands it only
  saves one re-hash on the first sync after a promotion, and it breaks the three assertions that hardcode the sizeless line
  format (`client_sync_spec.cr:465`, `:496`, `clustering_spec.cr:630`).

spuun (approved):

PR #2167, opened 2026-08-04
Fix wrong checksums for files appended to mid-content

spuun:

`stream_with_checksum` created a SHA1 digest for any file that lacked one, including files that already had content on disk.  Such a digest only covers the appended bytes, so persisting it on graceful shutdown claimed the hash of a tail was the hash of the whole file. On the next start `sync_files` compared it against the leader's hash, mismatched, and deleted and re-fetched a file we already had — for every active segment, ack file, consumer_offsets and definitions.amqp, on every restart.

A digest is only a valid checksum if it has seen the file from byte 0, and `Digest::SHA1` can't be seeded with a hash, so a file first seen mid-content is now left untracked (a nil entry in `@file_digests`) and its cached checksum dropped: the next sync re-hashes it locally instead of trusting a hash that doesn't match its content. Files created by appends and files installed by a replace are still tracked.

replace holds its digest aside until the rename has installed the file, adopting any running digest as the old file's checksum first, so an aborted replace leaves a checksum that still matches what's on disk instead of a hash of content that was never installed.

Specs

### DIFF
--- a/spec/clustering/client_sync_spec.cr
+++ b/spec/clustering/client_sync_spec.cr
@@ -3,6 +3,85 @@ require "lz4"
 
 module ClientSyncSpec
   extend ClusteringSpecHelper
+  # `extend` only copies methods, not the module's nested types.
+  alias TestClient = ClusteringSpecHelper::TestClient
+
+  # Runs one file-list comparison pass against a leader offering `leader_files`,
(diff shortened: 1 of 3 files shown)


kickster97 (approved):
Only note is that the comments are long, but also contain good info.. but my feeling is still that we should avoid documenting like this inside the code files

PR #2175, opened 2026-08-14
Fix O(N²) insert in DelayedRequeuedStore via min-heap

### LINKED ISSUE #2173: DelayedMessageStore: O(N²) insert causes broker startup to hang with large delayed-message backlogs
## Summary

`LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore::DelayedRequeuedStore#insert` has quadratic (O(N²)) behavior, which caused a production cluster to get stuck in a start/restart loop with a large backlog of delayed messages (millions, via `x-delay` on a delayed-exchange queue).

## Incident background

- Cluster (production) had ~4M messages when an OOM triggered a leader re-election.
- The newly elected leader got stuck during startup — no useful logs even at trace level.
- `strace -f -c` on the running process showed almost no syscall activity (mostly `epoll_wait`/`clock_nanosleep`), i.e. not I/O-bound.
- `perf top` on the pid showed:
  ```
  98.45% lavinmq [.] *LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore::DelayedRequeuedStore#insert<LavinMQ::SegmentPosition, Int64>:Nil
  ```
- Claude was asked to investigate assuming a large delayed-message backlog, and confirmed quadratic behavior by benchmarking the insert:
  - 100k inserts → 1.2s
  - 400k inserts → 18.8s (4× the messages, ~15× the time)
  - Extrapolated to 4M messages: ~30+ minutes just for this step.
  - Root cause: `Deque#insert` shifts up to half the deque element-by-element on every insert, so N random-delay inserts is O(N²).
- Confirmed in practice: **1.5M delayed messages alone was enough to stall a cluster restart.**
- Workaround used to recover the cluster: used `lmqrecover` to identify the queue with the delayed messages (`lmqrecover -D <file> -d -m show_sp` surfaces the AMQPTable, including the `x-delay` header) and deleted that queue (with customer's permission), which allowed the broker to start.
- systemd's `TimeoutStartSec` (customer bootstrap sets `TimeoutStopSec=600` for stop, but start timeout was still the default ~300s / whatever configured) was also killing the process mid-recovery, compounding the problem — the process needed more than the configured start timeout just to churn through the delayed store.

## Proposed fix

Since the delayed message store only ever needs the minimum entry (peek min / pop min / time-to-next-expiration), a `Deque` is the wrong data structure for this. A binary min-heap would give O(log n) insert and O(1) peek, fixing both:
1. Slow/stuck startup when recovering a large delayed-message backlog.
2. Slow publishing into a large delayed queue at runtime (same O(N²) insert path is presumably hit during normal operation too, not just startup).

Alternative structures discussed: skip list or B-tree (potentially better if working set fits in CPU cache).

## Troubleshooting guide (for future reference, added to this issue for visibility)

1. LavinMQ pinned at 1 core, 100% CPU, cluster/node stuck starting.
2. `strace -f -p <pid> -c` and `iostat` show no significant syscall/IO activity — process isn't stuck on I/O.
   - Trace-level logs can help but may show nothing (as in this case).
3. Run `perf top -p <pid>` to see where CPU time is actually spent.
4. If it points at the Delayed Store — need to identify *which* queue has the large delayed backlog.
5. `du -h | sort -h` under `/var/lib/lavinmq/` to find the largest queue directories.
6. `cat <largest queues>/.queue` to inspect.
7. `lmqrecover -D <file> -d -m show_sp` to dump segment positions/headers, which will include the `x-delay` AMQPTable entry identifying the delayed queue.
8. If customer agrees, delete the offending queue to unblock startup.

lukas8219:
## Summary

Fixes #2173: `DelayedMessageStore::DelayedRequeuedStore` used a `Deque` with
`bsearch_index` + `insert`, which is O(N) per insert (Deque shifts elements),
giving O(N²) total cost for N delayed messages. This caused broker restarts
to hang (rebuilding the in-memory delayed index) and made publishing into a
large delayed-exchange queue progressively slower.

This replaces the `Deque` with a binary min-heap (`LavinMQ::MinHeap`),
giving O(log n) insert and O(1) peek, while `shift?`/`first?` stay O(log n)
and O(1) respectively — no change to expiration behavior.

## Benchmark

Built `main` and this branch, declared a delayed exchange, and populated
1.5M messages with random `x-delay` (20–45 days — the `x-delay` header is
truncated to `UInt32` ms in `segment_position.cr`, capping any usable delay
at ~49.7 days) on identical on-disk data.

### Restart time (1.5M delayed messages already on disk)

| Branch | Restart time |
|---|---|
| `main` | **206.9s** — exceeded a 2-minute cap before even being tested further. Pinned at 99.96% CPU with `utime:13987`/`stime:13` (pure userspace, zero I/O) — matches the `perf top` output in the issue |
| this branch | **0.138s / 0.139s / 0.174s** (3 runs) |

**~1,200–1,500× faster.**

### Runtime insert throughput (publishing 1.5M messages into a live broker)

| Branch | Time for 1.5M inserts | Rate at 1.5M |
|---|---|---|
| `main` | 230.4s, decaying throughout (65k → 6.5k msg/s) | 6,510 msg/s |
| this branch | 14.8s, flat throughout | 101,417 msg/s (constant, confirms O(log n)) |

### Insert latency at a 1.5M-message baseline (2000 samples, random delay in-range)

| Branch | p50 | p90 | p99 | avg |
|---|---|---|---|---|
| `main` | 0.506ms | 0.742ms | 1.031ms | 0.514ms |
| this branch | 0.197ms | 0.219ms | 0.416ms | 0.207ms |

### Expiration/delivery latency at a 1.5M-message baseline (30 samples, fixed 2s delay)

| Branch | p50 overhead | p90 overhead | avg overhead |
|---|---|---|---|
| `main` | 3.53ms | 102.76ms | 23.09ms |
| this branch | 3.36ms | 101.91ms | 19.62ms |

Expiration latency is unchanged (as expected — `shift?`/`first?` were already
cheap on both structures; the ~100ms tail comes from `RoughTime`'s 100ms
clock-tick granularity, not the data structure). Only insert cost changes.

## Test plan

- [x] `spec/min_heap_spec.cr` — new unit tests for the heap itself
- [x] `spec/delayed_requeued_store_spec.cr` — updated for the new backing structure
- [x] Manual benchmark: 1.5M-message restart, runtime insert throughput, insert latency, expiration latency (above)

### DIFF
--- a/src/lavinmq/min_heap.cr
+++ b/src/lavinmq/min_heap.cr
@@ -40,8 +40,9 @@ module LavinMQ
       value = @heap[idx]
       while idx > 0
         parent = (idx - 1) // 2
-        break if @heap[parent] <= value
-        @heap[idx] = @heap[parent]
+        p = @heap[parent]
+        break if p <= value
+        @heap[idx] = p
         idx = parent
       end
       @heap[idx] = value
@@ -52,9 +53,13 @@ module LavinMQ
       size = @heap.size
       while (child = 2 * idx + 1) < size
         right = child + 1
-        child = right if right < size && @heap[right] < @heap[child]
-        break if value <= @heap[child]
-        @heap[idx] = @heap[child]
+        c = @heap[child]
+        if right < size && (r = @heap[right]) < c
+          child = right
+          c = r
+        end
+        break if value <= c
+        @heap[idx] = c
         idx = child
       end
       @heap[idx] = value
--- a/src/lavinmq/min_heap.cr
+++ b/src/lavinmq/min_heap.cr
@@ -1,0 +1,63 @@
+module LavinMQ
+  class MinHeap(T)
+    def initialize
+      @heap = Array(T).new
+    end
+
+    def size
+      @heap.size
+    end
+
+    def empty?
+      @heap.empty?
+    end
+
+    def first? : T?
+      @heap.first?
+    end
+
+    def push(value : T) : Nil
+      @heap << value
+      sift_up(@heap.size - 1)
+    end
+
+    def shift? : T?
+      return if @heap.empty?
+      min = @heap.first
+      last = @heap.pop
+      unless @heap.empty?
+        @heap[0] = last
+        sift_down(0)
+      end
+      min
+    end
+
+    def clear : Nil
+      @heap = Array(T).new
+    end
+
+    private def sift_up(idx : Int32) : Nil
+      value = @heap[idx]
+      while idx > 0
+        parent = (idx - 1) // 2
+        break if @heap[parent] <= value
+        @heap[idx] = @heap[parent]
+        idx = parent
+      end
+      @heap[idx] = value
+    end
+
+    private def sift_down(idx : Int32) : Nil
+      value = @heap[idx]
+      size = @heap.size
+      while (child = 2 * idx + 1) < size
+        right = child + 1
+        child = right if right < size && @heap[right] < @heap[child]
+        break if value <= @heap[child]
+        @heap[idx] = @heap[child]
+        idx = child
+      end
+      @heap[idx] = value
+    end
+  end
+end
--- a/spec/delayed_requeued_store_spec.cr
+++ b/spec/delayed_requeued_store_spec.cr
@@ -74,6 +74,27 @@ describe LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore::DelayedRequeu
       store.shift?.should eq sp2
       store.shift?.should eq sp3
     end
+
+    it "should shift by expiration then segment position for randomized inserts" do
+      store = LavinMQ::AMQP::DelayedExchangeQueue::DelayedMessageStore::DelayedRequeuedStore.new
+      timestamp = 1000i64
+      sps = Array(LavinMQ::SegmentPosition).new(500) do |i|
+        LavinMQ::SegmentPosition.new(Random.rand(3).to_u32, i.to_u32, 10u32,
+          delay: (Random.rand(5) * 100).to_u32)
+      end
+      sps.each { |sp| store.insert(sp, timestamp) }
+      store.size.should eq sps.size
+
+      expected = sps.sort do |a, b|
+        r = (timestamp + a.delay) <=> (timestamp + b.delay)
+        r.zero? ? a <=> b : r
+      end
+      shifted = Array(LavinMQ::SegmentPosition).new
+      while sp = store.shift?
+        shifted << sp
+      end
+      shifted.should eq expected
+    end
   end
 
   describe "#time_to_next_expiration?" do
--- a/spec/min_heap_spec.cr
+++ b/spec/min_heap_spec.cr
@@ -1,0 +1,100 @@
+require "spec"
+require "../src/lavinmq/min_heap"
+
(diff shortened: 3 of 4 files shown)


lukas8219:
## Follow-up benchmark: 4M messages / 15-minute publish cap

Same setup as the PR description, scaled up: `min-heap` was given a 4M-message
target, `main` was capped at 15 minutes of publishing (whichever it reached
first).

| | `main` (15-min cap) | `min-heap` |
|---|---|---|
| Messages reached | 2,970,000 (hit the cap) | 4,000,000 in 98.5s |
| Cumulative insert rate at cutoff | 3,297 msg/s, still falling | ~40.6k msg/s, flat |
| Insert latency (p50 / avg) | 0.801ms / 0.840ms | 0.196ms / 0.208ms |
| Restart time | **838.5s (~14 min)** | **0.33-0.34s** |

Scaling check against the original 1.5M numbers: `main`'s message count grew
1.98x (1.5M→2.97M) and its restart time grew 4.05x (206.9s→838.5s) — matching
1.98² = 3.92 almost exactly, i.e. O(N²). `min-heap`'s count grew 2.67x
(1.5M→4M) and restart time grew only 2.23x, consistent with O(N log N).

**Note on the ~40k vs ~100k msg/s baseline:** the flat insert rate for
`min-heap` here (~40.6k msg/s) is lower than the ~101k msg/s in the original
PR description. That's a benchmark-script artifact, not a server-side
change. What matter is the flat-ness

lukas8219:
RSS analysis

Screenshot 2026-08-14 at 12 23 00

spuun:
Should we use this in `DelayedRequeuedStore` too?
Maybe not necessary, I guess a requeued store won't contain that many messages, and it's not persisted so it's never "restored".

lukas8219:
@spuun The PR name was confusing. The change itself is on `DelayedRequeuedStore`

spuun:
> @spuun The PR name was confusing. The change itself is on `DelayedRequeuedStore`

I think it more was a matter of me doing to much concurrent work. For some reason I thought we had the same logic in two places.

Oh well, never mind my comment! :)

src/lavinmq/min_heap.cr
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    private def sift_down(idx : Int32) : Nil
+      value = @heap[idx]
+      size = @heap.size
+      while (child = 2 * idx + 1) < size
+        right = child + 1
+        child = right if right < size && @heap[right] < @heap[child]
+        break if value <= @heap[child]
+        @heap[idx] = @heap[child]
```
kickster97: Each `@heap[child]` read bounds-checks and copies the whole ~32-byte struct, and this loop body reads the same unchanged element three times. The loop runs once per heap level, and `shift?` runs on every delayed delivery. Load the element into a local once and you save alot:

```suggestion
        right = child + 1
        c = @heap[child]
        if right < size && (r = @heap[right]) < c
          child = right
          c = r
        end
        break if value <= c
        @heap[idx] = c
```

kickster97 (changes requested):

kickster97 (approved):

Issue #2177, opened 2026-08-17
Cluster followers syncfs at the replication stream's cadence, not the leader's confirm cadence


viktorerlingsson:
A follower can flush to disk far more often than the leader it replicates from, and in the common case (leader rarely or never using publisher confirms) the leader does zero `syncfs` while the follower does one every ~100 ms.

## Where it comes from

On a follower, `Clustering::Client#send_ack_loop` (`src/lavinmq/clustering/client.cr:596-608`) calls `sync_to_disk` before *every* ack batch it sends back to the leader, and `sync_to_disk` -> `sync_data_dir` (`client.cr:611-629`) is a filesystem-wide `syncfs(2)`.

The ack cadence is set by the replication stream, not by publishers:

- The follower acks each chunk as it is streamed to disk (`Client#ack`, `client.cr:584-588`, called from `stream_with_checksum`), so every arriving batch of replicated bytes produces at least one ack, and therefore at least one `syncfs`.
- The leader pushes those bytes on a timer regardless of confirms: `Follower#ack_loop` sets `@socket.read_timeout = 100.milliseconds` and flushes the LZ4 buffer on every read timeout (`src/lavinmq/clustering/follower.cr:91-106`). So a steadily-publishing leader with no confirms still hands the follower work ~10 times per second.
- Batching on the follower is only an emergent side-effect of `syncfs` blocking (acks accumulate in the channel while it runs), so the flush rate is "as fast as the disk allows", with no coalescing delay.

Meanwhile the leader's only publish-path barrier is `Persister#sync` from the publish-confirm drain (`src/lavinmq/persister.cr:47-54`, `75-113`) plus `tx_commit`. A workload that doesn't confirm costs the leader nothing, so the two nodes end up with wildly different flush rates for the exact same data.

Because each `syncfs` writes back every dirty folio of every file on the volume (see #2176), the follower's higher flush rate multiplies straight into disk write rate.

## Consequences

- Followers show a disk write rate that is a multiple of the leader's, for identical data, and the asymmetry is largest exactly when the leader looks healthy (few or no confirms).
- `--no-sync` does not propagate: `sync_to_disk` checks the follower's own `@config.sync?`, so a leader started with `--no-sync` still gets followers doing a filesystem-wide flush per ack batch unless every node is given the flag.
[...]

PR #2178, opened 2026-08-17
Leader controls follower's syncfs 

spuun:

This PR introduces "control packets" in the clustering protocol. A control packets is just a "file delete" packet with a special filename. This means that it's backward compatible as an old follower would treat these packets as noop.

By introducing control packets the leader will now send a "sync" packet when it want to do a syncfs, meaning that that followers won't do it periodically anymore, only on demand. The leader will wait for the sync packet to be acked before sending confirms to its publishers.

Addresses #2177

Specs, but should be tested and benchmarked manually.


carlhoerberg:
Also added in #2183 , but where we also track explicit files that needs to be persisted (`fsync`),  removing `syncfs` altogether.

PR #2179, opened 2026-08-17
Exit process if syncfs is blocked for 10 seconds

carlhoerberg:
`syncfs` can block, if it blocks for more than 10 seconds this PR exits the process so that another node in the cluster can take over.

### DIFF
--- a/src/lavinmq/config/options.cr
+++ b/src/lavinmq/config/options.cr
@@ -430,6 +430,17 @@ module LavinMQ
       @[EnvOpt("LAVINMQ_CLUSTERING_PORT")]
       property clustering_port = 5679
 
+      # Used on both the leader (Persister) and a follower (Clustering::Client)
+      # to bound how long a blocked syncfs(2) is tolerated before exiting so a
+      # standby node can take over. Only enforced while clustering is enabled:
+      # a standalone node has no follower to fail over to, so it logs instead
+      # of exiting (see Persister#wait_for_syncfs).
+      @[CliOpt("", "--clustering-syncfs-timeout=SECONDS",
+        "Seconds to tolerate a blocked syncfs(2) before exiting (default: 10)", section: "clustering")]
+      @[IniOpt(ini_name: syncfs_timeout, section: "clustering")]
+      @[EnvOpt("LAVINMQ_CLUSTERING_SYNCFS_TIMEOUT")]
+      property clustering_syncfs_timeout : Time::Span = 10.seconds
+
       @[IniOpt(section: "amqp")]
       property max_consumers_per_channel = 0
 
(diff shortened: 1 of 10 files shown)


src/lavinmq/persister.cr
```diff
@@ -6,45 +6,103 @@ require "./clustering/follower"
...
+      rescue Channel::ClosedError
+        break
+      end
+    end
+
+    private def wait_for_syncfs : Nil
+      select
+      when @syncfs_ok.receive     # syncfs completed
+      when timeout syncfs_timeout # syncfs blocked for too long
+        unless @replicator
```
viktorerlingsson, line 54: should this check if there's any connected followers (`@replicator.try(&.followers)`)? replicator would be `!= nil` regardless of if there's any followers to failover to, right?

src/lavinmq/config/options.cr
```diff
@@ -430,6 +430,17 @@ module LavinMQ
...
+      # Used on both the leader (Persister) and a follower (Clustering::Client)
+      # to bound how long a blocked syncfs(2) is tolerated before exiting so a
+      # standby node can take over. Only enforced while clustering is enabled:
+      # a standalone node has no follower to fail over to, so it logs instead
+      # of exiting (see Persister#wait_for_syncfs).
+      @[CliOpt("", "--clustering-syncfs-timeout=SECONDS",
+        "Seconds to tolerate a blocked syncfs(2) before exiting (default: 10)", section: "clustering")]
+      @[IniOpt(ini_name: syncfs_timeout, section: "clustering")]
+      @[EnvOpt("LAVINMQ_CLUSTERING_SYNCFS_TIMEOUT")]
+      property clustering_syncfs_timeout : Time::Span = 10.seconds
```
viktorerlingsson, line 442: I think we should validate to make sure `clustering_syncfs_timeout  > 0.seconds`

viktorerlingsson:
Minor and probably doesn't need any fixing, but noting anyway:  
I think **misbehaving** clients can break in a couple of new ways here.

If a client sends a second `tx.commit` before the first `CommitOk` arrives, `enqueue_commit_ok` never sends the second reply and the client hangs forever (the pending set collapses the two, and the 1-slot mailbox can drop one).

And since `CommitOk` now comes from `tx_commit_writer` instead of the read loop, a client that sends any other synchronous method before `CommitOk` arrives (`queue.declare`, `tx.rollback`) gets the replies out of order, which most libs raise on.

Both need clients that ignore the spec rule about not sending a synchronous method before the previous reply, so probably nothing that needs fixing, just noting it's sort of a regression since the inline send handled both. Maybe worth an extra row of comment.

Issue #2176, opened 2026-08-17
Leaders write 30-200x more to disk than they receive over the network


viktorerlingsson:
Investigation by Claude (Claude Code), driven by @chadknutson . Numbers below are from our two production clusters plus a controlled 3-node AWS test cluster. This is independent of clustering.

## TL;DR

Disk writes are dominated by a term that has nothing to do with message rate or message size:

```
disk_write  ~  hot_files x folio_size x syncfs_rate   +   message_bytes
               \____________ this dominates ____________/
```

We append messages by memcpy into `MAP_SHARED` mmaps, so the unit of a write is a page or folio, never a message. `syncfs(2)` is filesystem-wide and is paced by publisher confirms. And every queue has an `acks.*` file that takes 4 bytes per ack into that same page-sized unit. Multiply those together and you get 200x on a cluster doing 0.15 MiB/s.

It is not a leak, not replication, and not fan-out. Nothing is being written twice at the application level - we write each enqueued message exactly once.

## What we measured

Cluster A (prod), from Grafana + node tracing:

```
network in         6.16 MiB/s
disk write          204 MiB/s (2h mean); 76.4 MiB/s in a live 30s sample
disk read              0 B/s          <- rules out resync / re-read entirely
disk usage          flat at 2.4 GiB   <- steady-state churn, not accumulation
queues                    78
syncfs                10.4 /s
confirms       170.6/s of 3177 publishes  (~5% of publishers)
avg write req       72.9 KiB
```

Cluster B (prod): 167 KiB/s in -> 32.6 MiB/s written. Same fingerprint: zero reads, flat usage.

### 1. The write unit is a page/folio, not a message

`MessageStore#write_to_disk` (`src/lavinmq/message_store.cr:333-344`) is a memcpy into an mmap. There is no `write(2)` and no `msync` anywhere on the publish path (`MFile#flush`, the only `msync`, has no callers). When anything flushes, the kernel writes the whole dirty unit. On ext4 that's 4 KiB. On XFS with a modern kernel (cluster A is on 6.17) the page cache uses large folios up to 64 KiB - we measured 57 KiB per dirty event and 72.9 KiB average write request. So a 2 KB message can dirty 64 KiB.

### 2. syncfs is filesystem-wide and is paced by publisher confirms

[...]

Issue #2151, opened 2026-07-21
Clustering leader does not fail over when its data-dir disk stalls (zombie leader)


johanrhodin:
## Describe the bug

When the clustering **leader's** data-dir filesystem stalls on I/O — the backing
block device hangs while the network and process stay alive — the cluster does
**not** fail over. The stalled leader keeps its leadership indefinitely, all
publisher confirms hang, and no follower is ever promoted. The only recovery is
an operator manually killing/rebooting the leader.

### Root cause

Leadership is held via an etcd lease. The lease keepalive
(`Etcd::Lease#keepalive_loop`, `src/lavinmq/etcd/lease.cr`) runs on its own
`Fiber::ExecutionContext::Isolated` thread and only does `sleep` +
`Etcd#lease_keepalive` — a network-only HTTP POST. **It never touches the disk.**

Meanwhile the blocking `syncfs(2)` (`Persister#sync`, `src/lavinmq/persister.cr`)
runs on a *different* isolated thread. So when the disk wedges and `syncfs` blocks
in uninterruptible (`D`) sleep, the keepalive thread keeps renewing the lease.
etcd — and therefore every follower watching `{prefix}/leader` — still sees a
healthy leader, and no one campaigns.

The existing self-fencing paths don't cover this:

- `Controller` only `exit`s on **etcd** failures (lease expiry / not-found). A disk
  stall keeps etcd reachable, so the lease is happily renewed.
- `Persister#drain_pending_acks` only `exit 1`s when `syncfs` **returns an error**.
  A *hung* `syncfs` never returns, so that branch never runs.

Net result: the system stays **consistent** (no split brain, nothing falsely
confirmed) but **unavailable**, with no automatic recovery. This is a classic
gray-failure / "zombie leader" blind spot: liveness is proven over the network,
but the resource that actually failed (the disk) is never checked.

## Describe your setup

- LavinMQ clustering enabled, 3 nodes, etcd for coordination (default lease TTL 10s).
- Observed on a cloud block volume that experienced a silent
  I/O stall — `jbd2`/`ext4` tasks blocked in `D` state, `syncfs` wedged, **no**
  block-layer I/O errors in `dmesg` (the device stopped completing I/O rather than
  erroring).

## How to reproduce

1. Start a 3-node cluster with clustering + etcd; let one node become leader and
   the others reach the ISR.
2. Freeze the leader's data-dir I/O without killing the process or its network,
   e.g.:
[...]

PR #2183, opened 2026-08-19
Replace syncfs with per-file msync and replicated fsync requests

carlhoerberg:
## What

Publish confirms were gated on `syncfs(2)` of the whole data-dir filesystem — on the leader in the Persister and on followers before every ack batch. syncfs can stall (it flushes *every* dirty page on the filesystem) and amplifies writes: a single publish confirm also wrote out dirty pages of unrelated files (`acks.*`, other vhosts, definitions).

This replaces it with targeted durability:

- **Leader**: the `Persister` tracks which `MFile`s were written by confirmed or transactional publishes since the last sync (an atomic needs-msync flag dedupes registration) and msyncs only those via a new `MFile#fsync` (`msync(MS_SYNC)` — only actually-dirty pages are written, so no position tracking is needed). Ordinary durable publishes without confirms never enter the dirty set: a durability flag is propagated from publisher-confirm and transactional channels through routing into the message stores (classic, priority, stream, delayed).
- **Replication protocol**: instead of syncfs-before-ack, followers receive explicit fsync requests — records with a `$`-prefixed filename and zero length — and ack them only after fsyncing the named file. `wait_for_followers` therefore still means confirmed data is durable on every in-sync replica.
- **Large batches fall back to syncfs**: once a confirm batch has at least `syncfs_threshold` (default 10) live dirty files, one `syncfs` on a data-dir descriptor held for the persister's lifetime is cheaper than many serial msyncs. Closed and deleted files don't count toward the threshold.
- `acks.*` files are never eagerly synced for plain publish confirms: losing recent acks on crash only causes redeliveries, and `prune_orphaned_acks` already handles msgs/acks divergence. Transactional acknowledgments are the exception, see below.

## Publish confirm loop

- Every confirm — sync, no-sync and clustered alike — is routed through the publish confirm loop, which runs on an isolated thread so blocking msync only stalls that thread. It is the *only* thread that msyncs: `Persister#sync` (tx.commit) enqueues a `WaitGroup`, wakes the loop and blocks until the drain that made its writes durable — locally and on every in-sync follower — signals it. Commits batch with confirms instead of racing them, and each waiter is released only by its own batch (a later request belongs to the next drain).
- `Tx.CommitOk` is still sent by the connection fiber: it is a synchronous RPC response that must stay ordered with other responses on the channel, and the isolated loop must not write client sockets anyway. `tx.commit` now also waits for followers, so commit-ok carries the same durability guarantee as a publish confirm (previously leader-local syncfs only).
- The loop must never write follower sockets, so `$`-records are dispatched through each follower's existing `flush_loop`; `sent_bytes` is counted at enqueue time so a `wait_for_confirm` issued right after covers them. Queued fsync requests are written under the follower's write lock before any later record is counted, so the reserved byte count has the same order as records on the wire.
- Two orderings keep a `$`-record from overtaking an append it must cover: files are marked dirty only *after* the replicator dispatch, and the drain clears the flags *before* dispatching the requests.
- `MFile` holds a mapping lock across msync and any operation that unmaps (close, truncate), so a segment closed or shrunk concurrently can't be unmapped under a running msync.
- The data-dir descriptor used by syncfs is closed in `finalize`, not when the loop exits: a tx.commit racing shutdown syncs inline once the loop is gone and may still need it.

## Transactions and directory entries

- Transactional acknowledgments mark their `acks.*` files dirty like segments, so a tx.commit msyncs them locally and fences them on followers with the same `$` request. Segment removal is made durable (directory synced) before `deleted?` is published, since the persister skips deleted files as soon as it sees that flag.
- Message stores keep one directory descriptor per store (`FileSystem::Directory`) instead of opening and walking ancestor directories per sync. New segment names are fsynced through it before a durability ack, and only directories that actually changed are synced. Directory syncing lives with the owners (message store, consumer offsets) rather than in `MFile`. The descriptor is reopened once for queue-deletion teardown and reused for every removed segment.
- File replacements (`definitions.amqp`, users, parameters, vhosts, MQTT retain store, queue `.paused` marker) go through `FileSystem.durable_rename`, which fsyncs the parent directories after the rename. Followers do the same for received replacements before acking them, so promotion after a power loss can't expose the old name-to-inode mapping.

## Followers

- Followers fsync files received during full_sync (the baseline is durable at join, so syncing followers are skipped by the dispatch) and fsync replace tmp-files before the rename their deferred final ack implies. Durable definition changes dispatch a fsync request for `definitions.amqp` before fencing on followers.
- A follower on a pre-`$` leader still works: the strict rejection of legacy peers was tried and reverted, followers simply see no fsync requests from an old leader.
- Known trade-off: file-delete dir-entry durability on followers for non-transactional deletes (previously implied by syncfs) is dropped; deletes reconcile on re-sync.

## Configuration

- New `[main]` option `syncfs_threshold` (positive integer, default 10): number of dirty files in one confirm batch at which the persister uses one syncfs instead of per-file msync.

## Testing

- New specs: `MFile#fsync` + dirty-flag semantics and mapping-lock protection during msync; persister waiter release only after its own batch, empty transactions don't sync, syncfs threshold (below, at, closed/deleted excluded), data-dir descriptor kept open for its lifetime and usable for an inline sync after close; message-store durability flag and directory sync only for changed directories; follower handling of `$`-records (fsync+ack, missing file isn't resurrected, sync-disabled still acks, queued fences ordered with later records); client full_sync/replace fsyncs and parent-dir persistence; leader protocol spec asserting a confirm emits `$…msgs.…` and `$…definitions.amqp` but never any `acks.*` request, while a tx.commit does fence `acks.*`; tx.commit-ok is held until the follower acks.
- Full suite green (2226 examples), lint clean.
- Manual smoke under `strace`: zero `syncfs` calls below the threshold; one `msync(MS_SYNC)` of just the dirty segment per confirm batch; fsync only on definitions at declare.


carlhoerberg:
Fixes #2176 and #2177 and maybe #2151

viktorerlingsson:
I had been working on and testing this a bit as well, so I ran my tests against this PR. I think there's a couple of findings worth discussing and maybe adding to this PR. 

### 1 - There's a threshold where a full `syncfs` is *much* better

Confirmed publish rate, msg/s, mean of 2 reps. "depth" is the confirm window,  "100q" means the publisher round-robins 100 durable queues so a drain batch spans that many segments.

  | scenario | main | #2183 | mine |
  |---|---|---|---|
  | 1 queue, depth 1 | 202 | 402 | 400 |
  | 1 queue, depth 100 | 5,701 | 8,279 | 8,486 |
  | 100 queues, depth 1 | 202 | 402 | 401 |
  | 100 queues, depth 100 | 5,621 | 422 | 6,374 |
  | no confirms, 500B | 583,928 | 580,234 | 593,158 |

So there's a few wins in comparison to main here, but there's one big loss, and that's when there's a lot of confirms on a lot of different queues. Then the overhead of syncing each file one by one takes longer than just doing one `syncfs`. So I tried some different things and ended up on falling back to `syncfs` if there's more than `x` dirty files to sync. I haven't pinpointed what number `x` should be, and it might be different on different setups. Maybe that can be configurable. I tried with 4 in the example above, but that's too low. With a number too low, that means that in some other scenarios the fallback happens too early, so we don't get as big of a win. 


### 2 - Which files get synced

Not as big of an impact on performance, but still a pretty big impact on written bytes to disk. This will still sync all dirtied msg files if one `publish_confirm` arrives. In some cases, that will end up in a lot of disk syncs that aren't actually asked for. But if we keep track of which files we actually write `publish_confirm` to, we can sync only them. It leads to a bit more code because we need to thread down from `publish_confirm` to the actual dirty file tracking, but it does have more of an impact on written bytes.

100 durable queues. Queue 1 publishes 50 msg/s with confirms, queues 2-100 publish 100 msg/s between them without. All variants publish the same bytes, so anything above the no-confirm baseline is sync work charged to queues that never asked for it. `dirty` is `write_bytes` from `/proc/<pid>/io`, which counts a page again each time it's re-dirtied after writeback, so it tracks how often we rewrite the same page.

  | variant | dirty MiB/s | phys MiB/s |
  |---|---|---|
  | `main`, nobody confirms | 0.089 | 0.131 |
  | `main` | 0.663 | 2.394 |
  | `#2183` | 0.662 | 1.275 |
  | mine | 0.283 | 0.469 |

src/lavinmq/clustering/client.cr
```diff
@@ -588,46 +610,44 @@ module LavinMQ
...
-      # Make all replicated writes durable before acking the leader.
-      private def sync_to_disk : Nil
+      # Make a file durable before acking the leader's fsync request. Uses the
+      # open append handle when there is one (it wrote the data); otherwise a
+      # throwaway fd — e.g. a file received via full_sync. A file that's gone
+      # was deleted after the leader dispatched the request, so there's
+      # nothing left to sync (never open via the @files default block here,
+      # that would resurrect the deleted file as empty).
+      private def fsync_file(filename : String) : Nil
         return unless @config.sync?
```
spuun, line 655: We came here because the leader told us to fsync, so remove this?
```suggestion
```

src/lavinmq/clustering/client.cr
```diff
@@ -541,19 +550,32 @@ module LavinMQ
...
         File.open(path, "w") do |f|
           f.sync = true
           # The record's final ack tells the leader the replace is durable, so
           # it must not be sent while the new content only exists as the .tmp
-          # file; hold it back until the rename has installed the file.
+          # file; hold it back until the content is on disk and the rename has
+          # installed the file.
           deferred = stream_with_checksum(lz4, f, len, sha1, defer_final_ack: true)
-          f.rename f.path[0..-5]
+          f.fsync if @config.sync?
```
spuun, line 579: Should this be handled with sync frames from leader instead?

src/lavinmq/mqtt/retain_store.cr
```diff
@@ -86,7 +87,8 @@ module LavinMQ
           len = ::IO.copy(body_io, file, size)
           raise ::IO::EOFError.new("Copied only #{len} of #{size} bytes") if len != size
           final_file_path = File.join(@dir, msg_file_name)
-          file.rename(final_file_path)
+          file.fsync
```
spuun, line 94: Unconditional, even if config sync is false? What about sync in followers?

spuun (commented):
A lot going on in this PR, but I think I get most of it. 

I think it's worth thinking about doing syncfs instead of individual fsyncs if there are many files needing fsync.

src/lavinmq/persister.cr
```diff
@@ -59,20 +92,101 @@ module LavinMQ
...
+
+      dirty.each do |mfile|
+        next if mfile.closed? || mfile.deleted?
+        begin
+          sync_file(mfile)
+        rescue IO::Error
+          # Closed in the window since the check above — a deleted segment's
+          # content no longer needs durability. A real msync failure raises an
+          # errno-based RuntimeError, handled below.
+        rescue ex
```
viktorerlingsson, line 172: I think there's a very narrow window where we could also get a `RuntimeError` , if `close` manages to complete the `munmap`  between `check_open` and the actual `LibC.msync` in `MFile#msync` . Probably very unlikely, but maybe worth adding a rescue for that as well as cheap insurance? 

```suggestion
        rescue ex : RuntimeError
          # Unmapped between check_open and the msync syscall: ENOMEM means the
          # range is gone, and a closed segment's content needs no durability.
          raise ex unless ex.os_error == Errno::ENOMEM
        rescue ex
```

src/lavinmq/clustering/client.cr
```diff
@@ -541,19 +550,32 @@ module LavinMQ
...
       end
 
+      # fsyncing the replacement file before rename makes its contents durable,
+      # but not the directory entry changed by rename(2). Persist the parent
+      # directory before acking the replace so promotion after a power loss
+      # cannot expose the old name-to-inode mapping.
+      private def fsync_parent_dir(path : String) : Nil
+        return unless @config.sync?
+        File.open(File.dirname(path), &.fsync)
+      end
```
viktorerlingsson, line 595: Do we need to do something similar on the leader on `open_new_segment`? So we're sure that directory entries for any newly created files that a publish confirm depends on are synced to disk as well (we used to get that for free with `syncfs`).

viktorerlingsson:
I ran my performance tests against the latest changes and I think those are all looking good now! Everything is better-than or equal to main 👍 (Note that those tests only cover single node, I haven't done any testing of the clustering changes)

src/lavinmq/clustering.cr
```diff
@@ -4,7 +4,9 @@ require "digest/sha1"
 
 module LavinMQ
   module Clustering
-    Start = Bytes['R'.ord, 'E'.ord, 'P'.ord, 'L'.ord, 'I'.ord, 1, 0, 0]
+    # Version 2 requires explicit durability fences. Version 1 peers treat
+    # ordinary byte acknowledgments as durable and must not join this protocol.
+    Start = Bytes['R'.ord, 'E'.ord, 'P'.ord, 'L'.ord, 'I'.ord, 2, 0, 0]
```
viktorerlingsson: Do we need to change to version 2 here? I guess that means no rolling restarts on upgrades? Which feels more like a major version thing.

carlhoerberg: No, i think we can revert that commit

carlhoerberg:
Split into a stack of four PRs for review: #2238 (durable file replacements) → #2239 (per-file msync + replicated fsync requests) → #2240 (confirm-loop serialization, tx fences, syncfs_threshold) → #2241 (directory-entry durability). The tip of #2241 is tree-identical to this branch.

PR #2185, opened 2026-08-21
WIP: MQTT5 support

viktorerlingsson:
Adds MQTT 5.0 support alongside the existing 3.1 / 3.1.1, on the same listener.

A v5 client can connect, subscribe, publish and receive with properties intact,
gets an accurate reason code on every ack, gets a session whose lifetime it
controls, and gets a spec-correct rejection for every feature we do not
implement.

**Read `MQTT5.md` first.** It is committed in this PR and is the design and
status doc: section 2 is the correctness anchor, section 5 the remaining work,
section 8 the sequencing to ship.

The one thing to know before reviewing: MQTT 5.0 lets a server omit optional
features only if it advertises their absence in CONNACK *and* rejects clients
that use them anyway. Advertising alone is not compliance, and enforcing alone is
not compliance. That is what makes the deferrals here legal rather than broken,
so every deferred feature (QoS 2, topic aliases, shared subscriptions,
subscription identifiers, enhanced auth) has both an advertised capability and a
spec-cited rejection with a test. Section 2 has the matrix.

## Blocked on

84codes/mqtt-protocol.cr#9. `shard.yml` pins `branch: feat/mqtt5`, which cannot
ship, so that PR needs its version bump and tag first.

## Up for grabs

Independent of each other and of the shard work. `MQTT5.md` section 5 has the
detail and the files.

- **B, subscription options.** The shard already parses No Local, Retain As
  Published and Retain Handling; `Broker#subscribe` reads only `tf.qos`. Pure
  LavinMQ behaviour, no shard work, well bounded.
- **E, will properties and will delay.** `Client#publish_will` builds a PUBLISH
  with no properties at all, so every v5 Will Property is silently dropped.
- **F, properties on retained messages.** The retain store keeps only the body,
  so a retained v5 message reaches a later subscriber stripped. Needs a
  store-format change, so the most invasive of the three.

## Behaviour changes for existing v3.1.1 users

Both deliberate, both need release notes, both in `MQTT5.md` section 6.

- Delivery QoS is now the minimum of the publish and subscription QoS, not the
  subscription's alone ([MQTT-3.8.4-6]). This is a real bug fix, LavinMQ used to
  wait for a PUBACK on a fire-and-forget message and store it for an offline
  subscriber, but it is visible: a QoS 0 publish to a QoS 1 subscription is no
  longer upgraded. Fourteen existing specs had encoded the old behaviour.
- A v5 client sending Clean Start 0 with no Session Expiry Interval now gets a
  session that ends with its connection, where it used to get one that persisted
  forever. v3.1.1 is unaffected.

## Test plan

- [x] `make test TAGS=~etcd` - 2125 examples, 0 failures, 9 pending (pre-existing
      and unrelated to MQTT)
- [x] `make test SPEC=spec/mqtt` - 280 examples, 0 failures
- [x] `make lint` - 403 inspected, 0 failures
- [x] `crystal tool format --check` clean
- [ ] etcd-tagged specs - skipped locally, no etcd running
- [ ] Re-run the interop harness in `MQTT5-INTEROP.md`. The last run was
      2026-08-19 and predates the delivery-QoS fix, the unexpected-packet
      handling, the review fixes and session expiry, so its delivery-QoS and
      DISCONNECT `0x82` rows are expectations to confirm rather than results.

### DIFF
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -30,11 +30,25 @@ module LavinMQ
         @vhost.register_exchange(@exchange)
       end
 
-      def session_present?(client_id : String, clean_session) : Bool
-        return false if clean_session
+      # Clean Start = 1 always reports no session, because the stored one is about
+      # to be discarded [MQTT-3.1.2-4].
+      #
+      # The auto_delete? guard is for takeover: this runs before add_client, so a
+      # 0-interval session belonging to a still-connected client is visible here,
+      # and that session is ended by the takeover rather than resumed (3.1.4).
+      def session_present?(client_id : String, clean_start) : Bool
+        return false if clean_start
         session = sessions[client_id]? || return false
-        return false if session.clean_session?
-        true
+        !session.auto_delete?
+      end
+
+      # v3 has no expiry property, so its clean-session bit carries both meanings:
+      # 1 ends the session with the connection, 0 keeps it forever, which is what
+      # LavinMQ has always done. v5 reads the property, absent meaning 0
+      # [MQTT-3.1.2-11].
+      private def session_expiry_interval(packet : Protocol::Connect) : UInt32
+        return packet.clean_session? ? 0u32 : UInt32::MAX unless packet.version.v5?
+        packet.properties.session_expiry_interval || 0u32
       end
 
       def add_client(io, connection_info, user, packet) : Client
@@ -52,13 +66,20 @@ module LavinMQ
           packet.clean_session?,
           packet.keepalive,
           packet.will,
-          packet.properties.maximum_packet_size)
-        if client.clean_session?
+          packet.properties.maximum_packet_size,
+          session_expiry_interval(packet))
+        # Clean Start and the expiry are separate inputs: the first decides
+        # whether to discard the stored session, the second how long the session
+        # this connection ends up with will outlive it.
+        if packet.clean_session?
           sessions[client.client_id]?.try &.delete
         else
-          # If an existing session exists, reuse it. If no session exists
-          # it will be created on first subscribe
-          sessions[client.client_id]?.try &.client = client
+          # Reuse an existing session, adopting this connection's interval. No
+          # session yet means it is created on first subscribe.
+          if session = sessions[client.client_id]?
+            session.session_expiry_interval = client.session_expiry_interval
+            session.client = client
+          end
         end
         @clients[packet.client_id] = client
         @vhost.add_connection client
@@ -80,7 +101,7 @@ module LavinMQ
         if session = sessions[client_id]?
           if session.client.nil? || (session.client == client)
             session.client = nil
-            session.delete if session.clean_session?
+            session.delete if session.auto_delete?
           end
         end
         @clients.delete(client_id) if @clients[client_id]? == client
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -77,8 +77,11 @@ module LavinMQ
           # Reuse an existing session, adopting this connection's interval. No
           # session yet means it is created on first subscribe.
           if session = sessions[client.client_id]?
-            session.session_expiry_interval = client.session_expiry_interval
+            # Attach first: while the session has a client, wait_for_client
+            # cannot run, so narrowing the interval here can never expire the
+            # session this connection is about to resume.
             session.client = client
+            session.session_expiry_interval = client.session_expiry_interval
           end
         end
         @clients[packet.client_id] = client
--- a/src/lavinmq/mqtt/broker.cr
+++ b/src/lavinmq/mqtt/broker.cr
@@ -51,7 +51,8 @@ module LavinMQ
           packet.client_id,
           packet.clean_session?,
           packet.keepalive,
-          packet.will)
+          packet.will,
+          packet.properties.maximum_packet_size)
         if client.clean_session?
           sessions[client.client_id]?.try &.delete
         else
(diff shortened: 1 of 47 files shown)


src/lavinmq/mqtt/broker.cr
```diff
@@ -30,11 +30,25 @@ module LavinMQ
...
-        true
+        !session.auto_delete?
+      end
+
+      # v3 has no expiry property, so its clean-session bit carries both meanings:
+      # 1 ends the session with the connection, 0 keeps it forever, which is what
+      # LavinMQ has always done. v5 reads the property, absent meaning 0
+      # [MQTT-3.1.2-11].
+      private def session_expiry_interval(packet : Protocol::Connect) : UInt32
+        return packet.clean_session? ? 0u32 : UInt32::MAX unless packet.version.v5?
```
spuun, line 50: Can we move this in to mqtt-protocol instead? So a V3 packet has properties where session_expiry_interval is 0 or MAX?

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -12,28 +12,58 @@ module LavinMQ
...
-            session_present = broker.session_present?(packet.client_id, packet.clean_session?)
-            connack io, session_present, Protocol::Connack::ReturnCode::Accepted
-            broker.run_client(io, connection_info, user, packet)
+          # CONNECT carries the protocol version on the wire; read_connect
+          # bootstraps with a v3 IO and hands back one reframed to the negotiated
+          # version (v3.1 / v3.1.1 / v5) for every subsequent packet. It only
+          # rebinds io on success, so the v3 boot IO survives a parse error and
+          # the rescue below can still answer with a CONNACK.
+          io = Protocol::IO::V3.new(socket, @config.mqtt_max_packet_size)
+          packet, io = io.read_connect
```
spuun, line 32: It would be nice if we just had to do
```suggestion
          io = Protocol::IO.new(socket, @config.mqtt_max_packet_size)
          packet = io.read_connect
```
and that the IO checks the version in the connect packet and from that on uses that version.

(I have a working example, will try to clean up and make a PR against 84codes/mqtt-protocol.cr#9)

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -12,28 +12,58 @@ module LavinMQ
...
+          if packet.properties.authentication_method
+            logger.warn { "Enhanced authentication requested but not supported" }
+            reject_connack(io, Protocol::Connack::ReasonCode::BadAuthenticationMethod)
+            return socket.close
+          end
+          # We advertise maximum_qos = MAX_QOS, and 3.1.2.6 wants a Will above
+          # it refused rather than quietly clamped. v5 only, and deliberately:
+          # v3 has no return code meaning "QoS not supported", so a v3 Will at
+          # QoS 2 stays accepted and is clamped at delivery, as it always was.
+          if packet.version.v5? && packet.will.try { |w| w.qos > MAX_QOS }
```
spuun, line 46: Could this be moved into mqtt-protocol? Like validate this and raise an exception we handles here?

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -12,28 +12,58 @@ module LavinMQ
...
+          if packet.client_id.empty?
+            assigned_client_id = generated_client_id(user.name)
+            packet = packet.copy_with(client_id: assigned_client_id)
           end
+          validate_client_id!(packet.client_id, user.name)
+          if broker.connection_limit_reached?(packet.client_id)
+            raise Protocol::Error::ServerUnavailable.new(
+              "too many connections to vhost \"#{broker.vhost.name}\"")
+          end
+          session_present = broker.session_present?(packet.client_id, packet.clean_session?)
```
spuun, line 64: I wonder why we ever passed clean session to this method?

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -48,11 +78,54 @@ module LavinMQ
...
+
+      private def connack(io : Protocol::IO, session_present : Bool,
+                          return_code : Protocol::Connack::ReturnCode,
+                          assigned_client_id : String? = nil)
+        reason = Protocol::Connack::ReasonCode.from_v3_return_code(return_code)
+        # A v5 server must advertise which optional features it supports; an
+        # accepted v5 connection carries the capability set. On v3 the properties
+        # are ignored on the wire, so the v3 CONNACK is byte-for-byte unchanged.
+        properties =
+          if io.version.v5? && return_code.accepted?
```
spuun, line 96: Can move this somehow into mqtt-protocol? I'm thinking that we always construct the `ConnackProperties` then it's up to the IO (or `Packet#to_io`) to write the right format depending on the IO?

spuun (commented):
My idea with the comments is to question if we can get rid of all "version awareness", so we more or less make the code as if it's MQTT 5?

src/lavinmq/mqtt/connection_factory.cr
viktorerlingsson: Sounds good! Please give that a try :)

src/lavinmq/mqtt/connection_factory.cr
viktorerlingsson: Yeah, I guess that's a little confusing with the naming. It's been there since the first MQTT implementation 🤷 The check itself is needed though. Maybe we could do some renaming and refactoring and end up with something like

connection_factory.cr
```
  resuming_session = !packet.clean_session? && broker.resumable_session?(packet.client_id)
  connack io, resuming_session, Protocol::Connack::ReturnCode::Accepted, assigned_client_id
```

broker.cr (previously `session_present?`)
```
  def resumable_session?(client_id : String) : Bool
    session = sessions[client_id]? || return false
    !session.auto_delete?
  end
```

PR #2207, opened 2026-08-26
Moves add queue form to a overlay dialog

bengtmagnus:
Moves the "Add queue" form to a overlay dialog. Adds a button "Add queue" in the queue list card top right corner.

Look at queues. If this is accepted it will be implemented on more places.

Still has some fine tuning to do style wise. For example now it is an scroll overflow on the dialog. That is because how tooltips works so need to fix that too.

### DIFF
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -160,6 +160,17 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
   Table.renderCell(tr, 14, Helpers.nFormatter(item.total_bytes) + 'B', 'right')
 })
 
+const addQueueDialog = document.getElementById('add-queue-dialog')
+document.getElementById('show-add-queue').addEventListener('click', () => {
+  addQueueDialog.showModal()
+})
+document.getElementById('close-add-queue').addEventListener('click', () => {
+  addQueueDialog.close()
+})
+addQueueDialog.addEventListener('click', (evt) => {
+  if (evt.target === addQueueDialog) addQueueDialog.close()
+})
+
 document.querySelector('#declare').addEventListener('submit', function (evt) {
   evt.preventDefault()
   const data = new window.FormData(this)
@@ -177,6 +188,7 @@ document.querySelector('#declare').addEventListener('submit', function (evt) {
       queuesTable.reload()
       evt.target.reset()
       evt.target.querySelector('select[name="vhost"]').value = decodeURIComponent(vhost) // Keep selected vhost selected
+      addQueueDialog.close()
       DOM.toast('Queue ' + queue + ' created')
     })
 })
--- a/static/js/queues.js
+++ b/static/js/queues.js
@@ -161,15 +161,6 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
 })
 
 const addQueueDialog = document.getElementById('add-queue-dialog')
-document.getElementById('show-add-queue').addEventListener('click', () => {
-  addQueueDialog.showModal()
-})
-document.getElementById('close-add-queue').addEventListener('click', () => {
-  addQueueDialog.close()
-})
-addQueueDialog.addEventListener('click', (evt) => {
-  if (evt.target === addQueueDialog) addQueueDialog.close()
-})
 
 document.querySelector('#declare').addEventListener('submit', function (evt) {
   evt.preventDefault()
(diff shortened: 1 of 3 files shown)


views/queues.shtml
```diff
@@ -14,6 +14,9 @@
       <section class="card">
         <div class="table-header">
           <h2>Manage queues</h2>
+          <div class="btn-holder">
+            <button type="button" class="btn btn-green" id="show-add-queue">+ Add queue</button>
```
spuun:
```suggestion
            <button type="button" class="btn btn-green" id="show-add-queue" popovertarget="add-queue-dialog">+ Add queue</button>
```

views/queues.shtml
```diff
@@ -51,8 +54,12 @@
           </table>
         </div>
       </section>
-      <form method="put" id="declare" class="form card">
-        <h3>Add queue</h3>
+      <dialog id="add-queue-dialog" class="dialog">
+        <div class="dialog-header">
+          <h3>Add queue</h3>
+          <button type="button" class="dialog-close" id="close-add-queue" aria-label="Close">&#x2715;</button>
```
spuun:
```suggestion
          <button type="button" class="dialog-close" id="close-add-queue" aria-label="Close" popovertargetaction="hide" popovertarget="add-queue-dialog">&#x2715;</button>
```

static/js/queues.js
```diff
@@ -160,6 +160,17 @@ const queuesTable = Table.renderTable('table', tableOptions, function (tr, item,
...
+document.getElementById('show-add-queue').addEventListener('click', () => {
+  addQueueDialog.showModal()
+})
+document.getElementById('close-add-queue').addEventListener('click', () => {
+  addQueueDialog.close()
+})
+addQueueDialog.addEventListener('click', (evt) => {
+  if (evt.target === addQueueDialog) addQueueDialog.close()
+})
+
```
spuun, line 164: Use `command` and `commandfor` instead of javascript?
```suggestion

```

views/queues.shtml
```diff
@@ -51,8 +54,12 @@
           </table>
         </div>
       </section>
-      <form method="put" id="declare" class="form card">
-        <h3>Add queue</h3>
+      <dialog id="add-queue-dialog" class="dialog">
```
spuun:
```suggestion
      <dialog id="add-queue-dialog" class="dialog" popover>
```

spuun (commented):
Can we make the dialog a popover and get rid of the javascripts maybe?

bengtmagnus:
> Can we make the dialog a popover and get rid of the javascripts maybe?

Claude has this to say. Maybe makes sense? What do you think @spuun ?

I'd stick with <dialog> — I don't think popover is a good trade here.

showModal() gives you focus-trapping, an inert background, and a native ::backdrop for free; a popover element doesn't trap focus and isn't modal (background stays interactive), and it doesn't auto-center — you'd have to hand-roll the centering/dimming CSS that dialog already gives you. The only real savings would be the two open/close click listeners (~3 lines); the form-submit JS (the actual API call) isn't going anywhere either way. There is a newer declarative way to open a <dialog> without JS (command="show-modal" commandfor="..." on the button), but it's Chrome-only right now, so not safe for us yet.

spuun:
> I'd stick with — I don't think popover is a good trade here.
> 
> showModal() gives you focus-trapping, an inert background, and a native ::backdrop for free; a popover element doesn't trap focus and isn't modal (background stays interactive), and it doesn't auto-center — you'd have to hand-roll the centering/dimming CSS that dialog already gives you. The only real savings would be the two open/close click listeners (~3 lines); the form-submit JS (the actual API call) isn't going anywhere either way. There is a newer declarative way to open a without JS (command="show-modal" commandfor="..." on the button), but it's Chrome-only right now, so not safe for us yet.

Ah, true. But we can use `command` and `commandfor` to get rid two event listeners? And maybe make a generic thingy for closing the dialog when someone clicks outside? 

Like putting
```javascript
document.addEventListener('click', evt => {
  if (evt.target.matches('dialog')) evt.target.close()
})
```
in `dom.js` to get this globally? We could change the selector if we only want it for some dialogs?

spuun:
However, putting `popover` on the dialog and open it using popover API makes it works just with the dialog, but without javascript? hm

spuun:
Scratch what I've said...

No js needed with `Invoker API` (`command=` and `commandfor=`) together with `closedby="any"`.

bengtmagnus:
> Scratch what I've said...
> 
> No js needed with `Invoker API` (`command=` and `commandfor=`) together with `closedby="any"`.

Changes made :)

kickster97 (dismissed):
tried it out and clicked around and it works well, and looks even better! good job!

PR #2214, opened 2026-09-01
Reject HTTP error responses in request

### LINKED ISSUE #1727: Form resets on warnings from backend

snichme:

Changes the `HTTP.request` function to only call the `.then` callback of the promise of the response is good otherwise the reject part of the promise will be called ( `.then(fullfilled, rejectedCB)` or `.catch(rejectedCB)`
This changes the fullfilled callback so we don't need to check `.is_error` and handle that 

fixes #1727 

New specs and tested locally


kickster97:
this supersedes #1802, right?

snichme:
@kickster97 or rather, it's an alternate solution.

kickster97 (approved):
I think we should go with this PR. Request rejects on every error, so no caller can run its success code after a failed response, so we are solving the core issue with this.

PR #2215, opened 2026-09-01
Allow HTTP shovel destinations without queue or exchange

### LINKED ISSUE #2220: HTTP shovels cannot be created through the API since 2.7.0
## What happens

Creating a shovel with an HTTP destination through the API is rejected:

```
$ curl -u guest:guest -X PUT http://localhost:15672/api/parameters/shovel/%2F/webhook \
    -H 'Content-Type: application/json' \
    -d '{"value":{"src-uri":"amqp:///","src-queue":"events","dest-uri":"https://example.com/webhook"}}'
{"error":"bad_request","reason":"Shovel destination requires queue and/or exchange"}
```

An HTTP destination has no queue or exchange to name. It POSTs the message body to the endpoint, so there is nothing the caller can add to make the request succeed. `lavinmqctl add_shovel` goes through the same endpoint and fails the same way.

The management UI still works, but only by accident: it submits the hidden AMQP destination field as an empty string, and `""` is not `nil`, so the check passes and the shovel is stored with a stray `dest-queue: ""`.

## Expected

The request succeeds with 201 and the shovel starts POSTing to the endpoint, which is what `docs/shovels.md` documents under "HTTP Destination".

## Cause

`Shovel::Store.validate_config!` (`src/lavinmq/shovel/store.cr:61`) raises unless the config names a destination queue or exchange:

```crystal
raise ConfigError.new("Shovel destination requires queue and/or exchange") if dst.nil?
```

The check was added in #1670 and applies to every destination, including HTTP ones.

## Affected versions

2.7.0 through 2.9.2. #1670 landed 2026-02-06, first released in 2.7.0.

## Fix

- #2218 adds the specs, red
- #2215 makes the check apply only when every destination URI is AMQP, green

Found while working on #1423, which adds Standard Webhooks signing for HTTP shovels and is unreachable until this is fixed.

baelter:
Fixes #2220.

## Background

HTTP shovels cannot be created through the API. #1670 added a check that a shovel destination names either a queue or an exchange. An HTTP destination names neither: it POSTs the message body to the endpoint, so there is nothing the caller can add to make the request succeed. The CLI and the management UI go through the same endpoint, so webhook shovels have been unconfigurable everywhere since that release.

#2218 is the precursor to this PR. It adds the specs and is red on main. This PR is the fix and turns them green, so the diff here is 5 lines of `store.cr` and a changelog entry.

Found while finishing #1423, which adds request signing for HTTP shovels and is unreachable until this is fixed.

## Summary

The check now applies only when every destination URI is AMQP. It matches on the exact `http` and `https` schemes, the same way `Store#destination` decides which destination class to build, so validation cannot accept a config that construction then rejects.

A config that mixes an HTTP and an AMQP destination is still rejected, since the AMQP destination does need somewhere to publish.

## Test plan

No new specs, #2218 has them. Both of its failing specs pass with this change:

- `PUT /api/parameters/shovel/...` with an HTTP `dest-uri` returns 201
- `validate_config!` accepts an HTTP destination with no dest queue or exchange

Its three passing specs stay passing: an AMQP destination still needs a queue or an exchange, a mixed HTTP/AMQP destination list is still rejected, and `httpx://` is not treated as an HTTP destination.

`make test`, `make lint` clean.

## Review order

1. #2218, specs, red
2. This PR, the fix, green
3. #1423, Standard Webhooks signing for HTTP shovels

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- HTTP shovels could no longer be created through the API, config validation demanded a destination queue or exchange that an HTTP destination has no use for [#2215](https://github.com/cloudamqp/lavinmq/pull/2215)
 - Management UI OAuth2 login now works with identity providers that require a specific scope (e.g. Entra ID), via the new `mgmt_scopes` config option [#2127](https://github.com/cloudamqp/lavinmq/pull/2127)
 - Accept JWKS keys that omit the `alg` parameter when fetching keys for OAuth/OIDC [#2124](https://github.com/cloudamqp/lavinmq/pull/2124)
 - Document that OAuth/OIDC JWTs must include a `kid` header matching a JWKS key [#2107](https://github.com/cloudamqp/lavinmq/pull/2107)
--- a/spec/api/shovels_spec.cr
+++ b/spec/api/shovels_spec.cr
@@ -148,4 +148,20 @@ describe LavinMQ::HTTP::ShovelsController do
       end
     end
   end
+
+  describe "PUT api/parameters/shovel/:vhost/:name" do
+    it "should create a shovel with an HTTP destination" do
+      with_http_server do |http, s|
+        body = {
+          value: {
+            "src-uri":   s.amqp_server.url,
+            "src-queue": "events",
+            "dest-uri":  "https://example.com/webhook",
+          },
+        }
+        response = http.put("/api/parameters/shovel/%2F/webhook-shovel", body: body.to_json)
+        response.status_code.should eq 201
+      end
+    end
+  end
 end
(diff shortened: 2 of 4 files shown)


baelter:
Taken the scheme fix, `validate_config!` now matches on `in?("http", "https")` exactly like the dispatch in `destination`. Added specs for `httpx://` and for a mixed `["https://...", "amqp:///"]` destination, both still rejected.

The 500 does not happen though. `httpx://` on main gives a 400:

```
$ curl -X PUT .../api/parameters/shovel/%2F/x-shovel -d @-
{"error":"bad_request","reason":"Shovel destination requires an exchange"}
```

`ArgumentError` is caught in `src/lavinmq/http/handler/error_handler.cr:11` along with `JSON::Error` and `TypeCastError`, so it never reaches the client as a 500. The real problem was the misleading message, and validation drifting from construction. Both gone now.

snichme (approved):

Issue #2221, opened 2026-09-04
max-connections and max-queues are not enforced for MQTT


oskgu360:
Both vhost limits are enforced only on the AMQP path. MQTT clients bypass them.

### max-connections

`vhost.max_connections` is checked in one place, `src/lavinmq/amqp/connection_factory.cr:172`. `src/lavinmq/mqtt/` contains no reference to it, so MQTT connections are neither refused when the limit is reached nor counted toward it.

### max-queues

Asymmetric: MQTT sessions are counted, but never checked.

`VHost#queue_limit_reached?` (`src/lavinmq/vhost.cr:263`) includes sessions:

```crystal
@max_queues.try { |max| definitions.queues_size + definitions.sessions_size >= max } || false
```

Its only caller is `src/lavinmq/amqp/client.cr:769`. `MQTT::Sessions#declare` calls `@vhost.declare_queue` directly, and `VHost#declare_queue` applies the frame with no limit check.

The effect is that MQTT sessions consume an AMQP client's queue budget while growing without bound themselves. A vhost with `max-queues: 150` and 200 MQTT sessions refuses every AMQP queue declaration, while further MQTT sessions keep being created.

### Expected

MQTT connections counted against and refused by `max-connections`; MQTT session creation refused when `queue_limit_reached?`.

The counting side already accounts for sessions, so the gap looks like enforcement that was not wired up rather than a deliberate exemption.

spuun:
Fixed by #2222 and #2223

PR #2222, opened 2026-09-04
Validate connection limit for mqtt connections

spuun:
Fixes half #2221


kickster97 (approved):
lgtm, needs some docs though

viktorerlingsson (approved):

PR #2223, opened 2026-09-04
MQTT should respect vhost queue (session) limit

spuun:
Fixes part of #2221

### DIFF
--- a/src/lavinmq/mqtt/client.cr
+++ b/src/lavinmq/mqtt/client.cr
@@ -197,7 +197,13 @@ module LavinMQ
       end
 
       def recieve_puback(packet : Protocol::PubAck)
-        @broker.sessions[@client_id].ack(packet)
+        # No session means we never delivered anything to ack
+        unless session = @broker.sessions[@client_id]?
+          @log.warn { "Received PubAck from client without a session" }
+          close_socket
+          return
+        end
+        session.ack(packet)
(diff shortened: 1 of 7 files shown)


src/lavinmq/mqtt/client.cr
```diff
@@ -197,7 +197,13 @@ module LavinMQ
...
 
       def recieve_puback(packet : Protocol::PubAck)
-        @broker.sessions[@client_id].ack(packet)
+        # No session means we never delivered anything to ack
+        unless session = @broker.sessions[@client_id]?
+          @log.warn { "Received PubAck from client without a session" }
+          close_socket
+          return
+        end
+        session.ack(packet)
```
kickster97: Why is this needed in this pr?

kickster97 (approved):
also docs needeed here

src/lavinmq/mqtt/client.cr
spuun: It's an unexpected ack. Maybe it should have been in it's own pr.

spuun: Well, I'll keep it. I make some sense to introduce it in this PR. It's just a nice validation, preveting a KeyError to be raised.

kickster97: Go for it!

PR #2224, opened 2026-09-04
Track PROXY-protocol connections in ConnectionInfo

kickster97:
## Summary

`ConnectionInfo` gets a `proxied` flag and a `loopback?` method. The PROXY protocol parsers build their `ConnectionInfo` with `proxied: true`. `loopback?` is true only when the remote address is a loopback address and the connection is not proxied.

A header address describes the client as seen by the proxy, not a connection on the broker host, so it never unlocks the default user. This applies to every accepted header: empty trusted sources list, listed source, and cluster follower. The default user can only connect directly on the broker host.

The AMQP and MQTT connection factories read `connection_info.loopback?` and pass it to the auth context. MQTT over WebSocket and TLS now use the same loopback rule as AMQP. The `Auth::Context` overload that took an IO has no callers left and is removed.

The V1 parser returns nil for `PROXY UNKNOWN`, so the real socket address is used.

Docs in `docs/connections.md` and `docs/authentication.md` describe the rule, and CHANGELOG has an entry under Changed. Specs cover the empty-list, listed, and `UNKNOWN` cases for AMQP and MQTT.


src/lavinmq/connection_info.cr
```diff
@@ -11,13 +11,19 @@ module LavinMQ
     property ssl_key_alg : String?
     property ssl_sig_alg : String?
     property ssl_cn : String?
+    # Set when the addresses come from a PROXY header that any peer could have sent
+    property? untrusted_proxy : Bool = false
```
spuun: IMO remove the negation to get rid of some double negations in the code
```suggestion
    property? trusted_proxy : Bool = true
```

kickster97: Yeah good call

spec/mqtt/integrations/proxy_protocol_spec.cr
```diff
@@ -0,0 +1,48 @@
+require "../spec_helper"
+
+module MqttSpecs
+  extend MqttHelpers
+  extend MqttMatchers
+
+  private def self.with_proxy_protocol(&)
```
viktorerlingsson: nit: should we move `with_proxy_protocol` to a helper so we don't need `with_proxy_protocol` both here and in `auth_spec`?

src/lavinmq/connection_info.cr
```diff
@@ -11,13 +11,20 @@ module LavinMQ
...
 
     # Remote and local addresses from the server's perspective
-    def initialize(remote_address, local_address)
+    def initialize(remote_address, local_address, *, @proxied : Bool = false)
       @remote_address = IPAddress.new(remote_address)
       @local_address = IPAddress.new(local_address)
     end
 
+    # A proxied address describes the client as seen by the proxy, not a connection on this host
+    def loopback? : Bool
```
viktorerlingsson, line 24: It might be a little confusing that we have `loopback?` on both `ConnectionInfo` and `IPAddress` (`ConnectionInfo.remote_address`) that does different things. Maybe the one on `IPAddress` should be protected instead of a getter so it can't be accessed anywhere else 🤔

kickster97: Yep, smart!

src/lavinmq/mqtt/connection_factory.cr
```diff
@@ -53,7 +49,7 @@ module LavinMQ
         io.flush
       end
 
-      def authenticate(io : Protocol::IO, packet)
+      def authenticate(io : Protocol::IO, packet, connection_info : ConnectionInfo)
```
spuun: `io` can be removed here now.


```suggestion
      def authenticate(packet, connection_info : ConnectionInfo)
```

viktorerlingsson (approved):
I think this looks good now! 👍

carlhoerberg (approved):

src/lavinmq/proxy_protocol.cr
```diff
@@ -62,7 +62,7 @@ module LavinMQ
         header.split(' ') do |v|
           case i
           when 0 then raise InvalidSignature.new(v) if v != "PROXY"
-          when 1 then nil
+          when 1 then raise InvalidFamily.new("UNKNOWN") if v.starts_with?("UNKNOWN")
```
spuun: What about:
```suggestion
          when 1 then raise InvalidFamily.new(v) if !v.in?("TCP4", "TCP6")
```
In v1 these are the only valid values (except for UNKNOWN which we don't support).

src/lavinmq/mqtt/connection_factory.cr
kickster97: Removed, only connection_info is used now.

src/lavinmq/proxy_protocol.cr
kickster97: Done. The value is chomped in the error message since the header still carries its CRLF at that point. Also fixed the "can parse valid data" spec, which sent the family TCP, and added a spec for an unknown family name.

spuun (approved):

PR #2232, opened 2026-09-08
Default permission group instead of a mode switch, keep permission_check_enabled

kickster97:
Stacked on #2126, merges into `topic-permissions`.

Replaces the mode switch in `PermissionService` with a default group. Today a vhost allows every topic while no rule exists and flips to default deny when the first group is created, so an operator who adds a group for one team locks out every other client at once. Now a topic is allowed only when a rule grants it, and a vhost without `mqtt_permissions.json` gets a group named `default` in memory, member `*` and rule `#` for read and write. A new or upgraded vhost is open as before. The operator locks it down by deleting or narrowing `default` over the API. The group is not written to disk, the first put or delete writes `mqtt_permissions.json`, and from then on the groups on disk decide. `in_use?` and the empty flag on `Compiled` are gone.

A definitions import must not keep the automatic group next to the groups it brings. The service tracks with `untouched?` that its only group is the one it created itself, and while that holds the import deletes the automatic group before it puts the imported groups and skips nothing. After the first change by anyone the startup import behaves as for policies: it adds new names and never touches an existing one. This makes a `default` group in the `load_definitions` JSON apply on a fresh server, and an export of a locked down vhost stays locked down when imported into a fresh server.

The `permission_check_enabled` option and the AMQP checks it enables are restored in front of the topic checks, so a hosted customer with the option set sees no change at the upgrade. A client that fails the AMQP check is disconnected as before, a client that fails a topic check is acked and dropped.

`lavinmqctl definitions export` emits the same `default` group for a vhost without `mqtt_permissions.json`, so the offline export and `GET /api/definitions` agree.

The allow all rule check costs about 1 ns with no allocation, so there is no fast path flag for it.

### DIFF
--- a/src/lavinmq/definitions_generator.cr
+++ b/src/lavinmq/definitions_generator.cr
@@ -2,6 +2,7 @@ require "./version"
 require "../stdlib/slice"
 require "json"
 require "amq-protocol"
+require "./mqtt/permission_group"
 
 class LavinMQCtl
   class DefinitionsGenerator
@@ -55,7 +56,7 @@ class LavinMQCtl
             json.field("mqtt_permissions") do
               json.array do
                 each_vhost do |vhost, vhost_dir|
-                  mqtt_permissions(vhost_dir).each do |g|
+                  mqtt_permissions(vhost, vhost_dir).each do |g|
                     g.as_h.merge!({"vhost" => JSON::Any.new(vhost)}).to_json(json)
                   end
                 end
@@ -180,12 +181,14 @@ class LavinMQCtl
       File.open("users.json") { |f| JSON.parse f }.as_a
     end
 
-    private def mqtt_permissions(vhost_dir)
+    # A vhost without mqtt_permissions.json has the automatic default group in
+    # memory while the server runs, so the export shows the same group.
+    private def mqtt_permissions(vhost, vhost_dir)
       File.open(File.join(vhost_dir, "mqtt_permissions.json")) do |f|
         JSON.parse(f).as_a
       end
     rescue File::NotFoundError
-      Array(JSON::Any).new
+      [JSON.parse(LavinMQ::MQTT::PermissionGroup.default(vhost).to_json)]
     end
 
     private def vhosts
--- a/src/lavinmq/mqtt/permission_service.cr
+++ b/src/lavinmq/mqtt/permission_service.cr
@@ -6,8 +6,14 @@ module LavinMQ
   module MQTT
     # Every change rebuilds the compiled state and publishes it with one
     # reference assignment, so readers never see stale or partial state.
+    #
+    # A topic is allowed only when a rule grants it. A vhost is still open by
+    # default: the DEFAULT_GROUP grants every user every topic, and the
+    # operator locks the vhost down by deleting or narrowing that group.
     class PermissionService
       Log = LavinMQ::Log.for "mqtt.permission_service"
+
+      DEFAULT_GROUP = PermissionGroup::DEFAULT_NAME
 
       record CompiledRule,
         chain : TopicRuleSegment,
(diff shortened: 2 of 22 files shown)


snichme:
Why can't we do it like we do with the default guest user? If there are no permissions we create the default one which is allow-all. But we never delete it, it's up to the user to delete it. 
Now if I start up lavinmq I get the default "allow-all" rule, then I add one rule and then the "allow-all" rule is gone. 
Scenario i'm thinking about is: A user want to prepare a set of rules while the default rule is in place so that when the rules are in place it's safe to delete the default and no clients will be impacted. 
I also don't like things that happen automaticlly like rules just disapears. 
I think that this would mean that the `untouched?` flag is not needed. 


So my suggestion is: insert the default rule if there is none, then it's up to the user to remove it.

kickster97:
> Why can't we do it like we do with the default guest user? If there are no permissions we create the default one which is allow-all. But we never delete it, it's up to the user to delete it. Now if I start up lavinmq I get the default "allow-all" rule, then I add one rule and then the "allow-all" rule is gone. Scenario i'm thinking about is: A user want to prepare a set of rules while the default rule is in place so that when the rules are in place it's safe to delete the default and no clients will be impacted. I also don't like things that happen automaticlly like rules just disapears. I think that this would mean that the `untouched?` flag is not needed.
> 
> So my suggestion is: insert the default rule if there is none, then it's up to the user to remove it.

That is how it works now. The default group is created in memory when a vhost has no `mqtt_permissions.json`. A `PUT` of a new group does not remove it. A rule is only removed when someone deletes it over the API.

The untouched flag is only for the definitions import. The startup import runs with `skip_existing` and adds only new names. Without the flag, definitions from a locked-down server, imported into a fresh server, would keep the automatic default group next to the imported groups. The new server would then be open, and nobody would know. 

So on a vhost where nobody has made changes to the permission groups (untouched), the definitions import replaces the default group with the groups from the file. After the first change by anyone (touched), the import only adds new groups. It never changes or deletes an existing group. This is the same as how we do it for policies.

src/lavinmq/mqtt/permission_service.cr
```diff
@@ -160,6 +158,15 @@ module LavinMQ
         raise ex
       end
 
+      # Created only when mqtt_permissions.json is missing, and kept in memory
+      # only: the first real change writes mqtt_permissions.json, and from then
+      # on the groups on disk decide. A deleted default group stays deleted
+      # across restarts, because the delete leaves an empty list on disk.
+      private def create_default_group
+        put(PermissionGroup.default(@vhost), save: false)
```
spuun: Why `save: false`? Wouldn't it make sense to write it instantly?

snichme:
Okay, I saw the `.delete` call on the default group and thought it was removed but saw now that it was in definiton.

snichme (approved):

spuun:
Isn't `untouched` the same as `mqtt_permission.json doesn't exist`? It would be nice to get rid of `untouched`, it feels hacky.

The service isn't saved on a graceful shutdown, should it be? Shouldn't the default group be written on a graceful shutdown?

src/lavinmq/definitions_generator.cr
```diff
@@ -180,12 +181,14 @@ class LavinMQCtl
...
-    private def mqtt_permissions(vhost_dir)
+    # A vhost without mqtt_permissions.json has the automatic default group in
+    # memory while the server runs, so the export shows the same group.
+    private def mqtt_permissions(vhost, vhost_dir)
       File.open(File.join(vhost_dir, "mqtt_permissions.json")) do |f|
         JSON.parse(f).as_a
       end
     rescue File::NotFoundError
-      Array(JSON::Any).new
+      [JSON.parse(LavinMQ::MQTT::PermissionGroup.default(vhost).to_json)]
```
spuun, line 191: I can't wrap my head around this, if it makes sense or not...

Should something that doesn't exist in the data dir be included when generating definitions from a data dir?

kickster97: Hm, i get what you mean... but it would be an generation/export of what is actually active on the broker, and i think that is more right than generating nothing when there actually are permissions applied?

spuun:
> The service isn't saved on a graceful shutdown, should it be? Shouldn't the default group be written on a graceful shutdown?

I'd say it should be written on vhost close. The user has "accepted" the default group if it's still there on shutdown.

spuun (approved):

PR #2234, opened 2026-09-09
Add state filter to queue list endpoints

weistrand:

Adds a `state` query parameter to `GET /api/queues` and `GET /api/queues/:vhost` to filter queues by state, e.g. `?state=closed` or `?state=paused,closed`. An invalid state returns 400. This makes it possible to monitor closed queues without paging through all queues. 

Should probably add state to some prometheus metric as well, but starting with this. 

More info in https://github.com/84codes/ssh-monitor/issues/313

### How can it be tested?

make test or create some queues, get them into different states and then request queues with specific states.

### DIFF
--- a/src/lavinmq/http/controller/queues.cr
+++ b/src/lavinmq/http/controller/queues.cr
@@ -31,15 +31,13 @@ module LavinMQ
         get "/api/queues" do |context, _|
           vhosts = vhosts(user(context))
           itr = vhosts.flat_map(&.queues) + vhosts.flat_map(&.sessions)
-          itr = filter_by_state(context, itr)
           page(context, itr)
         end
 
         get "/api/queues/:vhost" do |context, params|
           with_vhost(context, params) do |vhost|
             refuse_unless_management(context, user(context), vhost)
-            queues = filter_by_state(context, vhost.queues + vhost.sessions)
-            page(context, queues)
+            page(context, vhost.queues + vhost.sessions)
           end
         end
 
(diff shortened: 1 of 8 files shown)


snichme (approved):

src/lavinmq/http/controller/queues.cr
```diff
@@ -31,13 +31,15 @@ module LavinMQ
         get "/api/queues" do |context, _|
           vhosts = vhosts(user(context))
           itr = vhosts.flat_map(&.queues) + vhosts.flat_map(&.sessions)
+          itr = filter_by_state(context, itr)
```
kickster97: Should `total_count` include queues excluded by the state filter? 
here we filter before calling page, so we send in a filtered `itr` to `page()`, then with 10 queues and for example 2 paused, we will return `total_count: 2`, `filtered_count: 2`

For context, the existing name filter in `controller.cr` preserves `total_count: 10` and returns `filtered_count: 2` 

i guess the consequence is that we show "2 of 2" instead of "2 of 10" in the ui

kickster97 (commented):
Other than the comment i think its good!

src/lavinmq/http/controller/queues.cr
kickster97: A solution could be to move `filter_by_state` into `controller.cr` and use the `query_params` in `page()` like we do for name filtering

weistrand: Good point!

kickster97 (approved):
nice! 

just documenting for future: Now, if other components like connections or shovels apply their state via parameters, we get a 400 error.. But since shovels or connections don't document `state` as a supported parameter I think its fine for now, but in the future we might want to have resource-specific filtering through method overloading in each component-controller.

PR #2235, opened 2026-09-10
Changes to tooltips to work better on different screen sizes

bengtmagnus:

 Replace the tooltip positioning implementation with native CSS Anchor Positioning to fix tooltips getting clipped or badly placed when their trigger is near an edge of the
  viewport or a scrollable/overflow container.

### How can it be tested?

Tooltips (.arg-tooltip/.prop-tooltip) were positioned with a fixed bottom: 100%; left: 50%; transform: translateX(-50%), computed relative to the trigger's own containing block. That has no awareness of viewport bounds, so a tooltip near the bottom or edge of a container (a scrollable table, a form near the bottom of the page, etc.) simply got clipped or ran off-screen, with no way to reposition itself.

CSS Anchor Positioning (anchor-name, position-anchor, position-area, position-try-fallbacks) lets a tooltip be positioned relative to its trigger while its actual containing block is a different, unclipped ancestor (e.g. the viewport), and lets the browser automatically flip the tooltip to the other side when there isn't room — all without JavaScript. This is now Baseline-supported across Chrome, Firefox, and Safari.

 What changed

- .arg-tooltip/.prop-tooltip triggers no longer establish their own containing block (position: static instead of relative) — the anchor can't also be the positioned element's containing block, so this lets it bubble up to a real (non-clipping) ancestor.
- Tooltip text spans use position-area: top, position-try-fallbacks: flip-block, and justify-self: anchor-center to center above the trigger and flip below automatically when clipped.
- Every tooltip trigger/text pair gets a unique anchor-name/position-anchor so each tooltip anchors to its own trigger.
- Applied consistently across all pages using this pattern: queues, exchanges, exchange, policies, operator-policies, queue, stream, plus the dynamically-rendered tooltips in nodes.js (GC stats) and shovels.js (per-row shovel errors).

No behavior change from a user's perspective (still hover-triggered, same markup/classes, no JS added), and no visual change in the common case — only the edge-of-viewport clipping behavior is fixed.

Test by clicking around on tooltips and change window size.

### DIFF
--- a/static/js/nodes.js
+++ b/static/js/nodes.js
@@ -62,12 +62,15 @@ const renderGCStats = (gc) => {
     const th = document.createElement('th')
     const tooltip = document.createElement('a')
     tooltip.className = 'prop-tooltip'
+    const anchorName = '--tt-' + field.key
+    tooltip.style.setProperty('anchor-name', anchorName)
     tooltip.append(document.createTextNode(field.heading))
     const icon = document.createElement('span')
     icon.className = 'tooltip-icon'
     icon.textContent = '?'
     const text = document.createElement('span')
     text.className = 'prop-tooltiptext'
+    text.style.setProperty('position-anchor', anchorName)
     text.textContent = field.info
     tooltip.append(icon, text)
     th.append(tooltip)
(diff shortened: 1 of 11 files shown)


spuun (approved):
Maybe get rid of the comments (or shorten them to one-liners).

PR #2238, opened 2026-09-15
Make file replacements durable with directory fsync

carlhoerberg:
Stack 1/5 (base: `main`). Split out of #2183. Next: per-file msync + replicated fsync requests.

## What

Files rewritten via a temporary file and `rename(2)` — `definitions.amqp`, users, parameters, vhosts, MQTT retained messages and their index, the queue `.paused` marker, compacted stream consumer offsets — fsynced the new content at best, never the directory entry the rename changes. After a power loss the directory could still point at the old inode.

- New `FileSystem.durable_rename`: rename, then fsync the parent directories (both when the rename crosses directories). Leader-side replacements go through it.
- Followers fsync the parent directory of a replacement received from the leader before acking it, so promotion after a power loss cannot expose the old name-to-inode mapping. Skipped when `sync` is disabled.

## Testing

- `spec/filesystem_spec.cr`: durable_rename semantics (same-dir and cross-dir).
- `spec/clustering/client_sync_spec.cr`: follower fsyncs the parent dir on replace, and does not when sync is disabled.


src/lavinmq/mqtt/retain_store.cr
spuun, line 122: Should we sync here too?
```suggestion
        @index_file.fsync
```

spuun (approved):
Maybe we should introduce a helper to replace a file, something like
```crystal
FileSystem.replace("file.ext") do |io|
  io.write "new content"
```
which will do the temp file dance.

src/lavinmq/amqp/stream/consumer_offsets.cr
```diff
@@ -110,6 +111,7 @@ module LavinMQ::AMQP
         @mfile.write_bytes offset
       end
       @mfile.rename(old_mfile.path)
+      File.open(File.dirname(@mfile.path), &.fsync)
```
spuun, line 114: `@mfile` should be synced before dir?

PR #2239, opened 2026-09-15
Replace syncfs with per-file msync and replicated fsync requests

### LINKED ISSUE #2177: Cluster followers syncfs at the replication stream's cadence, not the leader's confirm cadence

carlhoerberg:
Stack 2/5 (base: `durable-file-replacements`). Split out of #2183. Next: confirm-loop serialization + tx fences.

Fixes #2177
Addresses #2176

## What

Publish confirms were gated on `syncfs(2)` of the whole data-dir filesystem — on the leader in the Persister and on followers before every ack batch. syncfs can stall (it flushes *every* dirty page on the filesystem) and amplifies writes: a single publish confirm also wrote out dirty pages of unrelated files (`acks.*`, other vhosts, definitions).

This replaces it with targeted durability:

- **Leader**: the `Persister` tracks which `MFile`s were written by confirmed or transactional publishes since the last sync (an atomic needs-msync flag dedupes registration) and msyncs only those via a new `MFile#fsync` (`msync(MS_SYNC)` — only actually-dirty pages are written, so no position tracking is needed). Ordinary durable publishes without confirms never enter the dirty set: a durability flag is propagated from publisher-confirm and transactional channels through routing into the message stores (classic, priority, stream, delayed).
- **Replication protocol**: instead of syncfs-before-ack, followers receive explicit fsync requests — records with a `$`-prefixed filename and zero length — and ack them only after fsyncing the named file. `wait_for_followers` therefore still means confirmed data is durable on every in-sync replica.
- `acks.*` files are never eagerly synced for plain publish confirms: losing recent acks on crash only causes redeliveries, and `prune_orphaned_acks` already handles msgs/acks divergence.

## Details

- The publish-confirm loop runs on an isolated thread that must never write the follower sockets, so `$`-records are dispatched through each follower's existing `flush_loop`; `sent_bytes` is counted at enqueue time so a `wait_for_confirm` issued right after covers them. Queued fsync requests are written under the follower's write lock before any later record is counted, so the reserved byte count has the same order as records on the wire.
- Two orderings keep a `$`-record from overtaking an append it must cover on the wire: files are marked dirty only *after* the replicator dispatch, and the drain clears the flags *before* dispatching the requests.
- `MFile` holds a mapping lock across msync and any operation that unmaps (close, truncate), so a segment closed or shrunk concurrently can't be unmapped under a running msync.
- Followers fsync files received during full_sync (the baseline is durable at join, so syncing followers are skipped by the dispatch) and fsync replace tmp-files before the rename their deferred final ack implies. Durable definition changes dispatch a fsync request for `definitions.amqp` before fencing on followers.
- A follower of a leader that predates `$` records keeps working; it simply sees no fsync requests.
- Known trade-off: file-delete dir-entry durability on followers (previously implied by syncfs) is dropped; deletes reconcile on re-sync.

## Testing

- New specs: `MFile#fsync` + dirty-flag semantics and mapping-lock protection during msync; message-store durability flag; follower handling of `$`-records (fsync+ack, missing file isn't resurrected, sync-disabled still acks, queued fences ordered with later records); client full_sync fsyncs; leader protocol spec asserting a confirm emits `$…msgs.…` and `$…definitions.amqp` but never any `acks.*` request.
- Manual smoke under `strace`: zero `syncfs` calls; one `msync(MS_SYNC)` of just the dirty segment per confirm batch; fsync only on definitions at declare.


spuun:
Two quick thoughts:

1. Maybe we should make sure to fsync mqtt qos 1 (and later 2).
2. IMO it's a little bit ugly to have to pass the `needs_sync` everywhere. I wonder if we can change this, maybe by introducing a `ConfirmMessage` we pass around?

spuun:
>     2. IMO it's a little bit ugly to have to pass the `needs_sync` everywhere. I wonder if we can change this, maybe by introducing a `ConfirmMessage` we pass around?
Or maybe add a flag to Message that's in-memory only?

src/lavinmq/message_store.cr
```diff
@@ -30,12 +31,14 @@ module LavinMQ
     getter size = 0u32
     getter empty = BoolChannel.new(true)
 
-    def initialize(@msg_dir : String, replicator : Clustering::Replicator?, durable : Bool = true, metadata : ::Log::Metadata = ::Log::Metadata.empty)
+    def initialize(@msg_dir : String, replicator : Clustering::Replicator?, durable : Bool = true, persister : Persister? = nil, metadata : ::Log::Metadata = ::Log::Metadata.empty)
```
spuun, line 34: Maybe not for this PR, but IMO this could be a better method signature. 
```suggestion
    def initialize(@msg_dir : String, *, replicator : Clustering::Replicator? = nil, durable : Bool = true, persister : Persister? = nil, metadata : ::Log::Metadata = ::Log::Metadata.empty)
```
Then the caller must be explicit about what `true` or `nil`.

src/lavinmq/persister.cr
```diff
@@ -84,27 +144,21 @@ module LavinMQ
...
       return unless acks
 
+      sync_dirty_files
       # Ask each follower's flush fiber to push the pending replicated bytes,
-      # so they persist and ack them while our own syncfs runs. Only a
+      # so they persist and ack them while our own msync runs. Only a
       # request: this loop runs on an isolated thread and must never write
       # the follower sockets itself — their fds belong to the default
       # execution context's event loop (see Follower#flush_loop).
       @replicator.try &.followers.each &.request_flush
```
spuun, line 153: Do we need this `request_flush` when `wait_for_followers` will call `request_flush`?

src/lavinmq/persister.cr
```diff
@@ -69,7 +85,51 @@ module LavinMQ
       # @publish_confirm_requested is closed; flush anything that was persisted
       # but not yet confirmed before exiting.
       drain_pending_acks
-      LibC.close(@data_dir_fd) if @data_dir_fd >= 0
+    end
+
+    private def sync_dirty_files : Nil
```
spuun, line 90: This is called from different threads, so I think we may have a race condition...

1. client A publish confirm will wake the `publish confirm loop`, then yield on read handing over to client B
2. client B handles a tx.commit which means `sync_dirty_files`.
  a. It races with the `Publish confirm loop`'s `drain_pending_acks` and wins the `@dirty_files.replace`
3. `Publish confirm loop` sees empty `@dirty_files` which means `sync_dirty_files` returns instantly so `drain_pending_acks` will move on to enqueue confirmed acks, even though it can't be sure messages are synced yet.
4. client A's confirm writer may send an ack even though sync isn't done.

src/lavinmq/persister.cr
```diff
@@ -69,7 +85,51 @@ module LavinMQ
...
+      # the ack the confirm waits for.
+      dirty.each &.clear_needs_msync!
+
+      # Ask the followers to fsync the same files, so a follower ack past
+      # these requests means the data is durable there too. Dispatched before
+      # our own msync so followers sync in parallel with us. Only a request:
+      # this may run on an isolated thread that must never write the follower
+      # sockets itself (see Follower#request_fsync).
+      if replicator = @replicator
+        paths = dirty.reject(&.deleted?).map(&.path)
```
spuun, line 116: Is it worth trying to reduce allocations like this, to reduce pressure on GC? There will be one allocation per path too in `#fsync_files`.

src/lavinmq/persister.cr
spuun: Maybe this is fixed in the next PR...

PR #2240, opened 2026-09-15
Serialize msync in the publish confirm loop and fence transactions

carlhoerberg:
Stack 3/5 (base: `per-file-msync-fsync-requests`). Split out of #2183. Next: directory-entry durability.

## Publish confirm loop

- Every confirm — sync, no-sync and clustered alike — is routed through the publish confirm loop, which is now the *only* thread that msyncs: `Persister#sync` (tx.commit) enqueues a `WaitGroup`, wakes the loop and blocks until the drain that made its writes durable — locally and on every in-sync follower — signals it. Commits batch with confirms instead of racing them, and each waiter is released only by its own batch (a later request belongs to the next drain).
- `Tx.CommitOk` is still sent by the connection fiber: it is a synchronous RPC response that must stay ordered with other responses on the channel, and the isolated loop must not write client sockets anyway. `tx.commit` now also waits for followers, so commit-ok carries the same durability guarantee as a publish confirm (previously leader-local syncfs only).
- The data-dir descriptor used by syncfs is closed in `finalize`, not when the loop exits: a tx.commit racing shutdown syncs inline once the loop is gone and may still need it.

## Transactions

- Transactional acknowledgments mark their `acks.*` files dirty like segments, so a tx.commit msyncs them locally and fences them on followers with the same `$` request. Segment removal is made durable before `deleted?` is published, since the persister skips deleted files as soon as it sees that flag. Followers persist a delete record's directory change before acking it.

## Configuration

- New `[main]` option `syncfs_threshold` (positive integer, default 10): once a confirm batch has at least this many live dirty files, one `syncfs` on a data-dir descriptor held for the persister's lifetime is cheaper than many serial msyncs. Closed and deleted files don't count toward the threshold.

## Testing

- `spec/persister_spec.cr`: waiter released only after its own batch, empty transactions don't sync, syncfs threshold (below, at, closed/deleted excluded), data-dir descriptor kept open for its lifetime and usable for an inline sync after close.
- `spec/tx_spec.cr` + clustering specs: tx.commit-ok held until the follower acks; a tx.commit fences `acks.*` while a plain confirm never does.
- `spec/message_store_spec.cr`: only transactional ack writes are marked dirty; segment removal replicated before its ack history is discarded.


src/lavinmq/mfile.cr
```diff
@@ -112,15 +112,25 @@ class MFile < IO
     addr
   end
 
-  def delete(*, raise_on_missing = true) : Nil
-    return if @deleted.swap(true, :acquire_release) # avoid double deletes
-    if raise_on_missing
-      File.delete(@path)
-    else
-      File.delete?(@path)
+  def delete(*, raise_on_missing = true, durable = false) : Nil
```
spuun, line 115: `durable` here means  `durable delete`? I think `needs_sync` is more clear, since that's used in other places.

src/lavinmq/mfile.cr
```diff
@@ -112,15 +112,25 @@ class MFile < IO
...
+    @mapping_lock.synchronize do
+      return if deleted? # avoid double deletes
+      if raise_on_missing
+        File.delete(@path)
+      else
+        File.delete?(@path)
+      end
+      # Publish deleted? only once the removal is durable: the persister may
+      # skip this file as soon as it observes that flag.
+      fsync_parent_dir if durable
```
spuun, line 125: With this, when we purge a durable queue, we'll do this for each segment. Is this necessary, or can we batch it somehow?

PR #2245, opened 2026-09-15
Guard the channel's unacked deliveries behind an UnackedStore

### LINKED ISSUE #2206: Channel's unacked deque is not actually guarded by @unack_lock
**Describe the bug**

`Channel` exposes its unacked deque directly — `getter unacked = Deque(Unack).new`
(`src/lavinmq/amqp/channel.cr:48`) — so `@unack_lock` (`:60`) guards only the callers
that remember to take it. Four readers skip it, five mutators hold it while doing queue
or socket work, and one of those re-enters it on the same fiber. Two of the resulting
defects kill the client's whole connection with `INTERNAL_ERROR (541)`; one strands a
message permanently.

These are filed together because they are one root cause and one fix, not six:

| # | Defect | Location | Status |
|---|---|---|---|
| 1 | `basic.recover(requeue: false)` reaches `Consumer#deliver` → `Channel#next_delivery_tag` while holding `@unack_lock`. It is `Mutex.new(:checked)`, so this raises `Sync::Error::Deadlock`, which the read loop turns into a 541 that kills the connection. | `channel.cr:678-700` → `:743` | reproduced |
| 2 | Transactional `basic.ack`/`basic.nack` with delivery tag 0 and `multiple` set call `Deque#last` with nothing outstanding → `IndexError` → 541, connection killed. Outside a transaction the same frame is a successful no-op (`channel.cr:556`). | `channel.cr:542`, `:609` | reproduced |
| 3 | `basic.recover(requeue: true)` skips tx-pending entries with `next`, then `@unacked.clear` drops them anyway. The delivery is neither requeued nor left for `tx.commit` to find, so the message is unreachable and the queue's unacked count is permanently wrong. The `requeue: false` branch uses `reject!`, where `next` correctly keeps the entry — the two branches disagree. | `channel.cr:681-688` vs `:690` | reproduced |
| 4 | `notify_has_capacity` only ever sets the signal `true`. Lowering global prefetch with `basic.qos` cannot take capacity away, so deliver loops wake on a stale signal, recheck, and `flush` — taking the connection write lock — in a tight loop. | `channel.cr:782-790` | reproduced |
| 5 | `Consumer#accepts?` measures the global window with `@channel.unacked.size`, counting `basic.get` deliveries that `Channel#has_capacity?` correctly excludes. Two definitions of one window. Via `immediate_delivery?`, an `immediate` publish is returned `NO_CONSUMERS` while the consumer has room. | `consumer.cr:224` vs `channel.cr:775-779` | reproduced |
| 6 | `next_delivery_tag` reads `@unacked.size` outside the `synchronize` block that precedes it, and `has_capacity?` reads it with no lock at all. `Deque#size` tears while another fiber is mid-`shift`. `has_capacity?` also subtracts two counters sampled at different moments. | `channel.cr:749-751`, `:777` | read from source |
| 7 | `Queue#unacked_messages` iterates the channel deque unlocked, from an HTTP fiber, once per consumer — so two consumers sharing a channel make every delivery appear twice. Serves a live endpoint. | `queue/queue.cr:900-910` | read from source |
| 8 | `check_consumer_timeout` builds `Channel::Close` by hand instead of routing through `close_channel`, so `@running` stays true: unacked are not requeued until the client answers `CloseOk`, and the next sweep sends a second `Channel::Close` (a protocol violation). The frame is also written while holding `@unack_lock`. | `channel.cr:757-773` | read from source |

**How to reproduce**

Drop this in `spec/unack_lock_spec.cr` and run `make test SPEC=spec/unack_lock_spec.cr`.

```crystal
require "./spec_helper"

describe "unacked" do
  # Row 1
  it "recovers to a live consumer without killing the connection" do
    with_amqp_server do |s|
      with_channel(s) do |ch|
        q = ch.queue
        q.publish "m1"
        msgs = Channel(AMQP::Client::DeliverMessage).new
        q.subscribe(no_ack: false) { |m| msgs.send m }
        msgs.receive
        ch.basic_recover(requeue: false)
        msgs.receive.redelivered.should be_true
      end
    end
  end

  # Row 2
  it "treats tx ack-all with nothing outstanding as a no-op" do
    with_amqp_server do |s|
      with_channel(s) do |ch|
        ch.tx_select
        ch.basic_ack(0_u64, multiple: true)
        ch.tx_commit
        q = ch.queue # the connection must still be usable
        q.publish "m1"
        ch.tx_commit
        q.get(no_ack: true).not_nil!.body_io.to_s.should eq "m1"
      end
    end
  end

  # Row 3 — note: plain publish, publish_confirm would conflict with tx mode
  it "keeps a tx-pending delivery outstanding across basic.recover" do
    with_amqp_server do |s|
      with_channel(s) do |ch|
        q = ch.queue("txrec")
        q.publish "m1"
        msgs = Channel(AMQP::Client::DeliverMessage).new
        q.subscribe(no_ack: false) { |m| msgs.send m }
        msg = msgs.receive
        ch.tx_select
        ch.basic_ack(msg.delivery_tag)
        ch.basic_recover(requeue: true) # must not drop the tx-pending tag
        ch.tx_commit                    # applies tx acks before sending CommitOk
        sq = s.vhosts["/"].queues.find! &.name.== "txrec"
        sq.unacked_count.should eq 0 # stranded on main: neither acked nor requeued
        sq.message_count.should eq 0
      end
    end
  end

  # Row 4
  it "takes capacity away when global prefetch is lowered" do
    with_amqp_server do |s|
      with_channel(s) do |ch|
        ch.prefetch 10, true
        q = ch.queue
        3.times { q.publish "m" }
        msgs = [] of AMQP::Client::DeliverMessage
        q.subscribe(no_ack: false) { |msg| msgs << msg }
        wait_for { msgs.size == 3 }
        ch.prefetch 1, true
        channel = s.vhosts["/"].connections.first.channels.first.as(LavinMQ::AMQP::Channel)
        channel.has_capacity?.should be_false        # passes: recomputed live
        channel.has_capacity.value.should be_false   # fails: the signal is stuck true
      end
    end
  end

  # Row 5
  it "does not count basic.get deliveries against the global window" do
    with_amqp_server do |s|
      with_channel(s) do |ch|
        ch.prefetch 1, true
        q = ch.queue
        q.publish "get-me"
        q.get(no_ack: false).should_not be_nil # one unacked, via basic.get
        delivered = Channel(String).new
        returned = Channel(String).new
        q.subscribe(no_ack: false) { |msg| delivered.send msg.body_io.to_s }
        ch.on_return { |msg| returned.send msg.body_io.to_s }
        ch.basic_publish "now", "", q.name, immediate: true
        select
        when body = delivered.receive
          body.should eq "now"
        when returned.receive
          fail "immediate publish returned though the consumer had capacity"
        when timeout 2.seconds
          fail "neither delivered nor returned"
        end
      end
    end
  end
end
```

Observed on `main`:

```
1) unacked does not count basic.get deliveries against the global window   [row 5]
   Failure/Error: fail "immediate publish returned though the consumer had capacity"

2) unacked takes capacity away when global prefetch is lowered             [row 4]
   Failure/Error: channel.has_capacity.value.should be_false
     Expected: false
          got: true

3) unacked keeps a tx-pending delivery outstanding across basic.recover    [row 3]
   Failure/Error: sq.unacked_count.should eq 0
     Expected: 0
          got: 1

4) unacked treats tx ack-all with nothing outstanding as a no-op           [row 2]
   INTERNAL_ERROR - Unexpected error, please report (541)

5) unacked recovers to a live consumer without killing the connection      [row 1]
   INTERNAL_ERROR - Unexpected error, please report (541)
```

The two 541s are the broker turning an unhandled exception into a connection close:

```
ERROR - lmq.amqp.client: Unexpected error, while reading: Index out of bounds
Index out of bounds (IndexError)
  from src/lavinmq/amqp/channel.cr:542:39 in 'basic_ack'

ERROR - lmq.amqp.client: Unexpected error, while reading: Can't lock mutex recursively
Can't lock mutex recursively (Sync::Error::Deadlock)
  from src/lavinmq/amqp/channel.cr:743 in 'next_delivery_tag'
```

**Expected behavior**

- `basic.recover` in either mode completes and returns `RecoverOk`, and never closes the
  connection (rows 1, 3).
- Transactional `basic.ack`/`basic.nack` with tag 0 and `multiple` behaves as it does
  outside a transaction: a successful no-op when nothing is outstanding (row 2).
- A delivery already acked inside an open transaction stays outstanding until
  `tx.commit`, and is never left neither requeued nor acknowledged (row 3).
- `basic.qos(global:)` moves capacity in both directions (row 4).
- One definition of the global prefetch window, excluding `basic.get` everywhere (rows 5, 6).
- `GET /api/queues/:vhost/:name/unacked` reports each delivery once (row 7).
- The consumer-timeout sweep tears the channel down at the broker and emits exactly one
  `Channel::Close` per channel (row 8).

snichme:

Channel exposed its unacked Deque directly, so @unack_lock, the basic.get counter and the has_capacity signal only stayed consistent when every caller combined them correctly. They did not. Move the deque, the lock, the consumer-unacked counter and the global prefetch capacity signal into a new UnackedStore, where every mutation runs under the lock and flips the signal in either direction.

Fixes, all from #2206:

- basic.recover(requeue: false) re-entered the checked mutex through Consumer#deliver and killed the connection with a 541. Live entries are now redelivered outside the lock with their original delivery tag, so nothing is pushed and no re-entry is possible.
- Transactional basic.ack/nack with delivery tag 0 and multiple set raised IndexError on an empty deque, also a 541. It is now a no-op, as outside a transaction.
- basic.recover(requeue: true) skipped tx-pending entries but then cleared them, stranding the message. Both branches now use reject!.
- basic.qos(global) could only grant capacity, never take it away, so deliver loops spun on a stale signal.
- Consumer#accepts? measured the global window with the raw deque size, counting basic.get deliveries that has_capacity? excludes.
- GET /api/queues/:vhost/:name/unacked listed each delivery once per consumer on the channel.
- The consumer-timeout sweep sent Channel::Close by hand and left the channel running until the client answered, sending a second Close on the next sweep. It now goes through close(reason).

Consumer#deliver loses its recover flag; the redelivery path is a separate Consumer#redeliver. Queue ack/reject work still runs under the lock for multiple-tag paths, as before.

Closes #2206

### How can it be tested?

specs


src/lavinmq/amqp/channel.cr
```diff
@@ -493,60 +484,20 @@ module LavinMQ
...
-          @unack_lock.synchronize do
-            if frame.delivery_tag.zero? && frame.multiple # all msgs so far
-              @tx_acks.push(TxAck.new @unacked.last.tag, frame.multiple, false, false)
-              return
-            elsif @unacked.bsearch { |unack| unack.tag >= frame.delivery_tag }.try &.tag == frame.delivery_tag
-              check_double_ack!(frame.delivery_tag)
-              @tx_acks.push(TxAck.new frame.delivery_tag, frame.multiple, false, false)
-              return
+          if frame.delivery_tag.zero? && frame.multiple # all msgs so far
+            if last_tag = @unacked.last_tag?
```
spuun, line 493: Before we just assumed the `@unacked.last.tag` existed, now we actually nil check. I guess this means that instead of a runtime exception we silently do nothing.

What would be the right thing to do according to AMQP spec?

src/lavinmq/amqp/unacked_store.cr
```diff
@@ -0,0 +1,128 @@
...
+
+      # Removes and yields every delivery up to and including tag, all of them if tag is zero
+      def delete_upto(tag : UInt64, & : Unack -> Nil) : Nil
+        update_capacity do
+          count = if tag.zero?
+                    @unacked.size
+                  elsif (idx = @unacked.bsearch_index { |unack, _| unack.tag >= tag }) && @unacked[idx].tag == tag
+                    idx + 1
+                  else
+                    0
```
spuun, line 69: I managed to interpret this as "if we don't find the tag, delete all (because of comment `all of them if tag is zero`). Maybe just return to make it clear that we do nothing.
```suggestion
                    return
```

src/lavinmq/amqp/channel.cr
```diff
@@ -605,27 +554,27 @@ module LavinMQ
...
-          @unack_lock.synchronize do
-            if frame.delivery_tag.zero? && frame.multiple # all msgs so far
-              @tx_acks.push(TxAck.new @unacked.last.tag, true, true, frame.requeue)
-              return
-            elsif @unacked.bsearch { |unack| unack.tag >= frame.delivery_tag }.try &.tag == frame.delivery_tag
-              check_double_ack!(frame.delivery_tag)
-              @tx_acks.push(TxAck.new frame.delivery_tag, frame.multiple, true, frame.requeue)
-              return
+          if frame.delivery_tag.zero? && frame.multiple # all msgs so far
+            if last_tag = @unacked.last_tag?
```
spuun, line 558: Same here, what's the right call if nil?

src/lavinmq/amqp/channel.cr
```diff
@@ -677,29 +624,31 @@ module LavinMQ
...
-                consumer.reject(unack.sp, requeue: true)
-              end
-              unack.queue.reject(unack.sp, requeue: true)
+        if frame.requeue
+          @unacked.reject! do |unack|
+            next false if delivery_tag_is_in_tx?(unack.tag)
+            if consumer = unack.consumer
+              consumer.reject(unack.sp, requeue: true)
             end
-            @unacked.clear
```
spuun, line 689: Is this clear gone now? Now `@unacked` will contain any `unack` for a transaction?

PR #2246, opened 2026-09-15
Add users scoped to a single vhost

carlhoerberg:

Adds users scoped to a single vhost. A user can be either global, as before, or scoped to one vhost. Scoped users are stored in `users.json` in the vhost's directory, are deleted with the vhost, can only have permissions on that vhost, and can't have the `administrator` or `monitoring` tags. Their names only have to be unique within the vhost, so `alice` on vhost `tenant` can coexist with a global `alice`; for a connection to `tenant`, the scoped user wins.

To make that work for AMQP, credentials are now verified at `Connection.Open`, when the vhost is known, instead of at `Connection.StartOk`. Wrong credentials are refused with `ACCESS_REFUSED` in reply to Open. MQTT already selects the vhost from its `vhost:username` prefix and passes it into authentication. Over HTTP (API and management UI) a scoped user logs in as `username@vhost`, split at the last `@` so email style names keep working; the plain name is always tried as a global user first.

Scoped users are managed with `/api/vhosts/{vhost}/users[/{name}[/permissions]]` and `lavinmqctl -p vhost add_user` etc. They appear in `/api/users` and `/api/permissions` with a `vhost` field, and in definitions they are nested under their vhost's entry (and at the top level of the per-vhost definitions). The management UI shows the scope on the users, user and vhost pages and locks a scoped user to its vhost.

### How can it be tested?

```
make test SPEC="spec/users_spec.cr spec/auth_spec.cr spec/http/auth_handler_spec.cr spec/api/users_spec.cr spec/api/definitions_spec.cr spec/mqtt/multi_vhost_spec.cr"
```

Manually: create vhost `tenant`, run `lavinmqctl -p tenant add_user alice secret` and `lavinmqctl -p tenant set_user_tags alice management`. `amqp://alice:secret@localhost/tenant` connects, `amqp://alice:secret@localhost/` is refused at Open, and `curl -u alice@tenant:secret localhost:15672/api/vhosts` lists only `tenant`.

https://claude.ai/code/session_01SnsKeDYcy7CqeoJL6M62AU


carlhoerberg:
All three findings are addressed in 1c5f466:

1. `GET /api/vhosts/{vhost}/users/{name}/permissions` now returns a list (empty when the user has no permissions), matching the global endpoint, so the permissions table on the user page renders for scoped users. Docs updated accordingly.
2. The `default_user_only_loopback` gate now checks the resolved user: a vhost scoped user is never treated as the default user, whatever its name. Spec added.
3. `POST /api/users/bulk-delete` accepts `{"name": .., "vhost": ..}` objects for scoped users and returns 400 for a bare name that only matches a scoped user, instead of a silent no-op. `GET /api/users` keeps listing scoped users (with their `vhost`), since the UI relies on it.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lavinmq/auth/user_store.cr
```diff
@@ -15,94 +15,175 @@ module LavinMQ
...
           return user
         end
-        user = User.create(name, password, "SHA256", tags)
-        @users[name] = user
-        Log.info { "Created user=#{name}" }
-        save! if save
+        self.class.validate_scoped_tags!(tags) if vhost
+        user = User.create(name, password, "SHA256", tags, vhost)
+        store(user)
+        Log.info { "Created user=#{user.login_name}" }
```
spuun:
```suggestion
        Log.info { "Created user=#{user.login_name}, vhost=#{vhost}" }
```

PR #2247, opened 2026-09-15
Ack requeued messages in MessageStore#purge_all

### LINKED ISSUE #2249: Purged messages reappearing after Node upgrade
**Describe the bug**
During a node upgrade, messages that were previously purged reappeared in the queue.

**Setup**
LavinMQ Version 2.9.1 -> 2.9.2

**How to reproduce**
- Purge messages from a queue
- Trigger a Node Upgrade

**Expected behavior**
Messages are totally cleansed.

jage:

Ensure all purged messages don't re-appear on restart.

#### Requeued messages
`purge_all` dropped requeued entries from memory without writing an ack record. Requeued messages sit behind the read position, so the `shift?` loop never reached them. 

When they lived in the current read or write segment the segment survived the purge and the messages came back on the next restart, since the store rebuilds its state from the segment and ack files.

#### Underflow of `@size`
When `purge_all` drops all segments it decreased `@size` with the segment count. However, if some messages in the segment has been acked they are already accounted for, so only `[segment msg count] - [segment ack count]` should be subtracted from `@size`. 

This caused the loop that purges `@rfile` and `@wfile` abort in the middle because `delete` raised an error when `@size` had reached zero. The error was rescued and logged, counters reset and `purge_all` returned successfully, even though there could be more messages left in `@rfile` and `@wfile`.

***
 
Regression since #1957 (2.9.0).

Fix #2249

### How can it be tested?

Add a spec that requeues, purges, reopens the store and expects it empty.


lukas8219 (approved):

src/lavinmq/message_store.cr
```diff
@@ -230,10 +230,11 @@ module LavinMQ
     end
 
     def purge_all
-      # Drain @requeued and decrement @size/@bytesize for each entry
+      # Drain @requeued, acking each entry so it doesn't come back after a restart
       while sp = @requeued.shift?
         @size -= 1
         @bytesize -= sp.bytesize
+        delete(sp)
```
lukas8219: This one already does the whole clustering side-effects?

lukas8219 (approved):

src/lavinmq/message_store.cr
jage: Yeah this should ensure followers also delete the files.

src/lavinmq/message_store.cr
kickster97:
```suggestion
          @size -= msg_count - (@deleted[seg_id]?.try(&.size) || 0) if seg_id > @rfile_id
```

Not from this PR, but my agent found it as the same symptom: after a restart `@size` already excludes the acks on disk, so subtracting the full `msg_count` here counts them twice, the `shift?` loop underflows, and the last `wfile` message never gets its ack.

src/lavinmq/message_store.cr
```diff
@@ -230,10 +230,11 @@ module LavinMQ
...
       end
```
kickster97, line 238:
```suggestion
      while sp = @requeued.shift?
        @size -= 1
        @bytesize -= sp.bytesize
        begin
          delete(sp)
        rescue ex : Error
          @log.warn(exception: ex) { "purge_all: could not ack requeued #{sp}, skipping" }
        end
      end
```

`delete(sp)` raises `Error` when the segment is already gone, and `Queue#purge` then closes the queue. The `shift?` loop below swallows that error with a warning. We could do the same here, so one bad entry does not fail the whole purge.

kickster97 (approved):

src/lavinmq/message_store.cr
spuun: Are you sure about this? I'm not able to reproduce that scenario.

spuun: However, the `delete` should be guarded, e.g. like
```suggestion
      while sp = @requeued.shift?
        @size -= 1
        @bytesize -= sp.bytesize
        delete(sp) if @segments.has_key?(sp.segment)
      end
```
If there's no segment for the SP, the delete may (re)create an ack file which will be dangling after `purge_all`  (cleaned up on next message store load). If the segment exists the ack file will be deleted in the `reject!` loop below.

kickster97: Not sure!

kickster97: No you're right, delete(sp) does not raise, there are defaults that protect, tested it now. My bad!

PR #2248, opened 2026-09-16
Allow sorting on boolean columns in HTTP API

oskgu360:
## Problem

In the management UI, clicking the **TLS** header on the connections list to sort returned an error: `Can't sort on type Bool`.

The TLS column sorts on the `ssl` field (`static/connections.html:198`), which is a `Bool`. The generic sorting helper `sort_by_key!` in `src/lavinmq/http/controller.cr` only handled `Number`, `String`, `QueueState` and `Nil`, so any boolean column fell through to the `else` branch and returned `400 Bad Request`.

## Fix

Add a `Bool` branch to `sort_by_key!`, ordering `false` before `true` in ascending order (`sort_reverse=true` flips it as usual). This fixes every boolean column generically, not just TLS.

## Testing

Regression spec on `/api/connections?sort=ssl` — the endpoint the bug was actually reported against. It opens one plain and one TLS connection so `ssl` has both values, and asserts both orderings.

This needed a small spec-helper addition (`extra_tls_listener`), because a TLS listener has to be bound before the server starts listening.

Confirmed the spec guards the behaviour by mutating the implementation:

- Removing the `Bool` branch entirely → fails.
- Flipping the branch to `? 0 : 1` (true before false) → fails on ordering.

`make test TAGS='~etcd'`: 2195 examples, 0 failures. The etcd-tagged specs were excluded because `etcd` isn't installed locally — unrelated to this change. `make lint`: clean.


oskgu360:
Good catch — that was a real gap, the spec was passing for the wrong reason.

Replaced it with one on durable vs. transient queues (`spec/api/http_api_spec.cr`), which gives a genuine mix of `true`/`false` and asserts both directions:

```crystal
response = http.get("/api/queues/%2F?page=1&sort=durable")
items.map(&.["durable"].as_bool).should eq [false, true]

response = http.get("/api/queues/%2F?page=1&sort=durable&sort_reverse=true")
items.map(&.["durable"].as_bool).should eq [true, false]
```

Verified it actually guards the behaviour by mutating the implementation:

- Removing the `Bool` branch entirely → fails (400, no `items`).
- Flipping the branch to `? 0 : 1` (true before false) → fails on ordering.

So both the regression and the documented `false`-before-`true` ordering are now covered. It also sits next to the other sorting specs, which is a better home for a test of the generic helper than the connections spec.

spuun (approved):

PR #2250, opened 2026-09-16
Decrease the stream segment unmap interval

carlhoerberg:

Unmap unused stream segments more often, every 5s. When a consumer is consuming multiple segments fast, they are all brought into memory, and only `madvise` `DontNeed` every 60, so the memory chart showed a sawtooth pattern. Before consuming a large stream queue could use hundreds of MB of RAM. With this change it stays pretty flat at 40 MiB.

Skipping that rewrite of the consumer-offsets file when the stream's retention floor hasn't advanced since the last cleanup, size-triggered compaction is still handled separately by `store`.


carlhoerberg:
Addressed all three points, pushed as separate commits:

**2. IO/replication churn from the 12x sweep frequency** — confirmed: `ConsumerOffsets#cleanup` unconditionally fell through to `replace_file` (new mmap, full rewrite, rename, `@replicator.replace_file`) on every call once the file was non-empty. Fixed in d1bf59c by short-circuiting when the stream's retention floor (`lowest_offset_in_stream`) hasn't advanced since the last cleanup — size-triggered compaction is already handled independently by `store`'s own `full?` check, so this is safe. The secondary `madvise`-under-lock cost is inherent to sweeping more often and scoped to already-inactive segments; not changing that here.

**3. No spec** — added in 1156466: asserts a second `cleanup_consumer_offsets` call with an unchanged retention floor does not replace the underlying `MFile` (locks in the #2 fix). Full `stream_queue_spec.cr` run: 57 examples, 0 failures.

**1. PR description claiming "behavior unchanged"** — the live PR body doesn't say that; it already states the interval intentionally changed from 60s to 5s and why. Not sure what version of the description this was reviewing against.

---
_Generated by [Claude Code](https://claude.ai/code)_

carlhoerberg:
Good catch, and correct — you're right that `store`'s `full?` path delegates to the same guarded `cleanup`, and the floor-unchanged guard did block it. I reproduced it before fixing: reverting to the unguarded call and running the new regression spec reliably raises `IO::EOFError` at exactly `consumer_offsets.cr:101` via `store_consumer_offset`, as you predicted.

Fixed in b840ced with the `force` parameter you suggested — `cleanup(force: true)` from `store`, so the capacity-driven path always compacts while the periodic-sweep-driven path keeps skipping the no-op rewrite. Added the regression spec you described: fills the file, lets it compact once, then fills it again with no retention policy set (floor pinned at 1) — confirmed it reproduces the crash without the fix and passes with it. Full `stream_queue_spec.cr`: 59 examples, 0 failures.

---
_Generated by [Claude Code](https://claude.ai/code)_

snichme (approved):

PR #2252, opened 2026-09-18
Protect internal queues

weistrand:

Refuses passive declare, delete, purge, consume, basic_get, bind and unbind on internal queues (e.g. delayed exchange queues and soon the retry queue) over AMQP with ACCESS_REFUSED. Clients could previously consume, delete or purge broker-managed internal queues, silently dropping or stealing their messages. Internal queues stay visible and manageable through the HTTP API, which remains the operator escape hatch.

### How can it be tested?

make test (new specs in spec/internal_queue_spec.cr)


weistrand:
> Internal queues are only protected on the AMQP path - the HTTP management API still allows delete/purge/get/bind


Did this deliberately, but I'm not sure, maybe we should prevent the actions for the HTTP management API as well. WDYT?

src/lavinmq/amqp/client.cr
```diff
@@ -833,13 +837,9 @@ module LavinMQ
           end
         end
         return unless valid_q_bind_unbind?(frame)
+        return unless q = bindable_queue?(frame)
```
weistrand, line 840: FYI, had to do some refactoring due to cyclomatic complexity

spec/api/bindings_spec.cr
```diff
@@ -54,6 +54,20 @@ describe LavinMQ::HTTP::BindingsController do
   end
 
   describe "POST /api/bindings/vhost/e/exchange/q/queue" do
+    it "should refuse binding a delayed exchange to its internal queue" do
```
kickster97, line 57: The bindings controller only uses `validate_delayed_binding!`. That method only validates that the destination is not the source's own delayed queue. If we change the source to `amq.direct` the bind will go through with 201, even though the destination is still an internal queue. Over AMQP the same bind is refused. 

Maybe the controller should check `internal?` also

src/lavinmq/definitions.cr
```diff
@@ -337,6 +337,7 @@ module LavinMQ
         vhosts.each_value do |v|
           v.each_queue do |q|
             next if q.exclusive?
+            next if q.responds_to?(:internal?) && q.internal?
```
kickster97: `responds_to?` is only needed because `LavinMQ::Queue` is `AMQP::Queue | MQTT::Session` and Session has no `internal?`. 

Session has `internal: false` in its details tuple, so adding `getter? internal = false` to `Session` makes this a plain call like the others. It also lets the bindings controller use `q.internal?` on the result of `find_queue`, which returns a queue or a session.

src/lavinmq/amqp/client.cr
```diff
@@ -728,6 +728,8 @@ module LavinMQ
         q = @vhost.queue?(frame.queue_name)
         if q.nil?
           send AMQP::Frame::Queue::DeleteOk.new(frame.channel, 0_u32) unless frame.no_wait
+        elsif q.internal?
+          send_access_refused(frame, "Queue '#{frame.queue_name}' is an internal queue")
```
kickster97: This message is repeated in a lot of places. A helper next to `send_access_refused` (src/lavinmq/amqp/client.cr:608) would keep them in sync, and would we able to include the vhost name:

```
def send_internal_queue_refused(frame, name)
  send_access_refused(frame, "Queue '#{name}' in vhost '#{@vhost.name}' is an internal queue")
end
```
`channel.cr` can then call `@client.send_internal_queue_refused(frame, frame.queue)`.

```suggestion
          send_internal_queue_refused(frame, frame.queue_name)
```

weistrand: b94302595

src/lavinmq/definitions.cr
weistrand: 32ce5a1bd

spec/api/bindings_spec.cr
weistrand: https://github.com/cloudamqp/lavinmq/pull/2252/changes/c90a6bf67a5b560a3acc740b11a7c5b6bda7f4af

kickster97 (approved):

spuun:
> > Internal queues are only protected on the AMQP path - the HTTP management API still allows delete/purge/get/bind
> 
> Did this deliberately, but I'm not sure, maybe we should prevent the actions for the HTTP management API as well. WDYT?

I think it's fine to start by locking down AMQP. The HTTP API is used there to manage things, and some things should be managable when it comes to internal queues.

kickster97:
> > > Internal queues are only protected on the AMQP path - the HTTP management API still allows delete/purge/get/bind
> > 
> > 
> > Did this deliberately, but I'm not sure, maybe we should prevent the actions for the HTTP management API as well. WDYT?
> 
> I think it's fine to start by locking down AMQP. The HTTP API is used there to manage things, and some things should be managable when it comes to internal queues.

I guess the only thing we might want to block is bind? I assume we want to be able to delete(?), purge and get as management

PR #2254, opened 2026-09-18
Automatic retries with backoff

weistrand:

Adds broker-side automatic retries: a queue declared with `x-delayed-retry-min` parks messages rejected with `requeue=true` in an internal `amq.retry-<queue>` queue (backed by the delayed message store) and redelivers them after a growing backoff, until `x-delivery-limit` (default 20 when unset) dead-letters them. `x-delayed-retry-multiplier` selects linear/constant/exponential backoff and `x-delayed-retry-max` caps a single delay. Broker-initiated requeues (channel close, `basic.recover`) stay instant but still count towards the limit. The retry queue is recreated on demand if it disappears, and retry is refused in combination with message deduplication. Docs in `docs/automatic-retries.md`.

Stacked on #2252 (internal queue protection); the base retargets to main when that merges.

Fixes #1815

### How can it be tested?

make test (specs in spec/retry_queue_spec.cr)


src/lavinmq/amqp/queue/queue.cr
```diff
@@ -482,6 +492,20 @@ module LavinMQ::AMQP
...
+      @delayed_retry_min = parse_header("x-delayed-retry-min", Int).try(&.to_i64)
+      if @delayed_retry_min
+        if @deduper
+          raise LavinMQ::Error::PreconditionFailed.new(
+            "x-delayed-retry-min cannot be combined with x-message-deduplication")
+        end
+        @delivery_limit ||= DEFAULT_DELAYED_RETRY_DELIVERY_LIMIT
+        @effective_args << "x-delayed-retry-min"
+        @delayed_retry_multiplier = parse_header("x-delayed-retry-multiplier", Int).try(&.to_i32)
+        @effective_args << "x-delayed-retry-multiplier" if @arguments["x-delayed-retry-multiplier"]?
```
spuun: I think it will be the same result, but be consistent I guess this should be
```suggestion
        @effective_args << "x-delayed-retry-multiplier" if @delayed_retry_multiplier
```

src/lavinmq/amqp/queue/queue.cr
```diff
@@ -482,6 +492,20 @@ module LavinMQ::AMQP
...
+        @delayed_retry_max = parse_header("x-delayed-retry-max", Int).try(&.to_i64)
+        @effective_args << "x-delayed-retry-max" if @arguments["x-delayed-retry-max"]?
```
spuun:
```suggestion
        @effective_args << "x-delayed-retry-max" if @delayed_retry_max
```

